PR #46, opened 2026-05-26
Add LM Studio local OpenAI-compatible provider support

Ivan-Ryukendo:
## Summary

This PR adds LM Studio as a local LLM provider option for Karpathy LLM Wiki.

- Adds `LM Studio (Local)` to the provider registry.
- Adds settings UI support for:
  - editable LM Studio base URL, defaulting to `http://localhost:1234/v1`
  - optional token/key input for local endpoints that require one
  - freeform model identifier input
- Routes LM Studio through the existing OpenAI-compatible request and streaming client path.
- Keeps model selection fully user-controlled, with no hard-coded LM Studio model names.
- Updates the README with basic LM Studio setup instructions.

## Implementation Notes

LM Studio exposes an OpenAI-compatible local API, so this implementation reuses the existing OpenAI-compatible client behavior instead of adding a separate request stack.

The provider configuration adds `lm-studio` with a local default base URL and an empty default model. The model field remains freeform because LM Studio users may load any local model and the plugin should not assume or hard-code model identifiers.

The settings panel treats LM Studio similarly to other local providers where an API key is not required, while still allowing an optional token to be provided. When switching to LM Studio, any previous provider API key is cleared so the optional token field starts empty. Fetching models from `/v1/models` is supported, but fetched results do not overwrite the manually entered model identifier for LM Studio.

Trailing slashes are normalized from OpenAI-compatible base URLs before request construction to avoid malformed endpoint URLs.

## Test Plan

Verified locally with:

- `npm run build`
- `npm run lint`
- `npm test`

Result: TypeScript build completed successfully, lint completed successfully, and the test suite passed with 113 tests.

## Attribution

Implemented by Ivan Ryukendo in collaboration with OpenAI Codex as a software development assistant.

## Maintainer Note

This change is intentionally conservative: LM Studio support is added by extending the existing provider abstraction and OpenAI-compatible API flow. It uses LM Studio's local OpenAI-compatible API and does not include any hard-coded local model names, so users retain full control over the model identifier used by their local LM Studio server.

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,12 +106,14 @@ This project evolves rapidly — new features, bug fixes, and improvements are s
 ### 🔑 Configure an LLM Provider
 
 1. Open Settings → Karpathy LLM Wiki
-2. Pick a provider from the dropdown (Anthropic, Anthropic Compatible, Google Gemini, OpenAI, DeepSeek, Kimi, GLM, Ollama, OpenRouter, or custom)
-3. Enter your API key (not needed for Ollama)
+2. Pick a provider from the dropdown (Anthropic, Anthropic Compatible, Google Gemini, OpenAI, DeepSeek, Kimi, GLM, Ollama, LM Studio, OpenRouter, or custom)
+3. Enter your API key (not needed for Ollama; optional for LM Studio)
 4. Click **Fetch Models** to populate the model dropdown, or type a model name manually
 5. Click **Test Connection**, then **Save Settings**
 
 **🦙 Ollama (local, no API key):** Install [Ollama](https://ollama.com), pull a model (`ollama pull gemma4`), select "Ollama (Local)" in the provider dropdown.
+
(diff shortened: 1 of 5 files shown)


Ivan-Ryukendo:
Follow-up fix added after testing against an LM Studio REST API v1 local server.

The original implementation routed LM Studio through the OpenAI-compatible client with `response_format: { type: "json_object" }`. A plain Test Connection passed, but ingestion failed because ingestion uses JSON mode and the current LM Studio server returns HTTP 400 for `json_object`:

```text
'response_format.type' must be 'json_schema' or 'text'
```

The branch now maps internal JSON-object requests to LM Studio-compatible `json_schema` response format while keeping the existing `json_object` behavior for other OpenAI-compatible providers.

Additional verification:

- Direct LM Studio probe with `json_object`: reproduced HTTP 400.
- Direct LM Studio probe with `json_schema`: succeeded and returned JSON.
- `npm run build`
- `npm run lint`
- `npm test` (113 tests passed)

Ivan-Ryukendo:
Closing for now while OpenRouter privacy-routed local testing is completed. We will reopen or create a fresh PR after the LM Studio and OpenRouter changes are verified locally.

green-dalii:
@Ivan-Ryukendo 
Thank you for submitting the PR. I will consider providing support from LMStudio in the future. Please feel free to contact me if you have any questions.

PR #54, opened 2026-05-28
fix: prevent cross-folder entity/concept duplicates and sharpen classification

dmarchevsky:
## Summary
- **Cross-folder dedup**: \`resolvePagePath()\` now checks the opposite folder (entities ↔ concepts) after same-type matches fail. When a cross-type collision is found the existing page gets the new name as an alias and creation is skipped — no more duplicate pages for the same topic in both folders.
- **Classification Decision Tree**: Extraction prompt gains a 12-step ordered decision tree for entity vs. concept decisions. \`entity/other\` is now restricted to observable/instantiable things; a "prefer concept over entity" tie-breaker reduces mis-classification of techniques and paradigms.

## Test plan
- [x] \`pnpm lint\` — 0 errors
- [x] \`pnpm test\` — 165 tests pass
- [x] \`npx tsc --noEmit\` — 0 errors
- [x] \`pnpm build\` — clean exit
- [ ] Manual: ingest two sources that reference the same AI framework — verify only one wiki page is created, second run appends alias

### DIFF
--- a/src/wiki/page-factory.ts
+++ b/src/wiki/page-factory.ts
@@ -73,16 +73,19 @@ export class PageFactory {
 
   // Determine the actual file path for a new entity/concept, using slug-based
   // matching first and falling back to LLM semantic resolution.
+  // Returns null when a cross-type collision is detected (same name exists in the
+  // opposite folder) — callers must skip page creation in that case.
   private async resolvePagePath(
     name: string,
     pageType: 'entity' | 'concept',
     summary: string
-  ): Promise<string> {
+  ): Promise<string | null> {
     const folder = pageType === 'entity' ? 'entities' : 'concepts';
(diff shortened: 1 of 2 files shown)


green-dalii:
Good advice and PR. I will seriously consider it.

green-dalii:
Thank you for this excellent contribution! 🎉

**What we merged:**
- Cross-folder dedup in `resolvePagePath()` — clean defensive layer
- Classification Decision Tree — root cause fix for entity/concept misclassification

**Follow-up plan:**
We're planning a follow-up PR to improve user visibility of cross-type collisions:
- Add collision count to ingestion completion summary
- Display Notice when collisions occur
- Record collisions in log.md for traceability

This will complement your solid defensive implementation with user-facing feedback.

Reviewed-by: @greener-dalii

PR #77, opened 2026-06-02
feat: add useSubfolders setting for flat wiki layout

dmarchevsky:
## Summary

- Adds a `useSubfolders` toggle in Settings (default: `true`) so users can place all entity/concept pages directly in `wiki/` instead of `wiki/entities/` and `wiki/concepts/` subfolders
- In flat mode, page type is stored in frontmatter (`type: entity` / `type: concept`); index generation reads it back to classify pages
- Fixes a prompt bug found during review: `generateEntityPage` and `generateConceptPage` had hardcoded subfolder path examples (`[[entities/Qwen|Qwen]]`) that would cause the LLM to generate incorrect wiki-links in flat mode — replaced with a `{{link_format_instruction}}` placeholder injected at runtime with the correct format for the active mode

## Test plan

- [ ] Toggle `useSubfolders` off, ingest a source file — pages land in `wiki/` root, not `wiki/entities/` or `wiki/concepts/`
- [ ] Wiki-links in generated pages use `[[Slug|Name]]` format (no folder prefix) in flat mode
- [ ] Toggle `useSubfolders` on — pages land in subfolders as before, links use `[[entities/Slug|Name]]` format
- [ ] Generate Index works in both modes
- [ ] `pnpm lint && pnpm test && npx tsc --noEmit && pnpm build` all pass


green-dalii:
Hi Dima,

Thanks for the detailed proposal — the `useSubfolders` toggle is clearly a well-thought-out feature that you've tested with real usage. I really appreciate the effort you've put into this, including the prompt-level fix for hardcoded subfolder examples.

That said, after careful consideration, I won't be merging this one. Let me explain why.

**Karpathy's original vision for this plugin is built around the entities/concepts folder split.** The separation isn't just an organizational preference — it's the physical layer of a knowledge graph. Entities are concrete nodes you can point to (people, products, papers), concepts are abstract frameworks you reason with (methods, theories, paradigms). The folder structure carries this ontology directly into the file system so it survives even when you're browsing your vault without the plugin running. Obsidian's Graph View benefits from it visually, and the LLM's output quality is grounded by this clear distinction.

**The cost-benefit doesn't balance out for me.** Flattening the wiki saves a few path characters, but it doubles the conditional branching across the entire codebase — page creation, path resolution, index generation, prompt injection, and reporting all need `if (useSubfolders)` forks. Every future feature will need testing in both modes. The maintenance surface area grows permanently.

**Most critically, this is a breaking change that can't coexist cleanly with existing wikis.** Users who already have `wiki/entities/` and `wiki/concepts/` folders with hundreds of pages would face a migration nightmare — or worse, a mixed-mode mess where some pages are in subfolders and others aren't, with no clear upgrade path. A settings toggle doesn't solve backward compatibility when the physical file layout is the contract.

I want to keep the plugin's core architecture opinionated and simple. LLM auto-organization means the system makes classification decisions so the user doesn't have to. Adding a layout switch turns that into a user responsibility, which runs counter to the "just write, let the LLM organize" philosophy.

Again, I genuinely appreciate the contribution. The prompt-level cleanups in this PR are excellent, and your cross-type dedup work in #54 was a real quality improvement for the project. This particular feature just doesn't fit the architectural direction.

PR #78, opened 2026-06-02
fix: respect useSubfolders flag in mergePage prompt injection

dmarchevsky:
## Summary

- `mergePage` in `page-factory.ts` was missing the `useSubfolders`-aware link format injection that `createNewPage` already had
- When `useSubfolders=false`, the LLM merging into an existing page would receive hardcoded subfolder format instructions (`[[entities/Foo|Bar]]`) instead of flat format (`[[Foo|Bar]]`)
- Fixed by adding `linkFormatInstruction` injection in `mergePage` (mirroring the existing pattern in `createNewPage`)
- Fixed `mergeEntityPage` and `mergeConceptPage` prompts to use `{{link_format_instruction}}` placeholder instead of hardcoded subfolder examples; also replaced the schema rule lines that referenced `[[concepts/...]]` / `[[entities/...]]` directly

## Test plan

- [x] `pnpm lint` — 0 errors, 0 warnings
- [x] `pnpm test` — 400/400 pass
- [x] `pnpm build` — clean exit


green-dalii:
Hi Dima,

Following the decision on #77, I'm closing this one as well since it's a direct fix for the `useSubfolders` path introduced there.

The `mergePage` link format injection you identified *is* a real gap — it shows exactly why the flat layout feature has a larger blast radius than it appears on the surface. If the merge path needed a follow-up fix, the lint and query paths likely have similar gaps that weren't caught in initial testing. That kind of whack-a-mole is what I was worried about in #77.

The PR itself is well-scoped and cleanly implemented. If the flat layout direction ever becomes viable in the future, this patch would be a good starting point to revisit.

PR #82, opened 2026-06-02
fix: enforce wiki language for alias generation

dmarchevsky:
## Problem

Alias generation in both lint (`runAliasCompletion`) and ingestion (entity/concept/source page generation) was producing language-pair translations — e.g. adding English aliases for Chinese wikis and vice versa. This conflicts with the wiki language setting.

## Fix

- **`src/wiki/lint/fix-runners.ts`**: injects `buildWikiLanguageDirective` as the system prompt in `runAliasCompletion()` so the LLM knows the target language when suggesting aliases
- **`src/wiki/prompts/generation.ts`**: removes the hardcoded Chinese↔English translation instruction in 3 prompt sites (entity generation, concept generation, source generation); replaces with: *"All aliases MUST be in the same language as the wiki. Do NOT add translations into other languages."*
- **`src/wiki/prompts/lint.ts`**: updates alias suggestion instruction to reference "wiki language" rather than the prior hardcoded language pair

## Test plan

- [x] TypeScript: 0 errors
- [x] Tests: 400/400 pass
- [x] Build: clean exit


green-dalii:
@dmarchevsky — clean. Funneling the 4 alias paths through the same system prompt + reusing `buildWikiLanguageDirective` instead of a new abstraction is exactly right. 400/400 tests, build clean, no conflicts.

One nit: in the three `generation.ts` blocks, the new "use the page title itself" reads slightly ambiguous — an LLM could read it as "the only alias is the title." Reword to:

> If no natural alias exists, use the page title itself as the first alias. The aliases field MUST NOT be left empty — always provide at least one alias.

Push as a follow-up commit or amend this one — your call, I can merge either way. CHANGELOG entry's on me, targeting today's release.

dmarchevsky:
It worked fine on Gemma 4 E4B, so I suggest keeping it as is.
Thanks!

green-dalii:
Implemented in commit 4cec236 (and refined in 138ebb4 with English-as-linker-language). Thanks for catching the hardcoded Chinese↔English assumption — replaced with a single shared directive + 4 examples including the explicit 'do NOT invent translations' rule. Closing in favor of the in-tree version.

green-dalii:
Superceded by v1.16.0 (commit 4cec236 + follow-ups). We kept the injection of buildWikiLanguageDirective and the same-language-alias direction you established, refined with an 'English as linker language' rule and explicit examples to prevent LLMs from inventing non-existent translations (e.g. Transformer → 变换器). The original PR and this release share the same goal — alias quality has improved measurably. Thanks for the contribution.

green-dalii:
**We strongly recommend upgrading to v1.16.0.** The alias improvement ships alongside the sources normalization fix and startup quick fixes — all in one update. Just install via the Community Plugins market.

PR #87, opened 2026-06-03
fix+feat: enforce lowercase slugs and detect/fix case-variant duplicate wiki pages

dmarchevsky:
## Problem

Two related issues with case-variant duplicate wiki pages:

1. **Prevention gap**: `computeSlug()` was case-preserving, so parallel page generation could produce both `entities/Claude-Code.md` and `entities/claude-code.md` for the same concept — a silent data-corruption race condition on case-sensitive filesystems.

2. **No remediation**: Existing wikis with historical case-variant duplicates (e.g. from LLM output that varied capitalization across ingestion runs) had no way to detect or fix them.

## Fix

**Commit 1 — Prevention** (`fix: enforce lowercase slugs`):
- `computeSlug()` in `src/utils.ts` now lowercases its output
- `wiki-engine.ts`: deduplicates the `pagesToCreate` list by slug before the parallel `Promise.allSettled` fan-out, so two entities resolving to the same lowercase slug only generate one page
- Unit tests updated to assert lowercase slug output

**Commit 2 — Remediation** (`feat: detect and fix case-variant duplicate wiki pages in lint`):
- New **Phase -2** in the lint pipeline (runs before all existing phases) scans for pages whose lowercase-normalized filenames collide
- Detects all four wiki-link reference forms: full-path `[[entities/Claude Code]]`, bare-name `[[Claude Code]]`, and their case variants
- Merges duplicate content into the canonical (lowercase) page, rewrites all wiki-links pointing at the old casing, and renames standalone uppercase-named pages
- Shown in lint report modal with count of fixed duplicates

## Test plan

- [x] TypeScript: 0 errors
- [x] Tests: 452/452 pass (65 new lines in `scanners.test.ts` covering case-variant detection)
- [x] Build: clean exit


green-dalii:
Strong PR. Prevention fix is correct; Phase -2 ordering is the right call.

**`mergeDuplicatePages` semantics** — traced it (`lint-fixes.ts:482-619`). It's **LLM-driven, not concat**: the function calls the LLM with `PROMPTS.mergeDuplicatePages` asking for an "intelligent fuse", parses the JSON `{body, aliases}` response, and merges it back. Frontmatter (`sources`, `aliases`) is merged programmatically. Fallback: if the LLM call fails or returns empty, it appends `## From <sourceTitle>\n\n<sourceBody>` to the target body. So destructive behavior is bounded by LLM unreliability, not overwrite.

**Risk on "Fix All"** — Phase -2 calls `wikiEngine.mergeDuplicatePages` and `wikiEngine.normalizePageCase` directly with no preview. For users who intentionally kept case variants (rare but possible), this is silent destruction. **Suggestion**: gate behind a confirmation modal listing proposed merges/renames, matching #89's Phase 5 progress UI.

**Other**:
- `detectedUppercasePageNames` — deterministic across runs?
- Test coverage is on detection (`scanners.test.ts`). A test mocking `wikiEngine` and verifying the right merge/normalize calls would close the fix-path loop.

**Split**: Commit 1 (prevention) is tiny and safe — could ship standalone. Commit 2 (remediation) benefits from the confirm-modal work first.

**Sequencing with #88**: #88 → #87, so the new `bodyWordSet` gate in #88 doesn't over-report case variants until #87's detection is in place.

Approved contingent on confirm-modal for Commit 2.

dmarchevsky:
Thanks for the thorough review. Addressed in a new commit ([e6524e1](https://github.com/dmarchevsky/obsidian-llm-wiki/commit/e6524e1)).

**Confirmation modal (main blocker)**
Added `CaseVariantConfirmModal` in `src/ui/modals.ts`. It lists all proposed merges ("source → target, merge & delete source") and renames separately, with a "Confirm & fix" / "Cancel" button pair. The `onNormalizeCaseVariants` callback in `lint-controller.ts` now opens this modal first; the fix runner only fires on explicit confirmation.

**Fix All scoping**
Left Phase -2 in Fix All unchanged. Fix All is already an explicit opt-in action and every other phase runs without per-phase confirmation — inserting a blocking modal mid-run would break that contract. The FixReportModal at the end already surfaces what Phase -2 did.

**`detectedUppercasePageNames` determinism**
Detection iterates over vault file enumeration (non-deterministic order), but the *set* of detected pages is always the same. The modal will always list the same merges/renames, just possibly in a different order across runs. Benign.

**Fix-path loop test**
Agreed it would close the loop. `CaseVariantConfirmModal` is an Obsidian `Modal` subclass — not unit-testable without App mock — so no test for the modal itself. A test mocking `wikiEngine` and verifying the right merge/normalize calls fire on `runCaseNormalizationFixes` inputs would be the right addition; deferred to P1 backlog alongside the other mock-heavy tests.

green-dalii:
Thanks @dmarchevsky for raising this — your analysis was spot-on. computeSlug() needed lowercasing, and case-variant detection was a genuine gap.

We incorporated the approach locally with one design change: instead of adding a separate Phase -2 to the lint pipeline, we integrated the caseVariant signal into the existing generateDuplicateCandidates infrastructure (same function the bigram and crossLang signals use). This means case-variant pages are caught by the existing Tier 1 → merge flow, with no new i18n keys, modal UI, or settings toggle. The merge handling (wikiLink rewriting, source deletion, whole-vault dead link resolution) is the same.

Released as part of v1.16.1.

PR #88, opened 2026-06-03
fix: eliminate two lint false-positive classes

dmarchevsky:
## Problem

Two independent false-positive categories in the lint pipeline:

### Bug 1 — Duplicate detection (sharedLinks signal)

Pages were reported as duplicates solely because they shared the same set of outgoing wiki-links, even when their content was completely unrelated. Example: three pages all linking only to a single hub page (e.g. `[[Persistent-Wiki]]`) would score 100% link-set Jaccard similarity and be flagged as duplicates of each other.

### Bug 2 — Dead link scanner

Links were reported as dead when the target page existed under a slug with hyphens where the link used spaces (or vice versa). Example: `[[entities/Claude Code]]` was flagged as a dead link when the file existed as `entities/Claude-Code.md`.

### Bug 1 — Word-set body-similarity gate

Add a second signal: **word-set Jaccard similarity** of page body text. The sharedLinks candidate is only kept if `bodyWordSim >= 0.2`.

Implementation:
- `bodyWordSet(text)` — strips wiki-links, lowercases, removes stopwords and tokens ≤ 3 chars, returns `Set<string>` of meaningful words
- `computeJaccard(setA, setB)` — generic Jaccard utility (already used for link sets)
- Body words are pre-computed in the metadata pass (wiki-links stripped first so link text doesn't double-count)
- Threshold 0.2 is intentionally low — word sets are far more selective than bigram sets, so genuine duplicates easily clear it while different-topic pages (sharing only a hub link) do not

Both functions are exported pure functions with direct unit tests.

### Bug 2 — Slug-normalized fallback

Before declaring a link dead, check a `spaces→hyphens` normalized variant of the target path. 3-line change in `scanDeadLinks`.

## Files changed

- `src/wiki/lint/duplicate-detection.ts` — `bodyWordSet()`, `computeJaccard()`, word-set gate in `generateDuplicateCandidates()`
- `src/wiki/lint/scanners.ts` — slug-normalized fallback in `scanDeadLinks()`
- `src/__tests__/duplicate-detection.test.ts` — new file, validates `bodyWordSet()`, `computeJaccard()`, and the sharedLinks gate behavior
- `src/__tests__/scanners.test.ts` — 2 new cases for the space/hyphen dead-link scenarios

## Test plan

- [x] TypeScript: 0 errors
- [x] Tests: 454/454 pass
- [x] Build: clean exit

### DIFF
--- a/src/__tests__/duplicate-detection.test.ts
+++ b/src/__tests__/duplicate-detection.test.ts
@@ -47,6 +47,28 @@ describe('bodyWordSet', () => {
     const sim = computeJaccard(wA, wB);
     expect(sim).toBeGreaterThan(0.5);
   });
+
+  it('returns non-empty set for CJK text', () => {
+    // Regression: [^a-z0-9\s] stripped all CJK chars, producing empty sets
+    // that made computeJaccard return 0 and silently blocked CJK duplicate detection.
+    const words = bodyWordSet('深度学习是人工智能的核心技术之一 机器学习是基础');
+    expect(words.size).toBeGreaterThan(0);
+  });
+
+  it('produces high Jaccard for similar CJK texts', () => {
+    const shared = '深度学习是人工智能的核心技术之一 机器学习是深度学习的基础 神经网络架构';
+    const wA = bodyWordSet(shared + ' 图像识别卷积网络');
+    const wB = bodyWordSet(shared + ' 自然语言处理变换器');
+    const sim = computeJaccard(wA, wB);
+    expect(sim).toBeGreaterThanOrEqual(0.2);
+  });
+
+  it('produces low Jaccard for different-topic CJK texts', () => {
+    const wA = bodyWordSet('深度学习是人工智能的核心技术 神经网络用于图像识别任务');
+    const wB = bodyWordSet('历史是人类文明的记录 古代文化与现代社会的联系');
+    const sim = computeJaccard(wA, wB);
+    expect(sim).toBeLessThan(0.2);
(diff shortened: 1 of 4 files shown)


green-dalii:
Both fixes are targeted and address real false-positive classes. One thing to flag:

**CJK regression in `bodyWordSet`** — the strip regex `[^a-z0-9\s]` removes all non-ASCII. For Chinese/Japanese vaults, body word sets are empty for any non-trivial content, and `computeJaccard(empty, empty)` returns 0. The `bodySim < 0.2` gate fails and **CJK content is never flagged as duplicate** — a regression vs. the previous bigram-based detector.

Fix (preserve CJK, matching `normalizeForMatch` in the same file):
```typescript
text.toLowerCase().replace(/[^\w\s一-鿿]/g, ' ')
```

Ship as-is and track as a follow-up, or fold in — your call.

dmarchevsky:
Incorporated the suggested change to fix CJK regression

green-dalii:
Thanks @dmarchevsky. Both fixes were verified and merged as-is in v1.16.1:

1. Body similarity gate (bodyWordSet + 0.2 Jaccard threshold) — unit-tested with English and CJK edge cases. The stopword list (45 English function words) and the regex path for CJK [^a-z0-9\s] are working as designed.

2. scanDeadLinks space→hyphen normalization — also unit-tested with the space-variant and truly-unknown-target cases you provided.

Released as part of v1.16.1.

PR #89, opened 2026-06-03
feat: copy source pages with auto-linked wiki references

dmarchevsky:
## Summary

After ingestion completes, the plugin can optionally save a linked copy of each ingested source file to a configurable folder (default: `pages/`). Every mention of a known entity or concept in the copy is auto-linked to its wiki page using Obsidian wiki-link syntax.

The feature is **disabled by default** — zero impact on existing users unless they opt in.

## Motivation

The wiki pages accumulate structured knowledge, but the original source text loses its context once ingested. This feature bridges the gap: the copy in `pages/` becomes a "marked-up" version of the source where every named entity is a live wiki-link, useful for reviewing what was extracted and for navigating from source text directly into the wiki.

## What changes

**New module** — `src/core/wikilink-inserter.ts` (pure function, 71 lines):
- `insertWikiLinks(content, wikiPages)` — splits frontmatter from body, inserts `[[path|display]]` links for every entity/concept title and alias, processes candidates longest-first to avoid partial matches, skips text already inside `[[...]]`
- No Obsidian runtime dependency; fully testable

**Settings** — new toggle + folder input in the plugin settings panel:
- `copySourcePages: boolean` (default: `false`)
- `copySourcePagesFolder: string` (default: `"pages"`)

**Ingestion** — `wiki-engine.ts`: after all source pages finish ingesting, if `copySourcePages` is enabled, writes linked copies to the configured folder

**Lint Fix All** — `lint-controller.ts`: new Phase 5 refreshes all existing copies in `copySourcePagesFolder` whenever the wiki grows (new pages = new link opportunities)

**i18n** — 6 new string keys added to all 8 locale files (EN, DE, ES, FR, JA, KO, PT, ZH)

## Files changed

| File | Change |
|------|--------|
| `src/core/wikilink-inserter.ts` | New module |
| `src/__tests__/wikilink-inserter.test.ts` | 94-line test suite |
| `src/types.ts` | `copySourcePages`, `copySourcePagesFolder` fields |
| `src/ui/settings.ts` | Toggle + folder input |
| `src/wiki/wiki-engine.ts` | Post-ingestion copy pass |
| `src/wiki/lint-controller.ts` | Phase 5 refresh |
| `src/wiki/lint-fixes.ts` | Helper wired in |
| `src/texts/{en,de,es,fr,ja,ko,pt,zh}.ts` | 6 new i18n keys each |
| `src/__tests__/utils.test.ts` | Mock engine-context updated |

## Test plan

- [x] TypeScript: 0 errors
- [x] Tests: 456/456 pass (94 new lines in `wikilink-inserter.test.ts`)
- [x] Build: clean exit
- [x] Disabled by default — all existing tests unaffected

### DIFF
--- a/src/__tests__/__mocks__/engine-context.ts
+++ b/src/__tests__/__mocks__/engine-context.ts
@@ -80,6 +80,8 @@ export const DEFAULT_SETTINGS: LLMWikiSettings = {
   startupCheck: false,
   pageGenerationConcurrency: 1,
   batchDelayMs: 0,
+  copySourcePagesToWiki: false,
+  pagesFolder: 'pages',
   llmReady: true,
   maxTokensPerCall: 0,
 };
--- a/src/__tests__/utils.test.ts
+++ b/src/__tests__/utils.test.ts
@@ -662,6 +662,7 @@ describe('getGranularityInstruction', () => {
     enableSchema: true, autoWatchSources: false, autoWatchMode: 'notify',
     autoWatchDebounceMs: 5000, watchedFolders: [], periodicLint: 'off',
     startupCheck: false, pageGenerationConcurrency: 3, batchDelayMs: 500,
+    copySourcePagesToWiki: false, pagesFolder: 'pages',
     llmReady: false,
     maxTokensPerCall: 0,
   };
@@ -707,6 +708,7 @@ describe('getGranularityFixLimits', () => {
     enableSchema: true, autoWatchSources: false, autoWatchMode: 'notify',
     autoWatchDebounceMs: 5000, watchedFolders: [], periodicLint: 'off',
     startupCheck: false, pageGenerationConcurrency: 3, batchDelayMs: 500,
+    copySourcePagesToWiki: false, pagesFolder: 'pages',
     llmReady: false,
     maxTokensPerCall: 0,
   };
--- a/src/__tests__/wikilink-inserter.test.ts
+++ b/src/__tests__/wikilink-inserter.test.ts
@@ -92,6 +92,36 @@ describe('insertWikiLinks', () => {
     expect(result).toBe('[[entities/alan-turing|Alan Turing]] was great.');
   });
 
+  describe('CJK entity linking', () => {
+    const ENTITY_ZH: WikiPage = { title: '机器学习', wikiLink: '[[concepts/machine-learning|机器学习]]' };
+    const ENTITY_JA: WikiPage = { title: 'ニューラルネットワーク', wikiLink: '[[concepts/neural-network|ニューラルネットワーク]]' };
+    const ENTITY_KO: WikiPage = { title: '기계학습', wikiLink: '[[concepts/machine-learning|기계학습]]' };
+
+    it('links a Chinese entity name in plain text', () => {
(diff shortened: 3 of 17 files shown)


green-dalii:
Clean PR. Opt-in default, pure-function core, full 8-locale i18n. Approved with notes.

**Q1 (real)**: default folder collision. If a user has a `pages/` folder with non-wiki content, this writes into it silently. Two options:
- Default to `wiki/source-copies/` (namespaced)
- Keep `pages/` + surface a "folder contains N non-markdown files, continue?" warning in settings UI

Lean toward namespaced default + settings-UI note.

**Q2**: `insertWikiLinks` uses `\b` boundaries, which don't fire on CJK. If you have a fix in this PR, point me to the file. If not, open a follow-up — the same issue affects #92.

**Nit**: Phase 5 refreshes all copies. For 1000+ files that's slow. Incremental (only refresh copies whose source changed) would be friendlier; not a v1 blocker.

Ship after we agree on Q1.

dmarchevsky:
> **Q1 (default folder collision)**
> Changed default from `pages` to `wiki/source-copies` — lives inside the wiki folder by default so it can't collide with existing non-wiki content. Existing users who already saved a custom path in `data.json` are unaffected. Updated the settings description in all 8 language files to document the namespaced default.

> **Q2 (CJK word boundaries)**
> Fixed in this PR. [`src/core/wikilink-inserter.ts`](src/core/wikilink-inserter.ts) now detects CJK characters in the entity text (Hiragana, Katakana, CJK Unified/Ext-A, Hangul, CJK Compatibility) and skips `\b` for those — CJK scripts have no inter-word spaces so any occurrence is a valid match. Four new tests cover Chinese, Japanese katakana, Korean Hangul, and the double-wrap guard. Same fix applies to the `#92` codepath since it uses the same function.

> **Nit (incremental Phase 5)**
> Noted, will track as a follow-up. The fix would compare source-file and copy-file `mtime` via `app.vault.adapter.stat` and skip files whose source hasn't changed — happy to open an issue if that's the preferred workflow.

green-dalii:
Thanks @dmarchevsky for this well-structured PR — the wikilink-inserter implementation (longest-first matching, CJK-aware) is clean and well-tested.

After evaluation, we've decided to defer this one. The main concern is conceptual layering: with `sources/` (raw input) → `wiki/` (structured output), adding `pages/` (linked copies) creates ambiguity for new users about what each folder does. The core mission (user writes → LLM organizes → user queries) already covers the source→wiki flow without an intermediate copy.

The `wikilink-inserter.ts` module itself is valuable — we'll keep it in mind for future use cases where wiki-link injection into existing notes makes sense independently of a source-copy feature. If that scenario comes up, your implementation will be the foundation.

Given this decision, PR #92 will need a clean rebase from `main` (it currently inherits this branch's code).

green-dalii:
Closing this PR as per the deferral decision above. The wikilink-inserter implementation is solid and will be kept as reference for future wiki-link injection use cases.

If the project's direction changes, this PR can always be reopened. Thanks again for the thoughtful contribution, @dmarchevsky.

Issue #85, opened 2026-06-03
Feature: User-configurable entity/concept tag vocabulary


DocTpoint:
---

> **Title:** `Feature: User-configurable entity/concept tag vocabulary`
>
> **Problem**
> `VALID_ENTITY_TAGS` and `VALID_CONCEPT_TAGS` are hardcoded at lines 44–45:
> ```javascript
> var VALID_ENTITY_TAGS = ["person", "organization", "project", "product", "event", "location", "other"];
> var VALID_CONCEPT_TAGS = ["theory", "method", "technology", "term", "other"];
> ```
> The LLM is instructed to tag pages using this vocabulary. For domain-specific vaults this produces meaningless tags — a medical vault gets `product` for a drug, `theory` for a pharmacological mechanism, `term` for a receptor.
>
> **Request**
> Expose `VALID_ENTITY_TAGS` and `VALID_CONCEPT_TAGS` as user-configurable settings, with the current generic list as default. Users with specialized vaults (medical, legal, scientific) could define their own controlled vocabulary.
>
> **Example — medical vault:**
> ```
> Entity tags: Arzneimittel, Biochemie, Neurologie, Kardiologie, Mikrobiologie, ...
> Concept tags: Pharmakologie, Immunologie, Genetik, Ernährung, ...
> ```
> This would eliminate the need for manual tag correction after every ingest.

---

green-dalii:
Real use case. A few notes:

1. **Unblocks #90 and #91** — both downstream issues describe LLM tag generation violating an unstated vocabulary. Solving at the source is cleaner than three separate prompt-tweaks.
2. **Backward compat**: per `CLAUDE.md` §2 Three-No Principle, new settings need defaults. Keep `["person", "organization", ...]` and `["theory", "method", ...]` as defaults verbatim.
3. **i18n**: 8 locales × 2 new labels = 16 strings. Follow the v1.15.0 `isWikiInitialized` pattern (settings.ts + texts.ts barrel).
4. **Validation**: normalize user-typed tags the same way `computeSlug` does (lowercase, slugify), or user-vs-LLM tag mismatches creep in.

P1 given the downstream unblock. Queue #90/#91 to ship with it.

DocTpoint:
Thanks for the detailed notes and the P1 prioritization.

Agreed on all points — especially computeSlug normalization for validation. Tag mismatches between user input and LLM output are exactly the kind of silent bug that's hard to diagnose after the fact.

[...]

PR #115, opened 2026-06-10
fix: preserve tags on re-ingest and multi-source merge (Issue #114)

### LINKED ISSUE #114: Bug: tags field lost on re-ingest / multi-source merge
## Description

Manually set `tags:` on entity, concept, and source pages are silently lost when a note is re-ingested or when a new note triggers a merge of an existing page.

## Observed behaviour

After fixing empty tags on ~35 pages post-batch-1, running a second ingest batch caused the same pages to revert to empty tags. Two distinct failure modes:

**1. Source pages — always regenerated**

`createSummaryPage()` in `wiki-engine.ts` does not check whether a source page already exists. It regenerates the full page from scratch on every ingest of the same note. Any manually corrected `tags:` field on the source page is overwritten. The tag fallback path (`extractSourceTags()` → falls back to `analysis.concepts.map(c => c.name)`) produces concept names, not valid domain tags.

**2. Entity/concept pages — tags dropped in mergeFrontmatter**

`mergeFrontmatter()` in `utils.ts` (line 558) only emits the `tags:` field when `Array.isArray(fm.tags) && fm.tags.length > 0`. If the existing page has `tags:` empty (common after the first ingest when the LLM omits tags), the field is entirely absent from the merged frontmatter output. After the merge, the page has no `tags:` field — subsequent merges see `fm.tags = undefined` and the field is permanently missing.

## Minimal reproduction

1. Ingest a source note → entity page created with `tags:` empty
2. Manually set `tags: [Biochemie, Physiologie]` on the entity page
3. Ingest another note that mentions the same entity → `mergePage()` is called
4. Result: `tags:` field is either empty or missing

## Expected behaviour

- Source pages: preserve existing frontmatter fields (at minimum `tags:`, `aliases:`) when regenerating on re-ingest
- `mergeFrontmatter()`: always emit `tags:` line, even when empty, so the field is consistently present and can be populated by the LLM body or manual edits

## Suggested fix

**For source pages:** Before regenerating, check if the source page exists and merge its `tags:` and `aliases:` into the new content (similar to how entity/concept merges work).

**For `mergeFrontmatter()`:**
```typescript
// Always emit tags line so the field is never silently dropped
lines.push(Array.isArray(fm.tags) && fm.tags.length > 0
  ? `tags:${yamlStringify(fm.tags)}`
  : 'tags:');
```

## Context

Observed while maintaining a German medical wiki with 400+ source notes across multiple ingest batches. After each batch, tags require manual correction on 30–40 pages. Reproducible across multiple ingest runs with Gemma 4 26B (local) as the LLM provider.

DocTpoint:
## Problem

Manually-set `tags:` on entity, concept, and source pages were silently lost on every subsequent ingest. After fixing ~35 empty-tag pages post-batch-1, running a second batch caused all corrections to revert. Two independent failure modes:

**1. Source pages always regenerated**

`createSummaryPage()` does not check whether the source page already exists — it regenerates the full page from scratch on every ingest of the same note. Any manually corrected `tags:` field is overwritten. The fallback path (`extractSourceTags()` → concept names when source has no tags) produces concept names, not valid domain tags.

**2. `mergeFrontmatter()` drops empty tags field permanently**

`mergeFrontmatter()` only emits `tags:` when `fm.tags` is a non-empty array. A page with `tags:` empty (common after first ingest) produces `fm.tags = undefined` via `parseFrontmatter()`. The condition `Array.isArray(undefined) && ...` is false, so `tags:` is entirely absent from the merged output. Every subsequent merge sees no `tags:` field in the existing content — the gap is self-perpetuating.

## Fix

**`src/wiki/wiki-engine.ts` — `createSummaryPage()`**

Read existing source page frontmatter before regeneration. If `tags:` is already populated (manually corrected), carry it forward as `tagsValue` injected into the LLM prompt:

```
Priority: existing source-page tags → source-note tags → LLM concept name fallback
```

**`src/utils.ts` — `mergeFrontmatter()`**

Always emit `tags:` — even when empty. An empty field is visible and correctable; a missing field propagates invisibly across every future merge.

```typescript
// Before: silently omitted when empty
if (Array.isArray(fm.tags) && fm.tags.length > 0) {
  lines.push(`tags:${yamlStringify(fm.tags)}`);
}

// After: always present
if (Array.isArray(fm.tags) && fm.tags.length > 0) {
  lines.push(`tags:${yamlStringify(fm.tags)}`);
} else {
  lines.push('tags:');
}
```

## Test plan

- [x] `pnpm lint` — 0 errors, 0 warnings
- [x] `pnpm test` — 587/587 passed
- [x] `pnpm build` — production build succeeds
- [ ] Manual: ingest note, correct tags, re-ingest same note → tags preserved on source page
- [ ] Manual: ingest note with empty tags, trigger merge via second note → `tags:` field present in merged output

## Context

Reproducible across multiple batch ingests on a German medical wiki with 400+ source notes. Without this fix, every ingest batch requires manual re-correction of 30–40 tag fields.

Closes #114

🤖 Implemented with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+- **Tags silently lost on re-ingest and multi-source merge (Issue #114).** Two distinct failure modes caused manually-set `tags:` to be overwritten on every subsequent ingest:
+  - `createSummaryPage()` regenerated source pages from scratch on every re-ingest without checking for existing content. Manually corrected tags were discarded. Fix: read existing source page frontmatter before generation; if `tags:` is already populated, carry it forward as the `tagsValue` injected into the LLM prompt (existing tags > source-note tags > LLM concept name fallback).
+  - `mergeFrontmatter()` only emitted `tags:` when the field was a non-empty array. A page with `tags:` empty (common after first ingest) produced `fm.tags = undefined`, causing the field to be silently omitted from the merged output — permanently invisible to future merges. Fix: always emit `tags:` so the field is never dropped, even when empty.
+- **`enforceFrontmatterConstraints()` hardcodes fallback tag when deduped array is empty (Issue #114).** When the deduplication pass produced an empty tags array, the function silently replaced it with `[DEFAULT_ENTITY_TAG]` / `[DEFAULT_CONCEPT_TAG]`, overwriting the user's intent (e.g. intentionally empty `tags: []` on a reviewed page). Fix: preserve the empty array and only fall back to the default tag when `fm.reviewed` is not set.
+
 ## [1.18.1] - 2026-06-11
 
 ### Fixed
--- a/src/__tests__/root/utils.test.ts
+++ b/src/__tests__/root/utils.test.ts
@@ -570,13 +570,13 @@ describe('enforceFrontmatterConstraints', () => {
     expect(result).not.toContain('tags: [term]');
   });
 
-  it('non-reviewed page with empty tags array still gets fallback (control test)', () => {
-    // For non-reviewed pages, the existing v1.18.0 behavior is
-    // preserved: explicit `tags: []` triggers the default fallback.
-    // The reviewed-guard is opt-in via `reviewed: true`.
+  it('non-reviewed page with empty tags array emits empty tags field, not fallback', () => {
+    // Fix #114: explicit `tags: []` emits `tags:` (empty) rather than
+    // silently overwriting with DEFAULT_ENTITY_TAG. User intent must win.
     const input = '---\ntype: entity\ntags: []\n---\n\nBody';
     const result = enforceFrontmatterConstraints(input, 'entity');
-    expect(result).toContain('tags: [other]');
+    expect(result).toMatch(/\ntags:\s*(\n|$)/);
+    expect(result).not.toContain('tags: [other]');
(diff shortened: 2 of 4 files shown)


green-dalii:
Solid diagnosis on #114 — the "self-perpetuating gap" framing in `mergeFrontmatter()` is exactly right. Always emit `tags:` is the right call; silent omission is the worst kind of data loss because it never shows up in Lint.

One related bug to consider while you're at it: `enforceFrontmatterConstraints()` has the same shape — when the deduped tags array is empty, it hardcodes `tags: [DEFAULT_*_TAG]` as fallback. Same class of "user intent silently overwritten". Worth fixing in the same PR to close #114's class of bug more completely. Up to you.

Heads-up on the rebase: the v1.18.0 work (Issue #85 v6/v7, default vocab optimization) is currently in my local main but not yet pushed. The remote main you're rebasing against is at v1.17.0 (587 tests). Expect conflicts in 8 `texts/*` files once my v1.18.0 commits land. I'll push them in the next few days — you can wait for that and rebase onto the new main, or rebase now against v1.17.0 and re-rebase later. Re-check 4-gate after rebase.

green-dalii:
v1.18.0 is now live — includes the reviewed-guard + disableThinking fix. Your PR resolves #114's class of bug. The remote main now has the full v1.18.0 codebase (9 commits). Rebase your PR onto main and re-check 4-gate — ready to merge after that.

Issue #128, opened 2026-06-13
Plugin sends no sampling params — server-default temperature=1 causes script-mixing corruption in extracted text


DocTpoint:
## Summary

The plugin never sends sampling parameters (`temperature`, `top_p`, `top_k`, `min_p`) in any LLM request. All request-body construction in `src/llm-client.ts` includes only `model`, `max_tokens`, `messages`, `system`, and `thinking` — there are **0** occurrences of `temperature`/`top_*`/`min_p` in the whole file (5 body blocks: `AnthropicCompatibleClient.createMessage` ~L83, `.createMessageStream` ~L199, plus the `AnthropicClient`/OpenAI-compat bodies ~L318/L519/L634).

As a result, sampling is governed **entirely by the server-side default** (LM Studio / the model's preset). For most local setups that default is `temperature: 1`.

## Why this is a problem

For an **extraction/grounding** task (source analysis, Mentions/verbatim quotes), `temperature: 1` is the opposite of what you want — you want fidelity to the source, not diversity. At temp 1 the full logit tail gets sampled, and with large multilingual-vocab models this produces **script-mixing corruption**: stray CJK/Devanagari characters injected mid-word into otherwise-correct text.

Concrete example with a quantized Gemma model (Gemma 3/4-class, int4 QAT) ingesting German notes — corrupted Mentions written verbatim into the wiki:

- `Luftversch뚫zung` → should be `Luftverschmutzung`
- `Herzkrankさ` → `Herzkrankheit`
- `Verd谷ung` → `Verdauung`
- `Schleim야heilung` → `Schleimhautheilung`
- `entündungsförder드는` → `entündungsförderndes`

This silently degrades the corpus (search, RAG grounding) and is invisible until you grep for non-Latin codepoints. Lowering the server temperature to ~0.15 eliminates it — but that's a manual, out-of-band step a non-technical user will never know to do, and it can't be task-specific (see below).

## Why it belongs in the plugin (not the server default)

1. **Correct-by-construction / zero manual setup.** The end user shouldn't have to configure their inference server to get faithful extraction. This is the same philosophy as #125 — don't rely on external/manual state.
2. **Per-task temperatures.** A single server-global can't be right for everything: extraction wants low temp (~0.1–0.2) for fidelity, while chat/query wants higher temp (~0.7) for fluent answers. The plugin already separates these code paths (`source-analyzer`/`page-factory` vs `query-engine`), so it can send the appropriate value per call.
[...]

Issue #99, opened 2026-06-05
Thinking-token bleeding in append/page-fix mode: extract by expected format instead of stripping known tokens


DocTpoint:
We recently switched from a model without thinking mode (Qwen 2.5 32B) to **Gemma 4 26B MLX** (via LM Studio), which supports a reasoning/thinking mode. After a re-ingest triggering an append-reviewed call, the resulting wiki page was corrupted with the model's reasoning preamble written directly into the file content:

```
�-mm-im-en-en-en-A-B-C-...

Wait, I must write ALL content in Deutsch. The user provided the 
prompt in English but instructions state "You MUST write ALL content 
in Deutsch".

Plan:
1. Basic Information: Update type/sources/definition...
2. Description: Merge existing...
```

The actual page content followed only after ~80 lines of reasoning text.

**Current approach and its limitations**

The plugin currently uses `stripThinkingTokens()` to remove thinking-token wrappers from LLM responses. This approach is fundamentally model-specific: it requires knowing the exact token format (`<think>`, `<thinking>`, custom markers) for each model. As local model diversity grows, this list can never be complete.

In our case, Gemma 4's thinking tokens were partially stripped (the garbled characters on line 1 suggest the opening token was removed), but the reasoning content itself remained — because it was plain text, not wrapped in a recognizable token.

**Proposed concept**

Instead of *removing what doesn't belong*, the plugin could *extract only what is expected* — a model-agnostic approach:

- For **Markdown wiki pages**: find the first `---` (YAML frontmatter) or first `##` header; treat everything before it as preamble and discard it
- For **JSON responses**: this already works via `parseJsonResponse` / `extractBalancedJson` which scans for the first `{` — the same principle

This would handle thinking-token bleeding, reasoning preambles, apology text, and any other model-specific artifacts without needing to know anything about the model.

**Impact**

Affects any model that produces reasoning/planning text before the actual output — increasingly common with instruction-tuned local models. The append-reviewed and page-fix modes are most susceptible as they involve complex integration prompts that tend to elicit chain-of-thought.

**Workaround**

Disabling "Enable Thinking" in LM Studio before append-reviewed operations prevents the issue for now.

green-dalii:
Resolved with a three-layer defense in v1.16.2:

[...]

Issue #131, opened 2026-06-14
Ingest Stage 4 regenerates whole related pages via LLM even with no new info — ~34% of page-level calls, ~⅓ of them no-ops


DocTpoint:
## Summary

Stage 4 of ingest (related-page updates) regenerates the **entire body** of each related existing page via a full LLM call — even when there is no new information to add. Measured over 15 real ingests, Stage 4 accounts for **34% of all page-level LLM calls**, and roughly **a third of those have nothing to add** (the `new_info` slot resolves to the literal fallback string). This is the dominant *avoidable* cost in ingest, and it is also a content-corruption vector because it rewrites existing verbatim text on every pass.

## Where it happens

`PageFactory.updateRelatedPage()` builds the prompt and calls the LLM unconditionally:

```ts
.replace('{{new_info}}', JSON.stringify(
  analysis.entities.find(e => e.name === pageName)
  || analysis.concepts.find(c => c.name === pageName)
  || 'No directly relevant information'))
...
const updatedBody = await client.createMessage({ max_tokens: TOKENS_PAGE_GENERATION, ... });
```

`related_pages` are by definition **existing** pages that are topically related — a different set from the newly extracted `entities`/`concepts`. So the `.find()` frequently misses and `new_info` becomes `'No directly relevant information'`. The prompt (`merge.ts` `updateRelatedPage`) then still asks the model to *"add the new information"* and rewrite the whole body — i.e. a full page-sized generation that adds nothing.

## Measured impact (15 ingests)

| Metric | Value |
|--------|-------|
| Total page-level LLM calls | 399 (263 generation + 136 related-update) |
| Stage 4 share | **34%** |
| Stage 4 calls with no new info (fallback) | **~33%** of Stage 4 (~40 calls) |

Three distinct problems:
1. **No-op regenerations** — ~40 full-body LLM calls that add nothing.
2. **Double processing** — some pages are generated in Stage 3 and then regenerated again in Stage 4.
3. **Corruption vector** — every full rewrite of an existing page re-rolls its verbatim `## Mentions in Source` quotes through the model. With a quantized local model this is exactly how stray script-mixing characters get introduced into previously-clean pages (related to #128).

## Proposed fix (two tiers)

**Tier 1 — early return on no new info (trivial, zero risk).**
[...]

PR #132, opened 2026-06-14
Local-only control params: chat_template_kwargs (#99) + repetition_penalty

DocTpoint:
Completes #99 for local template-based reasoning models and adds a local-only repetition guard. **Both params are gated to local llama.cpp-style providers (LM Studio / Ollama); cloud providers never receive them**, so there is no cloud behavior change here.

### chat_template_kwargs (#99 completion)
The v1.18.0 fix wired `disableThinking` to all call sites, but it sends Anthropic-style `thinking: { type: 'disabled' }` — which LM Studio's OpenAI-compatible endpoint ignores/rejects for template-based models (Gemma 4, Qwen). The model then reasons until it exhausts the token budget and returns empty content (`finish_reason: length`), failing the ingest. The working mechanism on those runtimes is the chat-template kwarg `{ enable_thinking: false }`. This injects it for local providers only.

Also includes the `AnthropicCompatibleClient`/`AnthropicClient` try-catch fallback (on 400 + thinking-control error, cache `thinkingControlSupported=false` and retry without the thinking field), refactored into a shared `anthropicDoRequest` helper.

### repetition_penalty (local-only, default 1.1)
A llama.cpp extension honored by LM Studio / Ollama; heads off repetition loops some local models fall into. Auto-applied only for `LOCAL_LIKE_PROVIDERS`; cloud never sees it.

### Split note
Per your review, **per-task temperature (#128) is moved to its own PR** — its defaults change behavior for cloud providers too and deserve separate evaluation. This PR is now cloud-safe and ready for the v1.18.3 P0 window.

690 tests green, tsc + eslint clean.


green-dalii:
Thanks for tackling both #128 and #99. The chat_template_kwargs piece is a clean completion of #99 and fits the v1.18.3 P0 window. The temperature defaults, however, change behavior for cloud providers too — I'd prefer to split that into its own PR so we can evaluate default-impact separately. Happy to merge the #99 part quickly.

DocTpoint:
Updated per your review: **temperature is split out** into a separate PR (incoming), so this PR now contains only the cloud-safe, local-gated params — `chat_template_kwargs` (#99 completion) + `repetition_penalty`. Neither touches cloud providers. Should be safe for the v1.18.3 P0 window.

green-dalii:
DocTpoint, thanks for the quick split — that made the local vs. cloud question much clearer.

After more first-principles discussion, we want to take it one step further: rather than auto-injecting any parameter (even local-only), we want **zero default behavior change** and put all knobs in a user-controlled Advanced section. Provider runtimes vary too much for us to guess correctly, and automatic fallbacks create hidden state that's hard to debug.

Concrete direction:
- Drop the  classification entirely.
- , , and / all live in an Advanced panel.
- All params are only sent when the user explicitly sets them; defaults stay empty so existing behavior is preserved.
- For #99, instead of auto-injecting , we'll detect the empty-content-after-reasoning pattern and surface a clear error pointing the user to the relevant toggle.

We're going to implement this direction directly in the v1.18.3 branch since it touches the same surface and keeps the architecture consistent. Your analysis in #131/#132 was the direct input for this design — we'll credit you in the commit/release notes and close these PRs once the replacement lands. Really appreciate the clean split and the thoughtful write-ups.

green-dalii:
Closing this PR — the functionality (disableThinking 400 fallback + repetition_penalty handling) has been shipped in v1.19.0 as part of the unified Advanced LLM parameter settings panel. Users configure these via Settings → LLM Configuration → Advanced parameter settings → Custom mode. Thank you for the detailed analysis and clean PR split — the architecture decision to use user-controlled opt-in rather than auto-injection was directly informed by your approach.

See https://github.com/green-dalii/obsidian-llm-wiki/releases/tag/1.19.0 for release notes.

PR #135, opened 2026-06-15
Send per-task sampling temperature (#128)

DocTpoint:
Split out from #132 per review — temperature is sent to **all** providers (including cloud), so its defaults warrant separate evaluation from the local-only params in #132.

## Problem
The plugin currently sends no `temperature`, so the backend uses its server default (1.0 on LM Studio). At temperature 1.0 the logit tail gets sampled, and on quantized local models (observed: Gemma 4 26B A4B QAT) this injects stray CJK/Devanagari characters mid-word into otherwise-clean extracted German text — corrupting verbatim Mentions.

## Change
Two settings, forwarded per-request to every client:
- **`extractionTemperature`** (default **0.15**) — source analysis, entity/concept generation, verbatim Mentions; low to maximize fidelity.
- **`chatTemperature`** (default **0.7**) — query answers.

The `createMessage` wrapper injects `extractionTemperature` when the caller passes none; query-engine call sites pass `chatTemperature` explicitly. `temperature` is omitted from the body when unset (no behavior change for callers that don't opt in).

## Note on defaults / cloud impact
Because temperature now goes to cloud providers too, the defaults (0.15 / 0.7) change their behavior vs. today's implicit 1.0. Happy to adjust the defaults or gate per-provider — flagging it explicitly since that was the reason to split this from #132.

687 tests green (incl. 2 new temperature-forwarding tests), tsc + eslint clean.

### DIFF
--- a/src/__tests__/__support__/engine-context.ts
+++ b/src/__tests__/__support__/engine-context.ts
@@ -101,6 +101,8 @@ export const DEFAULT_SETTINGS: LLMWikiSettings = {
   customEntityTags: '',
   customConceptTags: '',
   slugCase: 'lower' as const,
+  extractionTemperature: 0.15,
+  chatTemperature: 0.7,
 };
 
 // ── Mock EngineContext ───────────────────────────────────────────
--- a/src/__tests__/root/llm-client.test.ts
+++ b/src/__tests__/root/llm-client.test.ts
@@ -147,6 +147,31 @@ describe('OpenAICompatibleClient.createMessage', () => {
     expect(mockRequestUrl).toHaveBeenCalledTimes(1);
   });
 
+  it('forwards temperature to the request body when set (#128)', async () => {
+    mockRequestUrl.mockResolvedValueOnce(makeOpenAIResponse('Hello world'));
+    const client = new OpenAICompatibleClient('key', 'https://api.openai.com');
+    await client.createMessage({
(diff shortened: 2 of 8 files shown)


green-dalii:
Thanks for splitting temperature out — having it as a standalone PR made the default-behavior question much easier to see.

We ended up deciding that **no provider should get temperature by default**, local or cloud. The right model is an opt-in Advanced panel where users explicitly set  and , with empty defaults that preserve today's behavior.

So rather than merging this PR as-is, we'll fold the temperature fields into the same Advanced-section refactor that handles #132's repetition penalty and the #99 thinking controls. Temperature will only be sent when the user has filled in a value.

Your measurements on the Gemma script-mixing corruption are the reason this setting exists at all — we'll make sure that's highlighted in the changelog and close this PR once the unified Advanced panel lands.

green-dalii:
Closing this PR — the temperature feature has been shipped in v1.19.0 as part of the unified Advanced LLM parameter settings panel (Settings → LLM Configuration → Advanced parameter settings → Custom mode). Your measurements on Gemma 4 script-mixing corruption are documented in the changelog and release notes. Thank you for the clean split and the performance measurements that justified this feature.

See https://github.com/green-dalii/obsidian-llm-wiki/releases/tag/1.19.0 for release notes.

PR #156, opened 2026-06-19
fix(wiki): fingerprint every source-page slug to prevent collisions (#155)

### LINKED ISSUE #155: Source provenance pages collide on duplicate filenames — sources/ slug derived from basename only, second ingest silently overwrites the first
## Summary

Source provenance pages (`sources/<slug>.md`) are named by `slugify(file.basename)` with **no path awareness and no collision check**. When two source notes in different folders share a filename (e.g. `About this course.md`, `README.md`, `index.md`, `notes.md`), the second ingest **silently overwrites** the first's source page. The first's entity/concept backlinks and any course `Curriculum` link — all `[[sources/<slug>]]` — then resolve to the *second* source's content. Pure data loss + cross-source contamination, with no warning.

## Impact / severity

- **Silent data loss** — the earlier source page is replaced wholesale (only its `tags:` are preserved, per #114).
- **Cross-link contamination** — every page that linked `[[sources/<slug>]]` now points at an unrelated source's content.
- **Blocks bulk ingest** of any multi-folder corpus where basenames repeat (course trees, per-project READMEs, dated notes, repeated `About`/`Overview`/`Glossary` pages, …).

## Root cause

The source slug is derived from the **basename only**, in two independent places that happen to agree (so links resolve today) but both drop the folder:

**1. Producer — `WikiEngine.createSummaryPage` (`src/wiki/wiki-engine.ts`, ~L655–658):**
```ts
const slug = slugify(file.basename, preserveCase);
const path = normalizePath(`${wikiFolder}/sources/${slug}.md`);
```
The pre-existing file at `path` is read **only to preserve tags** (#114); there is **no check** that it belongs to a *different* `source_file`, and it is then overwritten unconditionally via `createOrUpdateFile` (~L711).

**2. Consumer — `normalizeSourcePath` (`src/core/sources-normalizer.ts`, ~L58–64):**
```ts
const filename = s.split('/').pop() || s;   // drops the folder
s = `sources/${computeSlug(filename, preserveCase)}`;
```
This sets entity/concept `sources:` backlinks to the same basename-derived slug. It is a **pure, no-IO function by design**, so it cannot detect collisions — the fix cannot live here.

## Reproduction

1. Two source notes with the same filename in different folders:
   - `A/About this course.md`
   - `B/About this course.md`
2. Ingest `A/…` → creates `sources/about-this-course.md` (+ entities linking `[[sources/about-this-course]]`).
3. Ingest `B/…` → **overwrites** `sources/about-this-course.md` with B's content.
4. A's entities / curriculum now resolve to B's source page. A's provenance is gone, no warning emitted.

## Observed in the wild (abstracted)

A real corpus of **468 notes had only 341 distinct basenames → 127 colliding basenames**: `About this course` ×11, `Learning resources` ×8, `Technical prerequisites` ×5, `Check your knowledge` ×5, plus many ×2–3. Ingesting a second `About this course` fully replaced the first; the first node's `Curriculum → [[sources/about-this-course]]` silently retargeted to the second course's content.

Note: **entity/concept and course slugs are unique** (derived from page titles), so only the **`sources/` provenance pages** collide. That scopes the fix.

## Proposed fix

Decide a **single canonical, collision-aware source slug per ingest**, computed once before page generation:

1. `resolveSourceSlug(file, app, wikiFolder, preserveCase)`:
   - base = `slugify(basename)`;
   - if `sources/<base>.md` exists with a **different** `source_file` → disambiguate (prepend parent-folder slug, escalating up the path until unique) **or** append a short stable hash of the full path (order-independent);
   - if it exists with the **same** `source_file` → reuse that slug (re-ingest updates in place, never duplicates).
2. Compute it at the start of `ingestSource` and thread it into:
   - **page generation** (`page-factory`) — inject the canonical `[[sources/<slug>]]` instead of the raw path, so the LLM emits the correct backlink and `normalizeSourcePath` passes it through untouched;
   - **`createSummaryPage`** — use the passed slug, and on a same-slug/different-`source_file` hit, disambiguate rather than overwrite.

Backward-compatible: disambiguation only triggers on an actual collision, so existing single-folder vaults keep their current slugs and backlinks.

## Workaround

Rename colliding source files to unique basenames before ingest.

## Affected code

- `src/wiki/wiki-engine.ts` — `createSummaryPage` (~L655–658, ~L711), `ingestSource`
- `src/core/sources-normalizer.ts` — `normalizeSourcePath` (~L58–64)
- `src/wiki/page-factory.ts` — source reference passed into the generation prompt

Indexed-Apogrypha:
## Summary

Source provenance pages (`sources/<slug>.md`) were named `slugify(basename)` with no path awareness, so two source files sharing a filename in different folders (e.g. 11× `About this course.md` across Academy courses) collapsed to the same slug — the second ingest **silently overwrote** the first, and every `[[sources/<slug>]]` backlink then resolved to the wrong source. Fixes #155.

## Fix

Every source slug is now `<basename>_<6hex>`, where the fingerprint is an FNV-1a hash of the file's full path:

- **Unique** per file — no collision probing needed.
- **Order-independent** — derived from the path alone.
- **Re-ingest stable** — same file → same slug → updates in place, never duplicates.
- **Length-bounded** (≤ 80; basename trimmed, fingerprint never dropped).

Entity/concept/curriculum backlinks are threaded the canonical `[[sources/<slug>]]` (new-page prompt injection + `mergeFrontmatter` for merges); the write-path `normalizeSourcePath` passes the canonical link through unchanged. New pure module `src/core/source-slug.ts` with 9 unit tests.

## Backward compatibility

Because *every* source slug now carries a suffix, re-ingesting an existing vault renames its `sources/` pages (backlinks update on re-ingest). New vaults are unaffected. No settings / schema / API changes.

## Also (housekeeping)

Second commit bumps tsconfig `lib` → ES2021 (so ES2019 string APIs like `trimEnd` resolve under newer TS language servers) and drops the vestigial `baseUrl` (no `paths` map; clears a TS 6/7 deprecation).

## Out of scope (follow-up)

`conversation-ingest.ts` creates source pages via `slugify(source_title)` — the same collision class, different identity model (title, not file path). Flagged for a follow-up.

## Test plan

- [x] `pnpm lint` — 0 errors / 0 warnings
- [x] `pnpm test` — 788 pass (9 new for `source-slug`)
- [x] `pnpm build` — clean
- [x] Live end-to-end (incl. a small local model): two same-named "About this course" files → two distinct `about-this-course_<hex>.md` pages, no overwrite, backlinks resolve to the canonical slug.

## Gate 4: Performance

| Dim | Status | Notes |
|-----|--------|-------|
| CPU | OK | O(path len) fingerprint + a few string ops per ingest |
| Memory | OK | No new caches / arrays / listeners |
| IO | OK | No extra reads — fingerprint needs no collision probe |
| Network | OK | Zero new LLM calls; prompt size unchanged |
| Token | OK | N/A — substitutes a link of similar length |

### DIFF
--- a/src/__tests__/source-slug.test.ts
+++ b/src/__tests__/source-slug.test.ts
@@ -1,0 +1,56 @@
+import { describe, it, expect } from 'vitest';
+import { resolveSourceSlug, sourceFingerprint, sourceBaseSlug } from '../core/source-slug';
+
+describe('sourceFingerprint', () => {
+  it('is deterministic for the same path', () => {
+    expect(sourceFingerprint('raw/a/About this course.md'))
+      .toBe(sourceFingerprint('raw/a/About this course.md'));
+  });
+
+  it('is exactly 6 lowercase hex chars', () => {
+    expect(sourceFingerprint('raw/a/x.md')).toMatch(/^[0-9a-f]{6}$/);
+  });
+
+  it('differs for different paths with the same basename', () => {
+    expect(sourceFingerprint('raw/CourseA/About this course.md'))
+      .not.toBe(sourceFingerprint('raw/CourseB/About this course.md'));
+  });
+});
+
+describe('sourceBaseSlug', () => {
+  it('is the plain slugified basename, no path', () => {
+    expect(sourceBaseSlug('raw/Course X/About this course.md')).toBe('about-this-course');
+  });
+});
+
+describe('resolveSourceSlug', () => {
+  it('always appends a 6-hex fingerprint', () => {
+    expect(resolveSourceSlug('raw/A/Storage buckets.md')).toMatch(/^storage-buckets_[0-9a-f]{6}$/);
+  });
+
+  it('does NOT include a folder hint', () => {
+    const slug = resolveSourceSlug('raw/Course X/About this course.md');
+    expect(slug).toMatch(/^about-this-course_[0-9a-f]{6}$/);
+    expect(slug).not.toContain('course-x');
+  });
+
+  it('is stable for the same file (re-ingest determinism)', () => {
+    expect(resolveSourceSlug('raw/Course X/About this course.md'))
+      .toBe(resolveSourceSlug('raw/Course X/About this course.md'));
+  });
+
+  it('keeps two same-basename files in different folders unique', () => {
+    const a = resolveSourceSlug('raw/Agentic Prompt Engineering/About this course.md');
+    const b = resolveSourceSlug('raw/Working with Orchestrator resources/About this course.md');
+    expect(a).not.toBe(b);
+    expect(a).toMatch(/^about-this-course_[0-9a-f]{6}$/);
+    expect(b).toMatch(/^about-this-course_[0-9a-f]{6}$/);
+  });
+
+  it('never exceeds the length cap, keeping the fingerprint', () => {
+    const longName = 'A'.repeat(200);
+    const slug = resolveSourceSlug(`raw/Folder/${longName}.md`, { maxLen: 80 });
+    expect(slug.length).toBeLessThanOrEqual(80);
+    expect(slug).toMatch(/_[0-9a-f]{6}$/);
+  });
+});
--- a/src/core/source-slug.ts
+++ b/src/core/source-slug.ts
@@ -1,0 +1,76 @@
+// Collision-safe, length-bounded source-page slugs (Issue #155).
+//
+// Source provenance pages live at `sources/<slug>.md`. Historically the slug was
+// just `slugify(basename)`, so two source files with the same filename in
+// different folders (e.g. 11× "About this course.md" across courses) collapsed to
+// the same slug and silently overwrote each other.
+//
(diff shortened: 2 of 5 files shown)


green-dalii:
Approved. Clean diff — fingerprint is correctly computed once and threaded through all downstream calls (createSummaryPage, entity, concept, related-page updates). The "every slug gets a suffix" choice is the right call; collision probing on read would have been an O(n) cost we don't need.

One small note for a future PR (not blocking): `conversation-ingest` has the same collision class via `slugify(source_title)` per the "Out of scope" section. Worth tracking.

Will merge after #154 lands so we can ship both as v1.20.3.

green-dalii:
Thanks again — PR #154 and #158 just shipped together in v1.20.3 along with your source-slug fingerprint fix from PR #156. Three parity/latent-bug fixes in one hotfix, full changelog here: https://github.com/green-dalii/obsidian-llm-wiki/releases/tag/1.20.3

PR #158, opened 2026-06-19
fix(wiki): respect reviewed:true in Stage-4 updateRelatedPage (parity with createOrUpdatePage)

DocTpoint:
## Problem

`updateRelatedPage` (Stage 4 — where one ingested source updates a *different*, already-existing related page) rewrites the page body via the LLM **regardless of `reviewed: true`**. This is inconsistent with `createOrUpdatePage`, which routes reviewed pages to the minimal append path (`appendToReviewedPage`).

As a result, re-ingesting an unrelated note can silently overwrite a curated/locked page's body — the `reviewed` lock does not hold on the Stage-4 path.

## Fix

Add the same reviewed guard already present in `createOrUpdatePage`: when the related page is `reviewed: true`, route to `appendToReviewedPage` so the curated body is preserved exactly and only genuinely new content is appended.

```ts
if (parseFrontmatter(existingContent)?.reviewed === true) {
  await this.appendToReviewedPage(newInfo, sourceFile, existingContent, page.path);
  return true;
}
```

## Test

Adds a regression test: a `reviewed: true` page whose body is extracted by a *different* source. Without the guard the body is replaced by the LLM output; with it the curated body survives. The test fails on the unpatched code and passes with the fix. Full suite: **780/780 green**.

### DIFF
--- a/src/__tests__/wiki/page-factory.test.ts
+++ b/src/__tests__/wiki/page-factory.test.ts
@@ -167,3 +167,31 @@ describe('PageFactory — updateRelatedPage no-op skip (Issue #131)', () => {
     expect(vault.read('wiki/entities/Butyrat.md')!).toContain('New merged body.');
   });
 });
+
+describe('PageFactory — updateRelatedPage reviewed guard (Stage 4)', () => {
+  it('preserves a reviewed page body instead of LLM-rewriting it when another source extracts it', async () => {
+    const { ctx, vault } = createMockContext({
+      vaultFiles: {
+        'wiki/entities/Butyrat.md':
+          '---\ntype: entity\nreviewed: true\n---\n\n## Description\nCurated body, must stay verbatim.',
+      },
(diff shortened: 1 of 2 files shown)


green-dalii:
Approved. Same class of latent bug as #154/#156 — a guard that existed on one path but not its sibling. The test using `NO_NEW_CONTENT` as the LLM response is the right way to prove "curated body survives": the buggy path would write that string verbatim, the fixed path routes to appendToReviewedPage and the old body is untouched.

Will merge into v1.20.3 hotfix alongside #154 and #156 — three parity/latent-bug fixes in a single release keeps the change log coherent.

green-dalii:
Both PR #154 and #158 just shipped in v1.20.3 together with the source-slug fingerprint fix from PR #156. Three parity/latent-bug fixes in one hotfix — full changelog: https://github.com/green-dalii/obsidian-llm-wiki/releases/tag/1.20.3

PR #159, opened 2026-06-19
feat(i18n): add Italian (it) locale for UI and wiki output

FrancoTampieri:
Add Italian as the 9th supported language across the plugin.

- New src/texts/it.ts with the full 411-string UI translation, at exact key parity with en (no missing/extra keys).
- Wire 'it' into the UI language union (types.ts, llm-client.ts) and the settings dropdown.
- Register it in the wiki output path: WIKI_LANGUAGES.it + an 'it' block in SECTION_LABELS so generated pages get Italian section headings.
- Add languageIt: 'Italiano' to all existing locales.
- New docs/README_IT.md and an Italiano entry in every README language bar.
- New i18n-parity test: every locale must cover all en keys (guards the silent English fallback), and it is held to strict bidirectional parity.

All additive and backward-compatible; no fields removed, defaults unchanged. Gate 1 green: lint 0/0, tsc 0, 801 tests pass, build clean.


green-dalii:
Welcome — and thanks for the very clean first PR.👏

A few things stand out:

1. The `i18n-parity.test.ts` you added is worth more than the Italian strings themselves. `getText()` silently falls back to English on missing keys, so the previous 8 locales could drift and the build would still pass — your test makes that a CI failure. That's the kind of contribution that keeps paying for itself with every future locale.

2. Holding it to strict bidirectional parity (no orphan keys, no missing keys) is the right call for a *new* locale; the older locales carry historical orphan keys and we shouldn't block this PR on cleaning them up.

Approved. Will merge into v1.21.0 (not v1.20.3 hotfix — feature work belongs in a normal release, not a bug-fix hotfix).

Thanks again for setting the quality bar on the first PR.

green-dalii:
Shipped in main (commit 3936502). Will ride out with the next normal release (v1.21.0 — Schema Coherence Phase 1) along with the i18n-parity test guarding every future locale. Thanks again for the very clean first PR.

FrancoTampieri:
Thank you so much, and I really appreciate that you liked the test I posted 🙏

green-dalii:
Shipped in v1.21.0: https://github.com/green-dalii/obsidian-llm-wiki/releases/tag/1.21.0

The Italian locale is now the 9th fully-translated language in plugin UI and wiki output. Bidirectional parity test (`__tests__/root/i18n-parity.test.ts`) ensures every key in the EN source has a matching IT entry — so any future Italian-speaking contribution lands in the same shape as the existing 8 locales.

Special thanks for bringing Italian to the community — the localization effort is what makes this plugin usable beyond English-speaking users.

Issue #169, opened 2026-06-21
Better status reporting, watch files as they are generated


Indexed-Apogrypha:
**Is your feature request related to a problem? Please describe.**
yes. I'm never quite sure how long the ingestion task will take because the status indicator isn't intuitive or informative enough. 

**Describe the solution you'd like**
expand status indicator to show additional information: current file, current batch, current model, current granularity, expected time remaining, number of files remaining, number of tasks remaining, etc.

I'd also like to see the content of the wiki files as they are being generated. This would allow me to monitor how fast the files are generating, what changes are happening to the wiki as they happen, what wiki files will contain, etc.

**Describe alternatives you've considered**
Changing the status indicator to a live artifact that logs plugin tasks. Changing the status indicator to a popup. play notification sound when processing is complete. emit log files on each task. 

**Additional context**
i'll add additional context to comments as I work on this

<img width="322" height="82" alt="Image" src="https://github.com/user-attachments/assets/0d244f16-c302-445f-bd31-aae2d56b5dc6" />

<img width="556" height="113" alt="Image" src="https://github.com/user-attachments/assets/70b9c745-47e8-459f-b28e-e36df57bf7a2" />

green-dalii:
Splitting this into two scopes per your suggestion:

1. **Richer progress text** (v1.22.0 candidate): batch count, ETA, current model/granularity. The Operation History panel (#122) helps with post-hoc visibility in the meantime.
2. **Live file preview** (backlog): needs a new UI component design; deferring until we have a solid UX plan.

The immediate gap is the "rough progress" issue — agreed that's worth tackling sooner. Will prioritize the progress-text enhancement for v1.22.0.

green-dalii:
## Tracking: PPR for query retrieval

This issue (status reporting) shipped its scope 1 (document name + batch progress) as **PR #189** in v1.22.0 — the status bar now shows `<doc> · Ingesting... click to cancel` and `[4/10] My Note · ...` for batch. ETA + model/granularity in the status bar remains unimplemented (you said "more important than file preview" was the rough-progress gap, so we prioritized the doc-name + batch-progress visible feedback; ETA is queued but not scheduled).

For scope 2 (live file preview) — that needs a new UI component. Currently deferred; will revisit if / when we have a v1.23.0 UX pass.

[...]

PR #189, opened 2026-06-22
feat: show document name and batch progress in ingest status bar

YounianC:
Single-file ingest now shows "<doc> · Ingesting... click to cancel" instead of the bare label, and folder batch ingest shows "[current/total] <doc> · Ingesting... click to cancel".

demo:
image

- New pure-function core/status-bar.ts (buildIngestStatusBarText) composes the text from the existing localized ingestionStatusBar label — no new i18n keys, all 9 locales covered automatically.
- WikiEngine ingestion-start callback now passes the source basename (optional param, backward-compatible).
- main.ts tracks loop position via a batchProgress field, set per iteration in the folder loop and reset after.
- 7 new core/status-bar tests; 948 tests passing (was 941).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,15 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Changed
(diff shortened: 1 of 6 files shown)


green-dalii:
Excellent PR — clean pure function, 7 tests, 0 new i18n keys, all 9 locales covered automatically via the existing `ingestionStatusBar` label. Exactly matches the #169 scope 1 commitment ("Richer progress text: batch count, ETA, current model/granularity"). Will merge into v1.22.0 P1. Thanks @YounianC!

Issue #187, opened 2026-06-22
Related-link sections default to a sources/ prefix when the target isn't in the truncated existing-pages list


DocTpoint:
## Summary

Newly-generated entity/concept pages frequently emit their **Related Concepts / Related Entities** links with a `sources/` prefix (`[[sources/Hippocampus|…]]`) instead of `[[entities/…]]` / `[[concepts/…]]`. On a mature vault this is pervasive: in my vault **225 pages / ~450 related-links** are mis-prefixed, **76** of them dead (no such `sources/<slug>.md`); the rest resolve to the *source summary* page instead of the concept/entity, polluting the graph with `concept→source` edges.

## Root cause

The Related sections are LLM-formatted. The generation prompts (`generation.ts`) pass the related items as **bare names** and instruct the model to "use the exact full paths from the **Existing Wiki Pages** list". That list comes from `buildPagesListForPrompt()` and is capped:

- `src/wiki/page-factory.ts:206` — `const MAX_PAGES = 50;`
- `src/wiki/page-factory.ts:215` — when a concept page is being generated, the list is filtered to **concepts only**, so related **entities are never in it**.

On any vault with >50 pages the target name is usually not in the visible list, so the model has no path to copy and **guesses the prefix** — defaulting to the single most prominent path token in the prompt, `sources/` (it appears in the `sources:` frontmatter, the `## Source` section, and every "Mentions in Source" citation).

**Testable prediction that holds exactly:** on concept pages the related *entities* are almost always mis-prefixed (their folder is filtered out of the list entirely), while older pages from when the vault had <50 pages are correct. Both match the observed data.

## Impact

Scales with vault size — silently worsens as the wiki grows past the 50-page window. Produces dead links and a graph dominated by spurious links into `sources/`.

## Suggested fix (deterministic, no extra LLM calls)

The code already **knows the type** of every related item — `related_entities` are entities, `related_concepts` are concepts. So don't let the model guess: after generation, re-assert the folder for links inside the Related sections, scoped by section header so the legitimate `[[sources/<sourceSlug>]]` citations in "Mentions in Source" are never touched (their names often coincide with a related concept, e.g. a source `Gedächtnis` that is also a related concept).

A pure `correctRelatedLinkPrefixes(content, relatedEntities, relatedConcepts, …)` run after `createNewPage` and `mergePage` handles both freshly-generated and merge-carried-over links. This is the same "deterministic over guessing" direction as #185.

[...]

PR #196, opened 2026-06-23
feat: convert Query Wiki from Modal to Copilot-style right-docked side panel

YounianC:
## Summary
Converts Query Wiki from a centered popup (Modal) into a Copilot-style right-docked chat panel (Obsidian `ItemView`), so the conversation can stay open alongside notes instead of interrupting with a popup every time. **All functionality and interactions are strictly preserved** — this change only alters the host surface and visual styling.
- `QueryModal extends Modal` → `QueryView extends ItemView` (new `VIEW_TYPE_QUERY`, registered via `registerView`). The `query-wiki` command and a new `message-circle` ribbon icon activate/reveal a right sidebar leaf, reusing an existing leaf if one is already open.
- Styling moves from the fixed-size `.llm-wiki-query-modal` (800×780) to a flex `.llm-wiki-query-view` layout; all hardcoded colors (`#4caf50` / `#fff3e0` / `#2196f3` / `#f44336` / `#999` / `white`) are replaced with native `var(--…)` theme variables for automatic light/dark adaptation.
## Behavior Preserved
Three-tier retrieval, streaming generation + non-streaming fallback, collapsible thinking panel, save-to-wiki feedback loop, LLM save suggestion, history-round cap, clear, copy, and stop. The `renderThinkingBlocksUI` pure function and the post-query `SuggestSaveModal` are untouched.

## demo
image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,3 +22,4 @@ Thumbs.db
 docs/superpowers/
 
 .mimocode/
+data.json
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+- **Query Wiki moved from a centered Modal to a Copilot-style right-docked side panel.** `QueryModal extends Modal` became `QueryView extends ItemView` (registered via `registerView(VIEW_TYPE_QUERY, …)`). The command `query-wiki` and a new `message-circle` ribbon icon now activate/reveal a right sidebar leaf (reusing the existing leaf if already open) instead of opening a popup. All existing behavior is preserved unchanged: three-tier retrieval, streaming + non-streaming fallback, collapsible thinking panel, save-to-wiki feedback loop, LLM save suggestion, history cap, clear, copy, and stop. The `renderThinkingBlocksUI` pure function and the post-query `SuggestSaveModal` are untouched.
+- **Query panel styling now uses native theme variables.** All hardcoded colors (`#4caf50`, `#fff3e0`, `#2196f3`, `#f44336`, `#999`, `white`) were replaced with `var(--…)` tokens so the panel adapts to light/dark themes automatically. The fixed-size `.llm-wiki-query-modal` (800×780) rule was removed in favor of a flex `.llm-wiki-query-view` root.
+
+### Tests
+- **1003 tests passing (unchanged).** obsidian test mock extended with `ItemView` / `WorkspaceLeaf` stubs so `query-engine.ts` still imports cleanly under vitest; existing `renderThinkingBlocksUI` suite stays green.
+
 ## [1.22.0] - 2026-06-23
 
 ### Added
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -6,6 +6,9 @@
 
 ## Current Phase: v1.22.0 Released
 
+### In Progress (Unreleased) — Query Wiki Right-Docked Side Panel
+- 🔄 **Query Wiki: Modal → Copilot-style right side panel.** `QueryModal extends Modal` → `QueryView extends ItemView` (`VIEW_TYPE_QUERY`, registered via `registerView`). `query-wiki` command + new `message-circle` ribbon icon now activate/reveal a right sidebar leaf (reusing an existing leaf) instead of a popup. All existing behavior preserved (three-tier retrieval, streaming + fallback, thinking panel, save loop, LLM save suggestion, history cap, clear, copy, stop). `renderThinkingBlocksUI` pure fn + `SuggestSaveModal` untouched. Styles migrated to native `var(--…)` theme variables (0 hardcoded colors). Test mock extended with `ItemView`/`WorkspaceLeaf`. **1003 tests still passing.**
+
 ### Completed (v1.22.0) — Schema One-Click Apply + Dynamic Tag Sync + zh-Hant + Status Bar (2026-06-23)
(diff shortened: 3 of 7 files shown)


green-dalii:
## Review of the diff

Read the full diff (369 lines, 7 files). What's there:

- `src/main.ts` — `new QueryModal(this.app, this).open()` → `void this.activateQueryView()` + new `registerView(VIEW_TYPE_QUERY, …)` + new ribbon icon `message-circle` + `workspace.getRightLeaf(false)` reveal. Verified `new QueryModal` is only at `main.ts:646` (1 callsite), so the rename is safe.
- `src/wiki/query-engine.ts` — `class QueryModal extends Modal` → `class QueryView extends ItemView`. `onOpen`/`onClose` become `async` (per Obsidian ItemView contract). `renderThinkingBlocksUI` and `SuggestSaveModal` are untouched.
- `styles.css` — removes fixed `.llm-wiki-query-modal` (800×780) and replaces with flex `.llm-wiki-query-view`. All hardcoded colors (`#4caf50` / `#fff3e0` / `#2196f3` / `#f44336` / `#999` / `white`) replaced with native `var(--…)` theme tokens. This is genuinely a fix — the hardcoded colors were breaking light-mode rendering.
- `src/__tests__/__support__/setup.ts` — adds `ItemView` and `WorkspaceLeaf` stubs so the test mock can import the new class.
- `CHANGELOG.md` + `CLAUDE.md` — both updated to reflect the new In Progress section.
- **`.gitignore` — adds `data.json`**. Flagged: this addition is **dead noise** (`data.json` is Obsidian plugin runtime data stored at `<vault>/.obsidian/plugins/karpathywiki/data.json`, which lives in the user's vault, not the plugin repo — it cannot be committed even if we wanted to). The current `.gitignore` correctly doesn't list it because it's not a file this repo ever sees. Please drop this line in a follow-up commit; it's unrelated to the Modal → side panel goal and adds visual noise to the PR diff.

## Review notes

1. **Branch target.** This PR is opened against `main` directly. Per the project contribution rules, `main` is protected — feature work should go to a `feat/...` branch first. Can you retarget to `feat/query-side-panel`? (or do you want me to do that on the maintainer side?)
2. **`.gitignore` line.** As noted above, the `data.json` addition is unrelated and unnecessary — please drop it.
3. **Tests.** Body says "1003 tests passing" but I don't see new tests added for `QueryView` itself — only the mock was extended. The existing `renderThinkingBlocksUI` suite is what stays green. That's defensible ("all functionality preserved" claim), but for any future refactor it would be useful to have at least one test that asserts `QueryView extends ItemView` and the right-dock leaf is created on `activateQueryView()`. Optional for this PR.
4. **Threading.** `onOpen`/`onClose` are now `async`. The current `Modal` versions were sync. The diff doesn't show any such caller. ✅
5. **Leaf reuse.** The `existing = workspace.getLeavesOfType(VIEW_TYPE_QUERY)` pattern is correct — second click reuses the existing leaf instead of opening duplicates. Good UX.
6. **Cross-Origin** — none. All 7 files relate to the stated scope (Modal → side panel, style migration). The `.gitignore` line is unrelated to scope but trivial to drop.

## Where this fits in the v1.22.x / v1.23.0 plan

- v1.22.0 shipped today (commit 22750b8). v1.22.1 is the local PATCH cycle collecting in-the-wild P0 issues.
- This PR is a UX refinement, not a P0. It fits naturally into **v1.22.1** alongside the PATCH work (small bump, low risk), OR it can wait for **v1.23.0** if you'd rather land PPR + this in the same release.
- I'd lean **v1.22.1** because (a) the style fix (hardcoded colors → theme vars) is a real Obsidian-theme-compatibility fix, not just a UX choice, and (b) it would feel weird to leave popup-mode Query in the v1.22.x line once side panel is on the same release branch.

Your call on timing. The two things I'd ask before merge: **retarget to `feat/query-side-panel`** and **drop the `.gitignore` line**.

YounianC:
Good point on the normal installed-plugin layout. In our local dev workflow, though, we symlink the vault plugin folder directly to the source tree for fast testing, so Obsidian writes `data.json` into the repo working directory. Since that file can contain LLM API keys, I added it to `.gitignore` to prevent accidental commits.

I can add a short comment above the rule to make that context clear for future reviewers. @green-dalii

green-dalii:
## Follow-up: `.gitignore` — acknowledged, with context ask

Thanks for the explanation — the symlink-vault-to-repo workflow is a legitimate local dev pattern I hadn't considered. `data.json` containing LLM API keys makes the case for `.gitignore` protection clear. Adding a one-line comment above the rule clarifying this (e.g. `# Local dev: symlinked vault plugin dir writes Obsidian's runtime data here`) would make future reviewers' lives easier — I'll include that when I fix the `.gitignore` in a follow-up.

## Checklist before merge

Three things I'd need before merging, to respect the project's review process:

**1. Retarget base branch.** This PR's `baseRefName` is `main`. Per the project's contribution rules, `main` is protected — feature work should go through a review branch first. Two options:
- (a) If you have maintainer push access (or I can retarget for you), I'll retarget to a `feat/` branch on your side, rebase, and merge the resulting PR into main through the standard flow. I'd do this as a single maintainer commit with a clean `feat: ...` message.
- (b) If you'd prefer to handle it, you can do it from the GitHub PR page: click "Edit" at the top → change the base branch to a new `feat/query-side-panel` branch, then I'll rebase + merge.

**2. `.gitignore` context.** As noted above — happy to include the comment in my follow-up commit.

**3. Tests.** This is an optional ask, not a blocker — I'll note it as a "nice to have in this PR" but won't hold merge for it. The existing `renderThinkingBlocksUI` suite staying green covers the core logic.

## Timing

I'd like to land this in **v1.22.1 PATCH** (the local patch cycle currently collecting in-the-wild P0 fixes — CSS `:has()` fix just landed, #187 and #197 are queued). If the retarget + `.gitignore` context are in by **48 hours from now** (by **2026-06-25 15:00 UTC+8**), I'll merge this into v1.22.1. After that window, I may take over the retarget myself if the PR hasn't moved — just to keep the patch release on schedule.

Appreciate the clean diff — the `onOpen`/`onClose` async contract and `renderThinkingBlocksUI` untouched confirmation were exactly what I needed to see. Looking forward to getting this in.

Issue #188, opened 2026-06-22
merge.ts emits hardcoded English section headers, ignoring wikiLanguage


DocTpoint:
## Summary

`src/wiki/prompts/merge.ts` hardcodes **English** section headers in its output templates, ignoring the configured `wikiLanguage`. As a result, pages produced via the **merge** path get English headers (`## Related Concepts`, `## Related Entities`, …) even on a non-English wiki, while pages produced via the **create** path get localized headers — so a single vault ends up with inconsistent, mixed-language section headers.

## Root cause

`generation.ts` uses `{{section_*}}` placeholders that `applySectionLabels()` localizes via `getSectionLabels(settings)`. `merge.ts` does **not** — it writes the headers literally (v1.21.1):

- `src/wiki/prompts/merge.ts:92` — `## Related Entities`
- `src/wiki/prompts/merge.ts:95` — `## Related Concepts`
- `src/wiki/prompts/merge.ts:145` / `:148` — same, in the second template

(Other section headers in those templates are literal English too.)

## Reproduction

On a wiki with `wikiLanguage: "de"`, re-ingest a source so an existing entity/concept page is merged. The merged page comes out with `## Related Concepts` / `## Related Entities` (English), whereas a newly-created page on the same wiki has `## Verwandte Konzepte` / `## Verwandte Entitäten`.

## Impact

- User-visible inconsistency (mixed-language headers across the vault).
- Breaks any post-processing that locates sections by their localized label — it must special-case the English literals to handle merged pages.

## Suggested fix

Replace the literal headers in `merge.ts` with the same `{{section_*}}` placeholders used in `generation.ts`, so `applySectionLabels()` localizes them consistently across both code paths.

Happy to send a PR.

green-dalii:
Confirmed — `src/wiki/prompts/merge.ts` hardcodes `## Related Entities` / `## Related Concepts` (lines 56, 57, 92, 95, 108, 109, 145, 148). Fix is mechanical: replace literals with `{{section_*}}` placeholders + let `applySectionLabels()` localize. Picking up in v1.22.0 P1 alongside #187 / #186 (i18n bug batch). PR welcome.

green-dalii:
Fixed in v1.22.0: https://github.com/green-dalii/obsidian-llm-wiki/releases/tag/1.22.0

[...]

PR #200, opened 2026-06-23
fix(#187): deterministically correct sources/-prefixed related links

DocTpoint:
Closes the root cause behind #187.

Per the discussion, this is the deterministic post-pass rather than a list-visibility heuristic. The Related Concepts / Related Entities sections are LLM-formatted against a truncated (`MAX_PAGES`) existing-pages list, so the model often can't place a target and defaults to the most salient prefix in the prompt — `sources/`. Raising the cap or splitting the list by type shrinks the problem but can't cover the worst case: a page and a related sibling generated in the **same run**, where the sibling exists in the list at no `MAX_PAGES` because it doesn't exist yet when the prompt is built.

The code already knows each related name's type, so we re-assert it after generation instead of trusting the guess.

## Shape

- New pure module `src/core/related-link-corrector.ts` (~75 LOC), `correctRelatedLinkPrefixes(content, relatedEntities, relatedConcepts, relatedEntitiesLabel, relatedConceptsLabel, preserveCase)`.
- Wired into `page-factory.ts` after `createNewPage` and `mergePage`.
- Section-scoped by header label, so the legitimate `[[sources/<slug>]]` citations in *Mentions in Source* (whose names often coincide with a related concept) are never rewritten.
- Section-driven, so it also self-heals stale `sources/`-links carried through a `mergePage` from the existing body — coverage the (a)/(b) path wouldn't have given.
- Recognizes the localized headers and the canonical English ones, so it also heals pages merged before the #188 fix (which carry literal English headers).
- No LLM, no false-merge risk, `O(links)`.

## Named regression tests

In `src/__tests__/core/related-link-corrector.test.ts`, the two cases the (a)/(b) heuristic misses are named test IDs so they'll catch any future regression of that choice:

- **`[truncated-existing-pages]`** — target exists but fell outside the `MAX_PAGES` window.
- **`[co-created-siblings]`** — sibling generated in the same run, never in `existing_pages` at any `MAX_PAGES`.

Plus behavioural coverage (Mentions citations untouched, prose untouched, ambiguity-via-section, merge-carried self-heal, inflection/spacing via slug). Full suite green (1017).

## Validation

Validated on a ~900-page medical vault: a 46-page bulk ingest produced **0 pages** with a `sources/`-prefixed related-link entry, vs the standing backlog this was diagnosed from.

---

Note on target: `feat/v1.22.1-patches` doesn't exist yet, so this opens against `main`. Happy to retarget to the patch branch once you create it — or rebase as you prefer.

### DIFF
--- a/src/__tests__/core/related-link-corrector.test.ts
+++ b/src/__tests__/core/related-link-corrector.test.ts
@@ -1,0 +1,100 @@
+import { describe, it, expect } from 'vitest';
+import { correctRelatedLinkPrefixes } from '../../core/related-link-corrector';
+
+describe('correctRelatedLinkPrefixes (root-cause fix for sources/-prefixed related links)', () => {
+  const ENT = 'Related Entities';
+  const CON = 'Related Concepts';
+
+  // --- Named regression cases: the two failure modes the (a)/(b) list-visibility
+  //     heuristic cannot cover, and that this post-pass exists to catch. ---
+
+  it('[truncated-existing-pages] re-types a link the model prefixed sources/ because the target fell outside the truncated existing_pages window', () => {
+    // The related entity exists in the vault but outside the MAX_PAGES window, so the
+    // model could not see its path and guessed the most salient prefix — sources/.
+    const c = ['## Related Entities', '- [[sources/Hippocampus|Hippocampus]]'].join('\n');
+    const r = correctRelatedLinkPrefixes(c, ['Hippocampus'], [], ENT, CON, true);
+    expect(r).toContain('[[entities/Hippocampus|Hippocampus]]');
+  });
+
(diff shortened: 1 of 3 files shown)


green-dalii:
Accepting. Full diff read (246 lines, 3 files), matches what was asked for in #187:

- `correctRelatedLinkPrefixes` — pure, no LLM, O(links), section-scoped via localized headers. `[[sources/<sourceSlug>]]` citations in Mentions are never rewritten even when the source name is also a related concept.
- Known-type-wins-over-section; ambiguous names defer to section context.
- Bonus coverage beyond the (a)/(b) plan: co-created siblings, mergePage-carried stale links, and pre-#188 legacy pages with literal English headers. The last one is especially nice — it self-heals pages that were merged before the #188 fix without touching them individually.

Tests cover the two named regression cases I asked for (truncated-existing-pages + co-created-siblings) plus 6 behavioral ones. Going into v1.22.1 PATCH.

PR #263, opened 2026-07-09
feat: add Amazon Bedrock LLM provider with bearer + AWS Profile/SSO auth

dmsessions:
## Summary

Adds **Amazon Bedrock** as a new LLM provider with two auth modes:

- **Bearer API key** — long-lived Bedrock API key (portable across machines)
- **AWS Profile / SSO** — resolves credentials via `@aws-sdk/credential-providers` `fromNodeProviderChain` (~/.aws + SSO cache with 5-min-pre-expiry auto-refresh). Desktop-only.

**8 curated inference profile models** (2 `global.*` + 6 `us.*` covering Claude Sonnet 5 / Opus 4.8 / Haiku 4.5, Meta Llama 3.3 70B, Amazon Nova Pro / Lite). **8-region dropdown**. **Extended-thinking control** via `providerOptions.bedrock.reasoningConfig`. **9 new i18n keys** across all 10 locales.

## Motivation

Enables the plugin for AWS-native users, especially those on SSO federation who cannot use static IAM keys. Requested / dogfooded by @dmsessions (this PR's author) — end-to-end tested on live AWS with SSO federation (`aws sso login --profile ...` → plugin picks up refreshed credentials → real Bedrock calls succeed streaming).

## Changes

- `src/llm-sdk/bedrock-sdk-client.ts` — new client (~210 LOC, mirrors `AnthropicSdkClient`)
- `src/llm-sdk/create-llm-client.ts` — factory branches for both auth modes (async + sync), lazy-imports `@aws-sdk/credential-providers` in the async path
- `src/llm-sdk/provider-guards.ts` — new shared helpers (`requiresApiKey`, `isBedrockProfileMode`, `deriveBedrockAuthMode`, `BEDROCK_DEFAULT_REGION`). Production + tests import the same predicates so drift is impossible.
- `src/llm-sdk/openai-sdk-client.ts` — `mapAiSdkError` extended for `CredentialsProviderError`, with narrow SSO-hint matching (only appends "run aws sso login" when message actually indicates SSO expiry, not on profile typos)
- `src/types.ts` — `BEDROCK_REGIONS`, `BEDROCK_MODELS`, `PREDEFINED_PROVIDERS.bedrock`, optional `region`/`bedrockAuthMode`/`awsProfile` on `LLMWikiSettings`
- `src/ui/settings.ts` — Auth Method dropdown (Bearer ↔ AWS Profile), Region dropdown, Profile Name text input, mobile gate hides Profile option
- `src/main.ts` — plumbing + apiKey-empty guard exemption for Profile mode
- `src/schema/auto-maintain.ts` — same guard exemption in the welcome-note probe
- `src/texts/*.ts` (10 locales) — 9 new keys, natural translations
- `esbuild.config.mjs` — `platform: 'node'` + `node:*` builtin externals (required to bypass `@smithy/core`'s browser stubs which return `Symbol.for('node-only')` for filesystem helpers)
- `package.json` + lockfile — `@ai-sdk/amazon-bedrock@^4.0.131`, `@aws-sdk/credential-providers@^3.1083.0`

## Test plan

- [x] Six-Gate Gate 1 green: `pnpm lint` 0/0 · `tsc` 0 · **127 files / 1794 tests** · build clean · css-lint 0
- [x] Dogfooded end-to-end on real AWS SSO — Test Connection ✅, Query Wiki streaming ✅ with `global.anthropic.claude-sonnet-5`
- [x] Pre-merge 5-lens review (spec-compliance / code-quality / test-coverage / adversarial-refutation / correctness). Zero Critical. 9 Important + 8 Minor findings all fixed in commit `cd50420`.

## Key design decisions

1. **Bearer + credentialProvider mutually exclusive** in `BedrockSdkClient` — enforced at both the constructor (XOR check) and the SDK level.
2. **Error detection via `err.name === "CredentialsProviderError"`** — avoids the dual-package `instanceof` hazard with `@smithy/core`.
3. **Narrow SSO-login hint match** — only appended when the error message actually indicates SSO expiry (`SSO Token` / `SSO session` / `expired`). Profile typos / missing `~/.aws` get the raw AWS message instead of misdirecting advice.
4. **Mobile gate defense-in-depth** — settings UI hides Profile option, factory throws desktop-only error, `preloadLLMClientModules` skips `@aws-sdk/credential-providers` on mobile. All three layers verified.
5. **Backward compat** — `bedrockAuthMode`/`awsProfile`/`region` all optional and absent from `DEFAULT_SETTINGS`. Existing bearer-mode users (if you already had Bedrock configured somehow) load unchanged.
6. **esbuild `platform: "node"`** — required for `@smithy/core` CJS interop. Verified other `@ai-sdk/*` providers have zero Node builtins → mobile-safe.

## Bundle size

- Pre-branch: 3.5 MB
- Post-branch: 4.5 MB (+29%)
- Delta driven by `@aws-sdk/credential-providers` + `@smithy/*` transitives (~1 MB). Bearer users pay the full cost due to the sync-factory contract; lazy-import optimization deferred as follow-up if bundle size becomes a concern.

## Commit history

21 commits, structured incrementally: types → client → factory → main plumbing → i18n → settings UI → Gate 1, then SSO added in the same shape, then bug fixes surfaced during dogfooding (`errorNoApiKey` guard, `parseKnownFiles` bundling), then round-1 review polish.

Happy to squash on merge if you'd prefer — commits stand on their own but the granular history is more useful during review than in `main`.

## Docs

CLAUDE.md not modified. If you'd like a Bedrock-specific bullet under "In Flight (v1.24.0)" or a mention in the deferred-items section, happy to add in a follow-up commit — didn't want to touch project-standards prose without your input.

🤖 Generated with Claude Code


dmsessions:
Force-pushed to bring the branch into CONTRIBUTING.md compliance:

- Added CHANGELOG.md entry (Added / Unreleased).
- Added Amazon Bedrock bullet to CLAUDE.md under "In Flight (v1.24.0)".
- Added Bedrock to all 10 README language variants.
- Retroactively added `Signed-off-by: Darren Sessions <darrensessions@me.com>` (DCO) to all commits on the branch.

Gate 1 still green: pnpm lint 0/0, tsc 0, 1794 tests passing.

green-dalii (changes requested):
# Review on PR #263 — Amazon Bedrock LLM provider

**File:** `src/llm-sdk/bedrock-sdk-client.ts` (+ 30 files; `esbuild.config.mjs` global change)

## Summary

Substantial PR. Bedrock integration is genuinely well-architected, end-to-end dogfooded on real AWS SSO, and the bearer + credentialProvider mutually-exclusive XOR design is clean. However, two global changes in this PR violate this project's hard constraints and require rework before this can land.

## Blocking (must fix before merge)

### 1. `esbuild.config.mjs` global change to `platform: "node"` is a no-go

This is a single-line edit that affects every user's `main.js`. Mobile Obsidian users (iOS / Android) get served the same `main.js`; a global `platform: "node"` injection either breaks mobile outright or forces a node/mobile build split. Neither is acceptable.

**Right shape** (no global change needed):
- Wrap `@aws-sdk/credential-providers` and any `@smithy/*` modules behind a dynamic `await import('...')` inside the async factory branch only. Bearer-mode users never load these modules.
- If `@smithy/core` genuinely needs the platform node stub (your `Symbol.for('node-only')` comment in the diff), the right move is `external: ['@smithy/*']` for the Bedrock entry, not a global `platform` flip.
- Or — verify with the Vercel SDK docs whether `@ai-sdk/amazon-bedrock` actually pulls in the full `@smithy` chain or just its own dep set. Lazy import + tree-shaking may sidestep the platform issue entirely without any `external` config.

### 2. Bundle size: +1 MB (+29%) is not acceptable

Every plugin user pays this cost on download, even if they never pick Bedrock and even if they only want bearer mode. We treat bundle size as a project-wide constraint; v1.23.0 took a one-time 1.24 → 3.17 MB hit for AI-SDK v6 that we accepted explicitly, but +29% on top of that is a different conversation. A single provider should not move the bundle needle for users who don't use it.

**Right shape**: the lazy-import path in #1 above should drop the bundle delta to ~200-400 KB (only SSO users pay the `@aws-sdk/credential-providers` cost, and only on first SSO-call, not at startup).

## Non-blocking (recommend but not required for merge)

### 3. `mapAiSdkError` extension for `CredentialsProviderError`

The narrow SSO-hint matching (`SSO Token` / `SSO session` / `expired` substrings) is well-thought-out. Add a unit test that locks the specific message patterns so future AWS SDK upgrades don't silently break the SSO hint or misfire on profile typos. The `err.name === "CredentialsProviderError"` check is robust but brittle to upstream name changes — guard it.

### 4. `<!-- reviewed: keep -->` semantics verification

I'm flagging this as a separate concern from above because the docs/PR description doesn't make it clear which user-curation mechanism your code touches. For context: this codebase has **two** parallel reviewed mechanisms:
- Frontmatter `reviewed: true` — controls entire page-level read-only mode (`page-factory.ts:359, 379, 1218-1222` → `appendToReviewedPage` minimal-append path)
- HTML comment `<!-- reviewed: keep -->` — controls only the Mentions section (`core/mentions-injector.ts:12, 51`)

If the Bedrock path touches either, add a unit test that verifies the gate fires correctly under both markers.

## Approve (good as-is)

- `bedrock-sdk-client.ts` (~210 LOC mirroring `AnthropicSdkClient`) — clean
- `provider-guards.ts` shared predicates (`requiresApiKey`, `isBedrockProfileMode`, `deriveBedrockAuthMode`) — good DRY structure
- `BEDROCK_REGIONS` + `BEDROCK_MODELS` config — sensible defaults
- 9 i18n keys × 10 locales — no comments
- Mobile gate (UI hides Profile option, factory throws, `preloadLLMClientModules` skips import) — defense in depth done right
- Narrow SSO-login hint matching — well-designed
- Backward compat (new fields all optional + absent from `DEFAULT_SETTINGS`) — good

## Suggested path forward

1. Refactor in place on the same branch (no need to re-open PR)
2. Drop the global `esbuild` change
3. Lazy-import SSO deps
4. Re-run `pnpm build` and report the new bundle delta (target: ≤ ~300 KB above v1.24.0's 3.17 MB)
5. Once the cost model is right, this is a fast approve

**Cycle placement**: v1.24.1 PATCH is the right target — not a blocker for v1.24.0 (already shipped) and the user value (AWS-native + SSO) is real.

The CLS test plan you ran already covers the correctness side; we mainly need the cost side verified. Take the time you need.

---

🤖 Generated with Claude Code

dmsessions:
Thanks for the thorough review. Before I do the refactor, I want to push back on the technical premise behind blocker #1 — because if I'm right, the fix you're asking for won't actually reduce the bundle, and I'd rather we discuss the real trade-off up front than ship a change that costs a day and moves nothing.

## Bundle-size math (measured, not estimated)

I built at every configuration to isolate what each option actually costs:

| Config | main.js | Δ vs main baseline |
|---|---|---|
| main branch (v1.24.0, no Bedrock) | **3.1 MB** | — |
| Branch, `platform: 'node'` (current) | **4.3 MB** | +1.2 MB |
| Branch, `platform: 'browser'` (revert) | **4.1 MB** | +1.0 MB — **but parseKnownFiles is a Symbol stub → runtime crash for profile-mode users** |
| Branch, `conditions: ['node']` (no platform flip) | 4.4 MB | +1.3 MB, and `parseKnownFiles = no` (browser stub) still ends up in the bundle. `mainFields` override doesn't help either |
| Branch, drop `@aws-sdk/credential-providers` entirely | **3.4 MB** | +0.3 MB |

## Composition (with `platform: 'node'`, `pnpm build`)

```
1,282,677  @aws-sdk/credential-providers
  597,769  zod
  479,596  ai@6.0.214
  419,780  @smithy/core
  310,213  @ai-sdk/openai
  258,092  @ai-sdk/anthropic
  191,045  @aws-sdk/nested-clients
  127,869  @aws-sdk/core
   98,258  @ai-sdk/amazon-bedrock  ← the actual provider
   [zod/ai/anthropic/openai are already on main — not part of the delta]
```

The +1.2 MB is essentially all AWS SDK + Smithy transitives. `@ai-sdk/amazon-bedrock` itself is only 98 KB.

## Why `await import(...)` doesn't help here

I think this is the core disagreement, so I want to state it plainly:

**esbuild + `format: 'cjs'` + single-file bundle = dynamic `await import()` defers module *initialization*, not shipping.** The imported module's bytes are still inlined into `main.js`. Obsidian plugins ship as a single JS file (there's no chunking to a sibling), so there is no way for esbuild to "lazy-load" `@aws-sdk/credential-providers` at runtime without shipping a separate `.js` file that gets `require()`d from Electron.

I verified this in the current build:

```js
// Inside main.js, at the "create-llm-client.ts" section:
if (!import_obsidian2.Platform.isMobile) {
  const credProviders = await Promise.resolve().then(() => __toESM(require_dist_cjs19()));
  modules.fromNodeProviderChain = credProviders.fromNodeProviderChain;
}
```

`require_dist_cjs19()` is the fully-inlined credential-providers module. `await Promise.resolve().then(() => ...)` runs it on a microtask, but every byte is already in `main.js`. Bearer-mode users pay the full download cost because esbuild follows every `import()` statement it can statically resolve.

The same happens inside `fromNodeProviderChain` itself — the AWS SDK uses `await import("@aws-sdk/credential-provider-node")`, which uses `await import("@aws-sdk/credential-provider-sso")` etc. esbuild inlines all of them.

## Why `platform: 'node'` is (I believe) not the mobile risk you were worried about

I re-audited every path:

- **What `platform: 'node'` actually changes:** package export-condition resolution. It picks the `node` condition over `browser` when resolving `@smithy/core/config`, which is the difference between `parseKnownFiles = () => { throw }` (browser stub with `Symbol.for('node-only')`) vs the real CJS `parseKnownFiles = async fn(...)`. That's the fix.
- **What it does NOT change:**
  - No injection of `Buffer`, `process`, `__dirname` globals at the plugin-load boundary — the bundle still runs untouched top-level init identical to `platform: 'browser'` (verified by diffing the first 200 lines of both builds).
  - Node builtins (`node:fs`, `node:path`, etc.) are already `external` in `esbuild.config.mjs` — mobile-load never resolves them at plugin startup.
  - The AWS SDK's `require("node:fs")` calls happen only when profile-mode code executes. Mobile users can't reach that code path because we gate profile mode at three layers: settings UI (option hidden), `buildBedrockClient` (throws desktop-only error), `preloadLLMClientModules` (skips the credential-providers import on mobile).
- **What it does cost:** +200 KB. Under `platform: 'browser'`, `@smithy/core`'s browser stub submodules are ~200 KB smaller than their Node CJS versions (which contain real filesystem helpers). But the "smaller" ones don't work.

I'd like to defend `platform: 'node'` in the commit body more explicitly and add a comment in `esbuild.config.mjs` explaining the exact reason. If you still want it out after seeing this, I'll accept that.

## Options as I see them

Given the numbers above, the real trade-off is:

**Option A — Accept +1.2 MB as one-time cost.** Precedent: v1.23.0 took the 1.24→3.17 MB hit for AI-SDK v6 with your explicit sign-off. This is a similar shape. Pro: ship the feature. Con: bearer-only users pay the cost too because esbuild can't split the bundle.

**Option B — Split-bundle rework (~1 day).** esbuild config emits two files: `main.js` (~3.4 MB, no credential-providers) and `bedrock-sso.js` (~1 MB). The profile-mode path uses runtime `require('./bedrock-sso.js')` (Obsidian's Electron `require` is available at runtime for sibling files). Manifest still references `main.js` only; `bedrock-sso.js` gets shipped in the release ZIP. Pro: bearer-only users pay ~300 KB, not 1.2 MB. Con: novel pattern for this plugin, needs vitest fixture updates, manual verification on desktop + mobile, and potentially breaks Obsidian's plugin-install flow if it doesn't preserve sibling files (needs checking against Obsidian community plugin distribution rules).

**Option C — Drop SSO from this PR.** Land bearer-only Bedrock at +300 KB. Add SSO in a follow-up PR once we have a satisfactory design for the size constraint. Pro: cleanest. Con: I dogfooded SSO on live AWS federation — it's the feature I actually need, and dropping it means I go back to per-machine bearer keys.

## What I did this round

- ✅ **Non-blocking #3 — locked SSO-detection regex patterns.** Added 8 unit tests (`test(bedrock): lock SSO-detection regex alternatives + name-check fallback`, commit `a517eb3`). Each of the four regex alternatives (`SSO Token`, `SSO session`, `token .* expired`, `has expired`) now has a targeted lock; case-insensitivity is locked; `err.name === 'CredentialsProviderError'` primary path and the `AWS credential provider failed:` AI-SDK-wrapped fallback are both explicitly covered; false-positive guard for `.name`-matches without SSO-expiry substring is verified. **1794 → 1802 tests passing.**
- Non-blocking #4 — verified Bedrock path does not touch `page-factory.ts` `appendToReviewedPage` or `core/mentions-injector.ts`. Bedrock is an LLM client; it never reaches the write-gate. N/A for this PR.
- Kept `esbuild.config.mjs` `platform: 'node'` pending your decision above.

Gate 1 all green: `pnpm lint` 0/0, `tsc` 0, 1802 tests, `pnpm build` 4.3 MB, `pnpm css-lint` 0.

## Ask

Which option (A/B/C) do you want to pursue? If A, I'll add a defensive comment block in `esbuild.config.mjs` explaining the mobile-safety analysis and squash. If B, I'll cost it out more carefully first (Obsidian plugin distribution rules for sibling files need verification). If C, I'll strip the profile-mode diffs and open a follow-up issue for the split-bundle design.

Happy to hop on a shorter async back-and-forth if it's faster than the PR thread.

green-dalii (commented):
Thanks for the bundle-size pushback — your measured data forced me past a wrong assumption about `await import()` lazy-loading, and that's exactly how reviews should work. Walking through AWS docs + the dependency tree, I want to propose a different path.

## What I got wrong

My blocking #1 ("drop `platform: 'node'`") assumed `await import()` would lazy-load `@aws-sdk/credential-providers`. Your build-output trace is correct: in esbuild + CJS single-file bundle, dynamic `await import()` defers *initialization*, not *shipping* — the bytes are inlined. My framing was wrong.

## What changed my view

Three things from AWS docs + dependency tree inspection:

1. **`@ai-sdk/amazon-bedrock` itself is tiny.** v4.0.131 deps: `aws4fetch` (65 KB), `@ai-sdk/openai` (already on main), `@ai-sdk/anthropic` (already on main), two `@smithy/util-*` (light). The +1.2 MB delta comes *entirely* from `@aws-sdk/credential-providers` — your own drop-test confirmed this (branch, drop credential-providers → 3.4 MB, +0.3 MB).

2. **Bedrock Converse API supports bearer tokens directly.** Per https://docs.aws.amazon.com/bedrock/latest/userguide/api-keys.html, the official curl example is:
   ```
   curl -X POST "https://bedrock-runtime.<region>.amazonaws.com/model/us.anthropic.claude-sonnet-4-6/converse" \
     -H "Authorization: Bearer $AWS_BEARER_TOKEN_BEDROCK" \
     -d '{...}'
   ```
   So SSO is *not* required for Claude via Converse — it's an optional convenience for users already inside `aws sso login`. Different audiences, different bundle trade-offs.

3. **`bedrock-mantle` exposes the Anthropic Messages protocol verbatim.** Per https://docs.aws.amazon.com/bedrock/latest/userguide/bedrock-mantle.html and https://docs.aws.amazon.com/bedrock/latest/userguide/models-endpoint-availability.html, `bedrock-mantle.<region>.api.aws/v1/messages` accepts the *exact same* Anthropic Messages body schema as `api.anthropic.com` (with `anthropic_version: bedrock-2023-05-31`). Bearer ABSK key auth. **No SigV4, no AWS SDK.**

## The plan I'm proposing

I'd like to land Bedrock in **3 stages** rather than the all-at-once approach in PR #263. Stage 1 ships in v1.24.1 from my side; Stage 2 lands in v1.25.0+ if real demand shows up; Stage 3 is the SSO piece (which I think we should defer indefinitely unless we get 5+ explicit user requests).

### Stage 1 (v1.24.1 PATCH, ~750 LOC, ~+3 KB bundle delta) — **I'll land this myself**

Reuse the existing `AnthropicSdkClient` and `OpenAICompatSdkClient` factory paths. No new deps. ~10 LOC in the factory:

```ts
// In createLLMClientFromSettings / Sync
if (provider === 'bedrock-anthropic') {
  return new AnthropicSdkClient({
    apiKey,
    baseURL: `https://bedrock-mantle.${bedrockRegion}.api.aws`,
  });
}
if (provider === 'bedrock-openai') {
  return new OpenAICompatSdkClient({
    apiKey,
    baseURL: `https://bedrock-mantle.${bedrockRegion}.api.aws/v1`,
    provider: 'bedrock-mantle',
  });
}
```

New settings: `provider: 'bedrock-anthropic' | 'bedrock-openai'` + `bedrockRegion: string` (default `us-east-1`). New `PREDEFINED_PROVIDERS` entries. 5 i18n keys × 10 locales.

**Coverage** (per https://docs.aws.amazon.com/bedrock/latest/userguide/models-endpoint-availability.html):

| Models reachable via Stage 1 | Endpoint | Our client |
|---|---|---|
| Claude Sonnet 5 / Fable 5 / Haiku 4.5 / Opus 4.7 / Opus 4.8 | bedrock-mantle Messages | `AnthropicSdkClient` |
| GPT-5.4 / GPT-5.5 / GPT OSS 20B/120B | bedrock-mantle Chat Completions | `OpenAICompatSdkClient` |
| DeepSeek V3.1 / V3.2 | bedrock-mantle Chat Completions | `OpenAICompatSdkClient` |
| Mistral Large 3 / Devstral / Magistral / Ministral / Voxtral | bedrock-mantle Chat Completions | `OpenAICompatSdkClient` |
| Kimi K2 Thinking / K2.5 | bedrock-mantle Chat Completions | `OpenAICompatSdkClient` |
| Qwen3 family | bedrock-mantle Chat Completions | `OpenAICompatSdkClient` |
| GLM 4.7 / 4.7 Flash / GLM 5 | bedrock-mantle Chat Completions | `OpenAICompatSdkClient` |
| Grok 4.3 | bedrock-mantle Chat Completions | `OpenAICompatSdkClient` |
| Google Gemma 3 | bedrock-mantle Chat Completions | `OpenAICompatSdkClient` |
| NVIDIA Nemotron 3 | bedrock-mantle Chat Completions | `OpenAICompatSdkClient` |

**Bundle delta: ~+3 KB** (settings UI strings + factory branches; zero new npm deps).

**Known gap** (Stage 1 doesn't cover): Claude Sonnet 4 / 4.5 / 4.6, Claude Opus 4.1 / 4.5 / 4.6, Llama 3.x / 4.x — these are bedrock-runtime-only and require Converse / InvokeModel APIs. That's Stage 2.

### Stage 2 (v1.25.0+ MINOR, conditional on real demand) — bearer-only via `@ai-sdk/amazon-bedrock@^5`

If users start asking "why can't I use Claude Sonnet 4 on Bedrock?" or "where's Llama 4 Maverick?", we land Stage 2. Constraints:

- **Bearer-only.** Drop `@aws-sdk/credential-providers` entirely. Drop the SSO/profile code path you put in PR #263. Drop `platform: 'node'` global flip.
- **`@ai-sdk/amazon-bedrock@^5`** (not v4) — v5 swapped `@aws-sdk/signature-v4` for `aws4fetch` (~65 KB transitive), and the dep graph is lighter overall.
- Replace `bedrock-anthropic` / `bedrock-openai` Stage 1 paths with a single `bedrock` provider id that routes through `@ai-sdk/amazon-bedrock`. Two-stage consistency: Stage 1 factory branches become legacy fallback only.
- Bundle delta: ~+0.3 MB (matches your drop-test measurement). Mobile-safe without `platform: 'node'` (your analysis was correct on this).
- Models: 8-region × ~12-model (Sonnet 4.x, Opus 4.5-4.8, Llama 4 family, Cohere Command R+, etc.) — full Converse / Messages API coverage.

### Stage 3 (indefinitely deferred)

SSO/profile mode, IMDS refresh, the `@aws-sdk/credential-providers` chain — only land if 5+ users explicitly request SSO. Bundle delta +1.2 MB would be its own PR with full Gate-4 disclosure.

## What I'm asking you to do

Your PR #263 as written mixes Stage 2 + Stage 3. I'd suggest:

1. **Close PR #263** with a comment linking this discussion. The Stage 3 SSO piece (your main contribution — the SSO-detection regex tests + the dual-mode architecture analysis) is well thought-through and we want it documented for when Stage 3 is real, but shipping it now is not worth the bundle cost.
2. **Optionally**: open a new WIP PR titled "Bedrock SSO detection scaffolding (deferred to Stage 3)" that *only* contains the SSO-detection regex tests + your comments — so we don't lose the work. Close Stage 2 in the same PR with a WIP label.
3. **Don't block on Stage 1.** I'll land Stage 1 in v1.24.1 from my side (small PR, ~750 LOC, mostly i18n). When Stage 2 becomes real, your SSO regex tests + dual-mode analysis will accelerate it.

## Why this split

The bundle-math trade-offs across stages:

| Stage | Bundle delta | User coverage | Maintenance cost |
|---|---|---|---|
| 1 (now) | +3 KB | All Claude 5.x + GPT-5.x + Mistral + DeepSeek + Kimi + Qwen + GLM + Grok on Bedrock | Low (reuses existing factory paths + baseURL plumbing) |
| 2 (later) | +300 KB | + Claude 4.x + Llama 4 + Cohere + Jamba + Titan embeddings | Medium (one new client, drop SSO chain) |
| 3 (deferred) | +1.2 MB | + AWS SSO / profile / IMDS enterprise auth | High (AWS SDK chain to maintain, mobile gate, refresh logic) |

The exchange rate is: ~3 KB for 12+ models across 2 protocols; +300 KB for another 10 models; +900 KB for enterprise auth. Each stage pays only what it needs and only when demand justifies it.

If you'd like to be involved in Stage 2 when it lands, please flag it in the PR description — your AWS expertise is genuinely valuable there, and I'd rather pair than solo on the Converse protocol integration.

🙏 again for the data-driven pushback — it forced me to look up `bedrock-mantle` properly, which is a path I should have known about before the original review.

green-dalii:
Deferring to v1.26.0+ scope after review. Three reasons that I'd like you to weigh in on before we proceed:

1. **Architecture overlap with v1.24.1 Stage 1** — commit `5a65ad0` shipped Bedrock via bedrock-mantle (reuses `AnthropicSdkClient` + `OpenAICompatibleSdkClient`, ~+3 KB). Your #263 takes a different path via native `@ai-sdk/amazon-bedrock` + `@aws-sdk/credential-providers` (~+1.2 MB). These two approaches aren't directly in conflict, but having both land means we should pick a path deliberately rather than ship both side-by-side. Is your intent that #263 replaces Stage 1, or sits alongside it?

2. **Bundle size decision** — your Option A/B/C math is clean; the user-facing trade-off is whether AWS-SSO dogfooders earn a +1.2 MB hit that bearer-only users also pay (because esbuild can't split a single-file CJS bundle for Obsidian). My preference is to take Option B — split-bundle rework (~1 day, novel for this plugin but only realistic path given the SDK's `await import` is statically inlined). I'd want to see the Obsidian community plugin distribution rules verified before committing though. Want me to take the port pass for #263, or do you want to keep ownership of the bundle decision?

3. **Diff scope** — 31 files, +2081 / -126, 21 commits. I'd run a separate `code-review` skill on this (max effort) before any merge. Not blocking, just flagging that this won't slot into a fast PATCH cycle.

The 10-locale README sync, CLAUDE.md update, and DCO trailer compliance are all in good shape. Take your time — when you rebase onto v1.25.0 + resolve (1)-(2), I'll move it to v1.26.0 review.

Labels: `enhancement`. Milestone: `v1.26.0+`.

green-dalii:
Hi @dmsessions — really appreciate the work on this. Wanted to give you the full picture:

1. **Bedrock Stage 1 has shipped** via #280 on 2026-07-11 — uses the `bedrock-mantle` endpoint as a bearer-compatible drop-in. So the basic connectivity you covered is now in main.

2. **Stage 2 (the full `@ai-sdk/amazon-bedrock` SDK you're building toward, +0.3 MB)** is on the ROADMAP for v1.26.0 — conditional on ≥3 user-reported requests for Claude Sonnet 4 / Llama 4 on Bedrock.

Closing this PR because the diff is now in conflict with Stage 1's architecture, but **I'd love to reuse your work for Stage 2**. If you have bandwidth to rebase + adapt against the new `bedrock-mantle` design, I'm happy to review as a v1.26.0 candidate. Otherwise I'll port the relevant pieces (especially the SSO/Profile auth scaffolding you implemented) when v1.26.0 starts.

Either way: thank you for the contribution.

green-dalii:
Hi @dmsessions — following up on this with a decision update.

We've since re-examined the AWS SSO mechanism in light of our own Codex OAuth provider, which hand-rolls OAuth 2.0 device-code + PKCE with zero OAuth SDK — the same IAM Identity Center OIDC device-authorization-grant flow. The login layers are structurally identical; the one real difference is SigV4 request signing. Since the `bedrock-mantle` endpoint we already ship (Stage 1, +3 KB) accepts AWS credentials (SigV4) per AWS docs and speaks standard OpenAI/Anthropic protocols over plain SSE, an SSO user can authenticate to it with a hand-written SigV4 signer (~300 LOC) — no `@aws-sdk/credential-providers`, no +1.2 MB.

So we're **cancelling the "≥3 user requests" gate** and planning Bedrock Stage 2 as: hand-rolled IAM Identity Center OIDC (reusing the Codex device-flow skeleton) → `GetRoleCredentials` → temp IAM creds → hand-written SigV4 → existing `bedrock-mantle`. Target v1.26.x PATCH or v1.27.0, ~+10 KB, zero new npm deps.

Your #263 was the catalyst for this — especially the profile/SSO credential-chain scaffolding and the bundle-size math you laid out. If you'd like to contribute to Stage 2 under this design, it would be very welcome; the SSO credential chain maps almost 1:1 onto our Codex `auth-manager` pattern. Happy to open a design issue or PR to start.

PR #269, opened 2026-07-11
fix(mentions): non-lossy re-ingest — union accumulated mentions on merge (#267)

### LINKED ISSUE #267: Merge/skip path drops accumulated cross-source Mentions — assembleFinalContent re-emits only the new source's mentions (v1.24.0 regression from #244)
Follow-up to #244. The programmatic Mentions writer is the right direction — deterministic formatting removes the drift. But on the **merge/update path** it is non-conservative: it replaces a page's whole `## Mentions in Source` section with **only the newly-ingested source's mentions**, discarding quotes accumulated from previously-ingested sources.

**Impact:** silent data loss of verbatim provenance. This is exactly the provenance-backlink layer we discussed in #244 (walk one click back to each origin note) — it collapses to a single source on every re-ingest that touches an existing page.

**Mechanism (traced on `1.24.0`):**
- `mergePage` finalizes **all three** routes — `skip`, `complementary`, and full body-merge — through `assembleFinalContent` (`src/wiki/page-factory.ts:1030` and `:1080`).
- `assembleFinalContent` (`page-factory.ts:1100-1122`) passes only `info.mentions_with_provenance` — the extraction of the **new** source — to `injectMentionsSection`.
- `injectMentionsSection` (`src/core/mentions-injector.ts:42-99`) strips the entire existing `## Mentions in Source` section and re-emits from that new-source-only array.
- `info` never carries the existing page's mentions; those live only in the page body, which the injector removes.
- Note the `skip` route (`page-factory.ts:989-993`, "preserving existing body") shares the same finalizer, so it also rewrites the Mentions section — the widest-radius case, since duplicate re-ingest is the common path.

**Reproduce:**
1. Ingest source A that mentions entity/concept **P** → P gets A's Mentions.
2. Ingest source B that also mentions **P**.
3. P's `## Mentions in Source` now contains only B's quote; A's is gone.

Observed in practice on a 3-source concept page (Umweltbelastungen / Pharmakologie / Biochemie) reduced to one after an unrelated merge. Language- and model-agnostic (seen with a local Gemma / German wiki).

**Suggested direction:** keep the deterministic injector authoritative, but feed it the **union**. Before formatting, parse the existing page's `## Mentions in Source` back into `MentionWithProvenance[]`, merge with the new source's entries, dedup (by quote + source path), then `formatMentionsSection` the combined set — which already documents "multi-source dedup and sort". This keeps the #244 guarantees and restores cross-source accumulation. Prompt-level retention wouldn't help, since the injector overwrites deterministically afterward.

Happy to open a PR along these lines if the direction looks right — no rush, and it dovetails with the merge-conservation question in #220.

DocTpoint:
Fixes #267. Change 1 of the two-part plan we agreed on in the issue thread; Change 2 (consolidating the reviewed mechanisms) will follow as a second commit on this branch once we settle the locale-doc split I asked about — kept separate so the actual fix isn't held up.

**The bug.** On a merge, `assembleFinalContent` handed `injectMentionsSection` only the new source's mentions. Since the injector re-emits the whole `## Mentions in Source` block from the array it's given, every earlier source's accumulated mentions were dropped. `triage=skip`, `triage=complementary`, and the body-merge path were all affected — `skip` has the widest blast radius because it otherwise preserves the existing body verbatim.

**The fix.** Union at the `assembleFinalContent` choke-point, exactly where you asked — not at `mergePage` entry, so `info.mentions_with_provenance` keeps its meaning for every downstream reader.

- New `core/mentions-parser.ts`: `parseMentionsSection` recovers an existing block into `MentionWithProvenance[]`; `stripMentionsSection` shares the injector's section-boundary logic. The round-trip test pins the format↔parse contract so the two can't drift.
- `assembleFinalContent` parses the existing page's mentions and unions them with the new source's before injecting. The create path never reaches this method, so LLM-fabricated mentions on first write are still stripped.
- **Fail-safe:** if the existing block has hand-edited or linter-reflowed lines the parser can't structure, it's preserved verbatim and the merge is skipped for that pass — non-lossy even when the input isn't machine-clean.
- `dedupMentionsByProvenanceKey` (exported, reused by the union) replaces the quote-only dedup; same quote from different sources is now kept as distinct provenance.

**One note on the dedup key.** The constraint mentioned `MentionWithProvenance.position` as an existing optional field — in v1.24.0 the type carries `quote / translation / source_path / source_slug / extracted_at`, no `position`. So the key is `(quote, source_path)`, with a comment marking where an offset slots in if one is ever added. Round-trip is lossy only for fields the section never renders (`extracted_at`, `source_slug`), which the key ignores.

**Tests.** New `mentions-formatter-roundtrip.test.ts` (format↔parse fidelity, fail-safe, composite-key dedup, and a #267 cross-source regression proving all sources survive a merge); the batch-merger case is updated to the composite-key semantics (same-source dedups, cross-source stays distinct). Full suite: 1838 passing, `tsc` clean.

**CHANGELOG:** Fixed — Mentions accumulated across sources are no longer dropped when a page is re-ingested/merged (#267).


DocTpoint:
Change 2 is in — pushed as the second commit, so this PR now carries both the #267 fix and the reviewed-mechanism consolidation.

On the open question from the thread: I went ahead and wrote the "Upgrading from v1.24.0" note in all 10 README locales myself, each anchored to that file's existing terminology (来源提及 / Menciones en Source / vor Überschreiben geschützt, …) rather than a generic translation. The CJK ones especially are yours to sanity-check — I'd rather you catch anything off than assume they're perfect; if you'd prefer to rewrite any, say which and I'll drop mine.

Two places where the implementation diverged from the sketch, both because the code didn't match the assumption:

1. `pageIsReviewed` is wired at `appendToReviewedPage`, not `assembleFinalContent`. Reviewed pages route to the minimal-append path at `createOrUpdatePage` (`page-factory.ts:384/364`) and never reach `assembleFinalContent`, so the flag lives where reviewed pages actually get written. `injectMentionsSection` takes the flag and early-returns, exactly as proposed.
2. The dedup key is `(quote, source_path)`, not `(quote, source_path, position?)` — `MentionWithProvenance` has no `position` field in v1.24.0. Left a comment marking where an offset slots in if one is ever added.

CHANGELOG carries both entries. Full suite green (1838), `tsc` clean.

green-dalii:
Merged. While your PR was awaiting review I found two real bugs the original union masked:

1. Structured-branch mentions came in with empty source_path (the LLM-returned form), so the (quote, source_path) dedup key was 'quote\0' and never matched any existing mention.
2. Legacy-branch mentions carried sourceFile.path WITH the .md suffix, while parsed existing mentions had it stripped by buildBullets.

Re-ingesting the same source therefore duplicated the quote instead of collapsing it. I extracted a pure helper computeReingestMentions in core/mentions-parser.ts that owns source_path normalization (blank-fill from sourceFile.path + .md strip) and added 7 tests in mentions-reingest.test.ts (4 union scenarios + 3 defaultSourcePath fill scenarios, with insertion-order assertions).

Pushed as 2 commits on top of yours:
- f9b573d refactor(mentions): extract computeReingestMentions helper for testable union
- 10b1b74 fix(mentions): route re-ingest union through computeReingestMentions (#267)

Feel free to squash before final v1.24.1 release if you prefer a single authorship. No objections — your 2 commits stay authored by you with me as co-author on both hardening commits.

Thanks for the #267 fix and the reviewed-mechanism consolidation — high quality work, the fail-safe design is the right call.

DocTpoint:
Thanks — and both are fair catches I should have made before opening the PR. The `(quote, source_path)` key was only ever as reliable as the `source_path` normalization behind it, and the structured-empty vs legacy-`.md` split is exactly the cross-branch inconsistency the composite key cannot resolve on its own. Having `computeReingestMentions` own that normalization in one place is the right shape, and the insertion-order assertions in the new tests are a good addition.

On squash: please keep the commits as they are. Your two hardening commits are real work and should stay attributed to you, not folded into mine. The separate history is the accurate record.

On the locales: since the merge is already in, please treat my versions as a starting point. If any of the CJK or other strings reads wrong against that file's existing terminology, tell me the string and I will correct it.

PR #286, opened 2026-07-14
feat: native pdf support

AkaSakana:
## Summary

Adds native PDF source ingest without adding a PDF parser dependency.

PDF files are available from single-file, multi-file, folder, and active-file ingest paths. The plugin reads the original vault PDF through the configured provider's document capability, then sends the extracted text through the existing source analysis, duplicate detection, source-summary, entity, and concept workflows.

Kimi uses its official Files API: upload the PDF with `file-extract`, retrieve extracted text, then delete the temporary remote file.

## Provider support matrix

| Provider | PDF source ingest | Transport / behavior |
| --- | --- | --- |
| OpenAI | Supported | AI SDK native PDF `file` input |
| Anthropic | Supported | AI SDK native PDF `file` input |
| Anthropic-compatible | Endpoint-dependent | AI SDK native PDF `file` input; provider capability errors are surfaced as a skipped source |
| Gemini | Endpoint-dependent | AI SDK native PDF `file` input; provider capability errors are surfaced as a skipped source |
| GLM | Endpoint-dependent | AI SDK native PDF `file` input; provider capability errors are surfaced as a skipped source |
| OpenRouter | Endpoint-dependent | AI SDK native PDF `file` input; the selected routed model must support PDF input |
| Custom | Endpoint-dependent | AI SDK native PDF `file` input; the custom endpoint and selected model must support PDF input |
| Bedrock Anthropic / Bedrock OpenAI | Endpoint-dependent | AI SDK native PDF `file` input; selected Bedrock model capability determines success |
| Kimi | Supported | Moonshot Files API: `file-extract` upload → extracted-content retrieval → temporary-file deletion |
| DeepSeek | Unsupported | Chat Completions accepts text message content only; no documented PDF/file upload or document-input field |
| MiniMax | Unsupported | Current documented Files and Responses APIs do not define PDF document input |
| Ollama | Unsupported | Not treated as PDF-capable by the plugin |
| LM Studio | Unsupported | Not treated as PDF-capable by the plugin |

For endpoint-dependent providers, an unsupported-document response skips the source with a localized notice. Invalid, corrupted, or password-protected PDFs remain provider errors and are not misclassified as a capability failure.

## Behavior

- PDF transcriptions are cached only for one `ingestSource()` call, keyed by `${path}@${mtime}`.
- The generated source page retains the transcription under `## Extracted Text`.
- No PDF parser, OCR engine, persistent extraction cache, new plugin setting, or handwritten provider payload is added.
- DeepSeek, MiniMax, Ollama, and LM Studio reject PDF sources before any document request is sent.

## Validation

- `pnpm lint`
- `pnpm test` — 162 test files, 2098 tests passed
- `npx tsc --noEmit`
- `pnpm build`
- `pnpm css-lint`
- Manual end-to-end verification: PDF source ingest succeeded with the configured Custom provider.


green-dalii:
Hey @AkaSakana — really nice work on this. 49 files, 2 commits, Kimi Files API end-to-end with proper cleanup, error regex classifiers, transient-retry extension — that's the kind of focused execution this codebase needs more of. You clearly didn't just sketch an idea; you built the whole thing.

I want to be straight with you about where things stand on our side:

**v1.25.0 PDF Level 1 is already planned and scoped** — the plan just landed in [ROADMAP.md](https://github.com/green-dalii/obsidian-llm-wiki/blob/main/ROADMAP.md) (PR #291 merged minutes ago). So PDF is firmly on the roadmap and your timing is great. We're doing 3 PRs (Core / Ingest / UX+docs), API-native via the existing LLM SDK, 0 new npm deps.

**Why I'm not merging this PR as-is:** a few architectural mismatches where I want to be honest and invite your pushback, because you might be right and I might be wrong.

1. **Where the PDF logic lives.** Your design puts ~100 LOC of PDF branching inside `WikiEngine.readPdfSource()`. My local PR1 (committed on `feat/pdf-level-1-core`, not yet pushed) extracts a pure `core/pdf-converter.ts` + sha256 cache + metadata parser — keeps `WikiEngine` focused, lets the conversion be independently tested + cached + reused. **I might be over-engineering this** — if you see a reason inline-in-WikiEngine is actually cleaner, tell me. I'm not married to the split.

2. **Your inline output (writing extracted text into the existing source page under `## Extracted Text`).** I had planned a sidecar `<vault>/foo.pdf.pdf.md` next to the PDF. **I'm genuinely unsure which is better and want to think it through.** Sidecar pros: survives vault moves, easy to inspect/diff/delete independently, doesn't entangle with source-page rewrite. Inline pros: simpler pipeline (one file to track), no orphan management, the source-page rewrite already includes the extracted text naturally. If you have strong opinions here, I'd like to hear them before locking the design.

3. **Persistent cache.** Your PR has no disk cache — same PDF re-ingest re-bills the LLM. My local PR1 has a sha256+model cache with 30-day TTL. But honestly, if most users only ingest a PDF once, the cache might be over-engineering too. Tell me if you disagree.

4. **No escape hatch for `custom` / `anthropic-compatible` providers.** Some users route these to Kimi Coding Plan or GLM Anthropic-compatible gateways that DO support PDF even though our allowlist rejects them. I want a `forcePdfSupport: boolean` setting visible only for those two provider IDs. But if your `isDocumentUnsupportedError` regex catches these gateways gracefully on the 400 they return, maybe we don't need the setting at all. What do you think?

**What I definitely want to keep from your PR:**
- `kimi-document-reader.ts` — the multipart upload → content → delete flow is excellent work, and it's a provider my plan missed entirely
- `isDocumentUnsupportedError` / `isDocumentMissingResponse` regex classifiers — moving them to `core/pdf-error-classifier.ts` so they serve future document formats too
- The `shouldRetry` extension to `core/transient-retry.ts` — backward-compatible, useful beyond PDF
- The provider matrix with Kimi + Gemini + endpoint-dependent handling

**Plan:**
- I'll close this PR with credit (CHANGELOG + v1.25.0 Release Notes Contributors section — your name there)
- We land v1.25.0 as 3 PRs over the next few days — PR1 core is already committed locally and ready for your review on the diff
- After v1.25.0 lands on main, please open a follow-up PR on top — Level 2 multimodal, more provider coverage, OCR for scanned PDFs, encrypted PDF decryption, plus any of the design points above if you think I'm wrong

**One small question:** the `fileURLToPath` swap in `custom-instructions-injection.test.ts` — is that Node 22+ compatibility or a macOS path normalization fix? Want to know if we should adopt it globally.

Looking forward to your replies on the 4 design points. The fact that you built this end-to-end without prompting means you have real opinions — I want to hear them before I lock anything in.

AkaSakana:
Thanks for the thoughtful review and for being so clear about the v1.25.0 direction. I agree that the roadmap approach
 is better than merging this PR as-is, and I appreciate you preserving the useful pieces and crediting the contribution.

   On the design questions:

   1. **PDF logic placement**
      I agree with extracting this into a dedicated PDF conversion layer. `WikiEngine` should remain responsible for
 ingestion orchestration, while a `pdf-converter` module owns conversion, cache lookup/write, and PDF metadata handling.
 Provider-specific HTTP/API transport should remain in the LLM SDK clients. That boundary should scale much better once
 OCR, multimodal processing, encrypted PDFs, and more providers are added.

   2. **Extracted-text storage**
      I now prefer a sidecar over embedding `## Extracted Text` in the generated source page. The original PDF is already
 the canonical, directly readable Obsidian source; duplicating the full transcription in the source page makes that page
 large and mixes a regenerable implementation artifact with its summary/metadata.

      Rather than putting sidecars beside arbitrary PDFs or under `wiki/sources/`, I think a dedicated
 `${wikiFolder}/pdf-cache/` directory would be cleaner. It should be explicitly excluded from source discovery, ingestion,
 indexing, and link-generation scans. That keeps source pages clean while making the cache inspectable and manageable.

   3. **Persistent cache**
      I agree that a persistent cache is justified. Re-ingesting the same PDF otherwise repeats provider cost, latency, and
 quota usage. A SHA-256-based cache with a 30-day TTL sounds like a good default. The cache key should include the PDF
 content hash, provider/model or extraction configuration, and converter format/version so stale output is not reused after
 behavior changes. It can be refreshed when the PDF changes and cleaned when it expires or through an explicit
 cache-cleanup action.

   4. **`custom` / `anthropic-compatible` providers**
      I would avoid adding `forcePdfSupport`. For endpoint-dependent providers, the better behavior is to try the native
 document path and classify known capability failures as a clean source skip. A setting would shift capability detection
 onto users and create another long-term compatibility surface. Preemptive rejection should be limited to providers known
 not to support document input.

   For the `fileURLToPath` change: it is not specifically a Node 22+ compatibility fix. `import.meta.url` is a `file:` URL,
 and using `.pathname` is unreliable across platforms: it leaves URL encoding such as `%20` intact and can produce
 `/C:/...` rather than a valid Windows path. `fileURLToPath(...)` converts it correctly to the native filesystem path. I
 think it should be used anywhere test/tooling code converts a `file:` URL into a path.

   I’m happy for the kimi-document-reader, document-error classifiers, retry extension, and provider matrix to be carried into the planned implementation.

green-dalii:
@AkaSakana — thanks for the second-round follow-up. Your shifts (sidecar > inline, versioned cache keys, dropping `forcePdfSupport`) walked back cleanly when the argument landed. That kind of thinking is rare.

A few updates from our side, including a correction to my own draft.

## Cache-only architecture (new direction)

After your reply + some first-principles work on our side, we're pivoting the storage model:

- **No default sidecar write.** The LLM-converted markdown lives only in `.obsidian/plugins/karpathywiki/pdf-cache/{sha256}.json` — never touches the user's vault unless they opt in.
- **Markdown as a virtual source body.** `analyzeSource(file, { contentOverride })` accepts the converted markdown directly; the existing ingest pipeline runs unchanged. No sidecar file is created by default.
- **Advanced toggle `writePdfMarkdownToVault`** (default false) gives users the old sidecar behavior for audit / debugging. Same shape as your proposal, just opt-in instead of default-on.

The reasoning: the user-visible product of PDF ingest is the wiki pages (entities / concepts / summary), not the intermediate markdown. The original PDF is already the canonical, readable artifact in the vault. Duplicating extracted text into a sidecar pollutes the vault with an implementation artifact that has no end-state value — whether the sidecar lives next to the PDF or in a dedicated `pdf-cache/` directory. Cache in `.obsidian/` already gives us free re-ingest, so a second copy in the vault is pure overhead.

I considered your `${wikiFolder}/pdf-cache/` proposal carefully. Two things held me back: (a) it forces every source scanner to learn an exclusion rule (lint, index, link graph, `loadRelevantPagesForQuery`, `buildIngestedHashes`, `COMPATIBLE_SOURCE_EXTENSIONS`) — OCP violation waiting to happen, and (b) naming a sidecar directory `pdf-cache/` blurs "implementation artifact" with "sha256-keyed TTL cache" — two things with different lifecycles named the same. Happy to be wrong if you see a path I'm missing.

## Correction: `forcePdfSupport` semantics

My earlier draft proposed `forcePdfSupport` defaulting to `true`, scoped only to `anthropic-compatible`. On re-reading with our maintainer's eye, both were wrong: many compatible endpoints don't reliably support PDF, so defaulting to true would silently burn tokens for users whose gateway rejects the upload. **Final design:** `forcePdfSupport` defaults to `false`, scoped to **both** `custom` and `anthropic-compatible` as manual opt-in for users who know their gateway works. Your regex-classifier approach stays as the runtime escape hatch either way.

## Adopting from your reply

1. **`converterVersion` in cache key** — exactly your point. Prompt upgrades invalidate stale entries.
2. **Error classifiers as `core/pdf-error-classifier.ts`** — `isDocumentUnsupportedError` + `isDocumentMissingResponse` as a shared module for future document formats.
3. **`shouldRetry` extension to `core/transient-retry.ts`** — backward-compatible, useful beyond PDF.
4. **`fileURLToPath` everywhere** — adopting your point globally in test/tooling code where `import.meta.url` becomes a path. Thanks for the cross-platform explanation; we'll audit `.pathname` usages.

## Kimi Files API — your call to lead

Rather than us porting your Kimi work, **we'd like you to ship it as a follow-up PR on top of v1.25.0.** PR #286 already has the complete design. After we land PR2 + PR3 and merge to main, please open a follow-up — we can review and merge quickly since the architecture matches.

If your schedule slips, say the word and we'll take a 1-day port pass. The cleaner path is you owning the integration end-to-end so the design choices stay coherent.

## Plan

- **PR2 redo (1-1.5 days):** delete the orchestrator, feed converter output directly into `analyzeSource`, extend `PdfConversionCache` with `purgeExpired` + `enforceSizeLimit` (100MB / 1000-entry / 10MB-single-entry caps, LRU-by-mtime eviction, three-defense-layer design).
- **PR3 (1 day):** settings UI for `writePdfMarkdownToVault` (Advanced, default off) + `forcePdfSupport` for both `custom` and `anthropic-compatible` (default off, manual opt-in); 10-locale i18n; READMEs; CHANGELOG.
- **PR4 (optional, yours):** Kimi Files API on top of v1.25.0 baseline.
- **Final:** `pnpm build:dev` → HARD STOP → e2e on my vault → push.

I'll keep PR #286 open for reference. Once v1.25.0 lands on main and your follow-up merges, we close #286 with credit. Contributors section of v1.25.0 release notes will name you for Kimi Files API + provider matrix design + the cache-versioning insight — all durable contributions.

Looking forward to the follow-up.

green-dalii:
@AkaSakana — happy to report that v1.25.0 just shipped.

## What landed (matches what we agreed in this thread)

- **Cache-only architecture** — your sidecar-over-inline argument + my first-principles work both pointed the same direction. The LLM-converted Markdown lives only in `.obsidian/plugins/karpathywiki/pdf-cache/{sha256-16}.json`; never touches the user's vault unless they explicitly opt in via `writePdfMarkdownToVault`. `analyzeSource(file, { contentOverride })` runs the regular Markdown pipeline.
- **`converterVersion` in cache key** — exactly your point. Prompt upgrades invalidate stale entries.
- **`forcePdfSupport` default false, scoped to both `custom` and `anthropic-compatible`** — your regex-classifier argument landed; users who know their gateway works can opt in.
- **Provider matrix** — Anthropic / OpenAI / Bedrock-Anthropic / Bedrock-OpenAI native; DeepSeek / MiniMax rejected pre-flight; Ollama / LM Studio surfaced via Notice.
- **Error classifiers** — `isPdfRelatedLlmError` lives in `core/`, requires both rejection verb + PDF marker (no more 413 false-positives).
- **`shouldRetry`** — adopted into `core/transient-retry.ts`.
- **`fileURLToPath` audit** — completed; `.pathname` usages of `import.meta.url` replaced with `fileURLToPath`.

## v1.25.0 release notes

https://github.com/green-dalii/obsidian-llm-wiki/releases/tag/1.25.0

Contributors section names you for Kimi Files API + provider matrix + cache-versioning insight.

## PR4 — your Kimi work

v1.26.0 MINOR is the planned landing window for your Kimi Files API + non-routine PDF providers (PR4 per the original plan). Please open the follow-up PR against `main` — `feat/kimi-files-api-pdf` or similar — and ping me here. The architecture matches so review should be quick.

If your schedule slips, say the word and I'll take a 1-day port pass. But I'd rather you own it end-to-end so the design choices stay coherent with PR #286.

## Closing #286

I'll close this PR once your PR4 lands and merges. Until then it stays open for reference — it documents the Kimi Files API design intent and was the basis for several v1.25.0 decisions.

Thanks again for the focused execution. v1.25.0 is materially better for the round-trip we had here.

green-dalii:
Closing #286 now rather than waiting for PR4, on reflection. Two reasons:

1. **v1.25.0 is independently shippable without merging this PR's code.** The architecture on `feat/pdf-level-1-core` was built with the design points you raised — cache-only, converterVersion-keyed cache, regex-classifier + forcePdfSupport universal escape hatch — but the actual code on `feat/native-pdf-support` was never merged into that branch. Keeping #286 OPEN past v1.25.0 release makes `gh pr list --state open` show stale intent and complicates the next triage.

2. **Forcing PR4 to land before close would push a stale-base branch.** PR #286 is based on a v1.24.x snapshot. Shipping it now would mean pulling v1.25.0 changes forward (cache module, classifier, settings, i18n) into your branch, defeating the "fresh PR on top of v1.25.0 main" design we discussed.

**What I'd like from you on PR4 (Kimi Files API + non-routine PDF providers) when you're ready:**

- Open a new PR (`feat/kimi-files-api-pdf`) against `main` post-v1.25.0
- Target v1.26.0 MINOR (per ROADMAP)
- The 4 design points you contributed (cache key + version, classifier-first, persistent cache, sidecar/storage topology) are already in main, so the PR4 changes are purely additive on the provider-dispatch side

If your schedule slips past your own estimate, ping me and I'll take a 1-day port pass on Kimi Files API so we don't block v1.26.0.

Thanks again for the quality of the round-trip — the design review shaped v1.25.0 materially. Looking forward to PR4.

PR #272, opened 2026-07-11
fix: allow LM Studio ingest without API key

rkuzmin:
## Summary
- LM Studio users could pass Test Connection with an empty API key, but Ingest still showed `errorNoApiKey` because `initializeLLMClient()` only exempted `ollama`.
- Centralize the empty-key policy in `allowsEmptyApiKey` (`ollama` / `lmstudio`) and apply it to client init, `llmReady` migration, Test Connection, and `probeLlm`.
- Adds regression tests so ingest-gate and Test Connection cannot drift again.

## Test plan
- [ ] Settings → Provider = LM Studio (Local), leave API Key empty, set model, Save
- [ ] Test Connection succeeds
- [ ] Cmd+P → Ingest current/selected file — no "Please configure API Key first"
- [ ] Same flow with Ollama (empty key) still works
- [ ] OpenAI / Anthropic with empty key still blocked
- [ ] `pnpm lint && npx tsc --noEmit && pnpm test && pnpm build && pnpm css-lint`

Made with [Cursor](https://cursor.com)

PR #288, opened 2026-07-14
fix(related-page): stop persisting raw LLM output on the related path

### LINKED ISSUE #287: updateRelatedPage persists raw LLM output — accumulated Mentions can be destroyed
`updateRelatedPage` is the only write path that persists the model's reply verbatim. It runs none of the deterministic post-processing the other two paths run, and as a result it can silently destroy curated quotes.

I found this while re-ingesting a batch with a local model (Gemma, `wikiLanguage: de`). **A PR with the fix is attached** — I pushed it rather than waiting, because the failure is silent and destructive. Please close it without ceremony if you would rather solve it differently.

## The gap

| post-processing step | `createNewPage` | `mergePage` | `updateRelatedPage` |
|---|---|---|---|
| mentions assembly | ✅ | ✅ | ❌ |
| `canonicalizeSectionHeaders` (#241) | ✅ | ✅ | ❌ |
| `correctRelatedLinkPrefixes` (#187) | ✅ | ✅ | ❌ |

```ts
const cleanedBody = cleanMarkdownResponse(updatedBody);
const finalContent = `${frontmatter}\n\n${cleanedBody}`;
await ctx.createOrUpdateFile(page.path, finalContent);
```

The Mentions section lives in that body. So whether the curated quotes survive depends on whether the model happens to reproduce the section — a deterministic invariant resting on a probabilistic component.

## Impact

#267 established the non-lossy guarantee for re-ingest, but only on the merge path. A page touched as a *related* page — rather than as a directly extracted entity/concept — is unprotected.

Observed on one real ingest of 5 notes:

- `concepts/NF-κB-Signalweg` (2 sources) — **4 curated quotes destroyed**; the rewrite replaced `## Mentions in Source` with prose.
- `entities/Polyphenole` (3 sources) — **1 quote destroyed**.
- Pages touched via the merge path in the same run survived: `assembleFinalContent` unioned them, and where the section was unparseable the #267 fail-safe preserved it verbatim and logged a warning.

The merge path warns on the console when it skips. The related path is silent — it just overwrites.

The two missing correctors have the same shape: on this path a garbled section label stays garbled (and label-exact matching is what Tier-B retrieval needs), and a `sources/`-mis-prefixed link in a Related section is never re-typed.

DocTpoint:
Fixes #287.

`updateRelatedPage` wrote the model's reply straight to disk, running none of the deterministic post-processing the other write paths run. Because the Mentions section lives in that body, a rewrite that failed to reproduce it destroyed every accumulated quote on the page.

## Changes

- Route the body through `assembleFinalContent`, exactly as `mergePage` does — unions accumulated + new mentions, and preserves an unparseable section verbatim via the #267 fail-safe.
- Run `canonicalizeSectionHeaders` (#241) and `correctRelatedLinkPrefixes` (#187), so all three write paths line up.
- Strip the Mentions section from the rewrite prompt — the model cannot drift a format it never sees.

## Tests

Four regression tests, each verified red against the current implementation. Full suite green (2084 tests, `tsc --noEmit` clean).

---

I opened this without waiting for a go-ahead because the failure mode is silent data loss. If you would rather fix it your own way, close this — no hard feelings, the issue stands on its own.

### DIFF
--- a/src/__tests__/wiki/page-factory/related-page.test.ts
+++ b/src/__tests__/wiki/page-factory/related-page.test.ts
@@ -190,3 +190,123 @@ describe('updateRelatedPage — normal path rewrites via LLM', () => {
     ).rejects.toThrow(/LLM client not initialized/);
   });
 });
+
+// Issue #267 established a non-lossy re-ingest on the merge path. This path
+// never had it: the Mentions section lives in the body handed to the LLM, and
+// the reply was persisted verbatim, so a rewrite that dropped the section
+// destroyed every accumulated quote. Observed in the wild with a local model:
+// a 2-source hub lost 4 quotes, a 3-source hub lost 1.
+describe('updateRelatedPage — Mentions are never LLM-owned (#267 parity)', () => {
+  const PAGE_WITH_MENTIONS = [
+    '---',
+    'created: 2026-07-10',
+    'updated: 2026-07-10',
+    'sources:',
+    '  - "[[sources/old]]"',
+    'tags: []',
+    '---',
+    '',
+    '## Description',
+    'Old body.',
+    '',
+    '## Mentions in Source',
+    '',
+    '- "a curated quote from an earlier source" — [[sources/old|old]]',
+  ].join('\n');
+
+  function analysisWithMentions(): SourceAnalysis {
+    return {
+      ...makeAnalysis(PAGE_TITLE),
+      entities: [{
+        name: PAGE_TITLE,
+        type: 'other' as const,
+        summary: 's',
+        mentions_in_source: ['a fresh quote from the new source'],
+        related_entities: [],
+        related_concepts: [],
+      }],
+    };
+  }
+
+  it('preserves accumulated quotes when the LLM rewrite omits the Mentions section', async () => {
+    const ctx = makeCtx({
+      pageContent: PAGE_WITH_MENTIONS,
+      llmResponse: '## Description\nA rewritten body with no mentions section at all.',
+    });
+
+    const result = await updateRelatedPage(
+      ctx,
+      PAGE_TITLE,
+      analysisWithMentions(),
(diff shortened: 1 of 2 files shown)


green-dalii:
Closing #287 via #288.

Real bug + clean two-part analysis (the 3-row write-path coverage table makes the gap concrete). Shipping to v1.25.1 PATCH as priority HIGH — silent data loss is the most expensive class of bug because users don't see it until later.

Ask: rebase onto v1.25.0 + rerun Gate 1 — your 2084 tests claim is pre-v1.25.0; current main is 2182. After rebasing, ping here with the diff and I'll pull it through `code-review` + `simplify` and merge.

Labels: `bug`. Milestone: `v1.25.1 PATCH`.

DocTpoint:
Rebased onto v1.25.0 and re-ran Gate 1.

`upstream/main` currently points at the `1.25.0` tag exactly (no commits after it), so the rebase target is the release itself.

- Rebase was clean. No conflicts, no manual resolution, nothing moved — so there is no rebase diff to send.
- `tsc --noEmit`: 0 errors.
- Test suite: **165 files / 2186 tests passing** (2182 on current main + 4 from this PR).

Ready for review. On the follow-ups you listed (#292, #289, #290): I will send them as separate PRs shortly — details and current timing in #292.

green-dalii:
Rebased onto post-#272 main and merged — thanks @DocTpoint. Squashed into 65b5d1b; Gate 1 green (lint 0/0, tsc 0, 2195 tests, build clean, css-lint 0). The 3-paragraph fix description (strip → canonicalize → assembleFinalContent) is exactly the right structure — it forces a single root-cause change in each step instead of mixing concerns, which is what makes the regression class (silent mentions loss) structurally impossible to recur.

Re #292 / #289 / #290 — happy to review as separate PRs. Shipping in v1.25.1 PATCH alongside PR #272.

PR #304, opened 2026-07-19
fix(log): split created from updated pages by the pre-write existence check (#290)

### LINKED ISSUE #290: Ingest log lists already-existing pages under "Created pages"
`log.md` lists every page an ingest *wrote* under **Created pages** / **Erstellte Seiten**, including pages that already existed and were merged into.

Example from my log — a Curcumin ingest listed `concepts/Oxidativer-Stress.md` and `concepts/NF-κB-Signalweg.md` under "created". Both already existed, with 2–3 sources each. They were merges, not creations.

## Why it matters

The log overstates creations and hides merges. While auditing a batch, the log told me 45 pages had been created; the real split was 27 created and 18 merged.

That distinction is exactly the one that matters for review, because merges are where existing content can be lost — creations have nothing to lose. A log that reports every write as a creation makes the risky half of an ingest invisible.

## Suggestion

Count a page as created only when it did not exist before the write, and list the rest under "Updated pages".

DocTpoint:
Closes #290.

`createOrUpdatePage` already performs the existence check that decides between the create and the merge path, so the information was there and simply never surfaced. `PageCreationResult` now carries `created`, set from that same check, and the ingest callers route the path into `created_pages` or `updated_pages` accordingly — derived from what actually happened at the write, not from what the caller intended.

**On the i18n split you asked for: it turned out not to be needed.** The log emitter already writes both lines with localized labels from `TEXTS.logLabels`, and the log parser already has `UPDATED_RE` with all the locale variants. All 10 locales carry `updatedPages`. The "Updated pages" line was simply always empty for entity and concept pages, because every write was routed to `created_pages`. So this PR touches no locale file, and no header is hardcoded.

**One deliberate remainder.** The source summary page is still always counted as created, including on re-ingest when it already exists. Fixing it is the same one-line existence check — `createSummaryPage` already reads the file at `wiki-engine.ts:1256` — but it returns a bare `string`, and changing that signature churns three call sites in an unrelated test file. That is one page per ingest against the dozens this PR corrects, so I left the call to you rather than widening the diff. Say the word and I will add it here or as a follow-up.

**On the conversation path.** The PR also asserts both page lists right after the conversation-ingest JSON parse. They are bookkeeping we own, not content the model reports, and until now nothing pushed to `updated_pages` there — routing to it would have turned a model that omits the key into a `TypeError` several stages later, after pages had already been written to disk. The normal ingest path is unaffected: `batch-merger` constructs both arrays.

Five tests on the routing contract, three of them red when the result reports every write as a creation.

**Gate 2 (side effects):** One added field, set at all five return sites of the router. Six call sites updated; no other consumer of `PageCreationResult` exists. No IO, no new error paths — the conversation-path assertion removes one.

**Gate 3 (breaking changes):** `PageCreationResult` gains a required field. It is an internal type with no external implementers; every construction site is in `create-page.ts` and updated here. Log format, settings, commands and Obsidian API surface unchanged. The log gains entries under a header it already emitted.

**Gate 4 (performance):** No additional IO. The existence check is the one the router already performed; the field only carries its result outward.

**Gate 1 on current `main` (5e13ac5):** `pnpm lint` 0/0, `npx tsc --noEmit` 0, `pnpm test` 2203 → 2208, `pnpm build` clean, `pnpm css-lint` 0 violations.

Diff is 115 insertions across 5 files.

### DIFF
--- a/src/__tests__/wiki/page-factory/create-page.test.ts
+++ b/src/__tests__/wiki/page-factory/create-page.test.ts
@@ -212,3 +212,80 @@ describe('createNewPage — wraps errors with entity context', () => {
     ).rejects.toThrow(/Failed to create entity page "Karpathy"/);
   });
 });
+// ─── #290: the result reports create-vs-update ───────────────────────────
+//
+// The ingest log listed every written page under "Created pages", including
+// pages that already existed and were merged into. That hid the risky half of
+// an ingest: merges are where existing content can be lost, creations have
+// nothing to lose. The router already performs the pre-write existence check
+// that decides create-vs-merge, so it reports what actually happened rather
+// than what the caller intended.
+
+describe('#290 — PageCreationResult.created reflects the pre-write existence check', () => {
+  it('reports created:true when the page did not exist', async () => {
+    const ctx = makeCtx({ llmResponse: '## Description\nNew entity.' });
+    const result = await createOrUpdateEntityPage(
+      ctx,
+      createMockEntity({ name: 'NewEntity' }),
+      EMPTY_ANALYSIS,
+      { path: 'notes/article.md', basename: 'article.md' },
+    );
+    expect(result.created).toBe(true);
+  });
+
+  it('reports created:false when merging into a page that already existed', async () => {
+    const ctx = makeCtx({
+      files: { 'wiki/entities/X.md': EXISTING_FM },
+      llmResponse: '## Description\nMerged body.',
+    });
+    const result = await createOrUpdateEntityPage(
+      ctx,
+      createMockEntity({ name: 'X' }),
+      EMPTY_ANALYSIS,
+      { path: 'notes/article.md', basename: 'article.md' },
+    );
+    expect(result.path).toBe('wiki/entities/X.md');
+    expect(result.created).toBe(false);
+  });
+
+  it('reports created:false when appending to an existing reviewed page', async () => {
+    const reviewedContent = `---\ncreated: 2026-07-10\nupdated: 2026-07-10\nsources:\n  - "[[existing]]"\ntags: []\nreviewed: true\n---\n\n## Curated\nLocked.\n`;
+    const ctx = makeCtx({
+      files: { 'wiki/entities/X.md': reviewedContent },
+      llmResponse: '## New\nLLM body.',
+    });
+    const result = await createOrUpdateEntityPage(
+      ctx,
+      createMockEntity({ name: 'X' }),
+      EMPTY_ANALYSIS,
+      { path: 'notes/article.md', basename: 'article.md' },
+    );
+    expect(result.created).toBe(false);
+  });
+
+  it('reports created:false for the empty-name guard, which writes nothing', async () => {
+    const ctx = makeCtx({});
+    const result = await createOrUpdateEntityPage(
+      ctx,
+      createMockEntity({ name: '   ' }),
+      EMPTY_ANALYSIS,
+      { path: 'notes/article.md', basename: 'article.md' },
+    );
+    expect(result.path).toBeNull();
+    expect(result.created).toBe(false);
+  });
+
+  it('reports created:false for a concept merged into an existing page', async () => {
+    const ctx = makeCtx({
+      files: { 'wiki/concepts/Y.md': EXISTING_FM },
+      llmResponse: '## Description\nMerged concept.',
+    });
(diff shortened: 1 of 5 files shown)


green-dalii:
Thanks for the work, @DocTpoint — both the design and the field data behind it. I went through the diff and the audit, and I want to flag where this PR stands against the current `main`.

**Independent audit (verdict: READY_WITH_NOTES)**
- Gate 1: lint 0/0, tsc 0, build clean, css-lint 0.
- Gate 2 (side effects): clean — single required field on internal `PageCreationResult`, all 5 construction sites + 6 consumer sites updated, no IO, no new error paths. The `parsed.created_pages ?? []` defensive default in `conversation-ingest.ts` actually closes a latent TypeError path (model omits the key → crash several stages later, after pages already on disk). Good catch.
- Gate 3 (breaking): none — internal type, tsc 0 proves all sites updated. Settings / command IDs / log format / Obsidian API surface unchanged. The log gains entries under a header it already emitted (`updatedPages` was already in all 10 locale `TEXTS.logLabels` and `UPDATED_RE`).
- Gate 4 (performance): N/A — O(1) per call site, no new loops.
- Claim audit: i18n-split-unneeded is correct (verified all 10 locales); 5-tests-red verified by content-mutation assertions on `result.created`.

**Why this isn't going into v1.25.1**

When I tried to cherry-pick onto current `main` (post-#309 wiki-engine Phase C-PR1), `src/wiki/wiki-engine.ts` has a content conflict: PR #309 introduced `runBatchedWithRetry<T>` for the page-generation phase, which moved entity/concept handling into a single `execute` closure. Your patch is on the pre-#309 shape and updates 4 `analysis!.created_pages.push(...)` sites inside the old for-loop; on the new shape those push sites live inside `runBatchedWithRetry.execute`.

The fix is mechanical (4 lines, the same `(entityResult.created ? created_pages : updated_pages).push(...)` ternary inside the closure), but the patch is yours to write — that's contributor-owned work, and rebase + a fresh Gate 1 on current `main` is the right path.

**What I'd ask for the rebase**
1. `git fetch origin main && git rebase origin/main` (or merge — your call) on `pr/290-log-create-update-split`.
2. Resolve the `wiki-engine.ts` conflict: 4 `analysis!.created_pages.push(entityResult.path)` calls inside `runBatchedWithRetry.execute` (entity + concept + their retries) → ternary on `entityResult.created`.
3. Re-run Gate 1 against current `main` (2274 tests is the new baseline). I'll do an independent re-audit on the rebased diff before merge.
4. The summary-page deliberate remainder (`wiki-engine.ts:875` and `conversation-ingest.ts:232` unconditional push) — your call. I'd lean toward leaving it for a follow-up since changing `createSummaryPage`'s signature churns 3 call sites in test files, but I can do that as a sibling commit in the same PR if you'd rather close it now.

**Target: v1.25.2 PATCH.** v1.25.1 PATCH is targeted to ship with #303 and #302 only (legacy Mentions parse, schema-section drop protection) — both merged to `main` today. #290/#304 is a logging visibility improvement that doesn't block the write-path fixes, so it sits cleanly in the next PATCH.

No pressure on the timing — when you push the rebased PR I'll review and merge. If you'd rather hand off the conflict resolution to me, I can take it as a follow-up commit on a separate branch and you review; just say so and I'll set that up. Otherwise this is yours to drive.

DocTpoint:
Rebased onto current `main` (`509b214`). Gate 1 against that baseline: lint 0/0, tsc 0, build clean, css-lint 0, **2279 tests passing — baseline `main` is 2274, so the delta is exactly the 5 tests this PR adds.**

One note on the conflict resolution, since it differs from your description: it came out as **two** call sites, not four. `runBatchedWithRetry` re-invokes the same `execute` closure for the retry, so the entity and concept retry paths no longer carry their own `created_pages.push(...)` — the ternary in `execute` covers both the first attempt and the retry. The four sites of the pre-#309 for-loop collapsed into two under your refactor. Nothing was dropped; I'm flagging it so the re-audit doesn't look for two missing lines.

On the summary page: I've left it as the deliberate remainder, as you suggested. `wiki-engine.ts:875` and `conversation-ingest.ts:232` still push unconditionally, so a re-ingest that rewrites an existing summary page still logs as created. Happy to take it in a follow-up rather than churn `createSummaryPage`'s signature across three test files here.

Ready for your re-audit.

Issue #168, opened 2026-06-21
'auto' granularity setting - maps granularity to file size based on an equation


Indexed-Apogrypha:
**Is your feature request related to a problem? Please describe.**
yes. tiny files and huge files should be processed with different levels of granularity. If both of these files were processed as a batch, we would process at least one of these files with the wrong settings.

**Describe the solution you'd like**
granularity is determined on a file-by-file basis based on the size of the file.

**Describe alternatives you've considered**
granularity is determined on a file-by-file basis based on the size, type, or source of the file.

**Additional context**
we would need to determine an equation that maps file size to the 'minimum number of key entities'. I propose using a logarithmic relationship with a minimum value of 1 (the source file):

_minimum_knowledge_primitives_ =celieng(    a*log( _file_size_bytes_)  +  b    ); where a and b are constants such that _minimum_knowledge_primitives_ >= 1

Indexed-Apogrypha:
this could be added to the granularity dropdown in the plugin settings

DocTpoint:
Information density isn't necessarily correlated with text length. A short note can be very dense, and a long one can be mostly padding — so file size is a weak proxy for how many knowledge primitives a note should yield.

I started out with exactly this concern. Building a wiki from a mix of short and long notes, I wanted to stop the plugin from spawning tons of stubs out of a single short note. But I found the problem largely solves itself as the wiki grows: once there's enough existing context, the plugin's choice of how many pages to create becomes quite reasonable on its own. It links into what already exists instead of minting new stubs, and it scales granularity to the actual complexity of the material rather than to the byte count.

So in the end I decided to rebuild the wiki from scratch without any restriction, and that worked better. You don't want to build a wiki around artefacts that a restriction forced into existence — the constraint itself distorts what the graph would naturally become.

The only place where size genuinely matters is the degenerate one — a near-empty file shouldn't be forced to produce primitives at all (see [#164](https://github.com/green-dalii/obsidian-llm-wiki/issues/164)). But that's a floor against hallucination, not a granularity curve.

green-dalii:
Good pushback. After discussion, we're going with **two independent settings** (`singleFileGranularity` + `batchGranularity`) instead of an auto-curve. Rationale:

1. The single/folder asymmetry is real — single-file ingest is pre-vetted by user choice (expects depth), folder ingest is bulk (expects noise reduction). Two dropdowns express that directly without implicit defaults.
[...]

PR #320, opened 2026-07-20
fix(analyzer): route truncated responses into the halving retry (#305)

### LINKED ISSUE #305: Truncated batch responses bypass the halving retry: the batch is dropped and the partial result looks like a success
`analyzeSource` has a truncation retry that halves the batch size (`source-analyzer.ts:477`). It only runs inside the `catch` block, so it requires the provider call to **throw** with `truncated` or `max_tokens` in the message.

With an OpenAI-compatible provider (LM Studio here), a truncated response is not an error. It arrives as HTTP 200 with `finish_reason: "length"` and a body that stops mid-sentence. The flow is then:

1. `parseJsonResponse` fails (the JSON is incomplete, not malformed)
2. `analysisData` is `null`
3. `source-analyzer.ts:353` → `break`

The `catch` block is never reached, so the halving retry never runs.

## Observed

Ingesting one 10.5 KB German note, `standard` granularity, local `gemma-4-26b-a4b-qat`:

```
JSON parse completely failed (length 77436)
first 200 chars after normalization: {  "source_title": "Protein-Fehlfaltung", ...
last 200 chars after normalization: ... die Dissoziation in Monomere und die Aggregation im physiologischen Milieu spontan erfolgen
[Batch 2] JSON parse failed, skipping batch
```

The last 200 characters end mid-sentence, with no closing braces — the response was cut at the token limit. `Truncation detected, halving batch size` never appears in the log.

This reproduced on two separate runs, then succeeded unchanged on a third, so the trigger itself is stochastic. The retry gap is not: it is structural.

## Two consequences

**1. The repair callback cannot help here.** It asks the model to *"Only fix JSON syntax errors ... Do NOT change any values or content"* (`source-analyzer.ts:342`). That is the right instruction for malformed JSON, but truncated JSON is missing content, which no syntax repair can restore. The repair call also has to re-emit the full response, so it tends to hit the same limit again.

**2. The partial result is written and reads as a complete ingest.** With `break` on a non-first batch, the accumulated items from batch 1 are kept — which is reasonable in itself. But the source summary page is then created from that partial extraction. In our vault the note produced a 1989-byte source page with 1 derived page and no Mentions section; a successful run of the same note later produced 3065 bytes and 19 derived pages. Nothing downstream marks the page as partial, so a coverage check based on page existence reports the note as fully ingested.

## Suggested direction

Detect truncation where it actually surfaces — in the `!analysisData` branch — and route it into the existing halving retry instead of `break`. The signal is available without new plumbing: `finish_reason === 'length'`, or a response length at/near the effective `max_tokens`.

Optionally, marking a partially-extracted source page (frontmatter flag or log line) would let users see that a note needs a re-run. That is a separate design decision and I have no strong opinion on the shape.

Happy to open a PR for the retry routing if you want it — your call on the design.

DocTpoint:
Fixes #305.

## The signal was already arriving

The halving retry lived only in the `catch` block, so it required the provider call to throw. An OpenAI-compatible provider does not throw on truncation — it returns HTTP 200 with a body that stops mid-token. The batch then failed to parse and fell through to `break`, or to `return null` on the first batch.

What I did not expect when I opened the issue: the information needed to tell "finished" from "ran out of tokens" was already in the process. The AI SDK normalizes the provider's `finish_reason` into `result.finishReason`, and all three SDK-backed clients discarded it on the next line:

```ts
const result = await generateText({ ... });
return result.text;              // openai-compat:159, :186, :217
```

`finish_reason` appears nowhere in `src/` today, so no caller in the plugin can currently see why a generation ended. That is a wider gap than #305 — truncation is invisible everywhere, not just in the analyzer — but this PR only wires the one path the issue is about.

## What changed

- `LLMFinishReason` + an optional `onFinish` callback on the `createMessage` params (`types.ts`). **Additive on purpose:** the return type stays `Promise<string>`, so every existing caller and every mock client is untouched. Changing the return type would have reached every call site in the plugin, which felt wrong for a PATCH.
- The three SDK clients report it at every return site through one shared normalizer (`llm-sdk/finish-reason.ts`). `normalizeFinishReason` is total — anything unrecognized becomes `'unknown'`, so an unfamiliar value can never be mistaken for `'length'`. A client that never calls `onFinish` leaves the analyzer at `'unknown'`, i.e. exactly the pre-#305 behavior.
- In the analyzer's parse-failure branch: when the provider reported `length`, halve the batch and re-run instead of dropping it. The JSON-repair callback cannot help here — truncated JSON is missing content, not malformed syntax, and the repair call re-hits the same limit trying to re-emit everything.
- The halving block is hoisted out of the `catch` into a closure both paths call, as you suggested. The closure returns whether to re-run; the caller keeps the loop control (`batchNum--; continue;`) since a closure cannot `continue`.

## Two deviations from your suggested fix path, flagged deliberately

**1. I did not implement the response-length heuristic.** You suggested capturing raw length and comparing it against `maxTokensPerCall`. I left that out: the response is measured in characters and the cap in tokens, and the ratio between them varies by language — a German vault and a Chinese one would need different thresholds. That would be a tunable constant to maintain, which is the same trade you declined on the auto-curve in #168. `finishReason` is the provider's own statement and needs no constant, so I used it alone. Happy to add the heuristic as a fallback if you want coverage for providers that report nothing.

**2. The retry now also covers the first batch.** Your step 2 spoke of routing the parse-failure branch into the retry without distinguishing batches. On the first batch the current alternative is `return null`, i.e. the whole ingest fails, so one bounded attempt is strictly better. `retryingBatch` and `minBatchSize` bound it to a single extra call either way. Say the word if you want it narrowed to later batches.

## Scope left out

The second half of the issue — marking a partially-extracted source page so a coverage check can see it needs a re-run — is not in this PR. That is a design decision about frontmatter or logging, and I still have no strong opinion on the shape.

The streaming path (`createMessageStream`) is untouched; the analyzer does not use it.

## Tests

17 new tests. Four drive the analyzer through a client that reports `finishReason`, covering: truncation retried with a halved batch; a complete-but-malformed response *not* retried; a client that reports nothing behaving exactly as before; and the halving firing only once. The remaining 13 cover `normalizeFinishReason` against known and unrecognized inputs.

Gate 1 on `3578a9d`: lint 0/0, tsc 0, build clean, css-lint 0, **2296 tests passing — baseline `upstream/main` measured separately at 2279, so the delta is exactly the 17 tests added here.**

### DIFF
--- a/src/__tests__/wiki/source-analyzer-truncation-retry.test.ts
+++ b/src/__tests__/wiki/source-analyzer-truncation-retry.test.ts
@@ -1,0 +1,180 @@
+import { describe, it, expect, vi } from 'vitest';
+import { createMockContext, createMockFile } from '../__support__/engine-context';
+import { SourceAnalyzer } from '../../wiki/source-analyzer';
+import { normalizeFinishReason } from '../../llm-sdk/finish-reason';
+import type { LLMClient, LLMFinishReason } from '../../types';
+import { TFile } from 'obsidian';
+
+// Issue #305 — a truncated batch response is not an error.
+//
+// An OpenAI-compatible provider that hits the token limit returns HTTP 200
+// with a body that stops mid-token. The call does not throw, so the halving
+// retry that lived in the `catch` block never ran: `parseJsonResponse`
+// returned null and the batch was dropped silently (`break`), or the whole
+// ingest failed (`return null` on the first batch).
+//
+// These tests drive the analyzer through a client that reports
+// `finishReason: 'length'`, which is what the SDK normalizes the provider's
+// `finish_reason` field into.
+
+const TEST_PATH = 'sources/truncation.md';
+
+/** A response body cut off mid-token: valid JSON prefix, no closing braces. */
+const TRUNCATED = '{"source_title": "Protein misfolding", "entities": [{"name": "Amyloid", "summary": "Aggregation occurs spontane';
+
+const VALID = JSON.stringify({
+  source_title: 'Protein misfolding',
+  summary: 'A note about protein misfolding.',
+  entities: [{ name: 'Amyloid', type: 'other', summary: 'An aggregate', mentions_in_source: [] }],
+});
+
+interface ScriptedReply {
+  text: string;
+  /** Omitted for clients/paths that report nothing — the pre-#305 situation. */
+  finishReason?: LLMFinishReason;
+}
+
+/**
+ * Client that replays a script of *extraction* replies and reports
+ * `finishReason` through the optional `onFinish` out-channel, the way the
+ * SDK-backed clients do.
+ *
+ * `parseJsonResponse` may issue a JSON-repair call on the same client. Those
+ * are identified by their prompt and answered by echoing the broken text
+ * back unchanged — a repair cannot reconstruct content that was never
+ * emitted (that is the point of #305), and they carry no finish reason
+ * because the repair call site does not pass `onFinish`. Keeping them out of
+ * the script means the script indexes extraction attempts only.
+ */
+function scriptedClient(replies: ScriptedReply[]): {
+  client: LLMClient;
+  prompts: string[];
+} {
+  const prompts: string[] = [];
+  let idx = 0;
+  let lastText = '';
+  const client: LLMClient = {
+    createMessage: async (params) => {
+      const first = params.messages[0];
+      const prompt = typeof first.content === 'string' ? first.content : '';
+      if (prompt.startsWith('Fix the following malformed JSON')) return lastText;
+      prompts.push(prompt);
+      const reply = replies[idx++];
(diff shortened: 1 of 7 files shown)


green-dalii:
Closed by maintainer cherry-pick (v1.25.2 PATCH scope).

Your commit was preserved as the squash-style cherry-pick on `main`:
- `2e2bb78` fix(analyzer): route truncated responses into the halving retry (#305) — @DocTpoint (author preserved)

2296/2296 tests pass (Gate 1 clean). The 4 cleanup suggestions (LLMFinishReason placement, KNOWN_FINISH_REASONS as const, scriptedClient merge, `finish: { reason }` → bool) are tracked as a maintainer follow-up PR — your authorship for the core fix is preserved.

Thanks for the contribution — closing as merged.

PR #273, opened 2026-07-11
feat: add ChatGPT Plan Codex OAuth provider

CCRRAY-DESIGN:
## Summary

- add an experimental `ChatGPT Plan (Codex OAuth)` provider alongside the existing OpenAI API-key provider
- support browser PKCE and device-code authorization with credentials stored only in Obsidian SecretStorage
- send authenticated streaming Responses requests through the Codex backend without changing the OpenAI Platform path
- synchronize original model slugs from the signed-in account catalog without alias mapping, and prune models that the Responses endpoint reports unavailable
- add localized settings UI, documentation, migration handling, and lifecycle/security coverage

## Why

ChatGPT subscriptions and OpenAI Platform API billing are separate. This contribution lets users explicitly choose the experimental Codex OAuth path and use the allowance available to their ChatGPT plan, similar to other Codex-compatible clients, while keeping the existing API-key provider unchanged.

The implementation is deliberately marked experimental because availability remains controlled by OpenAI authentication, account, model, and allowance policies. It does not claim an OpenAI partnership or expose a general ChatGPT API.

## Security and compatibility

- OAuth uses PKCE S256, state validation, bounded polling, cancellation, and a localhost-only callback server
- access, refresh, ID tokens, and account IDs are never written to ordinary plugin settings
- cached model metadata is sanitized and bound to the current account inside SecretStorage
- model identifiers are used exactly as returned by the official account catalog; no model-name mapping is maintained
- authentication retries are limited to one refresh/replay, and Codex Responses requests use a single transport

## Validation

- `npm test -- --run` — 145 files, 2014 tests passed
- `npm run lint`
- `npx tsc --noEmit`
- `npm run build`
- `npm run css-lint`
- `git diff --check origin/main..HEAD`
- live Obsidian verification on macOS: browser/device authorization, account model refresh, unavailable-model fallback, and successful `gpt-5.6-sol` connection

## Documentation

The behavior and experimental status are documented in the README translations, CHANGELOG, CLAUDE guide, and `docs/contributions/openai-codex-oauth-proposal.md`.

PR #322, opened 2026-07-21
fix(lint): match dead-link targets against slug-normalized titles and aliases (#308)

### LINKED ISSUE #308: findDeadLinkTarget compares a slugified link target against unslugified titles and aliases
The pre-check in `fixDeadLink` resolves a dead link before any LLM call. It uses `findDeadLinkTarget` (`src/core/dead-link-detector.ts:29-41`):

```js
const targetBasename = targetName.includes('/') ? targetName.split('/').pop()! : targetName;
let match = pages.find(p => p.title.toLowerCase() === lowerTarget);
if (!match) match = pages.find(p => p.aliases?.some(a => a.toLowerCase() === lowerTarget));
```

The link target comes from a wiki link, so it is slugified: `Systemische-Inflammation`. Titles and aliases are not: `Systemische Inflammation`. `toLowerCase()` alone leaves a hyphen different from a space, so the comparison fails on a pair that refers to the same page.

**The same file already does this correctly one branch later.** The safety net before stub creation (`src/wiki/lint/fix-dead-link.ts:188-196`) compares both forms:

```js
slugify(p.title).toLowerCase() === safetySlug ||
p.aliases?.some(a => slugify(a).toLowerCase() === safetySlug)
```

`slugify` is already imported in that file. The pre-check does not use it.

**Measured effect** (one vault, 2789 pages, 417 source notes, German content, v1.24.1, 979 unresolved link targets):

| matching | targets resolved | links resolved |
|---|---:|---:|
| current: `toLowerCase()` only | 186 | 577 |
| `slugify()` on both sides | 361 | 1005 |

Examples that only the second form resolves: `[[concepts/Systemische-Inflammation]]` (22 links) to the alias `Systemische Inflammation`, `[[entities/Nervus-vagus]]` (7) to the alias `Nervus vagus`, `[[concepts/Kurzkettige-Fettsäuren]]` (7) to the alias `Kurzkettige Fettsäuren`.

**Two consequences beyond the miss itself.** Every target the pre-check misses costs an LLM round trip that a string comparison could have answered. And the safety net only guards the `create_stub` branch: if the model returns `action: "correct"` with a wrong link, nothing re-checks it.

**Caveats.** One vault, one model, German content — compound nouns and umlauts may make this more visible than in other languages. I approximated `slugify()` in the measurement script with lowercase plus whitespace-to-hyphen, so the exact gain from the real implementation may differ somewhat. The direction should hold.

I am happy to open a PR if you would like one.

DocTpoint:
Closes #308.

Branched from current `main` (`54bc128`) rather than `3578a9d` — `main` moved with the Codex OAuth merge (#273) while the issue was being triaged. No overlap with that change.

## What changed

`findDeadLinkTarget` (`src/core/dead-link-detector.ts`) gets two more lookup tiers: title and alias again, this time slug-normalized on both sides. Everything else in the function is untouched.

## Three deviations from the sketch in the issue thread

**1. `computeSlug` instead of `slugify`.** The safety net at `fix-dead-link.ts:190-196` runs on one string; the pre-check runs over every page in the vault. `slugify` logs a `console.warn` on empty input, and the comment at `slug.ts:12-16` marks `computeSlug` as the variant intended for exactly this ("matching 2141 existing pages"). It also lowercases already, so the comparison stays independent of the user's `slugCase` setting. Empty titles and aliases are mapped to `''` rather than to `computeSlug`'s `'untitled'` placeholder — otherwise a page with an empty alias would match a link to "untitled".

**2. Four tiers instead of one combined `find()`.** The safety net evaluates all four conditions inside a single `find()`, so the first *page* wins regardless of match quality. In the pre-check that would let a slug alias match on page B overtake an exact title match on page A. The slug tiers therefore run only after both exact tiers. There is a test for that ordering.

**3. One existing test asserted the defect and had to change.** `handles path in target (extracts basename)` expected `concepts/attention-mechanism` *not* to resolve to the page "Attention Mechanism" — that is the German hyphen-vs-space case in English clothing. Its expectation is updated; a second test keeps basename extraction itself covered. Flagging this because the issue estimated "4 lines + 1 test", and a reviewer would otherwise meet a changed assertion without explanation.

## Tests

A parameterized suite over `(link_target, expected_page_title)` as requested, including the German cases directly: `nervus-vagus` → "Nervus vagus", `kurzkettige-fettsäuren` → "Kurzkettige Fettsäuren" (umlaut survives), plus case variants. A second parameterized block asserts that near misses (`nervus-vagus-dorsalis`, `systemische`) stay unresolved, and single tests cover tier order and the two empty-string cases.

## Re-measured with this implementation

The issue declared a caveat: the measurement script approximated `slugify()` in Python. This run uses the real function on the reporting vault (2796 pages, German content). Same vault, but after a link-repair pass since the issue was filed, so the absolute numbers are lower than the ones in the issue text.

| pre-check matching | targets resolved | links resolved |
|---|---:|---:|
| before (`toLowerCase`) | 74 | 280 |
| after (slug-normalized) | 245 | 727 |

752 unresolved targets / 1507 links in total. **No target that resolved before resolves to a different page now** — the change is additive over the previous behavior; the script checked every target for that explicitly.

Examples that only the new tiers resolve: `concepts/Systemische-Inflammation` (22 links) via an alias, `entities/Nervus-vagus` (7) via the title, `entities/Akkermansia muciniphila` (5) — a link that carries a space where the page name carries a hyphen.

## Gate 1

`pnpm lint --max-warnings=0` 0 · `npx tsc --noEmit` 0 · `pnpm build` clean · `pnpm css-lint` 0 violations · `pnpm test` **2468 passed / 185 files**.

Baseline counted separately on `54bc128`: **2453 / 185** → +15, which is exactly the number of tests added here. One note on the baseline: `openai-codex-loopback-flow.test.ts` asserts against the built `main.js`, so a `pnpm build` has to precede `pnpm test` on a fresh checkout, otherwise that test fails on a stale bundle.

Diff: 2 files, +124/−2.

### DIFF
--- a/src/core/dead-link-detector.test.ts
+++ b/src/core/dead-link-detector.test.ts
@@ -41,8 +41,16 @@ describe('Dead Link Detector — Pure Functions', () => {
     });
 
     it('handles path in target (extracts basename)', () => {
+      // #308: the basename is still extracted; it now also resolves, because
+      // "attention-mechanism" and "Attention Mechanism" are the slugified and
+      // the unslugified form of the same name. Before #308 this returned
+      // undefined and the caller paid an LLM round trip for it.
       const result = findDeadLinkTarget(mockPages, 'concepts/attention-mechanism');
-      // Should extract "attention-mechanism" and not match any title
+      expect(result?.title).toBe('Attention Mechanism');
+    });
+
+    it('extracts the basename and does not match the folder part', () => {
+      const result = findDeadLinkTarget(mockPages, 'attention-mechanism/nonexistent');
       expect(result).toBeUndefined();
     });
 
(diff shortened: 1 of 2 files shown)


green-dalii:
Cherry-picked to main (292300b). Author: @DocTpoint. See #322 review for code-review + simplify assessment.

Issue #220, opened 2026-07-01
Source-revision awareness for merge: distinguishing self-updates from cross-source conflicts


DocTpoint:
Related to #216 — splitting this out as its own issue since it's a distinct proposal, not just an addition to the triage-routing discussion there.

Building on the dispatch trace posted in #216 (`createOrUpdatePage` branches purely on page state, not provenance) — I'd like to propose a concrete shape for closing that gap, split into pieces of clearly different difficulty and risk.

**Motivation**

The order-invariance argument for the structure layer (link topology, hub status) rests on treating the graph as a function of final content, not build chronology — and that holds up well, since topology is fully reconstructible from the end state regardless of order. The content layer doesn't have an equivalent: the merge prompts (`mergeEntityPage`/`mergeConceptPage`/`mergeAnalysis`, `merge.ts:48`) default to *"Existing content takes priority over new information (unless the new information is clearly more accurate)"* — a pure content judgment, with no notion of one source revising itself, and no signal from the `updated` dates both notes and pages already carry.

This surfaced from an ordinary editorial act: revising an existing source note to reflect newer information, and noticing the wiki page has no path to treat that revision as a revision rather than a new competing claim.

**Proposed shape — four tiers, increasing difficulty and decreasing confidence**

| Tier | Case | Proposed handling | Confidence |
|---|---|---|---|
| 0 | Same source note re-ingested, content changed | Deterministic: compare a content fingerprint of the source against what's already attributed to it in the page's Mentions block. If changed, treat as **replace this source's own block**, not as a new claim to merge/contradict. No new information needed — the source path already identifies self-revision. | High — purely structural, no LLM judgment involved |
| 1 | Automatic detection misses it (note renamed/split/etc.) | Optional frontmatter flag on the note (e.g. `supersedes: true`) as an explicit editorial signal that this edit corrects rather than extends. | High — user-declared, no inference needed |
| 2 | Two *different* source notes disagree on the same claim | Left open. Recency alone isn't evidence of correctness — a single revised note shouldn't casually override content several other sources corroborate. Not proposing an automatic rule here; genuinely asking whether one exists that isn't just "newest wins." | Open — no proposal, just flagging the gap |
[...]

Issue #216, opened 2026-06-30
Wire up the unused mergeAnalysis prompt as a classify-then-route merge triage


DocTpoint:
**Problem**

The page-merge path (`page-factory.ts:408`) always routes to `mergeConceptPage` / `mergeEntityPage`, whose `DESCRIPTION` rule is a *regeneration* instruction — "Integrate new understanding coherently with existing", "merge existing + new". Regeneration is intrinsically lossy for weaker models: they re-summarize the Description and silently drop detail.

**Measurement.** Same merge prompt, same lean existing page, two models:
- **Weak (local ~26B):** dropped a clinical clause from the existing Description (content *loss*, not addition) and fabricated two "verbatim" Mentions quotes — one was a table heading, one a doubled sentence fragment.
- **Strong:** preserved the existing prose verbatim and appended genuine quotes, as instructed.

So the full-rewrite framing sets a high capability bar. But the framing is a lever, not a fixed cost.

**Observation: the safe pattern already exists, half-built**

- `mergeAnalysis` (`merge.ts:5`) is a structured triage — it classifies each new piece as `new / duplicate / complementary / contradictory`, names the target section, and is explicitly forbidden from rewriting existing content. **It is never called anywhere in `src/`.**
- `appendToReviewedPage` already demonstrates the strict-preserve pattern ("existing AUTHORITATIVE, preserve exactly") — but only for `reviewed: true` pages.

**Proposal: classify → route**

1. **Classify** new info against existing (the existing `mergeAnalysis` analyzer; cheap, structured, low-risk).
2. **Route** by class:
   - `duplicate` / no-new → deterministic frontmatter-only update (this is the #131 Tier-1 early-return, already shipped — generalized).
   - `complementary` → targeted append into the named section, **not** a whole-page regeneration. Content-preserving by construction; low capability bar. (This is the unbuilt #131 **Tier-2** "deterministic append".)
   - `contradictory` → the one branch that genuinely needs careful reasoning ("preserve both with attribution") → the only place a strong model is required.

**Effect:** the Description is never blindly regenerated (corruption surface removed), and the expensive/strong path fires only on the contradictory minority (cost localized).

**Relationship to other issues**
- Complements #208 (separate models per operation): triage decides *whether* the expensive path fires; per-operation model selection decides *which* model runs when it does. They compose.
[...]

Issue #185, opened 2026-06-22
Propagate source-note frontmatter `aliases:` to the wiki pages they describe


DocTpoint:
### Summary

When a source note carries curated `aliases:` in its frontmatter, those aliases are not propagated to the entity/concept/source pages generated from it. The ingest reads the note body (`source-analyzer.ts`, `vault.read`), but there's no explicit path from a note's `aliases:` to the generated pages' `aliases:`. The LLM sees them in the raw text but incorporates them unreliably.

### Repro

A note `Exekutive Funktionen.md` with frontmatter:

```yaml
aliases: [Exekutive Funktionen, Executive Function, Exekutiv Funktionen, Exekutiven Funktionen, exekutive Funktionen]
```

After ingest, `concepts/Exekutive-Funktionen.md` has `aliases: [Executive Functions, Exekutivfunktionen]` — two LLM-invented forms, **none** of the five curated ones. Subsequent ingests then emit body links like `[[concepts/Exekutiven-Funktionen]]` and `[[concepts/Exekutiv-Funktionen]]` (inflection variants that *are* in the note's alias list) → dead links, because the canonical page doesn't carry those aliases.

### Language dependence (why this is easy to miss)

Severity scales with morphological complexity. In inflected languages, one concept surfaces in many grammatical forms in running text, and the model writes each link in whatever form the sentence requires — all grammatically correct, all different slugs:

- **German** — adjective declension varies with the determiner: `exekutive Funktionen` (bare) vs. `die **exekutiven** Funktionen` (weak declension after a definite article), plus the compound `Exekutivfunktionen`. The model correctly wrote `Die exekutiven Funktionen …`, which became `[[concepts/Exekutiven-Funktionen]]` → dead, even though `Exekutive-Funktionen` exists.
- **Isolating languages (Chinese)** barely exhibit this — 执行功能 stays 执行功能 regardless of syntactic role, so a concept has essentially one written form.

This means the defect is **least visible in zh/en testing** — yet the plugin ships wiki output for five strongly-inflected languages (de/fr/es/pt/it). It's the same "invisible in the default config" shape as #164 (which large models masked). And it's precisely why *curated* note aliases are the highest-value signal here: a human writing in an inflected language naturally enumerates the inflectional variants (the example note lists `Exekutive / Exekutiven / exekutive Funktionen`), whereas LLM-generated aliases tend to give the lemma plus a translation and miss the inflectional spread.

### Why this is high-leverage

[...]

Issue #330, opened 2026-07-22
The schema cannot fix what the ingest path never decides — a structural companion to #328


DocTpoint:
#328 asks where the line runs between user knowledge, runtime parameters and engine facts. That line is worth drawing. But it will not fix the problems that motivated it, because those problems are not in the schema layer. They are in the ingest path, and the schema has no way to reach them.

One data point first, because it is the reason this document exists.

I rebuilt my vault from scratch: 417 dense notes, folder ingest, roughly 5000 minutes of local inference, 2781 pages created, unattended. **The result was sobering.** My hope was that the knowledge in those notes would magically crystallise into a usable structure. It did not materialise.

What came out is large, well-formed and densely linked. It is also **worse to ask than the notes it was built from.** Running inference over the original notes gives better answers than running it over the wiki. That is an experience report from daily use, not a benchmark — but it inverts the premise the whole design rests on. The gist this plugin is named after says to ask the synthesised wiki *instead of* the raw sources. In my vault the raw sources win.

Three things happen to knowledge on the way in.

It gets **swallowed**. A truncated analyzer batch returned HTTP 200 and was dropped silently — one call in 24 in a field test, invisible until #305 was fixed. The lemma of a note frequently gets no page at all: `Coenzym-Q10` ends the full run with 41 incoming links and no page to receive them.

It gets **buried**. A statement survives, but under the wrong lemma or behind an identity claim that is not true — sugar filed as sucrose, iron deficiency as iron-deficiency anaemia. And the source that actually owns a topic has no privilege when merging (#312), so a page can open with the framing of a passing mention while the primary source is appended below it.

It gets **scattered**. One note that names a ranked list produced 20 nodes, 16 of them with that single source and nothing else to say. The note held the ranking; the wiki holds the items.

The pattern behind all three: **the note carried the connections, the wiki carries the fragments plus edges that assert only co-occurrence.** The connection was the information. Denser notes lose more, because density is exactly what does not survive being cut into pieces that no relation type reconnects.

That outcome is not a model problem. A stronger model writes better prose into the same structure. It is a design problem, and it sits below the schema layer that #328 proposes to re-cut.

[...]

PR #347, opened 2026-07-25
fix(merge): source ownership in the merge triage (#312)

### LINKED ISSUE #312: Merge triage decides what to keep without knowing where the content came from
## Summary

When a source is merged into an existing page, `mergeAnalysis` decides whether the new
information is redundant, complementary, or requires a rewrite. It makes that decision
without ever being told **which source the information came from**, or **what that source
is about**.

As a result, the note that a page is *about* carries exactly the same weight as a note
that mentions the topic in passing. Whichever source happens to reach the page first
defines it; the topic's own source, if it arrives later, can only append.

The information needed to avoid this already exists. `analyzeSource` produces a `summary`
field for every source (`src/wiki/source-analyzer.ts:91`). It is written to the source
page and then not passed to the triage step.

## What this produces

A concrete case from our vault. `Endotheliale Dysfunktion` (endothelial dysfunction) is an
8.8 KB note that defines the concept, its mechanisms, and its treatment options.

The **source page** generated from that note is accurate and complete. It names eNOS
uncoupling, SIRT1, the Nrf2 pathway, EndMT, and the therapeutic options, and links them
correctly. The model understood the note.

The **concept page** for the same topic contains none of this. It was created three days
earlier from an unrelated note about trans fats. When the note that actually defines the
concept was ingested, triage returned `complementary`, so its content was appended as
bullet points into the existing structure. The page now consists of 16 bullets sourced
mainly from notes about trans fats, pulmonary hypertension, malondialdehyde, and AGEs.

Write audit for that page:

    2026-07-15 21:20  concepts/Endotheliale-Dysfunktion   createNewPage      (unrelated source)
    2026-07-18 02:08  sources/Endotheliale-Dysfunktion    (own note ingested)
    2026-07-18 02:11  concepts/Endotheliale-Dysfunktion   mergePage:skip-or-complementary

The page describing a topic no longer contains that topic's definition, while the source
page beside it does. Nothing in the pipeline can detect this, because nothing in the
pipeline knows that one of those nine sources is the one the page is named after.

A second page, `Kognition` (cognition), shows the same pattern: the note's definition is
absent, and the page opens with a section about a gut bacterium contributed by a different
note.

## Root cause

`mergeAnalysis` (`src/wiki/prompts/merge.ts:30`) receives exactly five inputs:

- `page_name`
- `page_type`
- `existing_content`
- `new_info`
- `section_labels`

There is no source identity and no source summary. The only question the prompt can ask is
"is this new relative to what is already here?" It cannot ask "is this source authoritative
for this page?", because it has nothing to answer that with.

This is not a limitation of the model. The same prompt already asks for genuine editorial
judgement — redundancy, section fit, contradiction. The one input that would make that
judgement sound is computed one step earlier and dropped.

## Scope

In our vault (417 notes, 2374 knowledge pages):

- **160 of 370** notes that have a matching knowledge page saw that page created **before**
  their own source was ingested (43%). These are exposed to the pattern above.
- **817 of 2374** pages (34%) have two or more sources, median 3.

To be explicit about what these numbers are: the 160 measure *exposure*, not confirmed
damage. A later rewrite can restore some content. We verified two cases in detail by
reading them; we did not classify all 160. We attempted an automated text-overlap measure
and discarded it — it scored a known-good page at 0% and a known-damaged page at 43%, so
it measures term reuse rather than whether a definition survived.

## Secondary finding: append position

`updateRelatedPage` (`src/wiki/prompts/merge.ts`, last block) instructs only:

    Update the page by adding the new information without deleting existing content.

No output format, no position. New sections therefore often land at the **top** of the page,
above the description, and stay there permanently. On our cognition page the first thing a
reader sees is an incidental finding about a gut bacterium.

`appendToReviewedPage` gets this right — it specifies "append them in a New Information
section **at the end**" and gives a full output format. But that path only applies to pages
with `reviewed: true`, which is exactly what one must not set on pages that should keep
growing.

## Suggested fix

Two parts, the first one is the substantive one:

1. **Pass the source's `summary` and identity into `mergeAnalysis`**, and let the triage
   weigh relevance rather than novelty alone. No extra model call, no new data — the field
   already exists and is already paid for.

2. **Optionally, a deterministic guard**: when the source title or one of its curated
   aliases matches the page's lemma exactly, that source is authoritative for that page.
   This is a string comparison. It only helps corpora where one note corresponds to one
   topic, and it does nothing at all elsewhere, so it is safe to add.

The `updateRelatedPage` prompt could be given the same output format that
`appendToReviewedPage` already has.

## Caveats

One vault, one model (Gemma 4 26B via LM Studio), German content. The two detailed cases
were read by hand. The 43% figure is chronological exposure derived from frontmatter dates,
not a damage count. Verified against `upstream/main` at 1.25.0; none of our local patches
touch `merge.ts`, `merge-triage.ts`, or `source-analyzer.ts`.

Happy to open a PR for either part.

DocTpoint:
Closes #312.

## Problem

I corrected my own issue text before writing this (comment above): the triage does receive the source name — `buildNewInfoSummary` has written `Source: ${sourceFile.basename}` into `{{new_info}}` since #216. The issue body's "without ever being told which source the information came from" was wrong.

What is missing is narrower: `SourceAnalysis.summary` (what the source **document** is about, `types.ts:37`) is produced by the analyzer and never forwarded, and none of the four strategies is keyed to source at all. The identity is delivered; nothing asks for it to be used.

Four commits, each standing on its own. The last one is the only prompt-text change and can be dropped without touching the rest — see **Separability**.

## Change 1 — related-page output order, part 3 of the issue (`d6ca708`)

`updateRelatedPage` ended with a bare "Output ONLY the updated page BODY content" and no format block. Nothing told the model where new facts belong, so it routinely opened the page with the newest source's material and pushed the page's own definition down. `appendToReviewedPage` already carries the contract this path was missing: existing content keeps its position, new information comes after it.

Prompt text only, no code path touched.

## Change 2 — every `{{page_name}}` occurrence rendered (`3b128c6`)

Found while writing the test for Change 1. `String.prototype.replace` with a string pattern substitutes only the first match, and `{{page_name}}` occurs twice in the template, so every related-page rewrite reached the model as:

```
The new source file ("Omega-3") provides additional information about {{page_name}}:
```

The sentence that names the page's subject was the one carrying the literal placeholder.

Same class elsewhere in the prompt layer, all untouched here:

| template | placeholder | occurrences | rendered |
|---|---|---|---|
| `generateSummaryPage` | `{{date}}` | 3 | 1 |
| `preserveReviewedEntityPage` / `…ConceptPage` | `{{date}}` | 3 | 1 |
| `appendToReviewedPage` | `{{new_source}}` | 3 | 1 |
| `generateEntityPage` | `{{entity_name}}` | 2 | 1 |
| `generateSummaryPage` | `{{source_title}}` | 2 | 1 |

Happy to file that separately rather than widen this PR. `{{section_*}}` is unaffected — `applySectionLabels` already replaces globally.

## Change 3 — deterministic ownership guard, part 2 (`e6d539e`)

When the source's file name, its analyzer title, or one of its curated note aliases slug-matches the page name or one of the page's aliases, a `skip` verdict is routed to the body merge instead — which is the prompt's own stated default whenever the call is not clear-cut. No LLM, no extra call.

Deliberately narrow on three axes:

| axis | choice | why |
|---|---|---|
| strategies overridden | `skip` only | `complementary` already writes the new facts; rerouting it would trade a targeted append for a full rewrite |
| when it fires | only with a `SourceContext` | that is the ingest path; lint-side callers keep their exact previous behaviour (pinned by a test) |
| corpora without lemma-shaped notes | never matches | inert rather than harmful is the intended failure mode |

Plumbing: `createOrUpdateEntityPage` / `createOrUpdateConceptPage` already received the `SourceAnalysis` as their third parameter — typed `unknown`, named `_analysis`, dropped one line later. It is now typed and forwarded. No new call, no new data structure.

## Change 4 — source summary and the ownership question, part 1 (`1975817`)

Two things, and the second is the reason the first works:

1. The source summary under its own label `Source summary:`. Not `Summary:` — that label is taken by the extracted item, and two identical labels at different scopes would be worse than the missing field. The source **name** is not repeated; `{{new_info}}` already carries it.
2. One rule asking whether the source is primarily about this page or only mentions it in passing. All four strategies are keyed to novelty; without the question the summary is context the prompt has no reason to consult.

## Deviations from the fix path you proposed

| your sketch | what I did | why |
|---|---|---|
| extend `EngineContext` / `LintPhaseContext` | threaded on the page-factory path | `PROMPTS.mergeAnalysis` has one consumer, `merge-triage.ts:88`. `analysis-phase.ts` builds its own prompt and never references it, so the lint context would not have reached the triage |
| prepend `Source being merged: <title>` | summary only | it would print the source name a second time |
| `{ sourceTitle, summary, sourcePath }` | plus the curated note aliases (#185) | without them the deterministic guard misses every page living under an alias of its source note — the common case in a non-English vault |

`sourcePath` is in the type and deliberately not in the prompt, as you asked.

## Separability

Change 4 is the only one that alters prompt text. If you would rather leave the prompt untouched in a PATCH, drop it — Changes 1–3 stand alone, and the deterministic guard needs no prompt change.

Both new blocks in Change 4 are placeholder-rendered and empty without a `SourceContext`, so a caller without an ingest upstream gets the previous prompt character for character. A test asserts that, including that no blank line drifts in where the block would have been.

## Gate

Rebased onto current `main` (`7ef237a`). Baseline counted on `7ef237a` itself, with `pnpm build` run before `pnpm test`.

| check | `7ef237a` | this branch |
|---|---|---|
| `npx tsc --noEmit` | 0 | 0 |
| `pnpm build` | clean | clean |
| `pnpm test` | 2547 / 191 files | 2564 / 191 files |
| `pnpm lint --max-warnings=0` | 0 | 0 |
| `pnpm css-lint` | 0 | 0 |

+17 tests, no new file — exactly the tests added here. Both behavioural changes are mutation-checked: disabling the guard fails exactly the positive test and leaves the two backward-compatibility controls green.

## Not addressed

The harder half is untouched. When a source is **not** the page's subject, the triage still judges on novelty alone; the summary only makes a better judgement possible. The `supersedes:` flag from #220 is the part that decides rather than informs, and I see the Tier 0 / Tier 1 split is in the v1.26.0 scope now — the deterministic guard here is intentionally the weaker, PATCH-sized version of the same idea.

### DIFF
--- a/src/__tests__/wiki/page-factory/create-page.test.ts
+++ b/src/__tests__/wiki/page-factory/create-page.test.ts
@@ -51,7 +51,10 @@ function makeCtx(opts: {
   };
 }
 
-const EMPTY_ANALYSIS = {} as unknown;
+// #312: the third parameter used to be `unknown` and was dropped on the floor;
+// it now carries the source analysis. These cases exercise the no-analysis
+// path (lint-side callers), which is exactly what `{} as unknown` meant here.
+const EMPTY_ANALYSIS = undefined;
 
 describe('createOrUpdatePage — empty name guard', () => {
   it('returns path=null when name is empty', async () => {
--- a/src/__tests__/wiki/page-factory/merge-page.test.ts
+++ b/src/__tests__/wiki/page-factory/merge-page.test.ts
@@ -197,3 +197,84 @@ describe('appendToReviewedPage — happy path writes merged content', () => {
     expect(written).not.toContain('quote-A');
   });
 });
+// Issue #312 part 2 — deterministic ownership guard. A source that carries the
+// page's own lemma must not be dropped on a novelty judgement: that is how a
+// page keeps a definition written by the first source that mentioned it in
+// passing, while its actual subject source is skipped.
+describe('mergePage — source owning the page lemma overrides triage=skip (#312)', () => {
+  const OWNING_CONTEXT = {
+    sourceTitle: 'Caching',
+    summary: 'A note about caching.',
+    sourcePath: 'Notes/Caching.md',
+  };
+
+  it('routes to the body merge when the source basename is the page lemma', async () => {
+    const ctx = makeCtx(makeClient([
+      JSON.stringify({ strategy: 'skip', reason: 'no new info' }),
+      '## Description\nMerged text.',
+    ]));
+
+    const result = await mergePage(
+      ctx,
+      createMockEntity({ name: 'Caching' }),
+      'entity',
+      { path: 'Notes/Caching.md', basename: 'Caching' },
+      EXISTING,
+      [],
+      'wiki/entities/caching.md',
+      undefined,
+      OWNING_CONTEXT,
+    );
+
+    expect(result).toBe('wiki/entities/caching.md');
+    const written = ctx.written.get('wiki/entities/caching.md')!;
+    // The skip verdict was overridden — the merge path produced the body.
+    expect(written).toContain('Merged text.');
+  });
+
+  it('leaves triage=skip intact when the source does not carry the page lemma', async () => {
+    const ctx = makeCtx(makeClient([
+      JSON.stringify({ strategy: 'skip', reason: 'no new info' }),
+      '## Description\nMerged text.',
+    ]));
+
+    await mergePage(
+      ctx,
+      createMockEntity({ name: 'Caching' }),
+      'entity',
+      { path: 'Notes/Distributed Systems.md', basename: 'Distributed Systems' },
+      EXISTING,
+      [],
+      'wiki/entities/caching.md',
+      undefined,
+      { sourceTitle: 'Distributed Systems', summary: 's', sourcePath: 'Notes/Distributed Systems.md' },
+    );
+
+    const written = ctx.written.get('wiki/entities/caching.md')!;
+    // Unchanged behaviour: an incidental source is still skipped.
+    expect(written).toContain('Old text.');
+    expect(written).not.toContain('Merged text.');
+  });
+
+  it('does not fire without a source context — lint-side callers are unchanged', async () => {
+    const ctx = makeCtx(makeClient([
+      JSON.stringify({ strategy: 'skip', reason: 'no new info' }),
+      '## Description\nMerged text.',
+    ]));
+
+    await mergePage(
+      ctx,
+      createMockEntity({ name: 'Caching' }),
+      'entity',
+      // Same basename as the page: only the missing context keeps the guard off.
+      { path: 'Notes/Caching.md', basename: 'Caching' },
+      EXISTING,
+      [],
+      'wiki/entities/caching.md',
+    );
+
+    const written = ctx.written.get('wiki/entities/caching.md')!;
+    expect(written).toContain('Old text.');
+    expect(written).not.toContain('Merged text.');
+  });
+});
--- a/src/__tests__/wiki/page-factory/merge-triage.test.ts
+++ b/src/__tests__/wiki/page-factory/merge-triage.test.ts
@@ -9,6 +9,7 @@ import { describe, it, expect } from 'vitest';
 import {
   classifyMergeNeed,
   buildNewInfoSummary,
+  isSourceOwnPageLemma,
   type MergeTriageContext,
 } from '../../../wiki/page-factory/merge-triage';
 import { createMockEntity } from '../../__support__/factories';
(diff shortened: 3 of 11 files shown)


green-dalii:
Excellent work — the four-commit structure with documented separability (Change 4 can revert independently) is exactly what makes a multi-strategy refactor reviewable. I'll merge after the v1.25.9 Obsidian Bot review cycle settles (this PR was the trigger for that flow), so the merge lands cleanly without re-triggering review on an unrelated version.

Closes #312 (and reinforces #330's buried-failure-mode argument — your own issue).

Two things I'll do on the merge commit:
1. Tag the merge commit with `Closes #312` so auto-close fires.
2. Verify the prompt-text change (Change 4) doesn't accidentally shift behavior on the existing merge-triage fixtures; if it does, I'll drop it per your documented separability.

Issue #90, opened 2026-06-03
Source pages receive LLM-generated content tags instead of source note tags


DocTpoint:
When a source page (`wiki/sources/`) is created during ingest, the LLM assigns arbitrary content-derived tags (e.g. `Alzheimer-Demenz`, `Neuroprotektion`) instead of the tags from the original source note. This breaks tag consistency across the vault.

**Expected behavior:**
Source pages should either (a) inherit the tags from the linked source note, or (b) use the same controlled vocabulary as entity/concept pages.

**Connection to #[Tag-Vokabular-Issue]:**
Both issues share a root cause: the LLM has no constraint on which tags to use. A user-configurable tag vocabulary (`VALID_ENTITY_TAGS`, `VALID_CONCEPT_TAGS`) would solve both if it is also applied during source page generation.

**Workaround:**
Script to copy tags from source notes to source pages after each ingest. Related to #85

green-dalii:
Real consistency problem. (a) and (b) are complementary, both worth doing.

**(a) Inherit from source note** — low risk, ships immediately. At ingest, copy the source note's `tags:` frontmatter onto the generated source page. No LLM involvement.

**(b) Controlled vocabulary** — needs #85 first. Once `VALID_SOURCE_TAGS` is defined, the generation prompt can constrain tag selection.

I'd ship (a) first as a one-line fix and queue (b) behind #85. Users get consistency right away.

[Speculation, not verified] For (a) to help, the source note's `tags:` field needs to be reliably populated. If users routinely leave it empty, the fix doesn't help and you'd need a fallback. Quick spot check across a few sample vaults would confirm.

green-dalii:
Shipped in v1.17.0 following your (a) recommendation — programmatic tag inheritance from the source note's frontmatter via a new `extractSourceTags()` pure helper. The LLM-derived concept names are no longer written to the source page; if the source note has no tags, the system falls back to LLM-derived names (your suggested fallback).

Also added 3 new sections to the default Schema documenting the new contract: Source Page Template (mandates tag inheritance), Date Fields, Mentions Format. Existing user schema files are not overwritten.

Release: https://github.com/green-dalii/obsidian-llm-wiki/releases/tag/1.17.0

PR #349, opened 2026-07-25
fix(ingest): keep source-page tags inside the closed vocabulary

eucher:
## Problem

The `tags:` field of a generated source page falls back through three tiers:
tags already on the page, then the source note's frontmatter, then — until now
— every extracted concept name.

That last tier contradicts the rule the same code block documents (keeping
LLM-derived concept names out of the tag vocabulary), and it fires in the
normal case: a plain ASR transcript has no frontmatter at all.

Observed on a 9-source vault: **348 tags** written across the source pages, up
to **99 on a single page**, none of them members of `VALID_SOURCE_TAGS`. The
tag-vocabulary lint scanner flagged every page, and most values contained
spaces — so they were not valid Obsidian tags either, and the tag pane filled
with entries like `Джняна йога`. They were sticky too: tier 1 gives existing
tags priority, so re-ingesting preserved them.

## Fix

Source-page tags are a closed 7-value taxonomy (`VALID_SOURCE_TAGS`), and
anything outside it is stripped downstream regardless. So:

- tier 2 (inherited frontmatter tags) is filtered to that vocabulary;
- tier 3 falls back to `DEFAULT_SOURCE_TAG` instead of concept names.

A source whose type matters (e.g. `transcript`) declares it in its own
frontmatter — which is what tier 2 was built for.

| Source note frontmatter | Before | After |
|---|---|---|
| `tags: [transcript]` | `transcript` | `transcript` (unchanged) |
| `tags: [Джняна йога, philosophy, book]` | all three | `book` |
| no frontmatter | every extracted concept name | `other` |

Tier 1 is deliberately untouched, so manually-corrected tags on an existing
page still win (#114) — covered by a regression test.

## Tests

4 new tests in `src/__tests__/wiki/wiki-engine-summary-tags.test.ts`. Written
first: 2 of them fail on `main` for exactly the reported reason
(`'Jnana Yoga, Karma Yoga'` where `'other'` is expected), and the 2 regression
tests pass both before and after.

`{{tags}}` reaches the model through the summary prompt rather than being
written to the page directly, so `wiki-engine-harness` now records the
requests handed to the LLM stub and the tests assert on the prompt the engine
actually builds.

## Gate 1

| Check | Result |
|---|---|
| `pnpm test` | 2570 passed / 193 files |
| `pnpm build` | clean exit |

## Gate 4: Performance

| Dim | Status | Notes |
|-----|--------|-------|
| CPU | ✅ | One `Array.filter` over a source note's tag list (single digits) against a 7-element vocabulary, once per source page. |
| Memory | ✅ | No new retained structures; the filtered array replaces the unfiltered one. |
| IO | N/A | No vault reads or writes added. |
| Network | ✅ | No additional LLM calls. |
| Token | ✅ | Slight reduction — `{{tags}}` previously carried every extracted concept name (up to 99 values observed); now at most a few vocabulary terms. |

## Notes

- **Existing polluted pages are not rewritten.** Tier 1 still prefers tags
  already present on the page, so cleanup stays the lint retag runner's job.
  Happy to add a migration if you'd prefer it in this PR.
- **CHANGELOG not touched** — there is no `Unreleased` section and the target
  version is yours to pick. Tell me where it should land and I'll add the entry.
- **READMEs not touched** — this changes generated output, not a user-facing
  feature or workflow. Say the word if you want it documented.
- #90 is referenced in the surrounding code comment (pre-existing, unchanged).
  Since that issue is closed, this PR deliberately does not use a closing
  keyword — the fix covers a tier that the original change left open.

### DIFF
--- a/src/__tests__/__support__/wiki-engine-harness.ts
+++ b/src/__tests__/__support__/wiki-engine-harness.ts
@@ -14,8 +14,13 @@ import { WikiEngine } from '../../wiki/wiki-engine';
 import { parseFrontmatter } from '../../core/frontmatter';
 import { DEFAULT_SETTINGS } from './engine-context';
 
+type LLMRequest = Parameters<LLMClient['createMessage']>[0];
+
 export interface WikiEngineHarness {
   engine: WikiEngine;
+  /** Every request handed to the LLM stub, in order. Lets a test assert on the
+   *  prompt the engine built, not just on what the stub returned. */
+  llmRequests: LLMRequest[];
   /** Every path written by the engine (via onFileWrite), in order. */
   writtenPaths: string[];
   /** Every IngestReport delivered to onDone, in order. */
@@ -52,6 +57,7 @@ function mkFile(path: string): TFile {
 
 export function createWikiEngineHarness(opts: HarnessOptions = {}): WikiEngineHarness {
   const files = new Map<string, string>(Object.entries(opts.files ?? {}));
+  const llmRequests: LLMRequest[] = [];
   const writtenPaths: string[] = [];
   const reports: IngestReport[] = [];
   const stats = { llmCalls: 0, vaultMarkdownScans: 0 };
@@ -99,8 +105,9 @@ export function createWikiEngineHarness(opts: HarnessOptions = {}): WikiEngineHa
   } as unknown as App;
 
   const client: LLMClient = {
-    createMessage: async () => {
+    createMessage: async params => {
       stats.llmCalls++;
+      llmRequests.push(params);
       return opts.llmResponses?.[llmIdx++] ?? '{"entities":[],"concepts":[]}';
     },
   };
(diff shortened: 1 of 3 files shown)


green-dalii:
Thanks for the clean PR! Beautiful bug report — the "348 tags / 99 per page" measurement is exactly what makes this kind of pollution bug convincing.

The fix is right: tier 3 fallback to `DEFAULT_SOURCE_TAG` instead of unfiltered concept names is the minimum change that respects the closed-vocabulary rule.

Two things I want to verify on my end before merge:
1. `DEFAULT_SOURCE_TAG` exists and has a sensible value (I'll check the constants).
2. The new wiki-engine test covers the sticky case (re-ingest preserves tier-1 tags) — your commit message mentions this but I want to re-read the test before merging.

I'll merge after the current v1.25.9 Obsidian Bot review cycle settles.

Kicking off the merge — will close this PR's `Closes` reference in the merge commit message.

PR #350, opened 2026-07-25
fix(ingest): gate translation instruction on source frontmatter language

eucher:
## Problem

The extraction prompt asks the model to add a `translation` field alongside
every verbatim quote. It was gated on `wikiLanguage !== 'en'`, which is a proxy
for "this wiki is cross-language" — but it says nothing about the *source*.

Consequence: a Russian wiki ingesting a Russian transcript asked for a Russian
translation of Russian quotes. The output is meaningless, and it is not free —
the extra field is emitted per quote, inflating the batch JSON until it
truncates and gets force-halved.

## Fix

Read the source note's frontmatter `language:` and skip the translation
instruction when it matches the wiki language. When the source is untagged,
fall back to the legacy `!== 'en'` proxy, so existing cross-language wikis keep
their behaviour.

The value is matched against both the setting and its display name, so
`language: de` and `language: Deutsch` both resolve against a `de` wiki. A
custom `wikiLanguage` outside `WIKI_LANGUAGES` (e.g. `ru`) matches on the raw
value. Matching is case- and whitespace-insensitive.

| Source `language:` | Wiki language | Translation instruction |
|---|---|---|
| `ru` | `ru` | omitted (was: emitted) |
| `Deutsch` | `de` | omitted (was: emitted) |
| `en` | `de` | emitted |
| absent | `de` | emitted (legacy proxy) |
| absent | `en` | omitted (legacy proxy) |
| `ru` | `en` | emitted (was: omitted — see below) |

## Behaviour change worth calling out

An English wiki previously suppressed the instruction unconditionally, so
quotes from a foreign-language source arrived untranslated. A source that
declares `language:` now makes that cross-language case detectable, and the
instruction is emitted. Untagged sources in an English wiki are unaffected.
Happy to drop this half if you'd rather keep the change purely subtractive.

## Tests

7 new tests in `src/__tests__/wiki/source-analyzer-translation-hint.test.ts`,
asserting on the prompt handed to the LLM via the existing
`vi.spyOn(ctx.getClient()!, 'createMessage')` pattern. 4 of them fail on
`main`; the 3 legacy-behaviour tests pass both before and after.

## Gate 1

| Check | Result |
|---|---|
| `pnpm test` | 2573 passed / 193 files |
| `pnpm build` | clean exit |

## Gate 4: Performance

| Dim | Status | Notes |
|-----|--------|-------|
| CPU | ✅ | Two string comparisons per source, once per ingest. |
| Memory | ✅ | One short string retained. |
| IO | ✅ | No new reads — `getFileCache(file)` is already called on this path for `aliases`; the same frontmatter object is reused. |
| Network | ✅ | No additional calls. Fewer halve-and-retry rounds expected, since the truncation pressure this removes is what triggers them. |
| Token | ✅ | Removes a ~90-word instruction from every extraction batch of a same-language source, and stops the model emitting a `translation` field per quote — the larger saving, since it is output. |

## Notes

- `language:` in a source note's frontmatter is the new signal. It is optional
  and absent sources keep the old behaviour, so nothing breaks for existing
  vaults — but it is undocumented. Tell me if you want it added to the READMEs
  and I'll do it in this PR.
- **CHANGELOG not touched** — no `Unreleased` section, and the target version
  is yours to pick.

### DIFF
--- a/src/__tests__/wiki/source-analyzer-translation-hint.test.ts
+++ b/src/__tests__/wiki/source-analyzer-translation-hint.test.ts
@@ -1,0 +1,77 @@
+// The extraction prompt asks for a `translation` field alongside each verbatim
+// quote. That only makes sense when the source and the wiki speak different
+// languages — translating a Russian quote into a Russian wiki is wasteful and
+// bloats the batch JSON toward truncation.
+//
+// The source note's frontmatter `language:` is the explicit signal; the legacy
+// `wikiLanguage !== 'en'` proxy stays in place for untagged sources.
+
+import { describe, it, expect, vi } from 'vitest';
+import { createMockContext, createMockFile } from '../__support__/engine-context';
+import { SourceAnalyzer } from '../../wiki/source-analyzer';
+import { TFile } from 'obsidian';
+
+const SOURCE_PATH = 'sources/lecture.md';
+const TRANSLATION_MARKER = 'TRANSLATION (cross-language wikis)';
+
+const EXTRACTION_RESPONSE = JSON.stringify({
+  source_title: 'Lecture',
+  summary: 'A lecture.',
+  entities: [{ name: 'Foo', type: 'other', summary: 'bar', mentions_in_source: [] }],
+  concepts: [],
+});
+
+/** Run one extraction and return the user prompt handed to the LLM. */
+async function promptFor(body: string, wikiLanguage: string): Promise<string> {
+  const { ctx } = createMockContext({
+    vaultFiles: { [SOURCE_PATH]: body },
+    llmResponses: [EXTRACTION_RESPONSE],
+    settings: { wikiLanguage },
+  });
+  const spy = vi.spyOn(ctx.getClient()!, 'createMessage');
+  const analyzer = new SourceAnalyzer(ctx);
+  // eslint-disable-next-line obsidianmd/no-tfile-tfolder-cast
+  await analyzer.analyzeSource(createMockFile(SOURCE_PATH) as unknown as TFile);
+  expect(spy).toHaveBeenCalled();
+  return spy.mock.calls[0][0].messages[0].content as string;
+}
+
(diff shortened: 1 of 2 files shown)


green-dalii:
@eucher 
Спасибо за ваш вклад; я внимательно его изучу.

Кстати, я предполагаю, что вы русскоязычный пользователь. Если это так, не могли бы вы поработать над русским переводом i18n через новый запрос на слияние (PR)? Я заметил недавний рост числа русскоязычных пользователей, но в настоящее время поддержки русского i18n нет.

Конечно, вы можете отказаться, если время или обстоятельства не позволяют, или я могу помочь проверить качество перевода после того, как разработаю русскую версию. Вы не обязаны делать выбор.

---

EN Version:

Thank you for your contribution; I will review it carefully.

By the way, I gather you are a Russian speaker. If so, would you be willing to handle the Russian i18n translation via a new PR? I’ve noticed an increase in Russian-speaking users recently, yet there is currently no Russian i18n support.

Of course, please feel free to decline if your time or circumstances don't permit it—or perhaps you could simply help review the translation quality later on, once I’ve implemented the Russian localization. Please don't feel any pressure to commit.

green-dalii:
Thanks! The match-table at the bottom of your PR description (`ru` wiki + `ru` source → skip, etc.) is exactly the kind of deterministic spec that makes this safe to merge.

Two things I'll verify:
1. Custom `wikiLanguage` outside `WIKI_LANGUAGES` (your "raw value" case) — I want to confirm the fallback doesn't crash on a value not in the table.
2. The new test file's first failure scenario matches the truncation chain (extra field → truncation → halving → corruption) — your description hints at this but I want to read the assertion.

Will merge after the current v1.25.9 Obsidian Bot review cycle settles. Kicking off the merge — will close this PR's `Closes` reference in the merge commit message.

Issue #75, opened 2026-06-01
LM Studio: HTTP 400 on batch 2+ — hardcoded token limits exceed 8192-token context window


DocTpoint:
**Provider:** Custom OpenAI-compatible (LM Studio 0.4.x)
**Model:** google/gemma-4-26b-a4b
**Plugin version:** 1.14.0

**Problem**

Ingest fails silently on all batches after the first with HTTP 400. Batch 1 succeeds, Batch 2+ fails with:

> [Batch 2] Call failed: Error: Request failed, status 400
> [Batch 2] Non-first-round failure, keeping extracted items

**Root cause**

`MAX_TOKENS2` in `analyzeSource` is hardcoded to `16000`. LM Studio enforces a hard limit where `max_tokens` cannot exceed the model's context window (8192 for most local models). As the batch prompt grows (adding already-extracted entities as context), the total token count tips over the limit starting at batch 2.

The same issue affects `TOKENS_PAGE_GENERATION = 8000`, which also exceeds the 8192 context window once a system prompt is added.

**Workaround**

Patching the constants manually:
- `MAX_TOKENS2 = 3000` (in `analyzeSource`, ~line 9312)
- `TOKENS_PAGE_GENERATION = 3000` (constants block, ~line 220)

**Suggested fix**

Either cap `max_tokens` per call to `Math.min(requested, providerContextWindow - estimatedPromptTokens)`, or expose a "Max tokens per call" setting for custom providers.

green-dalii:
Hi @DocTpoint,

Thanks for this precise root-cause analysis. The  → 400 chain is clear.

You're right that we need a user-configurable ceiling rather than hardcoded constants. We're evaluating a  setting that caps all  values at runtime, so users with 4K/8K local models can set their own upper bound without us trying to guess provider capabilities.

One thing worth noting:  has a local  that shadows the centralized  constant — that's a consistency bug on our side that we'll fix as part of this work.

Deferring to v1.16.0 alongside #72/#76 since it touches the same LLM call layer. The manual constant patch you found is the right workaround for now.

green-dalii:
@DocTpoint — confirmed, plus one thing your repro didn't catch: there's a local `MAX_TOKENS` shadow at `source-analyzer.ts:113` that bypasses the centralized `MAX_TOKENS_BATCH` constant. My bad, that path never got finished when I centralized the constants.

Short-term fix landing today/tomorrow:
[...]

PR #352, opened 2026-07-25
fix(llm): stop silent truncation in chat and ingest

eucher:
## Problem

Reasoning models routinely hit the output cap, and both flows handled it
badly. In chat the answer was cut mid-sentence with nothing to indicate it. In
ingest the analyzer spent extra full-size calls trying to "repair" a response
that was merely incomplete, and could silently drop a batch. Reproduced
against llama-server b9205 + Qwen3.6-35B-A3B.

## Chat flow

Sibling of #305, whose finish-reason helpers were wired into
`source-analyzer.ts` only.

**Own token budget.** `TOKENS_QUERY_LLM_SELECT` (3000) was used as `max_tokens`
for the user-facing answer at all three Query call sites. On a reasoning model
the chain of thought is billed against the same budget and routinely eats a
third to a half of it. Added `TOKENS_QUERY_ANSWER = 8000`, matching
`TOKENS_PAGE_GENERATION`.

Those three sites were the *only* consumers of `TOKENS_QUERY_LLM_SELECT`, so it
is removed. Its name never matched its use: it was introduced in `7951173`
purely as a name for the literal `3000` already sitting on the answer path, and
the real selection steps have their own budgets (`TOKENS_QUERY_SEED_SELECT`,
`TOKENS_QUERY_PAGE_SELECT`, `TOKENS_QUERY_KEYWORDS`).

**Cap honoured on the streaming path.** `wrapWithAdvancedSettings` overrides
`createMessage` only — `createMessageStream` is inherited through the prototype
chain by design (v1.23.2 guard, with its own regression test). So a streaming
request ignored the Issue #75 `maxTokensPerCall` ceiling. At 3000 that rarely
showed; at the raised budget it would let a request exceed a local server's
context window (the exact HTTP 400 #75 exists to prevent). The streaming call
now applies `capMaxTokens` explicitly. Non-stream call sites are untouched —
they go through `createMessage` and are already capped.

**Truncation is observable.** No `onFinish` was passed in the Query flow, so
`finishReason: 'length'` was not observable at all. All three call sites now
pass it and log a warning, the way the ingest path already reports truncation.
`createMessageStream` gained an optional `onFinish`.

**Thinking stayed on for llama.cpp.** `enableThinking=false` mapped only to
`thinking: { type: 'disabled' }` (DeepSeek/Kimi/GLM dialect). `llama-server`
ignores that field — it expects `chat_template_kwargs: { enable_thinking: false }`.
Both dialects are now sent; each backend ignores the one it does not know.

## Ingest flow

**Token usage logging.** `LLMFinishMeta` now carries `usage`, and the openai +
openai-compatible clients forward `result.usage`. Logging real
prompt/completion tokens per batch turns truncation tuning into a measurement
instead of a guess.

**No repair call on truncation.** A truncated response is incomplete, not
malformed — repair cannot restore never-emitted content, and it re-hits the
same limit (observed: a ~48k-token repair that itself truncated). Note
`parseJsonResponse` attempts repair from *two* recovery strategies, so a
truncated batch burned up to two full `retryCap`-sized calls. Now it goes
straight to halve-and-retry. Providers reporting no finish reason keep the
repair path.

**Per-batch retry budget.** The first halve anywhere latched `retryingBatch`
for the whole source, so a later truncated batch could no longer halve and was
dropped — the source finalized as "complete" while missing those items. Reset
on the success path.

## Tests

5 new tests in `source-analyzer-truncation-retry.test.ts`, plus the updated
`chat_template_kwargs` assertion in the client test.

The skip-repair tests needed a new fixture. The existing `TRUNCATED` stops
before any closing brace, so neither recovery strategy finds a candidate and
`repairFn` is never consulted — a test written against it passes with or
without the fix. `TRUNCATED_MID_ARRAY` cuts off after several complete items,
which is what a real truncated batch looks like, and does reach the repair
path. Each new test was verified to fail on `main` for the right reason.

The existing `halves only once` test still passes and was not modified: it
scripts two truncations back-to-back, while the budget reset only fires after a
successful batch.

## Gate 1

| Check | Result |
|---|---|
| `pnpm test` | 2570 passed / 192 files |
| `pnpm build` | clean exit |

## Gate 4: Performance

| Dim | Status | Notes |
|-----|--------|-------|
| CPU | ✅ | One enum comparison per batch; one `Math.min` per streaming request. |
| Memory | ✅ | One small `usage` object per batch, not retained. |
| IO | N/A | No vault reads or writes in scope. |
| Network | ✅ | Strictly fewer calls: up to 2 repair calls saved per truncated batch, and a batch that used to be dropped now recovers via halve. |
| Token | ⚠️ | Two directions. Ingest: down — repair calls were `retryCap`-sized. Chat: `max_tokens` 3000 → 8000 raises the *ceiling*, not the spend, since billing follows tokens actually generated; but answers previously cut at 3000 will now legitimately generate their tail, which is a real cost increase for long answers on metered providers. `maxTokensPerCall` lets a user lower it, and now does so on the streaming path too. |

## Notes

- **Removed export.** `TOKENS_QUERY_LLM_SELECT` is deleted. History: introduced
  in `7951173`, value never changed, never referenced outside those three call
  sites and one test comment, no mentions in CHANGELOG/ROADMAP. An in-flight
  branch referencing it would see a one-line conflict. Happy to keep it as
  `@deprecated` instead if you prefer.
- **Truncation is logged, not surfaced in the UI.** `console.warn`, matching
  how the ingest path reports it. That is deliberate but not symmetric: ingest
  self-heals via halve-and-retry, whereas a truncated chat answer is final and
  the user still is not told. A `Notice` would need a TEXTS key across all 10
  locales and a wording decision that is yours — happy to do it here or in a
  follow-up.
- **Type shape.** `onFinish`'s inline `{ finishReason }` becomes the named
  `LLMFinishMeta`, additively extended with `usage?`. Callers ignoring it are
  unaffected.
- **CHANGELOG not touched** — no `Unreleased` section; the target version is
  yours to pick.

### DIFF
--- a/src/wiki/source-analyzer.ts
+++ b/src/wiki/source-analyzer.ts
@@ -29,6 +29,7 @@ import {
   MentionWithProvenance,
   WIKI_LANGUAGES,
   LLMFinishReason,
+  LLMUsage,
 } from '../types';
 import { PROMPTS } from '../prompts';
 import { parseJsonResponse } from '../core/json';
@@ -226,8 +227,13 @@ export class SourceAnalyzer {
     // closure so the parse-failure path and the catch path share one
     // implementation. Returns whether the caller should re-run this batch;
     // the caller owns the loop control (`batchNum--; continue;`).
+    // Whether a halve-and-retry is still available for the current batch.
+    // Gates both the retry itself and the decision to skip JSON repair on a
+    // truncated response — repair is the last salvage once halving is spent.
+    const canHalveBatch = (): boolean => !retryingBatch && currentBatchSize > limits.minBatchSize;
+
     const halveBatchAndRetry = (batchLabel: string, cause: string): boolean => {
-      if (retryingBatch || currentBatchSize <= limits.minBatchSize) return false;
+      if (!canHalveBatch()) return false;
       currentBatchSize = Math.max(limits.minBatchSize, Math.floor(currentBatchSize * 0.5));
       console.warn(`${batchLabel} Truncation detected (${cause}), halving batch size to ${currentBatchSize} and retrying`);
       retryingBatch = true;
@@ -340,7 +346,7 @@ export class SourceAnalyzer {
         // report it leave this at 'unknown', which keeps pre-#305 behavior.
         // Held in an object rather than a bare `let` so control-flow analysis
         // does not narrow it to its initializer across the callback.
-        const finish: { reason: LLMFinishReason } = { reason: 'unknown' };
+        const finish: { reason: LLMFinishReason; usage?: LLMUsage } = { reason: 'unknown' };
         const response = await client.createMessage({
           model: resolvedModel,
           max_tokens: batchMaxTokens,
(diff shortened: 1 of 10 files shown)


src/wiki/source-analyzer.ts
```diff
@@ -349,23 +350,41 @@ export class SourceAnalyzer {
...
-        }, { silentOnEmpty: true }) as Partial<SourceAnalysis> | null;
+        // Issue #305 follow-up: when the provider reports the response was
+        // truncated (finish_reason=length), the JSON is *incomplete*, not
+        // malformed — a syntax-repair pass cannot restore never-emitted
+        // content and just burns another retryCap-sized call re-hitting the
+        // same limit (observed: a ~48k-token repair that itself truncates).
+        // Skip repair on truncation and fall straight through to the
+        // halve-and-retry recovery below. Providers that don't report a
+        // finish reason keep 'unknown' and retain the repair path.
+        const repairFn = finish.reason === 'length'
```
DocTpoint: The premise is right: on `finish_reason=length` the repair call is usually futile and expensive, and skipping it is a sensible default. One case is missed, and there the cost is not one wasted call.

A truncated batch is a prefix of **complete** items followed by one cut-off item. Closing the brackets around the complete items is exactly what the repair prompt asks for ("missing brackets"). So repair can return the items that did arrive. That makes it the last salvage step when the halving retry is no longer available. `halveBatchAndRetry` returns `false` when `retryingBatch` is already set, or when `currentBatchSize <= minBatchSize`. What follows then is line 404: `if (isFirstBatch) return null` — the whole source is dropped, not just the batch.

Measured on both arms with the same script. The source truncates twice, so the halving budget is spent. The repair stub closes the brackets, which is what a repair pass does:

| | repair calls | `analyzeSource` result |
|---|---|---|
| `main` | 2 | `["Amyloid"]` — source survives |
| this PR | 0 | `null` — whole source dropped |

Your per-batch reset of `retryingBatch` narrows this, but does not close it: a second truncation inside one batch, and any truncation at `minBatchSize`, still reach line 404.

One note on the existing tests: the harness in `source-analyzer-truncation-retry.test.ts` answers repair calls by echoing the broken text back unchanged. That models repair as unable to help, so the current tests cannot observe this case.

A narrower gate keeps the saving and the salvage: skip the repair only when the halving retry can still run — that is the guard inside `halveBatchAndRetry` (`!retryingBatch && currentBatchSize > minBatchSize`). Happy to share the probe if it is useful.

eucher:
You're right, and I reproduced it before changing anything.

Same shape as your probe — a repair stub that closes the brackets rather than
echoing the broken text back, driven through two truncations so the halving
budget is spent:

| | repair calls | `analyzeSource` |
|---|---|---|
| `main` | 2 | `["Amyloid"]` |
| this PR (before) | 0 | `null` — source dropped |
| this PR (now) | 2 | `["Amyloid"]` |

Fixed as you suggested — the skip is gated on whether the halving retry can
still run, not on the finish reason alone. The condition already existed inside
`halveBatchAndRetry`, so I lifted it into a predicate and both sites use it:

```ts
const canHalveBatch = (): boolean => !retryingBatch && currentBatchSize > limits.minBatchSize;

const halveBatchAndRetry = (batchLabel: string, cause: string): boolean => {
  if (!canHalveBatch()) return false;
  ...
};

const repairFn = finish.reason === 'length' && canHalveBatch()
  ? undefined
  : async (malformedJson: string) => { ... };
```

The saving is unchanged in the common case — first truncation still halves
without paying for a repair — and repair comes back as the last salvage once
halving is exhausted, including the `minBatchSize` case you flagged.

Your note about the harness was the sharper half of the review: echoing the
broken text back models repair as unable to help, so no test written against it
could observe this. Added `closeBrackets` as an opt-in repair behaviour plus a
test that drives two truncations and asserts the source survives with the items
that did arrive. It fails on the unguarded version of this PR.

Gate 1 green: lint 0/0, tsc 0, 2571 tests, build, css-lint.

src/wiki/source-analyzer.ts
DocTpoint: Verified on `527a7a5`: the probe passes, and lifting the condition into `canHalveBatch()` is the right shape — the common case still skips the repair, and the salvage returns exactly where halving is exhausted. Gate reproduced here as well: tsc 0, lint 0/0, build clean, 2571 tests / 192 files.

The `closeBrackets` opt-in is the better half of the fix. It makes this case observable for whoever touches the path next.

green-dalii:
Thanks for the most impactful PR of the three — silent truncation in ingest is a data-loss class bug, not just a UX issue.

The `TOKENS_QUERY_LLM_SELECT` removal is well-justified (your audit of the 3 sites is in the PR body), and the truncated-vs-malformed distinction in the repair path is exactly the finish-reason hygiene #305 started.

Three things I'll verify on my side before merge:
1. `grep -rn TOKENS_QUERY_LLM_SELECT src/` confirms your 3-site audit — no orphan imports.
2. The new `TOKENS_QUERY_ANSWER = 8000` constant doesn't break the existing test that asserts `TOKENS_QUERY_LLM_SELECT === 3000` (you updated `constants.test.ts` to match — I want to read that diff closely).
3. The reasoning-model cap test covers at least one provider known to consume the chain of thought against `max_tokens` (Qwen3 / DeepSeek-R1) — I want to confirm the test isn't asserting on a friendly provider that masks the issue.

Will merge after v1.25.9 review settles.

I appreciate the quality of your PRs — every one of them ships with measurements, reproducible reproductions, and minimal fixes. That is rare.

Issue #348, opened 2026-07-25
Ingest extracts entities FROM a note but never a page FOR the note's own subject


DocTpoint:
`analyzeSource` extracts entities and concepts **out of** a note's text. That
the text is itself *about* something is not a category of the prompt, so a note
whose subject is a topic in its own right frequently produces a dozen derived
pages and none for its own lemma. Every `[[...]]` pointing at that lemma from
the derived pages then stays dead.

### Field data

Measured on a 416-note vault (German, medical), against the page titles and
aliases actually on disk, with `computeSlug` replicated character-for-character
from `core/slug.ts` so the comparison matches the plugin's own slugging:

- **43 of 416 notes (10.3%) have no knowledge page for their own subject.**
- Nine of those are domain containers (`Neurologie`, `Kardiologie`,
  `Biochemie-Signalwege` …) that legitimately should not get one.
- Example: `Notizen/Silent Inflammation.md` (19 KB) produced 38 derived pages
  and no page for "Silent Inflammation"; `Notizen/Inflammaging.md` (7 KB)
  produced 17 and none for itself.

Not a size effect: it happens at every vault size, and re-running does not heal
it — the analyzer simply is not asked the question.

One honest qualifier: it is **not strictly deterministic**. The same note under
the same granularity produced `entities/Klotho` in one run and only
`entities/alpha-Klotho` / `entities/beta-Klotho` in another (different builds,
so the comparison is confounded). That argues for a guarantee rather than a
retry.

### Proposed fix

A deterministic step after the analyzer and before page generation:

1. Derive slug keys from the note's file name plus its curated frontmatter
   aliases.
2. If none of them appears among the extracted entities/concepts (title or
   alias), add the lemma as a candidate.
3. **Skip domain containers**: a note whose title or alias equals one of the
   configured domain tags is a container, not a subject.

The decision *whether* the lemma is missing needs no model. Only the type
choice (entity vs. concept) is a model call — a two-element classification,
~32 tokens, roughly 30 extra calls on a full run. Every failure path is a
no-op: no summary, unreadable answer, failed call → nothing happens.

The guard matters as much as the addition: the lemma is often extracted on its
own, and blindly creating it would duplicate pages.

### Note on overlap with #312

The matching predicate is the same one #312 needs for its deterministic
ownership guard ("does this source carry the page's own lemma?"). If the #312
PR lands first, this builds on the `slugKeys` helper it introduces in
[...]

Issue #328, opened 2026-07-22
Rethinking the schema layer: separating user knowledge from engine internals


green-dalii:
## Rethinking the schema layer: separating user knowledge from engine internals

**Status:** Open for discussion. No PR is open yet — I want to converge on the direction with the people who actually use this before writing code.

### 1. What the schema does today

The wiki schema (`schema/config.md` in your wiki folder) is the layer that tells the LLM *how to shape your wiki*. It describes page structure — what sections an entity page should have, what frontmatter fields concept pages carry, how mentions should be formatted, how merges should resolve contradictions. Think of it as the "editorial guidelines" for the LLM that writes your pages.

In the current implementation, the schema is a markdown file with sections like `## Entity Page Template` and `## Classification Rules`. The LLM sees these sections as part of its system prompt whenever it ingests, generates, or merges a page. You can also customise the section list per page type in `## Entity Page Template > **Sections:**`, and the system uses that to shape the LLM's output template.

At least, that's the idea. In practice, a few things get in the way.

### 2. The principle: organising knowledge against entropy

A long-lived knowledge base has one enemy that always wins if you let it: **drift**. Same concept, different section structure this month than last. Same type of page, different naming convention. Same tag, different spelling. It starts small, and a year later the wiki is hard to trust — and a wiki you can't trust is one you stop using.

The schema layer is the tool we have to fight that, and it fights it with two hands:

- **Flexibility** so you can adapt the wiki to your domain — a medical vault needs "drug" and "trial", a legal vault needs "statute" and "party", a learning-funnel vault needs "journal" and "map". A schema that can't flex to what *you* need is a schema that doesn't really work.
- **Constraint** so the wiki stays consistent over time — that every page of the same type has the same sections, that tags come from the same controlled list, that every ingest adds to the structure rather than inventing a new one. A schema with no constraint is not really a schema at all.

Neither wins alone. Pure flexibility produces chaos in months. Pure constraint locks the plugin at whatever three page types the initial developer guessed at, and every new domain needs a feature request.

The way out is not a compromise somewhere in the middle. It's a clean split between the things that are *the user's job to decide* and the things that are *the engine's job to decide*:

| Layer | Owned by | What it is | The principle |
|---|---|---|---|
| **User domain knowledge** | You | The schema file: page templates, content rules, naming conventions, merge policies | This is *your* structure, *your* domain. Edit freely; the LLM defers to what you wrote. |
[...]

PR #357, opened 2026-07-26
feat(ingest): guarantee a page for the source note's own subject (#348)

DocTpoint:
Draft PR for #348 — opening early in parallel as you invited, so it can be reviewed alongside the v1.26.0 ingest-pipeline change set.

## What it does
After `buildSourceAnalysis`, if the source note's own lemma is not already among the extracted pages, emit one for it. Closes the dead-link class where a note's own subject never gets a node (10.3% of notes in the measurement in the issue).

## Approach — the three-step deterministic fix from the issue
- **New pure module `src/core/source-lemma.ts`** (no IO, no model call, fully unit-tested): `slugKeys` / `isLemmaExtracted` / `isDomainContainer` / `parseTagList` / `decideSourceLemma`. Matches title + curated frontmatter aliases + the analyzer's `source_title`, reusing the existing slug/alias helpers.
- **Only the entity-vs-concept choice goes to the model**, as a two-element classification. Every failure path (missing summary, unparseable answer, failed call) is a no-op — nothing is emitted.
- **Named after the file**, because incoming dead links point at the filename; a divergent model title can suppress the addition but never rename the page.
- **Domain-container guard**: notes whose title/alias equals a *configured* domain tag (not a hand-written list) are skipped — they aren't lemmas.

The guard checks whether the lemma is *missing* — it is often self-extracted already, in which case this is a no-op.

## Tests
21 unit tests for the pure module + 5 wiring tests driving `analyzeSource` over the mock harness (add-when-missing / no-duplicate-when-present / unusable-type-is-noop / concept-lands-in-concepts).

## Gate
Cherry-picks cleanly onto current `main` (`7e22848`, after #349/#350/#352). tsc 0 · lint 0 · build clean · 2631 tests pass (197 files).

Ready for review — per the v1.26.0 scope lock in #330. Happy to adjust scope.

### DIFF
--- a/src/__tests__/core/source-lemma.test.ts
+++ b/src/__tests__/core/source-lemma.test.ts
@@ -3,26 +3,29 @@
 // vault observation, noted next to the assertion.
 import { describe, it, expect } from 'vitest';
 import {
-  slugKeys,
   isLemmaExtracted,
-  isDomainContainer,
-  parseTagList,
   decideSourceLemma,
 } from '../../core/source-lemma';
+import { slugKeys } from '../../core/slug';
 
-describe('slugKeys', () => {
-  it('collects title and aliases, case-insensitively comparable', () => {
-    const keys = slugKeys('Silent Inflammation', ['Stille Entzündung']);
-    expect(keys.has('silent-inflammation')).toBe(true);
-    expect(keys.has('stille-entzündung')).toBe(true);
+describe('slugKeys (imported from core/slug — Issue #366 regression guard)', () => {
+  // Regression: PR #357 originally re-implemented slugKeys locally without
+  // threading the `turkishFold` opt-in, which would have re-introduced the
+  // duplicate-page bug Issue #366 fixed. source-lemma.ts must use the canonical
+  // version. Compare keys between Turkish-folded inputs to confirm they share
+  // at least one common slug.
+  it('folds Turkish İ→i so İnsülin and insülin share a comparison key', () => {
+    const a = slugKeys('İnsülin', [], { turkishFold: true });
+    const b = slugKeys('insülin', [], { turkishFold: true });
+    const intersection = [...a].filter(k => b.has(k));
+    expect(intersection.length).toBeGreaterThan(0);
   });
 
-  it('ignores empty and whitespace-only aliases', () => {
-    expect(slugKeys('Klotho', ['', '   ']).size).toBe(1);
-  });
-
-  it('yields no keys for an empty name', () => {
-    expect(slugKeys('', []).size).toBe(0);
+  it('without Turkish fold, the same inputs yield distinct keys', () => {
+    const a = slugKeys('İnsülin');
+    const b = slugKeys('insülin');
+    expect(a.has('insülin')).toBe(false);
+    expect(b.has('insülin')).toBe(true);
   });
 });
 
@@ -52,47 +55,13 @@ describe('isLemmaExtracted', () => {
   });
 });
 
-describe('isDomainContainer', () => {
-  const tags = parseTagList('Erkrankung, Neurologie, Kardiologie, Immunologie, Biochemie');
-
-  it('recognises a note named after a configured domain', () => {
-    expect(isDomainContainer(slugKeys('Neurologie'), tags)).toBe(true);
-  });
-
-  it('recognises it through a curated alias', () => {
-    // `Notizen/Biochemie-Signalwege.md` carries `Biochemie` as an alias.
-    expect(isDomainContainer(slugKeys('Biochemie-Signalwege', ['Biochemie']), tags)).toBe(true);
-  });
-
-  it('leaves a topic note alone', () => {
-    expect(isDomainContainer(slugKeys('D-Manose'), tags)).toBe(false);
-  });
-
-  it('never fires when no domain vocabulary is configured', () => {
-    expect(isDomainContainer(slugKeys('Neurologie'), [])).toBe(false);
-  });
-});
-
-describe('parseTagList', () => {
-  it('splits, trims and drops empties', () => {
-    expect(parseTagList(' A ,, B , ')).toEqual(['A', 'B']);
-  });
-
-  it('treats undefined as no vocabulary', () => {
-    expect(parseTagList(undefined)).toEqual([]);
-  });
-});
-
 describe('decideSourceLemma', () => {
-  const tags = parseTagList('Neurologie, Kardiologie, Biochemie');
-
   it('adds the lemma when the extraction missed it', () => {
     // The observed b1 case: alpha-/beta-Klotho extracted, Klotho itself not.
     expect(decideSourceLemma({
       sourceTitle: 'Klotho',
       entities: [{ name: 'alpha-Klotho' }, { name: 'beta-Klotho' }],
       concepts: [{ name: 'Phosphathaushalt' }],
-      domainTags: tags,
     })).toEqual({ action: 'add', name: 'Klotho' });
   });
 
@@ -101,7 +70,6 @@ describe('decideSourceLemma', () => {
       sourceTitle: 'Papain',
       entities: [{ name: 'Papain' }],
       concepts: [],
-      domainTags: tags,
     })).toEqual({ action: 'skip', reason: 'already-extracted' });
   });
 
@@ -110,28 +78,7 @@ describe('decideSourceLemma', () => {
       sourceTitle: 'Silent Inflammation',
       entities: [],
       concepts: [{ name: 'Silent-Inflammation' }],
-      domainTags: tags,
     })).toEqual({ action: 'skip', reason: 'already-extracted' });
-  });
-
-  it('skips a domain container even when its lemma is missing', () => {
-    expect(decideSourceLemma({
-      sourceTitle: 'Neurologie',
-      sourceAliases: ['Neurowissenschaften'],
-      entities: [{ name: 'Hippocampus' }],
-      concepts: [],
-      domainTags: tags,
-    })).toEqual({ action: 'skip', reason: 'domain-container' });
-  });
-
-  it('prefers the container reason over the extraction reason', () => {
-    // Both would fire; the reported reason must be stable for the log.
-    expect(decideSourceLemma({
-      sourceTitle: 'Neurologie',
-      entities: [{ name: 'Neurologie' }],
-      concepts: [],
-      domainTags: tags,
-    })).toEqual({ action: 'skip', reason: 'domain-container' });
   });
 
   it('skips when no title is available', () => {
@@ -139,7 +86,6 @@ describe('decideSourceLemma', () => {
       sourceTitle: '   ',
       entities: [],
       concepts: [],
-      domainTags: tags,
     })).toEqual({ action: 'skip', reason: 'no-title' });
   });
 
@@ -151,11 +97,10 @@ describe('decideSourceLemma', () => {
       sourceAliases: ['LAMA/LABA'],
       entities: [{ name: 'LAMALABA' }],
       concepts: [],
-      domainTags: tags,
     })).toEqual({ action: 'skip', reason: 'already-extracted' });
   });
 
-  it('adds nothing for a vault with no domain vocabulary but an extracted lemma', () => {
+  it('adds nothing for a vault with an extracted lemma and no aliases', () => {
     expect(decideSourceLemma({
       sourceTitle: 'Rutosid',
       entities: [{ name: 'Rutosid' }],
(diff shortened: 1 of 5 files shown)


green-dalii:
Hi @DocTpoint 👋

Thanks for the source-lemma PR — the design idea (deterministic guarantee for the source note's own subject) is sound and the test coverage looks comprehensive at first glance. **But running Gate 1 + /simplify (4 angles) + /code-review (max effort) on the diff surfaced 4 P0 correctness bugs that block merge.** Posting them here so you have visibility before I push fixes.

All findings have been independently verified against the source — including throwaway probe tests that reproduced P0 #1 in real production code paths (not against mocks). Each has a one-line fix.

---

## 🔴 P0 #1 — `buildSystemPrompt('compact')` throws in production → feature never executes

`src/wiki/source-analyzer.ts:731`

```ts
system: await this.ctx.buildSystemPrompt('compact'),
```

**Root cause (verified):**

- `SchemaTask` (`src/schema/schema-manager.ts:15`) = `'analyze' | 'summary' | 'entity' | 'concept' | 'related' | 'conversation' | 'index' | 'lint' | 'merge' | 'full'` — **no `'compact'`**
- `wiki-engine.ts:478` signature is `buildSystemPrompt(task: SchemaTask)`
- `types.ts:665` erases the type: `buildSystemPrompt: (task: string) => Promise<string | undefined>` → TypeScript can't catch this
- `schema-manager.ts:294` `TASK_SECTIONS['compact']` → `undefined` → `wanted.includes(...)` → **`TypeError`**

The `'compact'` string you may have been thinking of is the **page-factory** union (`'full' | 'compact' | 'merge'`) — a completely unrelated type that never feeds `buildSystemPrompt`.

**Why tests pass:** `createMockContext.buildSystemPrompt` is stubbed to `async () => undefined` — real `TASK_SECTIONS` is never invoked.

**Fix:** `await this.ctx.buildSystemPrompt('analyze')` (semantically correct since this is part of source analysis).

---

## 🔴 P0 #2 — Misleading log disguises our code bug as "model indecision"

`src/wiki/source-analyzer.ts:690, 741-742`

The catch block swallows the `TypeError` from #1 and returns `null`; the caller then logs `(type undecided)` — pushing maintainers/users to switch provider when the actual cause is our own schema-selection typo.

**Fix:** Hoist `buildSystemPrompt` outside the try (our bug should be loud); distinguish `(classification call failed)` vs `(model returned an unusable label)`.

---

## 🔴 P0 #3 — Bypasses `custom` granularity cap

`src/wiki/source-analyzer.ts:628` + `:701/:703`

`ensureSourceLemma` runs **after** the `extractionGranularity === 'custom'` cap slice at `:602-607`, then `push`es directly to `analysis.entities` — exactly the invariant the slice was written to guarantee. Reproduction: `customEntityLimit: 1` → user expects 1 → gets 2 + extra LLM round-trip. `mergeBatchResults.remaining` accounting also bypassed.

**Fix:** Check cap **before** `classifyLemmaType` to avoid a wasted call.

---

## 🔴 P0 #4 — `slugKeys` duplicate reintroduces Issue #366 regression for Turkish vaults

`src/core/source-lemma.ts:35-42` vs `src/core/slug.ts:75-96`

`core/slug.ts` already exports `slugKeys(name, aliases, { turkishFold })` — added in v1.25.10 PATCH for #366. Local copy lacks `opts.turkishFold`. Reproduction: Turkish vault, note `İnsülin.md`, extraction produces `insülin` → `computeSlug('İnsülin')='i̇nsülin'` ≠ `'insülin'` → guard misses → duplicate page created.

**Fix:** Delete local `slugKeys`, `import { slugKeys } from './slug';`. Bonus: this also fixes #5 (punctuation-only names get `untitled-<Date.now()>`, non-deterministic keys).

---

## 🟡 P1 #5 — `parseTagList`绕过 `tagVocabularyMode` + 语义错位

`src/wiki/source-analyzer.ts:668-671`

Two independent issues:
1. **Mode gate bypass** — `tag-vocab.ts:3-19` `getActiveEntityTags(settings)` checks `tagVocabularyMode === 'custom'`; raw `parseTagList(customEntityTags)` skips it. Built-in-mode users with stale CSV silently get lemmas suppressed with no UI explanation. Conversely, built-in-mode users with empty CSV get `domainTags=[]` → the "Neurology/Pharmacology" guard is **dead for the vast majority of users**.
2. **Semantic mismatch** — `customEntityTags/customConceptTags` are **entity subtype vocabulary** (`person, organization, theory, method`), not domain names. Matching a note title against `person`/`method` is not the "field of study" test the doc-comment claims. Violates Schema Phase 1 Option A invariant (no dual-source problem between runtime params and settings).

**Recommendation:** Drop the domain-container heuristic from this PR. Reuse `isSourceOwnPageLemma` from `merge-triage.ts` as a guide; wait for #328 Phase 2 to land a real folder/domain registry. Happy to discuss alternatives if you have a concrete user scenario this would catch.

---

## 🟡 P1 #6 — `max_tokens: 32` + missing `silentOnEmpty` for thinking models

`src/wiki/source-analyzer.ts:730, 736`

`max_tokens: 32` is far below peer short-answer calls (`TOKENS_DEDUP_RESOLUTION=1000`, `TOKENS_QUERY_KEYWORDS=1000`, `TOKENS_QUERY_MODEL_DETECT=2000`). `disableThinking` defaults to `false`, so reasoning models exceed 32 tokens with their thinking block → empty body → noisy `console.error` ("JSON parse completely failed...") per ingested note.

**Fix:** Add `TOKENS_LEMMA_CLASSIFY = 2000` constant in `constants.ts` (peer-aligned); pass `silentOnEmpty: true` to `parseJsonResponse`.

---

## 🟡 P1 #7 — Hardcoded `type: 'other'` triggers lint violation under custom tag vocabulary

`src/wiki/source-analyzer.ts:696`

`getActiveEntityTags(s)` in custom mode returns only the user's CSV, no `'other'`. `scanTagViolations` flags the generated page. The feature manufactures the lint debt it asks users to fix.

**Fix:** Pick first active tag, or expose `'other'` as a documented fallback.

---

## 🟢 Cosmetic (optional, will fold in)

Dead code `keys.size === 0` (unreachable) · public surface reduction · 3 skip-and-log blocks → one site · `matchAliases` redundant `!== fileBasename` Set check · `sourceTitle` field rename · test `describe('slugKeys')`/`describe('parseTagList')` blocks (deleted with P0 #4 / P1 #5) · `analyzerWith` unused 2nd param · 2 redundant tests merge.

---

## Meta-finding — Mock fidelity

P0 #1 reached review only because **two mocks jointly conceal the failure**:
1. `createMockContext.buildSystemPrompt` → `async () => undefined` — never invokes real `TASK_SECTIONS[task].includes()`
2. `createMockClient` returns a default past end-of-array instead of throwing — makes "correctly skipped" indistinguishable from "never called"

Hardening those two mocks (suggest: `buildSystemPrompt` delegates to real `SchemaManager` + throws on non-`SchemaTask`; `createMockClient` throws on over-read) would convert this class of bug into an immediate test failure. Out of scope for this PR — separate cleanup.

---

## Plan

I'll push fixes to your branch (`fix/348-source-lemma-guarantee`) preserving your commit authorship (amend) and re-run Gate 1 + simplify + code-review. **For P1 #5 (domain-container heuristic), my recommendation is to drop it from this PR** — happy to discuss a different design if you have a concrete scenario in mind. Everything else I'll implement per the snippets above.

If you'd rather handle any of these yourself, just let me know which ones — I'll skip them in my push.

green-dalii:
Update: pushed the 2 P0/P1 fix commits to your branch as independent commits (so each one is independently auditable / revertable). Your original commit is preserved untouched on the PR head.

```
95268eb feat(ingest): guarantee a page for the source note's own subject (#348)  ← yours
825e21c fix(source-lemma): use canonical slugKeys; drop domain-container heuristic  ← fix #1
47bc328 fix(source-analyzer): correct buildSystemPrompt task; classify-first-then-cap; active-vocab type  ← fix #2
```

Quick rundown of what each commit changes (full rationale in the messages and my earlier review comment):

**Commit #1 — source-lemma.ts**

- Imports `slugKeys` from `core/slug` instead of having a local duplicate that lacked `turkishFold` (P0 #4 / Issue #366 regression). Added a regression test pinning the İ→i fold.
- Drops the `domain-container` skip reason and the `parseTagList` helper entirely (P1 #5 — semantic violation; that heuristic read `customEntityTags` as a "field-of-study" list which it isn't). Wait for #328 Phase 2 to land a real folder/domain registry before reintroducing.
- The wiring adjustment in `source-analyzer.ts` (removing the now-orphaned `parseTagList` import and `domainTags` field) is included in this commit because the diff wouldn't build green without it.

**Commit #2 — source-analyzer.ts + constants.ts**

- `buildSystemPrompt('compact')` → `buildSystemPrompt('analyze')` (P0 #1 — `'compact'` is not a `SchemaTask`, so the call threw on every real vault; only the mock-context stub passed tests).
- Hoisted `buildSystemPrompt` outside the try in `classifyLemmaType` and reworded the catch log so a `TypeError` from our own code is no longer reported as "type undecided" (P0 #2 — the misleading log pushed debugging in the wrong direction).
- Added an `exceedsCustomCap` check after `classifyLemmaType` (P0 #3 — the push used to bypass the extraction-phase cap slice, so `customEntityLimit: 1` got 2 entities + a wasted LLM call). Pinning test in `source-analyzer-lemma-guarantee.test.ts`.
- `type` field on the new candidate is now `firstActiveTag(target)` (entity/concept) instead of hard-coded `'other'` (P1 #7 — custom-vocabulary users don't have `'other'` in their active vocabulary, so the page was flagged out-of-vocab by `scanTagViolations`).
- New `TOKENS_LEMMA_CLASSIFY = 2000` constant in `src/constants.ts` (P1 #6 — peer-aligned with `TOKENS_QUERY_MODEL_DETECT = 2000`; the original `32` was unreachable for thinking-capable models). Also passed `silentOnEmpty: true` to `parseJsonResponse`.

Gate 1 is green on both commits: lint 0/0, tsc 0/0, 2731 tests pass, build clean, css-lint 0.

---

**Two things I deliberately did NOT change** (you may want to revisit):

1. **The `classifyLemmaType` LLM call itself.** Reviewers (and the review) flagged that this could be folded into the existing first-round extraction prompt's schema (ask for `subject_kind` in the same response where entities/concepts are extracted). I left this as-is because it's a real design change that touches the prompt registry / i18n — happy to take it on as a follow-up if you agree.
2. **`ensureSourceLemma` placement at the end of `analyzeSource`.** The Altitude agent suggested hoisting it into a `POST_PROCESS_POLICIES` array in `core/batch-merger.ts`. Same reasoning — it's a real refactor and deserves its own discussion.

**One thing I want to flag separately** (not blocking this PR, but worth tracking):

The mock-fidelity bug that hid P0 #1 — `createMockContext.buildSystemPrompt` is stubbed to `async () => undefined`, and `createMockClient` returns a default past the end of the response array rather than throwing. Together they made a never-executing feature look fully tested. Hardening those two mocks would convert this whole class of bug into an immediate test failure. I'll file a separate issue for it.

Let me know if you want any of the design-side changes (1) or (2) folded into this PR, or kept for follow-up.

DocTpoint:
Thank you for the review — all four P0 findings are correct, and I have no objection to any of the fixes.

**On P0 #1, one measurement to add.** I checked our deployed build, which carries this patch as a local commit on top of `1.25.5`. `grep` finds `buildSystemPrompt("compact")` once in the shipped `main.js`. So the feature has been inert in a production vault since 25 July, and it did not surface because that vault has not ingested since. The lemma guarantee never wrote a page, and the `(type undecided)` line reported our own `TypeError` as model indecision — exactly as you describe. Your review caught this before our smoke test did.

**Your meta-finding is the more valuable one.** Two mocks made "correctly skipped" indistinguishable from "never called". That is the same failure mode as measuring a merge on the resulting file instead of in the code path: both states produce identical evidence. Hardening those two mocks is worth its own issue, and I support it.

**Question 1 — folding `classifyLemmaType` into the extraction round: yes, and the reason is not the saved call.**

The separate call receives only `name` and `summary`. The full source text — which the model has just read in order to extract from it — is gone by then. For a note titled `DORA` or `MEMS`, deciding entity vs. concept without that text is close to guesswork; with it, the answer is usually obvious. The context exists at extraction time and never again afterwards.

There is also a consistency argument. The extraction already classifies into exactly these two classes, for every node it returns. The lemma is the one node we ask about separately, in a second conversation, and nothing makes the two answers agree.

The saved round-trip matters too — one call per note over a full-corpus run — but it is the weakest of the three reasons.

One thing that should not change with the fold: `decideSourceLemma` must keep deciding *whether* the lemma is missing, after all batches have accumulated. Only the type answer moves into the first response, where it rides along at no extra cost even in the majority of cases where it is never used.

**Question 2 — hoisting into `POST_PROCESS_POLICIES`: I would leave it where it is.**

That array does not exist in `core/batch-merger.ts` yet. Introducing it now means building a policy structure for exactly one policy, and the call site in `analyzeSource` is currently readable as a single line with a comment explaining why it runs after accumulation. If a second post-processing step appears, the abstraction will have two users and a shape derived from both rather than guessed from one. I would rather wait for that.

PR #370, opened 2026-07-28
fix(ingest): scope folder ingest on the folder boundary (#364)

### LINKED ISSUE #364: Folder ingest has no folder boundary — sibling folders sharing a name prefix are ingested too
**Plugin version:** current `main` (`2d26620`)

## Summary

"Ingest from folder" collects files with a bare prefix comparison, so selecting a folder also pulls in every **sibling** whose name starts with the same string.

`src/main-commands/ingest-commands.ts:136`:

```js
const files = this.app.vault.getFiles()
  .filter(f => f.path.startsWith(folder.path) && allowedExts.includes(f.extension.toLowerCase()));
```

`"Notizen-temp/x.md".startsWith("Notizen")` is `true`. There is no separator check and no root special-case.

## Repro

Create `Foo/` and `Foo-bar/`. Ingest from folder → select `Foo` → the files in `Foo-bar/` are ingested as well, with no warning and no indication in the confirmation dialog.

## Impact

This is a data-safety issue rather than a cosmetic one: the user asks for one folder and silently gets several. In my own vault, selecting `Notizen/` (8 notes staged for a batch) also pulled in `Notizen-temp/` — 409 notes. Because collection is alphabetical and underscore sorts first, it began with `_Handbuch.md`, my own documentation file, and generated wiki pages about the plugin's own UI concepts before I noticed.

The blast radius scales with model cost and with how much unwanted content reaches the graph, and undoing it means deleting pages by hand.

## Fix

A boundary check, with the root folder handled explicitly:

```js
const folderPrefix = folder.isRoot() ? '' : `${folder.path}/`;
const files = this.app.vault.getFiles()
  .filter(f => f.path.startsWith(folderPrefix) && allowedExts.includes(f.extension.toLowerCase()));
```

Happy to open a PR with a regression test if useful.

DocTpoint:
## Problem

`selectFolderToIngest` filtered candidates with `f.path.startsWith(folder.path)`. That is a string-prefix test, not a descendant test, and two things leak through it:

| Picked folder | Also swept in | Why |
|---|---|---|
| `Notizen` | `Notizen-temp/x.md` | sibling folder sharing a name prefix |
| `Notizen` | `Notizen.md` | a file sitting beside the folder |

The first is the reported case. The second falls out of the same root cause and is fixed by the same change.

## Change

Anchor the comparison on a trailing slash. The vault root keeps an empty prefix, so every path remains a descendant of it.

The rule lives in `src/core/folder-scope.ts` as a pure, IO-free helper rather than inline at the call site, because `selectFolderToIngest` opens a modal and cannot be unit-tested without an Obsidian vault. The helper takes two primitives, so the boundary cases are testable directly. This mirrors the shape used elsewhere in `core/` for decision rules.

## Tests

11 new tests in `src/__tests__/core/folder-scope.test.ts` — 4 on the prefix derivation, 7 on the predicate.

**Mutation-checked rather than only green:** removing the anchoring fails exactly 6 of them — both prefix cases, sibling folder, neighbouring file, the folder itself, and nested name-prefixed neighbours. The 5 controls stay green: root handling, true descendants at depth, empty path, and the folder name appearing later in a path.

## Related call sites — deliberately not in this PR

The same unanchored comparison appears in seven other places. I am not touching them here, because this is a PATCH cycle and they split into two different failure directions that deserve their own issue:

| Site | Direction | Consequence with `wikiFolder: "wiki"` |
|---|---|---|
| `wiki/lint/delete-empty-stubs.ts:10` | over-include | a vault folder `wiki-archive` enters the **stub deleter** |
| `wiki/lint/get-existing-pages.ts:12` | over-include | counted as wiki pages |
| `wiki/lint/phases/preparation.ts:23` | over-include | counted as wiki pages |
| `schema/auto-maintain.ts:407` | over-include | scanned by Phase 2 quick-fixes |
| `core/orphan-matcher.ts:81` | over-include | treated as a wiki path |
| `ui/modals/suggest-modals.ts:29` | over-exclude | user's `wiki-archive` notes vanish from the ingest picker |
| `ui/modals/MultiFileSuggestModal-class.ts:101` | over-exclude | same, multi-file picker |

Four other sites already anchor correctly (`ensure-welcome-note.ts:179`, `prompt-builders.ts:126`, `auto-maintain.ts:550`, `lint/scanners.ts:268`), so the codebase is inconsistent rather than uniformly wrong. Happy to open that as a separate issue if you want it tracked.

## Gate

Baseline counted on `2d26620` itself, with `pnpm build` before `pnpm test`.

| | baseline | this branch |
|---|---|---|
| `tsc --noEmit` | 0 | 0 |
| `pnpm lint --max-warnings=0` | 0 | 0 |
| `pnpm css-lint` | 0 | 0 |
| `pnpm build` | clean | clean |
| `pnpm test` | 2629 / 197 | 2640 / 198 |

+11 tests / +1 file — exactly the new ones.

Closes #364.

### DIFF
--- a/src/__tests__/core/folder-scope.test.ts
+++ b/src/__tests__/core/folder-scope.test.ts
@@ -1,0 +1,60 @@
+import { describe, it, expect } from 'vitest';
+import { folderScopePrefix, isInFolderScope } from '../../core/folder-scope';
+
+describe('folderScopePrefix', () => {
+  it('anchors a folder path on a trailing slash', () => {
+    expect(folderScopePrefix('Notizen', false)).toBe('Notizen/');
+    expect(folderScopePrefix('a/b', false)).toBe('a/b/');
+  });
+
+  it('returns an empty prefix for the vault root', () => {
+    expect(folderScopePrefix('/', true)).toBe('');
+    expect(folderScopePrefix('', true)).toBe('');
+  });
+
+  it('does not double the separator on an already-trailing slash', () => {
+    expect(folderScopePrefix('Notizen/', false)).toBe('Notizen/');
+    expect(folderScopePrefix('Notizen//', false)).toBe('Notizen/');
+  });
+
+  it('treats an empty non-root path as unscoped rather than emitting a bare slash', () => {
+    expect(folderScopePrefix('', false)).toBe('');
+  });
+});
+
+describe('isInFolderScope', () => {
+  it('matches true descendants at any depth', () => {
+    expect(isInFolderScope('Notizen/a.md', 'Notizen', false)).toBe(true);
+    expect(isInFolderScope('Notizen/sub/deep/a.md', 'Notizen', false)).toBe(true);
+  });
+
+  // Issue #364, the reported case.
+  it('does not match a sibling folder sharing a name prefix', () => {
+    expect(isInFolderScope('Notizen-temp/a.md', 'Notizen', false)).toBe(false);
+    expect(isInFolderScope('Notizen2/a.md', 'Notizen', false)).toBe(false);
+  });
+
+  // Same root cause, second symptom: an unanchored prefix also swallows a
+  // file that merely starts with the folder's name.
+  it('does not match a file sitting beside the folder', () => {
+    expect(isInFolderScope('Notizen.md', 'Notizen', false)).toBe(false);
+  });
+
+  it('does not match the folder itself', () => {
+    expect(isInFolderScope('Notizen', 'Notizen', false)).toBe(false);
+  });
+
+  it('scopes nested folders without leaking into name-prefixed neighbours', () => {
+    expect(isInFolderScope('a/b/x.md', 'a/b', false)).toBe(true);
+    expect(isInFolderScope('a/bc/x.md', 'a/b', false)).toBe(false);
+  });
+
+  it('accepts every path at the vault root', () => {
+    expect(isInFolderScope('a.md', '/', true)).toBe(true);
+    expect(isInFolderScope('deep/nested/a.md', '/', true)).toBe(true);
+  });
+
+  it('is unaffected by the folder name appearing later in the path', () => {
+    expect(isInFolderScope('Archiv/Notizen/a.md', 'Notizen', false)).toBe(false);
+  });
+});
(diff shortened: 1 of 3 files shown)


green-dalii:
Reviewing your PR now. Your mutation tests + the second case (`Notizen.md` beside the folder) + the four-correct-sites audit are stronger than my local #364 fix on `feat/v1.25.10-patch`. I'll cherry-pick your version into the branch and credit you as co-author on the consolidated commit. Closing this PR in favor of the merged result — let me know if you spot anything I miss in the consolidation.

green-dalii:
Merged in #373 — adopted your version in full (with attribution: `Co-authored-by: DocTpoint <tatamis.gruppe_6k@icloud.com>`). Two specific gains over my local helper that I'm glad to keep:

  1. The `Notizen.md` mutation test — the third case my local fix covered only by accident, now pinned.
  2. The audit of the seven other call sites with leaky `startsWith` — flagged as out-of-scope for this PATCH as you noted; we will pick that up in a follow-up.

Thanks for the cleaner separation of `folderScopePrefix` (pure derivation) from `isInFolderScope` (descendant predicate) — it makes the test file read top-to-bottom without setup.

green-dalii:
Hi @DocTpoint — your PR #370 has shipped in **v1.25.10**: https://github.com/green-dalii/obsidian-llm-wiki/releases/tag/1.25.10

Adopted in full into commit `98afe42` with `Co-authored-by: DocTpoint <tatamis.gruppe_6k@icloud.com>`. The 11 mutation tests (including the `Notizen.md` beside the folder case) are in `src/__tests__/core/folder-scope.test.ts` and pin the third case my local helper had covered only by accident.

Two specific gains over my local #364 fix I'm glad to keep:
1. The `Notizen.md` mutation test — pinned, no longer accidental.
2. The audit of seven other call sites with leaky `startsWith` — flagged as out-of-scope for this PATCH. If you want to drive that consolidation as a v1.26.0 item, please open an issue and I'll link it from #328 Phase 2 alongside the per-type registration work.

Thanks for the cleaner separation of `folderScopePrefix` (pure derivation) from `isInFolderScope` (descendant predicate) — the test file reads top-to-bottom without setup, which is the read order that matches how the helper is taught.

Issue #363, opened 2026-07-27
Mentions in Source renders `— [[|]]` when `mentions_with_provenance[].source_path` is blank


Chris-Hestro:
**Plugin version:** 1.25.9 · **Model:** claude-sonnet-5 · **Obsidian:** desktop (Windows)

## Summary

Every quote under `## Mentions in Source` can render with an **empty wikilink** — `— [[|]]`. In
graph view all of them collapse into a single nameless ghost node; clicking it raises Obsidian's
*"File name cannot contain any of the following characters"* notice, because the link target is an
empty string.

In my vault this hit **134 links across 72 pages** (older build, near-100%). On 1.25.9 it is much
rarer but still reproducible: pages *created during a re-ingest of the same source* still get it —
4 of 19 pages in one run.

## Root cause

`formatMentionsSection()` — the structured branch takes `source_path` straight from the LLM output
and never falls back to the `sourcePath` argument the function already receives:

```js
const entries = structured ? mentions.map((m) => ({
  quote: m.quote,
  translation: m.translation,
  sourcePath: m.source_path,      // <-- blank when the LLM omits it
  extractedAt: m.extracted_at
})) : mentions.map((quote) => ({ quote, sourcePath }));
```

`buildBullets()` then does `const leftPath = e.sourcePath.replace(/\.md$/, "")` and emits
`[[${leftPath}|${displayName}]]` → `[[|]]`.

A blank `source_path` is an expected state, not a malformed response — the auto-provenance builder
sets it deliberately:

```js
const provenance = quotes.map((quote) => ({
  quote,
  source_path: "",   // filled by page-factory at write time
  source_slug: "",   // filled by page-factory at write time
  extracted_at: now2
}));
```

…and the prompt documents `mentions_with_provenance` as **OPTIONAL** ("When omitted, the system
auto-generates provenance from `mentions_in_source`").

The sibling function already applies the fallback correctly:

```js
// computeReingestMentions
const normalize = (m) => ({ ...m, source_path: (m.source_path || defaultSourcePath || "").replace(/\.md$/, "") });
```

## Evidence it is not a configuration problem

One page, one generation pass — the source is known in the frontmatter and blank in the citations:

```yaml
sources: ["[[sources/<source-slug>]]"]
- "<verbatim quote from the source>" — [[|]]
```

Another page from the same run is the exact mirror: empty `sources:` frontmatter, **correct**
citation link. Two independent code paths, each failing on its own.

[...]

Issue #289, opened 2026-07-14
Pre-#244 Mentions sections are permanently frozen by the #267 fail-safe (no migration path)


DocTpoint:
Before `1c1db3a` (#244) the prompt asked the model to write the Mentions section as a blockquote grouped by source:

```
> **Source: [[sources/X|X]]**
> - "quote"
```

Since #244 the section is written programmatically and flat:

```
- "quote" — [[sources/X|X]]
```

The #267 parser (`BULLET_RE` in `core/mentions-parser.ts`) only accepts the flat form. A legacy section is therefore unparseable, so `computeReingestMentions` returns `preserveRaw`, the fail-safe fires, and this source's mentions merge is skipped.

**The fail-safe is right** — dropping curated quotes is worse than skipping the merge. But there is no migration path: the legacy format is never upgraded, so the fail-safe fires again on the next ingest, and the next. Those pages never accumulate another quote for the rest of the vault's life. Silently — the only trace is a `console.warn` that is easy to miss.

## Scale

In my vault, 45 of 324 pages (14%) were in the legacy format. Any vault first ingested before 2026-07-06 is affected, in proportion to how much of it predates #244.

## Two ways out

(a) Teach the parser to *read* the legacy shape and re-emit it flat. Self-healing, needs no user action, and the page is upgraded the first time it is merged again.

(b) Ship a one-shot migration command.

I lean towards (a) — it needs nothing from the user and cannot be forgotten. I wrote a deterministic migration for my own vault (blockquote → flat, invariance-gated, every emitted line re-checked against `BULLET_RE` itself) and am happy to contribute in either direction.

green-dalii:
Real bug + clean two-option analysis. (a) parser self-healing is clearly better than (b) one-shot migration — the user's vault should not have to know about internal schema changes, and (a) means the fix is automatic the next time a legacy page merges.

Targeted at v1.25.1 PATCH alongside #288 / #272 / #292 / #290. Please open a follow-up PR implementing (a). Rough budget ~150-200 LOC including the parser extension + invariance-gated re-emit tests. After Gate 1 green on a v1.25.0 rebase, ping here.

Labels: `enhancement`. Milestone: `v1.25.1 PATCH`.

DocTpoint:
Gate 1 is green on a rebase onto current `main` (5e13ac5): lint 0/0, tsc 0, 2203 → 2213 tests, build clean, css-lint 0. PR: #303 — implements option (a).

[...]

PR #371, opened 2026-07-29
fix(mentions): tolerate an empty or absent citation target (#363)

DocTpoint:
Closes the parser half of #363, as scoped in ROADMAP v1.25.10 ("DocTpoint contributes parser side"). The formatter one-liner on your side (`sourcePath: m.source_path || sourcePath`) is deliberately **not** in this PR.

### Why this is not cosmetic

`BULLET_RE` required a non-empty link target. A line ending in `— [[|]]` therefore fails to parse, `computeReingestMentions` returns `preserveRaw`, the #267 fail-safe fires, and the mentions merge for that page is skipped. Not once — on every later ingest. The page never accumulates another quote.

So repairing the producer stops new pages from being poisoned, but it does not release the pages that already carry a `[[|]]`. This PR makes the reader tolerant, which is how the other two failure shapes in this section were handled before: the garbled header (`canonicalizeSectionHeaders`, `c4e47ba`) and the legacy grouped blockquote (#289).

### The two halves are one fix

* **Formatter:** an empty target is no longer expressible. A blank `source_path` is a legitimate state — the auto-provenance builder writes `source_path: ''` for the page-factory to fill at write time — so the quote is emitted **without** a citation instead of with a link that points nowhere. Same rule applied in conversation mode.
* **Parser:** accepts both that citation-less shape and the `[[|]]` already sitting in vaults, recovering an empty `source_path`. `computeReingestMentions` fills it from `defaultSourcePath`, so the page repairs itself on its next merge.

Shipping only the formatter half would trade one unparseable shape for another: the citation-less bullet would itself fail `BULLET_RE` and freeze the page.

**One judgment call, stated plainly.** Filling the blank from the source being merged is an *inference*, not a recovery — an empty target carries no attribution, so the quote is attributed to the source currently ingesting. That is the same rule #289 already applies to a quote under a group header, and it is strictly more than `[[|]]` conveyed. If you would rather leave such quotes unattributed, the change is one line in `computeReingestMentions` and I will follow your call.

### Measured on my corpus, re-measured today

The 4.7 % (119/2586 pages) I reported in the thread came from the full-corpus run of 9272 writes. **On today's vault the same probe gives 23 frozen pages (0.9 %), of which this patch releases 22.** The difference is repair passes run since that measurement, not a correction to it — the two numbers measure different vault states.

The one page that stays frozen is a quote reflowed across two lines. That is a genuine hand-edit and the fail-safe should keep holding it.

| | pages |
|---|---|
| with a Mentions section | 2586 |
| frozen, current parser | 23 (0.9 %) |
| frozen, this PR | 1 |
| released | 22 |

Probe self-checked against six fixtures with known answers before the corpus numbers were taken.

### Gate

`tsc --noEmit` 0 · `lint --max-warnings=0` 0 · `css-lint` 0 · `build` clean · **tests 2638 / 197 files against a self-counted baseline of 2629 / 197 on `0c9b0b6` = +9, exactly the new ones** (`pnpm build` before `pnpm test`).

Mutation-checked, both halves separately: reverting the parser tolerance fails exactly 5 tests and leaves the 4 controls green; reverting the formatter guard fails exactly 2. Under the formatter mutation alone the round-trip test still passes — which is the interlock above, visible as a test result.

### Not in scope

The other line-level failure I reported in that thread — nested parentheses inside the gloss group, `\(([^)]*)\)` forbidding an inner paren — is untouched here. It no longer occurs anywhere in my current corpus, so I have no live measurement that would justify widening the regex in a PATCH. Happy to open a separate issue if you want it tracked.

### DIFF
--- a/src/__tests__/core/mentions-formatter-roundtrip.test.ts
+++ b/src/__tests__/core/mentions-formatter-roundtrip.test.ts
@@ -301,3 +301,117 @@ describe('#289 — legacy grouped-blockquote Mentions sections', () => {
     ]);
   });
 });
+
+// ─── #363: citation with no resolvable target ────────────────────────────
+//
+// An empty wikilink `[[|]]` reached pages through `formatMentionsSection`
+// taking a blank `source_path` straight from the model. It reads as cosmetic
+// and is not: `BULLET_RE` required a non-empty target, so the line failed to
+// parse, `computeReingestMentions` returned `preserveRaw`, the #267 fail-safe
+// fired, and the page stopped accumulating quotes on every later ingest.
+// Measured on a 417-note corpus (9272 writes): 119 frozen pages, 104 of the
+// 144 unparseable lines were empty backlinks.
+//
+// The two halves below are one fix, not two: the formatter stops writing an
+// empty target (it emits the quote without a citation), and the parser accepts
+// both that shape and the `[[|]]` already sitting in vaults. Shipping only the
+// formatter half would trade one unparseable shape for another.
+
+describe('#363 — empty and absent citations', () => {
+  const frozen = [
+    `## ${LABEL}`,
+    '',
+    '- "ein eingefrorener Beleg" — [[|]]',
+    '- "mit Übersetzung" (with translation) — [[|]]',
+  ].join('\n');
+
+  it('parses a bullet whose citation target is empty', () => {
+    const r = parseMentionsSection(frozen, LABEL);
+    expect(r.found).toBe(true);
+    expect(r.fullyParsed).toBe(true); // the fail-safe no longer fires
+    expect(r.mentions.map(m => m.quote)).toEqual(['ein eingefrorener Beleg', 'mit Übersetzung']);
+    expect(r.mentions.map(m => m.source_path)).toEqual(['', '']);
+    expect(r.mentions[1].translation).toBe('with translation');
+  });
+
+  it('parses a bullet that carries no citation at all', () => {
+    const body = [`## ${LABEL}`, '', '- "a quote with no source"'].join('\n');
+    const r = parseMentionsSection(body, LABEL);
+    expect(r.fullyParsed).toBe(true);
+    expect(r.mentions).toEqual([mention({ quote: 'a quote with no source', source_path: '' })]);
+  });
+
+  it('[control] a real target is still required to be recovered verbatim', () => {
+    const r = parseMentionsSection([`## ${LABEL}`, '', '- "q" — [[notes/A|A]]'].join('\n'), LABEL);
+    expect(r.fullyParsed).toBe(true);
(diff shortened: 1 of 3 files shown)


green-dalii:
Reviewing your PR now. The round-trip interlock test + the conversation-mode fallback + the `renderCitation` single-render-gate are stronger than my local #363 fixes (which I had split into two commits and only patched the data-layer entry). I'll cherry-pick your version into `feat/v1.25.10-patch` and credit you as co-author on the consolidated commit. Closing this PR in favor of the merged result — your attribution is preserved via Co-authored-by trailer.

green-dalii:
Merged in #373 — adopted your version in full (with attribution: `Co-authored-by: DocTpoint <tatamis.gruppe_6k@icloud.com>`). Three specific gains over my local two-commit fix that I'm glad to keep:

  1. `renderCitation(leftPath)` as a single render gate — my data-layer fallback (`m.source_path || sourcePath`) silently rewrote the attribution of an empty-sourcePath mention to the current source. Your render-layer fix keeps the empty value, so a later re-merge fills it from the real source.
  2. Conversation mode handled in the same `renderCitation` path — I had not touched that branch.
  3. The round-trip interlock test — under my two-commit split, shipping only the formatter half would have left the citation-less bullet unparseable (and freeze the page). Your test would have caught that.

Thanks for the single `renderCitation` function — conversation mode, regular quote, and truncated-head all share it, which is exactly the deduplication I needed.

green-dalii:
Hi @DocTpoint — your PR #371 has shipped in **v1.25.10**: https://github.com/green-dalii/obsidian-llm-wiki/releases/tag/1.25.10

Adopted in full into commit `292d42e` with `Co-authored-by: DocTpoint <tatamis.gruppe_6k@icloud.com>`. The round-trip interlock tests in `src/__tests__/core/mentions-formatter-roundtrip.test.ts` (`#363 — empty and absent citations` describe block) pin the formatter ↔ parser contract so neither half can ship alone again.

Three specific gains over my local two-commit fix I'm glad to keep:
1. `renderCitation(leftPath)` as a single render gate — my data-layer fallback (`m.source_path || sourcePath`) silently rewrote the attribution of an empty-sourcePath mention to the current source's path. Your render-layer fix preserves the empty value, so a later re-merge fills it from the real source. This is strictly more correct than the data-layer approach.
2. Conversation mode handled in the same `renderCitation` path — I had not touched that branch, and unifying it removed a duplicate `displayName.split('/').pop()` line.
3. The round-trip interlock test — under my two-commit split, shipping only the formatter half would have traded one unparseable shape (`[[|]]`) for another (citation-less). Your test would have caught that.

Thanks for the diagnosis in the #363 thread (the `preserveRaw → fail-safe → no further accumulation` chain) — that's what made the freeze mechanism visible and shaped the fix scope.

Issue #65, opened 2026-05-31
response_format: json_object` causes API errors with LM Studio 0.4.15+


DocTpoint:
---

When using an OpenAI-compatible local API endpoint (LM Studio ≥ 0.4.15), all ingest and analysis calls fail with an API error. LM Studio 0.4.15 dropped support for `response_format: { type: "json_object" }` — it now only accepts `{ type: "json_schema", json_schema: {...} }` or no `response_format` field at all.

The plugin sends `response_format: { type: "json_object" }` in every structured-output request, which causes LM Studio 0.4.15+ to return an error response and abort the ingest.

**To Reproduce**

1. Install LM Studio ≥ 0.4.15
2. Configure the plugin to use the local OpenAI-compatible API
3. Trigger any ingest

**Expected behavior**

Ingest completes normally.

**Actual behavior**

API returns an error (`response_format type 'json_object' is not supported` or similar). Ingest fails.

**Suggested fix**

For the OpenAI-compatible client, remove `response_format` from the request body entirely. JSON output is already reliably enforced by two existing mechanisms in the plugin:

1. The system prompt instructs the model to output JSON
2. The assistant message is prefilled with `"{"` (already implemented in the plugin at the request-building step)

These two mechanisms are sufficient — `response_format: json_object` is redundant for well-instructed models and breaks compatibility with LM Studio 0.4.15+.

Alternatively, a setting like **"Structured output mode"** (`json_object` / `none`) would let users with older LM Studio versions or other endpoints keep the current behavior.

**Environment**

- Plugin version: 1.12.6
- LM Studio: 0.4.15+
- Endpoint: OpenAI-compatible local API

---

Issue #358, opened 2026-07-27
[design] v1.26.0 design — complementary memory model + consolidation kernel


green-dalii:
## What this issue tracks

This is the canonical design anchor for v1.26.0 MINOR. It synthesizes the four open items that all touch the same architectural decision — **how the plugin relates to its source notes** — and serves as the integration surface for Phase 2 / Phase 3 reviews.

## Why one issue instead of four

| Issue | Surface area | Architectural decision |
|---|---|---|
| [#220](https://github.com/green-dalii/obsidian-llm-wiki/issues/220) | source-revision awareness | how the plugin tracks that a source note changed after wiki pages were derived from it |
| [#285](https://github.com/green-dalii/obsidian-llm-wiki/issues/285) | OKF + typed edges | how the plugin exposes typed predicates on wiki links |
| [#328](https://github.com/green-dalii/obsidian-llm-wiki/issues/328) | schema three-layer split | what belongs in user schema vs runtime params vs engine |
| [#330](https://github.com/green-dalii/obsidian-llm-wiki/issues/330) | ingest-path critique | where decisions are made and whether they can be revisited |

Each is a distinct surface, but the underlying decision is shared. This tracking issue is where the integration lives.

## Design philosophy — complementary memory model

Source notes are **episodic memory**: sequential, lossy-never-intended, preserves authorial voice, hesitation, retraction. They serve the query "what did the source say, verbatim?".

Wiki pages are **semantic memory**: consolidated, abstracted, indexed, graph-traversable. They serve the query "across many sources, what is the relation between X and Y?".

Neither replaces the other. Both are necessary. The plugin's design target is to expose a **complementary query surface**, not to maximise fidelity to source.

The framing was developed in the cross-disciplinary discussion of #330, with two analogies (signal-processing and neuroscience) tested for fit. A signal-processing analogy (time-domain ↔ frequency-domain projection, with explicit caveats about Parseval, linearity, canonical basis) sharpens the trade-off but the analogy breaks on the lossy / non-linear / multi-basis axes. A neuroscience analogy (episodic ↔ semantic memory, consolidated via the hippocampal-cortical transfer during sleep) maps better onto the wiki / source relationship. The discussion on #330 captures the pushback and the limits of both analogies.

### Implications

- Wiki pages do not need to win every query to be valuable. They win the queries they are designed for; source notes win the rest.
[...]

Issue #275, opened 2026-07-12
selectSeedsWithLLM returns empty body with deepseek-v4-pro + json_object + thinking


green-dalii:
## Summary

Query Wiki's PPR (Personalized PageRank) path silently falls back to lex-only when `selectSeedsWithLLM` receives an empty response from the LLM. The user observes correct answers via lex, but the LLM-augmented PPR seed selection never contributes.

## Reproduction

1. Settings → LLM Provider = DeepSeek, model = `deepseek-v4-pro`, base URL = `https://api.deepseek.com/v1`
2. Keep **Enable Thinking** ON (default)
3. Obsidian → Query Wiki → ask any multi-hop question

Expected: PPR cascade selects 5 pages via (Index + LLM) augmentation.
Actual: `arm: index` only — PPR returns 0 seeds.

## Observed logs (2026-07-12)

```
[selectSeedsWithLLM] query model: deepseek-v4-pro
[LLM response] Seed selection: length=0, first100=""
parseJsonResponse parsing started... response length: 0
JSON parse completely failed (length 0)
[LLM retry] Seed selection attempt 1/3 failed: parseJsonResponse returned null or non-array seeds
... 3 retries, all length=0 ...
[LLM retry] Seed selection giving up after 3 attempts: parseJsonResponse returned null or non-array seeds
[QueryView PPR] LLM returned 0 seeds
[Step 2] PPR cascade selected 5 pages (arm: index)
```

Same provider URL → chat streaming endpoint works (1280 chunks OK). Only the non-streaming `/v1/chat/completions` with `response_format: { type: 'json_object' }` returns empty.

## Workaround

Settings → LLM Provider → uncheck **Enable Thinking**. The seed-selector already passes `disableThinking: true → enableThinking: false` (line 95 of `src/wiki/query-engine/pipeline/seed-selector.ts`), so this restores the JSON output for DeepSeek-V4.

## Root-cause hypothesis

DeepSeek-V4's non-streaming + `response_format: json_object` mode returns an empty body when thinking is enabled and the model is asked for a structured JSON output within a small token budget (200 max_tokens). Same model returns normal streaming chat responses, so this is a DeepSeek-side limitation, not an Obsidian-side bug.

## Related

- PR #267/#244 (commit `1d943ea`) addressed a related case — "DeepSeek in JSON mode returns empty body if no system message is provided". This issue is the post-fix next layer: system message is now present, but json_object + thinking + small max_tokens still produces empty body.
[...]

Issue #76, opened 2026-06-01
resolvePagePath: empty JSON response with thinking models — TOKENS_DEDUP_RESOLUTION too small


DocTpoint:
**Provider:** Custom OpenAI-compatible (LM Studio 0.4.x)
**Model:** google/gemma-4-26b-a4b (thinking mode ON)
**Plugin version:** 1.14.0

**Problem**

During page generation, `resolvePagePath` repeatedly logs:

> JSON parse completely failed (length 0)
> first 200 chars after normalization: [empty]

**Root cause**

`TOKENS_DEDUP_RESOLUTION = 300`. With thinking-capable models (Gemma 4, QwQ, DeepSeek-R1), the model emits a `<think>...</think>` block before the actual response. At 300 tokens, the entire budget is consumed by the thinking block. After `stripThinkingTokens` removes it, the remaining content is an empty string.

**Workaround**

Patch `TOKENS_DEDUP_RESOLUTION = 1000`.

**Suggested fix**

Increase the default to at least 1000, or dynamically add a thinking buffer when a thinking-capable model is detected (ties into the per-call thinking budget feature requested in #72).

green-dalii:
Hi @DocTpoint,

Great catch on the token budget exhaustion. The  vs thinking block interaction is a real edge case.

We've mapped this as a symptom of the same root cause as #72 — without a way to disable thinking on JSON calls, the budget gets eaten by reasoning tokens. Bumping to 1000 helps your immediate case but the systematic fix is per-call thinking control (#72) plus a  safety net that strips think blocks before JSON parsing.

The workaround you found () is solid for now. We'll address the full fix in the same v1.16.0 batch as #72/#75.

Thanks again for the detailed reports across all three issues.

green-dalii:
@DocTpoint — quick follow-up. The `parseJsonResponse` safety net I mentioned earlier: adding `extractJsonObject` fallback that scans for the first balanced `{ ... }` when strict parse fails. Won't fix every "thinking ate the budget" case but recovers the middle ground (Gemma 4's "Yes. Check..." preamble is parseable).

Root fix is #72, same PR as #75. Bumping `TOKENS_DEDUP_RESOLUTION` to 1000 as a workaround for anyone stuck, but I'm not making it the default — papers over the bug for non-thinking models.

[...]

PR #372, opened 2026-07-29
feat(tools): headless ingest CLI, with the sampling controls it needs

eucher:
Running the ingest engine outside Obsidian, and being able to say what changed between two runs.

## What this adds

A headless CLI (`tools/wiki-ingest-cli`) that runs the real `WikiEngine.ingestSource` against a vault on disk under Node. Nothing is reimplemented: the engine, the analyzer, the page factory, the schema manager and the LLM client are the production ones. Obsidian is supplied by a shim next to the CLI, and esbuild rewrites every `obsidian` import to it, so plugin code and CLI code share one `TFile` class and `instanceof` keeps working.

`--dry-run` keeps every write in memory; `--extract-only` stops after extraction and implies it, so a run that cannot write cannot touch the vault by forgetting a second flag.

## What changed after review

This branch originally corrected the `providerOptions` key, which decides whether raw fields reach the request body at all. @DocTpoint showed that doing so re-enables `response_format` on the OpenAI-compatible path, and measured against LM Studio 0.4.20 by curl — outside both the plugin and the SDK — that `{"type":"json_object"}` answers `400 'response_format.type' must be 'json_schema' or 'text'`, while `json_schema` answers 200. That is #65, and it is not historical.

`752576e` reverts the key correction rather than gating it, and removes `response_format`. The key decides delivery of three raw fields — `thinking`, `chat_template_kwargs`, `repetition_penalty` — to every provider on that path at once, and there is no backend for which delivering them is known to help: the same test shows LM Studio accepting `chat_template_kwargs` with a 200 while the model keeps reasoning, `ollama/ollama#16240` is an open feature request for the same thing, and llama-server documents `repeat_penalty` rather than the spelling this client sends. That belongs with the structured-output work, where the direction is the SDK's own `responseFormat` argument with `supportsStructuredOutputs` set — which needs no raw key at all.

So this branch no longer changes what any provider receives for an existing configuration. `extractionTopP` and `samplingSeed` are new and default to unset; the one behaviour change is extraction now honouring `disableThinking`.

## Fixes underneath

`top_p` was not threaded while `temperature` was. A sampling preset is the pair, so overriding one and leaving the other to the server runs on half of each. It is now threaded through the OpenAI, Anthropic, Codex and OpenAI-compatible clients. `seed` is threaded where it means something, which is fewer places: Anthropic has no such parameter, the Codex adapter omits it deliberately, and the `openai` provider builds the Responses model, which reports `seed` unsupported and leaves it out of the body. All three are said in the clients, where the argument is assembled.

Extraction did not consult `disableThinking`. It was not the only such call site — the lint alias and tag runners and the PDF converter still do not, and the welcome note and the connection probe omit it deliberately — but it is the one this branch is about. On the paths whose provider keys are correct this now takes effect; on the OpenAI-compatible path it does not, for the reason above.

## Testing what reaches the wire

The existing test for `providerOptions` asserted the argument handed to `generateText`, which is exactly the value the SDK then discards — it recorded the intent and was blind to the outcome, which is how the wrong key stayed correct-looking. The new test stubs `fetch` and reads the request body. It is written in the direction the branch now takes: `seed` and `top_p` are asserted present, and `thinking`, `chat_template_kwargs`, `repetition_penalty` and `response_format` asserted absent, with the reason. Correcting the key fails this test, which is the point — it cannot land unnoticed. A second case covers the ordinary install: `response_format` requested, nothing else configured, nothing added.

## Type-checking `tools/`

`tsconfig.json` covers `src/**`, and `run-ingest.mjs` bundles with esbuild, which strips types without reading them, so nothing checked this directory. Adding a config for it surfaced three real errors, including `parseArgs` being absent from the pinned `@types/node@16` — the CLI needs Node 18. A `typecheck:tools` script keeps it checked. Neither it nor `pnpm lint` runs in CI — the only workflow is `release.yml` — and `pnpm lint` covers `src/` only in any case.

## Scope

`extractionTopP` and `samplingSeed` are settings and types only — this branch adds no UI, so they are reachable from the CLI or by editing `data.json`. `--context-window` and `--json-object` are held back; they depend on settings this branch does not add.


DocTpoint:
The key bug reproduces on one more path, and there is a consequence I think is worth pinning down before this lands.

**`provider: 'custom'`.** `providerOptionsName` is `config.provider.split('.')[0]`, fed from `createOpenAICompatible({ name: this.provider })` with `provider: settings.provider`. A local LM Studio setup configured as the built-in `custom` provider resolves to the name `custom`, so the fixed `openaiCompatible` key misses there exactly as it does for the named providers. Some field data: a 2824-page vault built in a 3.5-day run, 9272 page writes, on `custom` + LM Studio. `response_format`, `thinking`, `chat_template_kwargs` and `repetitionPenalty` were dropped for the entire run, and nothing about the result looked wrong from the outside.

**The part I would flag: correcting the key re-enables `response_format` on the OpenAI-compatible path, and that is what #65 was about.** #65 was closed on 2026-06-01 by `5851cc8`, whose message reads: "LM Studio 0.4.15+ compatibility (Issue #65): Remove `response_format: json_object` from OpenAICompatibleClient — Prompt instruction + prefilled `{` is sufficient for JSON enforcement — Only affects OpenAI-compatible local endpoints (LM Studio, Ollama, etc.)". The AI-SDK migration in `6be9258` reintroduced the field under `openaiCompatible`, where it has been inert since. So today's behaviour matches the #65 decision by accident rather than by intent, and fixing the key is precisely what ends that.

**I tested this against LM Studio 0.4.20 today**, with `curl` straight at `/v1/chat/completions` so that neither the plugin nor the SDK is in the path. Model `google/gemma-4-26b-a4b-qat`:

| sent | result |
|---|---|
| control, no extra fields | 200 |
| `response_format: {"type":"json_object"}` | **400 — `'response_format.type' must be 'json_schema' or 'text'`** |
| `response_format: {"type":"json_schema", …}` | 200, correctly structured |
| `chat_template_kwargs: {"enable_thinking": false}` | 200, accepted — but the model still emits reasoning tokens |

So #65 is not historical: with the key corrected, every extraction call against LM Studio would fail at the first request. Two ways out, and the choice is yours — the small one is to leave `response_format` out of the corrected object, so the other three fields start arriving and this one keeps its current behaviour. The larger one is to stop routing it through `providerOptions` at all and use the SDK's own `responseFormat` argument, which already emits `json_schema` when `supportsStructuredOutputs` is set and a schema is present, and falls back to `json_object` otherwise. That is more than this PR's scope, so I mention it only as the direction.

The other fields have no such history. `thinking`, `chat_template_kwargs` and `repetitionPenalty` were never part of #65, so for those the key fix is a straight repair. Two notes from the same test, in case they save someone time: `chat_template_kwargs` is accepted by LM Studio but did not stop this model from reasoning, so the delivery fix alone may not make `disableThinking` effective on that backend. And with a small `max_tokens`, reasoning consumes the whole budget and the response comes back with empty content — which looks like #76 and #275.

On the test change: reading the request body instead of the argument passed to `generateText` is the part I would keep even if everything else changed. The old assertion measured exactly the value the SDK then discards, which is how the wrong key stayed correct-looking since the AI-SDK migration.

eucher:
You measured it, and the measurement changed my mind about the shape of the fix — though not to either of the two options you offered. I took a third, and it is more conservative than what you proposed, so the call is worth reviewing.

**What landed:** `752576e` reverts the key correction entirely and removes `response_format`; `c98cdc9` replaces the hedged comments with your 0.4.20 numbers.

Your second option is where this ends up — `responseFormat` as the SDK's own argument, with `supportsStructuredOutputs` set *and a schema present*, which is the whole of your qualification — drop either half and the SDK falls back to `json_object`, which is your 400. That is the follow-up.

**Where I went further than you suggested, and why.** You wrote that for `thinking`, `chat_template_kwargs` and `repetitionPenalty` the key fix is a straight repair, and on the 400 question you are right — your `chat_template_kwargs` row shows LM Studio accepting it with a 200. What stopped me is the second half of that same cell: accepted, and the model still emits reasoning tokens. Delivering the field to that backend does not make `disableThinking` work there; it only makes the request larger. Ollama looks the same (`ollama/ollama#16240` is still an open feature request), and on `repetition_penalty` llama-server documents `repeat_penalty` instead. So the repair delivers three fields to ten providers, and I could not name a backend where a user would notice an improvement. That seemed like the wrong thing to ship inside a CLI PR — but it is a judgement call about scope, not a defect you missed, and if you or the maintainer prefer the narrow version I will restore it.

**What the revert does not cost.** `temperature` and `top_p` are standard SDK arguments and never depended on the key, so they still land; `seed` lands on this path too, though not on Anthropic, Codex or the `openai` provider, each for its own reason. The extraction call now reads the thinking setting, which takes effect on the paths whose provider keys were already correct.

**Still open, and it is two calls, not the battery I nearly asked you for:** `thinking: {"type":"disabled"}` alone, and the pair together — the plugin would send both in one request once the key is corrected, so a result on the pair is what actually describes our traffic. Only if you are already at that machine; the other rows you have covered.

**On your `max_tokens` note** — reasoning eating the whole budget and returning empty content is the failure I have been chasing from the other end, and it lines up with #76 and #275. That is the substance of the follow-up rather than a footnote.

The test keeps reading the request body, and it is now inverted: it asserts those three fields are *absent*, so whoever corrects the key has to rewrite this test rather than discover the delivery in production.

DocTpoint:
Both measurements you asked for, plus one that turned out to matter more.

Setup: LM Studio 0.4.20, `google/gemma-4-26b-a4b-qat` (MLX 4-bit), `temperature: 0`, `max_tokens: 1200`. The task is a small arithmetic question that reliably triggers reasoning.

| request | status | completion tokens | reasoning chars | visible answer |
|---|---|---|---|---|
| no extra field (control) | 200 | 334 | 703 | 4 |
| `thinking: {"type":"disabled"}` alone | 200 | 334 | 696–704 | 4 |
| `chat_template_kwargs` alone | 200 | 334 | 703 | 4 |
| both together | 200 | 334 | 703 | 4 |

So your reading holds: the fields are accepted and reasoning continues unchanged. The pair behaves exactly like each field alone.

**A measurement warning first.** My initial run produced four bit-identical results with latency dropping from 16s to 5s — that was prompt caching, not agreement between arms. Every request above carries a unique nonce in the prompt. With that, values scatter between 687 and 704 characters, which is real generation. Anyone repeating this should bust the cache first, otherwise the arms are indistinguishable by construction.

**The finding that changes the diagnosis.** I ran `enable_thinking: true` against `enable_thinking: false`. They are identical to each other, and identical to sending no field at all.

Gemma's chat template reads `enable_thinking` in five places, and on `true` it injects a `<|think|>` token into the first system turn. If the parameter reached the template, the two rendered prompts would differ. They do not. So `chat_template_kwargs` is not reaching the template rendering stage in LM Studio at all — it is accepted with a 200 and dropped.

That is a stronger version of your argument. You wrote that delivering the field does not make `disableThinking` work on that backend. On this backend the model never sees the field, even though its template would act on it.

**A second detail, in the same direction.** The template's logic is inverted relative to the name:

```jinja
{%- if not enable_thinking | default(false) -%}
    {{- '<|channel>thought\n<channel|>' -}}
```

`enable_thinking: false` opens the thought channel rather than closing it. Even with correct delivery, sending `false` would not be "reasoning off".

**One caveat I cannot rule out from the outside:** LM Studio may use its own precompiled template on the MLX path instead of the one in `tokenizer_config.json`. If so, the cause sits one layer lower than I describe. The effect for the user is the same.

**On scope: I am not asking you to restore the narrow version.** The revert is the better call, and these numbers support it rather than the fix. A field that is accepted, dropped before rendering, and inverted in meaning where it does render is not something to ship delivery for. If it is ever revisited, the prerequisite is a backend where the switch is demonstrably read — not merely accepted.

Your inverted test is the right shape for this. It pins the current, honest state instead of a delivery that does nothing.

green-dalii:
Hi @eucher 👋

Thanks for the CLI PR — the design (headless Node ingest with sampling + thinking controls) is sound, the request-body test (`openai-compat-request-body.test.ts`) is exactly the kind of high-fidelity test we needed, and the LM Studio measurements + DocTpoint's chat_template_kwargs measurement together pinned the `disableThinking` partial-fix honestly rather than overpromising. Two follow-ups from my review that I think are useful regardless of where this lands (v1.25.11 or v1.26.0):

---

## Pushed as 3 fix commits on your branch

I ran simplify + code-review (max effort) on the diff and pushed the items below to `pr/ingest-cli` as independent commits, so each one is auditable / revertable on its own. **Your 6 original commits are preserved untouched at the head of the PR.**

```
0fb614f docs: two claims the branch had already disproved of itself    ← yours
c98cdc9 docs(llm): cite the measured LM Studio result, not the old report  ← yours
752576e fix(llm): leave the providerOptions key alone, drop response_format  ← yours
2944bff feat(cli): sampling and thinking controls for the ingest CLI   ← yours
ce5d247 fix(tools): resolve plugin sources from the repo root          ← yours
c9e326b feat(tools): headless ingest CLI for running the engine under Node  ← yours
837e6d4 fix(codex): correct top_p/temperature comment to match adapter strip  ← fix #1
393e307 fix(sdk): re-indent seed spread in token-key retry paths        ← fix #2
63f8176 test(wrapper): move llm-client-wrapper.test.ts to conventional test dir  ← fix #3
```

**Fix #1 (P0-2):** `src/llm-sdk/openai-codex-sdk-client.ts` had a comment claiming "`top_p` and `temperature` it does accept, and both are forwarded`. This is incorrect — `openai-codex/request-adapter.ts:normalizeResponsesBody` calls `delete normalized.top_p` (and `.temperature`) before the wire request, a behaviour that predates this client (added by PR #273). The SDK forwards them, but the adapter removes them — Codex users setting `top_p` see a silent no-op with no log. Fix is comment-only: removing the adapter strip would re-test Codex's response shape, which is out of scope for a CLI PR. Comment updated to point at the adapter strip as the authoritative behaviour.

**Fix #2 (P0-1 cosmetic):** two sites in `openai-sdk-client.ts:191,367` had the new `seed` spread line indented at 8 spaces while sibling `topP` was at 10 spaces — mechanical copy-paste artefact. Pure indentation, no behaviour change.

**Fix #3 (P1-6):** `src/llm-client-wrapper.test.ts` landed at `src/` root but every other test in the repo lives under `src/__tests__/`. Vitest include pattern still picks it up; pure relocation + import path fix.

---

## Items deliberately NOT in the pushed commits — these are v1.26.0 follow-up scope

Per our discussion, these belong to the v1.26.0 MINOR redesign (especially around typed edges and per-provider dialect tables), not to this PATCH-bug PR. I tracked them so we don't lose them:

| Item | What | Why v1.26.0 |
|---|---|---|
| **#372-followup: per-provider `providerOptions` dialect table** | The `openaiCompatible` key the SDK does not read raw fields from → split into `providerDialectKey(provider)` resolver + per-dialect raw-fields object. Current workaround is "send under a fixed key, leave the comment" — the right fix is "dialect-table by provider". | Real architectural change; tied to #358 design track |
| **#372-followup: settings UI for `extractionTopP` / `samplingSeed`** | The new settings are persisted but have no UI input. Currently settable only via CLI flags or hand-editing `data.json`. Either add to advanced-section.ts (paired with extractionTemperature) or re-document as CLI-only in the type JSDoc. | UX scope, not bug-fix |
| **#372-followup: per-client `pickSamplingOpts` helper** | 16 hand-written `(temperature\|top_p\|seed\|repetition_penalty) !== undefined ? { … } : {}` spreads across 4 SDK clients. One `pickSamplingOpts({...})` helper in `src/llm-sdk/sampling-opts.ts` collapses the duplication. | Refactor scope |
| **#372-followup: comment-essay trim** | `openai-compat-sdk-client.ts:263-322` is 52 comment lines / 17 code lines. Each block keeps 2-3 lines of constraint note; measurement log moves to issue. | Cleanup |
| **#372-followup: `responseFormat` symmetry between openai + openai-compat** | The compat client removes `responseFormat` while openai keeps it. Either restore on compat or add a request-body test that pins the current divergence. | Touches #65 history + provider compat matrix |
| **#372-followup: `typecheck:tools` integration into Gate 1** | New sub-project script not invoked by the 5-Gate or pre-release-gate. Either integrate or document the exemption. | CI scope |
| **#372-followup: `chmod +x` for `run-ingest.mjs`** | Shebang `#!/usr/bin/env node` is decorative because the file isn't executable. | Doc / packaging |
| **#372-followup: vault.ts caching** | `cachedRead` doesn't cache; `frontmatterOf` re-reads on every call. `WeakMap<TFile, …>` would bring it in line with Obsidian's `metadataCache`. | Refactor scope |
| **#372-followup: esbuild incremental build** | `run-ingest.mjs` re-bundles the entire plugin graph on every CLI invocation. Add mtime check or `incremental: true`. | Perf scope |

---

## Process notes

- I ran the same audit as PR #357 (4 simplify angles + 3 code-review angles, 7 subagents in parallel) — Gate 1 stayed green throughout (lint 0/0, tsc 0/0, 2638 tests pass, build clean, css-lint 0).
- I held off on merging per your conversation with green-dalii: this PR is v1.26.0 scope and should stay on PR through the v1.25.11 PATCH ship window.
- Happy to discuss any of the follow-up items — especially the dialect-table one, which is the load-bearing design decision behind the partial-fix comment you already wrote.

🤝

eucher:
Thanks — and fix #1 caught a real error of mine, not a cosmetic one. I checked `request-adapter.ts:28-29` after your note: `delete normalized.temperature` and `delete normalized.top_p` are both there. My commit message for `2944bff` says `top_p` is "threaded through every client that accepts it", and for Codex that is false — the SDK forwards it, the adapter removes it. The comment you replaced was mine and it asserted the opposite of what the code does.

Agreed on the window. This adds a tool and settings that did not exist, so it is not a patch and v1.25.11 is the wrong place for it.

One note about ordering within v1.26.0, for whenever that gets sequenced. Everything else I have is stacked behind this branch — structured extraction output, the token budget derived from the context window, the observability that measuring any of it needs. Not because of merge conflicts, but because the CLI is how those get tested: without it, comparing two configurations means editing `data.json` between runs and hoping nothing else moved, which is the failure the flags exist to remove. So this one wants to be early in that milestone rather than late, and it is deliberately small — the `providerOptions` key is left as it was on `main`, and `extractionTopP` / `samplingSeed` have no UI, so an ordinary install is byte-identical.

## On the dialect table

I do not have a recommendation, and I would rather log what surfaced than argue a shape from one backend.

What surfaced is that the same intent takes a different field per backend, and the client sends one set to all ten providers on this path — so each gets fields it does not read and misses the one it does. Two concrete instances: OpenRouter's form is `reasoning: {enabled: false}`, which the comment in this file names and the code has never sent; and the repetition penalty is `repeat_penalty` for llama.cpp, Ollama and LM Studio but `repetition_penalty` for vLLM and OpenRouter, so the field name itself is per-dialect.

Worth knowing before the shape is chosen: this was not a table before v1.23.0. The old client probed three thinking-control dialects — `thinking.type='disabled'` → `reasoning_effort='none'` → nothing — learning which one a baseURL accepted from its 400 body and caching that per baseURL, with a companion `unsupportedFields` set built the same way and a `paramStripped` notice so the user knew a setting of theirs was being dropped. The tier-2 entry carries the annotation "(Gemini OpenAI-compat endpoint)", which is what Google documents today. The AI-SDK migration removed all of it, and the fixed `openaiCompatible` key meant nothing reached a server afterwards, so the loss was invisible.

And one thing that bears on any shape: neither a table nor a probe detects "accepted and ignored". DocTpoint measured exactly that — LM Studio answers 200 to `chat_template_kwargs` and reasons anyway. A table would record it as supported because it did not error; a probe would conclude the same. Telling that case apart needs the reply inspected, not the status code.

So: table, probe, or something else is future work that wants measuring per backend before it is written down. **Current behaviour is unchanged either way** — this branch leaves the key as `main` has it, so no provider receives anything new. What changed is that the gap is now written down instead of being hidden behind a key that dropped everything.

## Two of your follow-up items are already resolved in the stacked branch

So the work is not done twice:

- `responseFormat` symmetry: the compat client delivers it again, in the `json_schema` form rather than the `json_object` that #65 was about, carrying an extraction schema. The divergence you flagged is closed in that direction.
- the `providerOptions` key: corrected there, keyed by provider name. That is the prerequisite for the dialect table and not the table itself — the raw fields are still identical for every provider on the path.

Not resolved, to be exact: the comment volume in `openai-compat-sdk-client.ts` is down from 62 lines to 46 in that region as a side effect of rewriting them, which is not the same as trimming it. The settings UI for `extractionTopP` / `samplingSeed` is untouched, and you are right that they are reachable only through the CLI or by editing `data.json`.

DocTpoint:
Two things from my side: one on the probe question, one check on the stacked branch.

**A probe can tell "accepted and ignored" apart, if it is differential rather than status-based.**

You are right that a status-code probe cannot, and the one you describe is blind to that case by construction — learning from the 400 body means it only ever learns what a backend *rejects*. But the measurement I ran is itself a probe shape, and it needs no semantic reading of the reply. Send the same request twice with opposite values of the same flag and compare one observable. For `enable_thinking` I compared reasoning-token count and the presence of the thought channel: `true` and `false` came back indistinguishable from each other and from sending no field at all, so the parameter never reached template rendering. The signal is the difference between the arms, not the value of any single arm.

Three conditions, from having got each of them wrong first:

- **Bust the cache.** My first run produced four bit-identical arms with latency falling from 16s to 5s. That was prompt caching, and it makes every arm agree by construction. A unique nonce per request fixes it.
- **Test for a difference, not for a direction.** Gemma's template opens the thought channel on `enable_thinking: false` — the branch is `{% if not enable_thinking %}`. A probe asserting "false must reduce reasoning" would have recorded a correct delivery as a failure.
- **The result is per model, not per baseURL.** The template belongs to the model. The same LM Studio with a different model can answer differently, which is an argument for measuring rather than tabulating.

What the differential test does not give you is *where* the field was dropped — my caveat about LM Studio possibly using a precompiled template on the MLX path still stands. For the user the effect is identical, but for a table entry "dropped by the server" and "dropped by the runtime for this model" are different rows.

For the two field-name cases you list, the same shape works and is easier: `repeat_penalty` against `repetition_penalty`, both at a value extreme enough to visibly degrade the output. If exactly one arm degenerates, that backend's dialect is settled by observation rather than by documentation.

**One check on the stacked branch, before it lands.** You wrote that the compat client delivers `responseFormat` again in the `json_schema` form, carrying an extraction schema. Worth pinning that both halves are set: in `@ai-sdk/openai-compatible` 2.0.62 the request builder emits `json_schema` only when `supportsStructuredOutputs === true` **and** `responseFormat.schema != null`. With the schema present and the flag unset, it falls back to `{type: "json_object"}` and pushes an `unsupported` warning ("JSON response format schema is only supported with structuredOutputs") — the request still goes out. On LM Studio 0.4.20 that fallback is the 400 from my earlier table, and a warning in the warnings array is easy to miss. A request-body test asserting the `json_schema` shape would catch it at the point where the divergence is cheap.

That also puts a second instance under what you described: `json_object` against `json_schema` is itself a per-backend dialect, in a field that is not raw and does not travel through `providerOptions` at all.

**On ordering, a data point rather than an opinion.** The run I mentioned earlier — 2824 pages, 9272 writes, `custom` + LM Studio, 3.5 days — dropped four fields for its entire duration with nothing looking wrong from the outside, and every measurement in this thread had to be run with `curl` outside the plugin, because varying one setting between runs otherwise means editing `data.json` and hoping nothing else moved. That is the failure you describe, from the other side. Where it belongs in the milestone is green-dalii's call, but the instrument argument matches what I hit.

green-dalii:
Hi @eucher 👋

Quick follow-up on the CLI split planning. The earlier comment on this thread described a v1.0 plan (submodule + 5-symbol shim); I deleted it and replaced with this v2.0 plan after re-evaluating from first principles. Your merged work is unchanged either way — the corrections are on our side, not yours.

## TL;DR

Your CLI is staying where it is for the next 1–3 releases. Long-term, it will live in a sibling repo as a **fully independent project — no submodule, no shared dependencies, no `obsidian` imports**. There will be a dual-repo coexistence window; nothing is getting yanked on existing users.

## What happened

v1.26.0 was submitted to the Obsidian official marketplace pre-review. The Bot reported ~60 Warnings on `tools/llm-wiki-cli/`. None block the submission, but they all hit the score, so this is now a real ongoing cost rather than a one-off.

The deep dive surfaced three things that decided the shape of the fix:

1. **The Bot scans the whole repo `.ts` tree, not just `main.js`.** Local `pnpm lint` = `eslint src/`, and the root `tsconfig.json` includes only `src/**` — both are blind to `tools/`. So the warnings surface only from the Bot, never from local development. The eslint.config.mjs comment that said the Bot "inspects only `main.js`" is now fixed (the assumption pre-dated your CLI landing in `tools/`).
2. **`eslint-disable` is not an escape hatch.** Obsidianmd rules (`no-nodejs-modules`, `no-global-this`, `hardcoded-config-path`) are no-disable — the Bot rejects `eslint-disable` comments. The ~43 `console.log` warnings are a Bot-side heuristic check, not an eslint rule at all (the plugin ships zero console-related rules), so comments cannot suppress them either.
3. **Only 4 of ~60 are fixable, and they are pure type-safety cleanups** — `JSON.parse(...) as ...` annotations, drop a redundant `as`, `globalThis.crypto.subtle` → `crypto.subtle`. The other ~56 are structurally unfixable: a Node CLI importing `node:fs` / `node:path` / `node:console` at the top level, `console.log` as the actual output interface, `globalThis` shimming `window` for the engine, `.obsidian` literal in the path the plugin reads from, and a `fetch`-backed `requestUrl` because Node has no Obsidian `requestUrl`. None of these are fixable without destroying what the CLI is.

## Why the move, from first principles

A Node CLI will always collide with a ruleset written for a renderer running in an Electron browser. The only durable answer is to put the CLI outside the Bot's scan scope.

We considered four shapes:

- **Path A — rename `tools/` to `cli/` / `.build-tools/` / etc.** Doesn't help. The Bot scans the `.ts` tree regardless of directory name; `.gitignore` does not hide files from a clone-based review.
- **Path A2 — git submodule from a new sibling repo.** What we first picked (the earlier deleted comment on this thread described it). Failed review: the submodule still lives at `src-link/obsidian-llm-wiki/` in the main repo's `.ts` tree, so the Bot keeps scanning it. The ~60 Warnings did not decrease. The submodule + alias plugin approach also forced users to know about submodules, made `npm install -g` impossible, and produced two parallel implementations (one type-time ambient, one runtime shim) for the same Obsidian contract. Rejected.
- **Path B2 — pnpm monorepo workspace.** Splits the main repo (`packages/plugin/src/`), not the CLI. Scope far exceeds the need; doubles CI and release complexity for the plugin.
- **Path B3 — extract `@green-dalii/llm-wiki-core` as a published npm package.** Adds an OSS publishing attack surface (npm tokens, registry permissions) and a dual-release coordination problem. Useful one day, not for this.

**Path B1 — independent repo + copy + CI sync. What we settled on.** The CLI is its own repo from day one. Zero Obsidian imports anywhere in `src/`. Layer 1 (pure algorithms, no `obsidian` deps) and Layer 2 (algorithms taking `TFile` / `{ app, file }`, with signatures rewritten to take `{ path, content, frontmatter }` instead) are copied from the main repo. Layer 3 (full Obsidian integration — `App` instance, vault adapter, `Notice` calls) is **rewritten from scratch** using Node `fs/promises` + `gray-matter` + `fast-glob`. Sync is via GitHub Actions `repository_dispatch` on main-repo release — see Phase 2 below.

## Why the CLI doesn't actually need `obsidian`

From a first-principles pass on what the CLI's loop actually does, it doesn't need the `obsidian` npm package at all. v1.0 framed this as "we use 5 symbols, here's a shim that exposes them". v2.0 reframes it: the CLI's actual work is filesystem IO on `.md` files + LLM HTTP + string manipulation. None of those steps require an Obsidian runtime.

So instead of `import { TFile } from 'obsidian'` plus a shim that fakes `TFile`, the engine receives plain objects: `{ path: string; content: string; frontmatter: Record<string, unknown> }`. Grep `from 'obsidian'` in the new repo's `src/` returns zero matches. There is no shim, no `App`, no `TFile`, no `normalizePath`, no `Modal`, no `Notice`. The CLI is a plain Node tool that happens to read and write Markdown — which is what it always should have been.

(For completeness: I checked [Obsidian Headless](https://obsidian.md/help/headless) — it's a Sync/Publish CLI client that doesn't expose plugin runtime API. Per their docs: "It isn't designed for plugin code." Closing that path.)

## 4-phase gradual migration (zero destruction to your merged work)

| Phase | When | What |
|---|---|---|
| **1. Boot** | v1.26.x PATCH cycle, 5–6 days | `gh repo create green-dalii/obsidian-llm-wiki-cli --public`; copy Layer 1 + Layer 2 (with signature rewrites) from main repo; write Layer 3 orchestration + `FsVaultReader` / `FsVaultWriter` from scratch (Node `fs/promises` + `gray-matter`); CI lint + typecheck + test + bundle; `npm publish v0.1.0` |
| **2. Coexist** | v1.26.x PATCH → v1.27.0 MINOR (1–3 months) | main repo `tools/llm-wiki-cli/` keeps working; new repo absorbs PR #408 + PR #409 follow-on work; sync workflow (`repository_dispatch` on main-repo release tag) live; users can choose either install path |
| **3. Deprecate** | v1.27.0 MINOR design stage, ~1.5 days | DEPRECATION marker on `pnpm llm-wiki`; 11 READMEs rewrite Tools H2; CHANGELOG DEPRECATION entry |
| **4. Delete** | v1.28.0 MINOR or later, ~0.5 days | `git rm -rf tools/llm-wiki-cli/ src/__tests__/tools/`; drop `bin.llm-wiki` / `scripts.ingest`; CHANGELOG BREAKING |

**Phase 2 sync, concretely.** The main-repo release workflow adds one line: `gh api repos/green-dalii/obsidian-llm-wiki-cli/dispatches -F event_type=upstream-released -F client_payload[tag]=$TAG`. The CLI repo's `repository_dispatch` trigger fires `.github/workflows/sync-from-upstream.yml`, which pulls the new tag, copies Layer 1 verbatim and Layer 2 with signature rewrites (the rewrite is mechanical: `TFile` → `{ path, content, frontmatter }`), runs full CI, and opens a PR if anything changed. A user going from `pnpm llm-wiki` to `npm install -g @green-dalii/obsidian-llm-wiki-cli` gets the same engine behaviour, because the same algorithm files are running.

## Effort

~5–6 focused work days across 1–3 months. Main-repo PR flow / CI / release flow stay unchanged through Phase 4.

## What this means for you

- **Your PR #372 work is preserved as-is.** It ships in v1.26.0 from the main repo exactly as merged.
- **PR #408 and PR #409** (still open against this repo) — the right home for follow-on work is the new repo, not this one. When the new repo is up, I'll notify you and DocTpoint; you can transfer ownership of follow-up PRs at your discretion.
- **Main-repo maintenance cost: low.** v2.0's sync is one-way (main repo is the producer, CLI repo is the consumer) and automated via CI on tag push. The producer's release workflow gains one line, not a new gate.
- **If you'd rather drive the new-repo bootstrap yourself**, say so — I'd hand it over rather than write it for you, since you know this code better than anyone.

Planning details live in the new repo's `SPEC.md` (https://github.com/green-dalii/obsidian-llm-wiki-cli — to be created) and the discovery in [[feedback_obsidian_bot_tools_cli_warnings]]. The full scoring & timing rationale is on ROADMAP v1.27.0 MINOR design track row 7.

Thanks again for the CLI — the `openai-compat-request-body.test.ts` test was the foundation that let PR #410 / #411's wire-level verification land cleanly, and the LM Studio + DocTpoint measurements you both ran pinned the disable-thinking fix honestly rather than overpromising. The split is the right cleanup, not a retreat from the work.

🤝

PR #385, opened 2026-08-01
fix(scope): anchor the wikiFolder prefix checks (Closes #383)

### LINKED ISSUE #383: Seven unanchored wikiFolder prefix checks — same class as #364, and one of them deletes
Follow-up hygiene noted during the #364 consolidation. Line numbers are against `29e2402`.

`path.startsWith(wikiFolder)` is a string-prefix test, not a containment test. With the default `wikiFolder: wiki`, both `wiki-archive/note.md` and `wikipedia.md` pass it. `src/core/folder-scope.ts` — merged with #364 — already states this rule and provides the primitive; these call sites predate it.

**Over-including (a non-wiki path is treated as a wiki page):**

| File | Line | Effect |
|---|---|---|
| `src/wiki/lint/delete-empty-stubs.ts` | 10 | **deletes** — see below |
| `src/wiki/lint/get-existing-pages.ts` | 12 | user notes enter the page inventory |
| `src/wiki/lint/phases/preparation.ts` | 23 | same, in the lint preparation phase |
| `src/schema/auto-maintain.ts` | 407 | same, in auto-maintain |
| `src/core/orphan-matcher.ts` | 81 | a relative page path starting with the folder name is mistaken for an absolute one |

**Over-excluding (a real user note is treated as a wiki page and hidden):**

| File | Line | Effect |
|---|---|---|
| `src/ui/modals/suggest-modals.ts` | 29, 56 | files and folders vanish from the ingest picker |
| `src/ui/modals/MultiFileSuggestModal-class.ts` | 101 | same, multi-file picker |

**Why `delete-empty-stubs.ts:10` is the sharp one.** The prefix test runs first; the `/schema/`, `/sources/`, `contradictions/` and `log.md` exclusions below it only narrow an already-wrong set. A user folder named `wiki-archive` next to `wiki` therefore enters the stub deleter. Anything in it that `isPageEmpty()` considers empty — short notes qualify — and that does not carry `reviewed: true` is deleted. This is user data outside the wiki folder.

**Not uniformly broken.** `wiki-engine.ts` (`isInWikiContentFolder`), `ensure-welcome-note.ts` and `auto-maintain.ts:550` all anchor with `${wikiFolder}/` and are correct. The inconsistency is the reason this is easy to miss in review.

**Fix.** Six of the seven are a direct substitution: `isInFolderScope(f.path, wikiFolder, false)` from `src/core/folder-scope.ts`. `orphan-matcher.ts:81` has a different shape — it asks "is this path already absolute?" — so it needs ``startsWith(`${wikiFolder}/`)`` rather than the helper.

Happy to open a PR if you want it as one change; equally happy to leave it as a tracked note if it belongs in the v1.26.0 hardening pass (#382).

DocTpoint:
Closes #383.

Thanks for cross-checking every line in the issue — the fix follows exactly the form you confirmed, with three deviations I want to flag rather than bury.

## What changed

Six sites move to `isInFolderScope(path, wikiFolder, false)` (the helper from #364), and `orphan-matcher.ts` gets the anchored plain prefix:

| file | direction of the leak |
|---|---|
| `wiki/lint/delete-empty-stubs.ts` | included too much — **and deletes** |
| `wiki/lint/get-existing-pages.ts` | included too much |
| `wiki/lint/phases/preparation.ts` | included too much |
| `schema/auto-maintain.ts` (startup Phase 2) | included too much — **and writes** |
| `ui/modals/suggest-modals.ts` (file + folder picker) | excluded too much |
| `ui/modals/MultiFileSuggestModal-class.ts` | excluded too much |
| `core/orphan-matcher.ts` | produced a link to a path that does not exist |

## Three things that are not mechanical

**1. `FolderSuggestModal` needed an identity check on top of the anchoring.** The set to exclude there is the wiki folder *itself* plus its descendants, and `isInFolderScope` answers "descendant of" — which a folder is not of itself. A plain substitution would have anchored the boundary correctly and, in the same move, put the wiki folder back into the folder picker. It now reads `path === wikiFolder || isInFolderScope(...)`.

**2. `orphan-matcher.ts` keeps the plain prefix form**, as you suggested. The question there is "is this path already absolute?", not "is it in scope", and a scope-shaped predicate would misname it. Only the anchor was missing.

**3. An existing test asserted the bug and had to be inverted.** `orphan-matcher.test.ts` contained:

```ts
it('handles paths starting with wiki folder name', () => {
  const result = normalizeOrphanPagePath('wiki-concepts/page.md', 'wiki');
  // startsWith('wiki') is true, so treated as full path
  expect(result).toBe('wiki-concepts/page.md');
});
```

That is the unanchored behaviour written down as intent. It now asserts `wiki/wiki-concepts/page.md`, with a comment recording that it was inverted and why. Worth a second look from you, since inverting a green test is the one change in this PR that a reviewer cannot verify by reading the diff alone.

## Tests

New `src/__tests__/wiki/wiki-folder-boundary.test.ts`, one regression per call site. `folder-scope.test.ts` already covers the helper, so these pin the call sites — which is where the bug lived. Every case asserts **both** leak shapes, sibling folder (`wiki-archive/Old.md`) and neighbouring file (`wiki.md`), and keeps a real wiki page as the control so the fix cannot pass by simply excluding everything.

I verified the tests actually catch the bug: with the source changes stashed and the tests in place, all nine fail; restored, all pass.

Two notes on the test setup:

- `__support__/setup.ts`: the `FuzzySuggestModal` stub did not assign `this.app`, unlike the real base class, so subclass `getItems()` threw on `this.app.vault` and could not be tested at all. The stub now takes the app in its constructor.
- **One call site has no test**: `MultiFileSuggestModal`. Its filter line is character-identical to the tested `FileSuggestModal.getItems`, but it sits inside `onOpen()`, which builds the modal DOM. Driving that needs the jsdom environment, and under jsdom the `obsidian` import does not resolve — `server.deps.inline` in `vitest.config.ts` only takes effect in the node environment. Rewiring that felt out of scope for a fix on this timeline; happy to do it separately if you want the coverage.

## An eighth site with the same shape — and why anchoring is the wrong fix there

`wiki/lint/merge-duplicates.ts:168` carries the same unanchored form:

```ts
const allWikiFiles = ctx.app.vault.getMarkdownFiles().filter(
  f => f.path.startsWith(wikiFolder) && f.path !== sourcePath
);
```

It is not in #383, and it should not be folded into this PR — but not because the boundary is fine there. It is because anchoring alone would make it *worse*.

`mergeDuplicates` deletes the source page at the end (`deleteFile(sourcePath)`, line 182). Every link that is not rewritten therefore points at a file that no longer exists. So the defect at this site is the mirror image of the seven: the rewrite is scoped **too narrowly**, not too widely. Notes outside the wiki folder never get their links carried over and are left with dead links. The unanchored prefix happens to catch an arbitrary subset of them — exactly those whose path starts with the folder's name — which is neither wiki-only nor vault-wide.

Anchoring that line in isolation would make the scope coherent and take the accidental rescue away, turning those links dead for real.

A proper fix has a second axis, too: the search term is `[[entities/Foo]]`, the wiki-relative form. A link written outside the wiki folder usually carries a different shape (full path, or just the title), so even a vault-wide sweep with only this one form would still miss most of them.

Both axes together are a scope decision about the merge path, not a boundary repair, so I would rather raise it as its own issue than attach it to a fix you want landed before the v1.26.0 hardening. Say the word and I will write it up.

For a comparable reason I left the neighbouring `configDir` prefix checks alone, though they have the identical shape.

## Gate

Built on current `main` (`29e2402`). `pnpm lint --max-warnings=0`, `npx tsc --noEmit`, `pnpm test`, `pnpm build`, `pnpm css-lint` — all clean. Test suite: 207 files / 2772 tests, up from 206 / 2762 on the base.

### DIFF
--- a/src/__tests__/__support__/setup.ts
+++ b/src/__tests__/__support__/setup.ts
@@ -138,9 +138,15 @@ vi.mock('obsidian', () => ({
     loadData() { return Promise.resolve({}); }
     saveData() { return Promise.resolve(); }
   },
-  // Suggest modals (stubbed — used in settings.ts FolderSuggestModal)
+  // Suggest modals (stubbed — used in settings.ts FolderSuggestModal).
+  // The real FuzzySuggestModal takes the app in its constructor and exposes it
+  // as `this.app`; subclasses read `this.app.vault` in getItems(). The stub has
+  // to do the same or those methods cannot be exercised at all.
   FuzzySuggestModal: class {
-    constructor() {}
+    app: unknown;
+    constructor(app: unknown) {
+      this.app = app;
+    }
     open() {}
     close() {}
     onOpen() {}
--- a/src/__tests__/wiki/wiki-folder-boundary.test.ts
+++ b/src/__tests__/wiki/wiki-folder-boundary.test.ts
@@ -1,0 +1,289 @@
+// Issue #383 — regression tests for the unanchored `wikiFolder` prefix checks.
+//
+// `path.startsWith(wikiFolder)` is a string test, not a containment test. With
+// `wikiFolder: "wiki"` two kinds of foreign path pass it:
+//
+//   * a sibling folder sharing the name prefix — "wiki-archive/note.md"
+//   * a file sitting beside the folder          — "wiki.md"
+//
+// The call sites leak in both directions, so every case below is asserted
+// twice, once per leak shape:
+//
+//   * INCLUDING too much — lint, startup fixes and the page listing treat
+//     foreign notes as wiki pages (deleteEmptyStubs then deletes them).
+//   * EXCLUDING too much — the ingest pickers subtract those same foreign
+//     notes from the candidate set, so the user cannot select their own files.
+//
+// The helper itself (`isInFolderScope`) is covered in
+// `__tests__/core/folder-scope.test.ts`; these tests pin the CALL SITES, which
+// is where the bug lived.
+
+import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
+import { TFile, TFolder } from 'obsidian'; // mocked in __support__/setup.ts
+
+import { deleteEmptyStubs } from '../../wiki/lint/delete-empty-stubs';
+import { getExistingWikiPages } from '../../wiki/lint/get-existing-pages';
+import { runPreparationPhase } from '../../wiki/lint/phases/preparation';
+import { FileSuggestModal, FolderSuggestModal } from '../../ui/modals/suggest-modals';
+import { AutoMaintainManager } from '../../schema/auto-maintain';
+import { createMockContext } from '../__support__/engine-context';
+import type { LintPhaseContext } from '../../wiki/lint/types';
+import type { LLMWikiSettings } from '../../types';
+import type { WikiEngine } from '../../wiki/wiki-engine';
+
+const WIKI = 'wiki';
+
+/** A real wiki page — the control. Every fix must keep this one in scope. */
+const REAL_PAGE = 'wiki/entities/Real.md';
+/** Leak 1: sibling folder sharing the name prefix. */
+const SIBLING_FOLDER_FILE = 'wiki-archive/Old.md';
+/** Leak 2: a file sitting beside the folder. */
+const NEIGHBOUR_FILE = 'wiki.md';
+
+/** Short enough to count as an empty stub (< MIN_SUBSTANTIVE_CHARS). */
+const EMPTY_STUB = '---\ntitle: X\n---\n\nTBD\n';
+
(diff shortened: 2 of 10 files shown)


green-dalii:
@DocTpoint 
I sincerely apologize for the inconvenience. A user @Hashim1999164  submitted PR #384 to fix Issue #383 earlier, and that PR has already been merged into the `main` branch. To avoid duplicate work, please review the existing PR; if you wish to make further improvements, you can rebase your branch onto the latest version and submit a new PR.

DocTpoint:
Superseded by #384, which landed the same set of fixes while this branch was being built. Closing to avoid a duplicate.

PR #392, opened 2026-08-02
fix(merge): retarget links vault-wide and in every link form before the delete

### LINKED ISSUE #386: `mergeDuplicates` deletes a page and leaves dead links in the user's own notes — the rewrite misses both the vault outside `wiki/` and the bare-title link form
Split out of #383 / #385, where I found this while auditing the unanchored `wikiFolder` prefix checks. It carries the same `startsWith` shape, but the defect points the other way, so anchoring it would have been the wrong move and I kept it out of that PR.

## What happens

`mergeDuplicates` (`src/wiki/lint/merge-duplicates.ts`) merges a duplicate page into its target, rewrites wiki-links from the old target to the new one, and then **deletes the source page**:

```ts
const sourceRel = sourcePath.replace(wikiFolder + '/', '').replace('.md', '');
const allWikiFiles = ctx.app.vault.getMarkdownFiles().filter(
  f => f.path.startsWith(wikiFolder) && f.path !== sourcePath
);
// ... rewrites [[sourceRel]] and [[sourceRel| ...
await ctx.deleteFile(sourcePath);   // line 182
```

Because the page is deleted, any link that is *not* rewritten now points at a file that no longer exists. The rewrite, however, only ever visits files inside the wiki folder. A user note that links to the merged page — the ordinary case, since source notes live outside the wiki — keeps a link into nothing.

## Two independent gaps — and the second one is the big one

**1. Radius.** The filter is limited to the wiki folder. Since the target of the link is being deleted, there is no reason to stop at the folder boundary; the deletion affects the whole vault.

**2. Link form.** The search term is `sourceRel`, the wiki-relative form `[[entities/Foo]]`. But a link written *outside* the wiki folder is normally a bare title — `[[Foo]]` — because that is what Obsidian's own autocomplete inserts, and Obsidian resolves it into the wiki by suffix match. The rewrite never sees those.

I measured this on a vault with 2824 wiki pages and 473 notes outside the wiki folder:

| links from outside the wiki that resolve *into* it | count |
|---|---|
| written as a bare title `[[Foo]]` | **1762**, across 340 notes |
| written with a folder prefix (`entities/Foo`, `wiki/entities/Foo`) | **0** |

(The only prefixed occurrences anywhere outside the wiki were 30 quotations inside my own bookkeeping files — documentation about the plugin, not navigation.)

So in that vault the current rewrite would miss **every single** incoming link from a user note, and a vault-wide sweep that kept the wiki-relative search term would miss every single one as well. Fixing only the radius would look like a fix and change nothing.

## Reproduction

With `wikiFolder: wiki`:

1. Two duplicate pages `wiki/entities/Foo.md` (target) and `wiki/entities/Foo-2.md` (source).
2. A user note `Notes/Study.md` containing `[[Foo-2]]` — the form Obsidian's autocomplete produces.
3. Another wiki page containing `[[entities/Foo-2]]`.
4. Run lint and let it merge the duplicates.

Result: the link in the wiki page is rewritten; the link in `Notes/Study.md` is untouched and now dead, because `wiki/entities/Foo-2.md` has been deleted. Nothing reports this — the page is gone, so no dead-link scan inside the wiki folder can see the reference either.

## Relation to #383

The same line also carries the unanchored prefix, so today it accidentally rescues an arbitrary subset of outside files — exactly those whose path happens to begin with the folder's name (`wiki-archive/...`, `wiki.md`). That is neither wiki-only nor vault-wide. It is worth noting that simply anchoring this line, as #385 does for the other seven sites, would remove that accidental rescue and turn those links dead as well. The boundary is not the problem here; the scope is.

## What users already expect here

Obsidian itself rewrites links across the **entire** vault when a file is renamed — that is the baseline behaviour every user of this plugin has been trained on by the app. A merge is the harsher operation: the page is not renamed, it is deleted. Doing less than a rename does on an operation that destroys more is the part that will surprise people.

## Three ways to fix it

**A — rewrite vault-wide, in every link form.** Before the delete, collect every form under which the source page was addressable (`wiki/entities/Foo`, `entities/Foo`, `Foo`, each with an optional `|alias` or `#heading`) and retarget every occurrence in the vault.

One safeguard is required: a bare `[[Foo]]` must not be replaced by string match. Resolve it first and rewrite only where it actually resolves to the page being deleted — otherwise a same-named note elsewhere in the vault gets its link bent to a page it never referenced.

Keeps the vault navigable with no user action. Cost: the plugin edits files outside its own folder.

**B — do not rewrite, report.** Collect the incoming references before the delete (afterwards they are unfindable, since the target is gone) and surface them as a lint finding: "these 12 notes referenced X, now merged into Y".

The plugin stays out of foreign files, but navigation is broken after every merge until the user repairs it by hand.

**C — do not delete; leave a redirect.** The source page stays as a stub whose body is a link to the target. Every existing link stays valid without touching a single foreign file.

Two follow-ups come with it: `deleteEmptyStubs` would reap such a stub on the next lint run (short body, no `reviewed: true`), and duplicate detection has to skip redirects or it will merge them in circles. It also leaves one tombstone per merge in the graph.

## Recommendation

**A, with the resolve-before-replace safeguard.** It is what the surrounding application already does on the weaker operation, and it is the only option that keeps a vault navigable all the way from a wiki page back to the originating note without asking the user to repair anything.

**C** is the honest fallback if you would rather the plugin never wrote into a user's own notes — it preserves navigability too, at the price of tombstones and two extra code paths.

**B** I would not pick on its own: it converts a silent failure into a visible one, which is better, but it still leaves the vault broken after every merge.

Happy to prepare the PR once you have picked a shape. If it is A, I would also add the vault-wide sweep as a separate, testable function rather than inlining it in `mergeDuplicates`, since the same "retarget every form of a page" logic is what a future rename/redirect feature would need.

DocTpoint:
Closes #386 — option A as picked in the issue, with the resolve-before-replace safeguard, and as a separate testable function rather than inlined in `mergeDuplicatePages`.

## What changed

New `src/core/link-retarget.ts` with one exported function, `retargetLinksToPage(deps, fromPath, toPath)`. `mergeDuplicatePages` calls it immediately before `ctx.deleteFile(sourcePath)`; the old in-place loop is gone.

The module carries no wiki vocabulary — no `wikiFolder`, no page types, no lint context — so the same primitive is what a rename or redirect feature would call.

## The safeguard is the app's own resolver, not a string match

Every candidate goes through `metadataCache.getFirstLinkpathDest(linkpath, linkingFile.path)`, and only a link that actually lands on the page being deleted is touched. A bare `[[Foo]]` in a note that sits next to the user's own `Foo.md` resolves to that note and is left alone.

One thing I checked before building, because it decides whether this bug exists at all: the merge adds the source page's title to the target's `aliases`, so it would be reasonable to assume a bare `[[Foo-2]]` survives the deletion by resolving through the alias. It does not. Obsidian's resolver reads its file lookup and never consults frontmatter aliases — that is why community plugins exist that monkey-patch `getFirstLinkpathDest` to add exactly this ([forum report on 1.12.7](https://forum.obsidian.md/t/wikilink-resolution-does-not-honor-frontmatter-aliases-1-12-7/113902), [Alias Linker](https://community.obsidian.md/plugins/alias-linker)). The alias is good for the suggester, not for resolution. The same fact is why this module must not consult aliases either: doing so would rewrite links that were never pointing at the page.

## Two decisions worth your eye

**1. Foreign notes are written through `vault.process`, not `createOrUpdateFile`.** The wiki's write gate normalizes the `sources:` frontmatter field and auto-corrects link pollution on every write. That is right for a generated wiki page and wrong for someone's own note — the plugin should change the one link it came for and nothing else. Edits are applied at the offsets the metadata cache reports, after verifying the text at that offset still matches what the cache promised; a note that changed since indexing is counted as `stale` and reported rather than spliced blind. Links quoted inside code blocks are untouched, since the cache does not report them as links.

**2. The rewrite keeps the link's shape.** A folder-prefixed `[[entities/Foo-2]]` inside the wiki becomes `[[entities/Foo]]`, a bare `[[Foo-2]]` in a user note becomes `[[Foo]]`. The candidate forms are generated from the target path (`Foo`, `entities/Foo`, `wiki/entities/Foo`) and each is verified to resolve back to the target from that particular file, preferring the original's depth. Rewriting everything to the shortest resolving form would have been simpler, but it would convert a wiki page's link convention as a side effect of a merge, and it would fight `correctRelatedLinkPrefixes`.

## Not in this PR

`stale` currently surfaces as a `console.warn`. If you would rather have it in the lint report as a finding, say so and I will wire it — it is the one case where the merge knowingly leaves a dead link behind, so it arguably belongs on screen.

The sweep is one pass over the markdown file list per merged pair, using in-memory cache lookups. On a lint run merging many pairs that is O(pairs × files) cache hits; measurable but not hot. If you want it batched across a whole lint run, that is a different shape and I would rather do it as a follow-up than fold it in here.

## Verification

Gate 1 measured against `69f3357`, both sides on this machine rather than quoted:

| | base | branch |
|---|---|---|
| `pnpm lint --max-warnings=0` | 0/0 | 0/0 |
| `npx tsc --noEmit` | clean | clean |
| `pnpm test` | 2863 / 212 files | 2874 / 214 files |
| `pnpm build` | — | clean, production |
| `pnpm css-lint` | — | 0 |

11 new tests: 9 on the primitive (bare-title case, resolve-before-replace, shape preservation, display text / subpath / embed marker, code block, stale offsets, self-skip, several links in one write, no-op) and 2 at the merge site — the latter being the leak-direction coverage PR #389 deliberately left to this issue.

Each of the three load-bearing tests was verified by mutating the production code: removing the resolve guard, removing the shape preference, and removing the stale check each fail exactly their own test and nothing else.

### DIFF
--- a/src/__tests__/__support__/link-vault.ts
+++ b/src/__tests__/__support__/link-vault.ts
@@ -1,0 +1,134 @@
+// A vault fake with the two things `src/core/link-retarget.ts` depends on and
+// the production mock context does not model: a link cache with positions, and
+// a linkpath resolver.
+//
+// The resolver deliberately mirrors Obsidian's own behaviour, because the
+// safeguard under test ("only rewrite links that actually resolve to the page
+// being deleted") is only meaningful against a realistic resolver:
+//
+//   * a linkpath may be a full path, a partial path, or a bare basename
+//   * frontmatter aliases are NOT consulted — Obsidian's resolver reads its
+//     file lookup, not aliases, so a bare link matching only an alias does not
+//     resolve
+//   * when several files match, the one in the linking file's own folder wins,
+//     then the shortest path
+//
+// Links inside fenced code blocks are not reported, which is also what the real
+// metadata cache does — a quoted `[[Foo]]` in documentation is not a link.
+
+export interface FakeReference {
+  link: string;
+  original: string;
+  position: { start: { offset: number }; end: { offset: number } };
+}
+
+export interface FakeLinkVault {
+  vault: {
+    getMarkdownFiles(): Array<{ path: string }>;
+    process(file: { path: string }, fn: (data: string) => string): Promise<string>;
+  };
+  metadataCache: {
+    getFileCache(file: { path: string }): { links?: FakeReference[]; embeds?: FakeReference[] } | null;
+    getFirstLinkpathDest(linkpath: string, sourcePath: string): { path: string } | null;
+  };
+  read(path: string): string;
+  write(path: string, content: string): void;
+  /** Paths whose content was handed to `vault.process`, in call order. */
+  processed: string[];
+}
+
+const FENCE = /^\s*(```|~~~)/;
+
+function fencedLineRanges(content: string): Array<[number, number]> {
+  const ranges: Array<[number, number]> = [];
+  let offset = 0;
+  let openedAt: number | null = null;
+  for (const line of content.split('\n')) {
+    if (FENCE.test(line)) {
+      if (openedAt === null) openedAt = offset;
+      else {
+        ranges.push([openedAt, offset + line.length]);
+        openedAt = null;
+      }
+    }
+    offset += line.length + 1;
+  }
+  if (openedAt !== null) ranges.push([openedAt, content.length]);
+  return ranges;
+}
+
+function scanReferences(content: string): { links: FakeReference[]; embeds: FakeReference[] } {
+  const fenced = fencedLineRanges(content);
+  const inFence = (offset: number): boolean =>
+    fenced.some(([start, end]) => offset >= start && offset < end);
+
+  const links: FakeReference[] = [];
+  const embeds: FakeReference[] = [];
+  const pattern = /(!?)\[\[([^\][]+)\]\]/g;
+  let match: RegExpExecArray | null;
+  while ((match = pattern.exec(content)) !== null) {
+    if (inFence(match.index)) continue;
+    const inner = match[2];
+    const pipe = inner.indexOf('|');
+    const link = pipe >= 0 ? inner.slice(0, pipe) : inner;
+    const reference: FakeReference = {
+      link,
+      original: match[0],
+      position: {
+        start: { offset: match.index },
+        end: { offset: match.index + match[0].length },
+      },
+    };
+    (match[1] === '!' ? embeds : links).push(reference);
+  }
+  return { links, embeds };
+}
+
+export function createFakeLinkVault(initial: Record<string, string>): FakeLinkVault {
+  const files = new Map<string, string>(Object.entries(initial));
+  const processed: string[] = [];
+
+  const resolve = (linkpath: string, sourcePath: string): { path: string } | null => {
+    const wanted = linkpath.replace(/\.md$/, '');
+    const sourceFolder = sourcePath.slice(0, sourcePath.lastIndexOf('/') + 1);
+    const matches = [...files.keys()].filter(path => {
+      const withoutExt = path.replace(/\.md$/, '');
+      if (withoutExt === wanted) return true;
+      if (withoutExt.endsWith(`/${wanted}`)) return true;
+      return false;
+    });
+    if (matches.length === 0) return null;
+    matches.sort((a, b) => {
+      const exact = (p: string): number => (p.replace(/\.md$/, '') === wanted ? 0 : 1);
+      if (exact(a) !== exact(b)) return exact(a) - exact(b);
+      const local = (p: string): number => (p.startsWith(sourceFolder) ? 0 : 1);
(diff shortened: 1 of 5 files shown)


green-dalii:
This is exactly the fix I wanted to see for #386 — `vault.process` preserves the user's own files from the wiki's write gate, the resolve-before-replace closes the silent-data-loss class, and the shape preference keeps `correctRelatedLinkPrefixes` honest.

Merged in `3f27810`. Per our coordination, #393 needs a trivial rebase on top — single-line diff expected on the `merge-duplicates.ts` touchpoint.

For the `stale` counter — leave it on `console.warn` for now. A separate lint-report path belongs in its own PR; bundling it here would over-scope the bug fix.

Thank you for the design discipline.

PR #405, opened 2026-08-03
fix(#399): duplicate `sources:` frontmatter key on stub-created concept pages

### LINKED ISSUE #399: Duplicate `sources:` frontmatter key on stub-created concept pages breaks Properties render
## Bug

`buildStubContent` (in `src/wiki/lint/fix-dead-link.ts`) emits inline flow-style frontmatter:

```yaml
sources: ["[[${referringPageRel}]]"]
```

Later, `appendSourceSlugToFrontmatter` (in `src/wiki/page-factory/related-page.ts` area) tries to add a new source entry. It looks for an existing sources block via `lines.findIndex((l) => /^sources:\s*$/.test(l))` — a regex that only matches **block-style** (`sources:` on its own line). The inline flow-style form doesn't match, so it falls through to the insert branch:

```js
lines.splice(insertAt, 0, `sources:\n  - ${sourceEntry}`);
```

Result: the file ends up with **two top-level `sources:` keys** — one inline, one block-style. YAML rejects duplicate keys, so Obsidian's Properties parser bails and dumps the raw frontmatter into the note body.

## Reproduction

1. Trigger a Fix Dead Links flow that creates a stub concept page.
2. Later, run an ingest that appends a new source to that same concept page.
3. Open the file — YAML now has two `sources:` entries. Properties panel doesn't render; raw YAML shows in the body.

## Impact

In my vault (running v1.25.11), **321 wiki-concept pages** had this exact duplicate. All render broken.

## Minimal fix

Emit block-style in the stub template so `appendSourceSlugToFrontmatter` finds it:

```diff
- sources: [\"[[${referringPageRel}]]\"]
+ sources:
+   - [[${referringPageRel}]]
```

That single change in `buildStubContent` prevents new duplicates. Alternative: broaden the detection regex in `appendSourceSlugToFrontmatter` to also match flow-style and merge into it.

## Cleanup for existing files

For anyone whose vault already has the duplicates, a Python one-liner works:

```python
# See gist / attached script — merges both source lists into a single block
# then re-writes the file. Idempotent.
```

I've cleaned my own 321 files locally with a script; happy to share if useful.

---

Environment:
- Plugin: karpathywiki v1.25.11
- Obsidian: 1.13.4
- macOS

Thanks for a great plugin!

borthwick:
Fixes #399.

Two-sided fix so the hole stays closed even if a future writer emits inline flow-style `sources:` again:

## 1. `buildStubContent` (src/wiki/lint/fix-dead-link.ts)

Emit `sources:` as block-style instead of inline flow-style:

```diff
- sources: ["[[${referringPageRel}]]"]
+ sources:
+   - [[${referringPageRel}]]
```

Stops the specific collision that shipped in v1.25.11.

## 2. `appendSourceSlugToFrontmatter` (src/wiki/page-factory/create-page.ts)

Detect inline flow-style `sources:` BEFORE falling through to the insert branch. When flow-style is found, parse its wikilink entries, add the new slug if absent, and re-emit as block-style (canonical shape).

This is the durability piece @DocTpoint flagged in the issue thread — any future inline writer no longer reopens the hole. The function is now exported (was previously module-private) so it can be unit-tested directly; still not part of the plugin's public API surface.

## Test coverage

- Updated the existing `fix-dead-link.test.ts` assertion to match the new block-style stub template.
- New `append-source-slug.test.ts` with 5 cases:
  1. Merges into single flow-style entry — re-emits block-style, exactly one `sources:` key.
  2. Merges into multiple flow-style entries.
  3. Idempotent when the new slug is already present.
  4. Regression guard: block-style existing key still merged into (not duplicated).
  5. Baseline: absent `sources:` key still inserted correctly.

All 2924 tests pass (up from 2919; +5 new).

## Repair for already-affected users

Users on v1.25.11 who already accumulated duplicate keys will keep seeing broken Properties render until the files are repaired. A short Python script that merges the two keys into one block-style list is at the top of #399 — I ran it against 132 affected files in my own vault (out of 321 flagged). Happy to add a bundled cleanup command (`Repair duplicate sources: keys` in the plugin's lint menu) if you'd like — say the word and I'll add it in this PR or a follow-up.

## Environment

- Plugin fork: `borthwick/obsidian-llm-wiki` @ `fix/399-duplicate-sources-key`
- Plugin: karpathywiki v1.25.11
- Obsidian: 1.13.4 (macOS)
- Node: v24.13.0

Generated with Claude Code.

### DIFF
--- a/src/__tests__/wiki/lint/fix-dead-link.test.ts
+++ b/src/__tests__/wiki/lint/fix-dead-link.test.ts
@@ -52,9 +52,12 @@ describe('fixDeadLink stub construction (#197 — honest placeholders)', () => {
         wikiFolder: 'wiki',
         referringPageRel: 'entities/SomeSource',
       });
-      // sources array points at the referring page — better than nothing, and
-      // the placeholder text explicitly notes this is not a real source yet.
-      expect(out).toContain('sources: ["[[entities/SomeSource]]"]');
+      // sources points at the referring page — better than nothing, and the
+      // placeholder text explicitly notes this is not a real source yet.
+      // Emitted as block-style so the v1.25.11 provenance-stamp writer in
+      // create-page.ts merges into it instead of appending a duplicate
+      // top-level `sources:` key. See issue #399.
+      expect(out).toContain('sources:\n  - [[entities/SomeSource]]');
     });
 
     it('produces a concept-shaped stub when stubType is concept', () => {
--- a/src/__tests__/wiki/lint/fix-dead-link.test.ts
+++ b/src/__tests__/wiki/lint/fix-dead-link.test.ts
@@ -57,7 +57,11 @@ describe('fixDeadLink stub construction (#197 — honest placeholders)', () => {
       // Emitted as block-style so the v1.25.11 provenance-stamp writer in
       // create-page.ts merges into it instead of appending a duplicate
       // top-level `sources:` key. See issue #399.
-      expect(out).toContain('sources:\n  - [[entities/SomeSource]]');
+      //
+      // PR #405 review: wikilink value MUST be double-quoted. Bare
+      // `- [[value]]` YAML-parses as a nested flow sequence, not a string,
+      // breaking Obsidian's Properties panel + backlinks + graph edges.
+      expect(out).toContain('sources:\n  - "[[entities/SomeSource]]"');
     });
 
     it('produces a concept-shaped stub when stubType is concept', () => {
--- a/src/__tests__/wiki/page-factory/append-source-slug.test.ts
+++ b/src/__tests__/wiki/page-factory/append-source-slug.test.ts
@@ -1,0 +1,95 @@
+// Coverage for the #399 fix in appendSourceSlugToFrontmatter — the
+// v1.25.11 provenance stamp previously inserted a duplicate block-style
+// `sources:` key when the existing key was flow-style, producing invalid
+// YAML. The fix detects flow-style and merges into it (re-emitting as
+// block-style canonical shape).
+
+import { describe, it, expect } from 'vitest';
+import { appendSourceSlugToFrontmatter } from '../../../wiki/page-factory/create-page';
+
+describe('appendSourceSlugToFrontmatter (#399)', () => {
+  it('merges into an existing FLOW-style sources key (single entry) — re-emits as block-style, no duplicate key', () => {
+    const before = [
+      '---',
+      'type: concept',
+      'sources: ["[[sources/Existing]]"]',
+      'tags: [term]',
+      '---',
+      '',
+      '# Title',
+      '',
+    ].join('\n');
+    const out = appendSourceSlugToFrontmatter(before, 'New-Source_abcdef');
+    // Exactly ONE sources: key
+    const sourcesKeys = out.split('\n').filter(l => /^sources:/.test(l));
+    expect(sourcesKeys.length).toBe(1);
+    // Block-style with both entries preserved, in insertion order
+    expect(out).toContain('sources:\n  - [[sources/Existing]]\n  - [[sources/New-Source_abcdef]]');
+  });
+
+  it('merges into an existing FLOW-style sources key (multiple entries)', () => {
+    const before = [
+      '---',
+      'type: concept',
+      'sources: ["[[sources/A]]", "[[sources/B]]"]',
+      'tags: [term]',
+      '---',
+      '',
+      '# Title',
+      '',
+    ].join('\n');
+    const out = appendSourceSlugToFrontmatter(before, 'C');
+    const sourcesKeys = out.split('\n').filter(l => /^sources:/.test(l));
+    expect(sourcesKeys.length).toBe(1);
+    expect(out).toContain('sources:\n  - [[sources/A]]\n  - [[sources/B]]\n  - [[sources/C]]');
+  });
+
+  it('is idempotent when the new source is already present in flow-style', () => {
+    const before = [
+      '---',
+      'type: concept',
+      'sources: ["[[sources/Existing]]"]',
+      'tags: [term]',
+      '---',
+      '',
+      '# Title',
+      '',
+    ].join('\n');
+    const out = appendSourceSlugToFrontmatter(before, 'Existing');
+    // Unchanged
+    expect(out).toBe(before);
+  });
+
+  it('still handles the block-style existing sources key (regression guard)', () => {
+    const before = [
+      '---',
+      'type: concept',
+      'sources:',
+      '  - [[sources/Existing]]',
+      'tags: [term]',
+      '---',
+      '',
+      '# Title',
+      '',
+    ].join('\n');
+    const out = appendSourceSlugToFrontmatter(before, 'New');
+    const sourcesKeys = out.split('\n').filter(l => /^sources:/.test(l));
+    expect(sourcesKeys.length).toBe(1);
+    expect(out).toContain('sources:\n  - [[sources/Existing]]\n  - [[sources/New]]');
+  });
+
+  it('inserts a fresh block-style sources: when the key is absent', () => {
+    const before = [
+      '---',
+      'type: concept',
+      'tags: [term]',
+      '---',
+      '',
+      '# Title',
+      '',
+    ].join('\n');
+    const out = appendSourceSlugToFrontmatter(before, 'First');
+    expect(out).toContain('sources:\n  - [[sources/First]]');
+    expect(out).toContain('tags: [term]');
+  });
+});
--- a/src/__tests__/wiki/page-factory/append-source-slug.test.ts
+++ b/src/__tests__/wiki/page-factory/append-source-slug.test.ts
@@ -180,4 +180,108 @@ describe('appendSourceSlugToFrontmatter (#399)', () => {
       ]);
     });
   });
+
+  // PR #405 review — @DocTpoint notes A and B.
+  //
+  // A. "Flow heals, block doesn't": the block-append branch previously
+  //    only spliced one new line, so a v1.25.11-stamped file (with bare
+  //    `- [[x]]` continuation entries) ended up with a correct new entry
+  //    next to a broken bare one. The block branch now re-emits the whole
+  //    block from a Set, normalizing legacy entries on touch.
+  //
+  // B. The quote-tolerant dedup path had no test feeding legacy unquoted
+  //    block input.
+  //
+  // These cases feed the exact shape a real v1.25.11 vault carries today
+  // and assert both the healing behavior AND idempotency against it.
+  describe('PR #405 note A/B: legacy bare block-style entries heal on touch', () => {
+    it('bare entries get re-emitted as quoted when a new entry is appended', () => {
+      const before = [
+        '---',
+        'type: concept',
+        'sources:',
+        '  - [[sources/Existing]]',  // legacy bare
+        'tags: [term]',
+        '---',
+        '',
+        '# Title',
+        '',
+      ].join('\n');
+      const out = appendSourceSlugToFrontmatter(before, 'New');
+      // Both entries end up quoted — the old one heals.
+      expect(out).toContain('sources:\n  - "[[sources/Existing]]"\n  - "[[sources/New]]"');
+      // No bare `- [[x]]` continuation lines survive.
+      expect(out).not.toMatch(/^\s*-\s+\[\[/m);
+      // Parser-shape guard: sources is a flat string array, both entries strings.
+      const parsed = parseFrontmatter(out);
+      expect(parsed.sources).toEqual([
+        '[[sources/Existing]]',
+        '[[sources/New]]',
+      ]);
+    });
+
+    it('bare + quoted mixed continuation lines heal on touch (no duplication)', () => {
+      // The exact shape a v1.25.11 vault could carry: an old bare entry
+      // plus a newer quoted one, from two separate provenance stamps.
+      const before = [
+        '---',
+        'type: concept',
+        'sources:',
+        '  - [[sources/OldBare]]',
+        '  - "[[sources/NewerQuoted]]"',
+        'tags: [term]',
+        '---',
+        '',
+        '# Title',
+        '',
+      ].join('\n');
+      const out = appendSourceSlugToFrontmatter(before, 'ThirdEntry');
+      expect(out).toContain(
+        'sources:\n  - "[[sources/OldBare]]"\n  - "[[sources/NewerQuoted]]"\n  - "[[sources/ThirdEntry]]"',
+      );
+      const parsed = parseFrontmatter(out);
+      expect(parsed.sources).toEqual([
+        '[[sources/OldBare]]',
+        '[[sources/NewerQuoted]]',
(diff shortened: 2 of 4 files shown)


green-dalii:
Thanks for the fix — root-cause analysis is right (flow/block mismatch → duplicate key → invalid YAML → Properties panel breaks), and the two-sided fix is the right shape (close the immediate hole + make `appendSourceSlugToFrontmatter` durable against future inline writers).

Two things before we can land this:

**1. Have you actually loaded the stub in an Obsidian vault and eyeballed the Properties panel?**

The motivation in the PR body is the duplicate-key YAML problem. The fix replaces inline flow-style with block-style, which is a correct direction — but I want to be sure the replacement itself parses cleanly in Obsidian's frontmatter parser (js-yaml under the hood), since the merge target is the Obsidian Properties panel. A fix that closes the duplicate-key hole but opens a new "Properties shows `sources:` as a nested array" hole would be a net loss on user-visible output. Could you confirm what the Properties panel actually shows for a stub created with the new template on your vault?

**2. The block entries in your new format are unquoted wikilinks, which is not how the plugin's canonical writer emits `sources:`**

`src/core/frontmatter.ts` `yamlStringify()` (line 104-108) writes block entries as `  - "[[value]]"` with double quotes:

```ts
return '\n' + value.map(v => `  - "${v}"`).join('\n');
```

Your `buildStubContent` and the new flow→block branch in `appendSourceSlugToFrontmatter` both emit unquoted:

```ts
// buildStubContent
`sources:\n  - [[${referringPageRel}]]`

// appendSourceSlugToFrontmatter (the new branch in PR #405)
const blockLines = ['sources:', ...entries.map(e => `  - [[${e}]]`)];
```

I parsed both shapes with the `yaml` package (YAML 1.2, same spec as Obsidian's js-yaml):

```
unquoted: sources:\n  - [[entities/SomeSource]]
  → sources: [[["entities/SomeSource"]]]   ← nested array

quoted:   sources:\n  - "[[sources/Foo]]"
  → sources: ["[[sources/Foo]]"]            ← correct string array
```

The unquoted `- [[x]]` opens a flow sequence on the `[`, and the closing `]` matches the opening `[` of the wikilink, producing a nested flow sequence instead of a string. So the new template's `sources:` value is structurally `[[["entities/SomeSource"]]]` — not a wikilink string.

I have NOT been able to test this in Obsidian myself; my evidence is the YAML parser output. If Obsidian's Properties panel actually still shows the wikilink correctly (because Obsidian does some custom rendering over the parsed value), then the user-visible result is fine and this is a non-issue. If it shows the nested array as a plain nested list, it's a P0 regression on user-visible output.

A separate consequence: `src/core/sources-normalizer.ts` `extractRawSourcesEntries` (line 156) only matches `- "..."` (quoted) block entries:

```ts
const m = line.match(/^\s*-\s*"([^"]*)"/);
```

So the unquoted stub format silently slips past `scanPollutedSources` (returns empty array → reports not polluted) and `fixPollutedSources` (no-op). That's a pollution-detection gap for the lifetime of these stub pages, until a normal `mergeFrontmatter` cycle writes them back in quoted canonical form. Not destructive (the writer doesn't touch them), but the linter is blind to them.

Suggested fix (only if my YAML-parse observation is correct): quote the wikilink in both write sites:

```ts
// buildStubContent
`sources:\n  - "[[${referringPageRel}]]"`

// appendSourceSlugToFrontmatter (new branch)
const blockLines = ['sources:', ...entries.map(e => `  - "[[${e}]]"`')];
```

Plus a regression test that parses the resulting frontmatter with a YAML parser and asserts `sources` is `string[]`, not `Array<Array<...>>`. The current 5 tests are string-`contains` assertions — they'd pass either way and wouldn't catch this.

Happy to land this as-is if you've confirmed the Properties panel renders correctly on your vault; otherwise I'd rather get the quoted version in. Either way this isn't blocking — we're parking it in the v1.26.x PATCH milestone rather than the current MINOR (Batch 5 + release prep are next), so there's room for a revised PR.

DocTpoint:
@borthwick @green-dalii — data for the open Properties question, since I have a vault at hand.

Test note with both shapes in a single frontmatter block:

```yaml
---
sources:
  - [[sources/Test-Unquoted]]
sources_kontrolle:
  - "[[sources/Test-Quoted]]"
---
```

What the Properties panel shows (Obsidian on macOS):

| Property | Type icon | Rendered value |
|---|---|---|
| `sources` (unquoted) | **?** — type not recognized | literal text `[[["sources/Test-Unquoted"]]]` |
| `sources_kontrolle` (quoted) | list | `sources/Test-Quoted` as a link chip with a remove × |

So there is no custom rendering that rescues the nested value. @green-dalii's parser observation carries all the way through to the UI: the unquoted block form loses both the property type and the link.

For this key the second part weighs more than the render. The unquoted entry is not parsed as a link, so it produces no backlink and does not appear in the graph — and for `sources:` that edge *is* the provenance. (Observed: no link chip and no link styling. The backlink/graph consequence follows from that; I did not separately open the backlinks panel.)

Independently reproduced the parse (Ruby/psych, YAML 1.2):

```
unquoted  {"sources":[[["entities/SomeSource"]]]}
quoted    {"sources":["[[sources/Foo]]"]}
```

So the quoted variant in @green-dalii's suggestion looks necessary rather than optional. My vault is on 1.25.10 and carries none of the #399 duplicates, so I cannot test the repair path — this covers only how the two shapes render.

DocTpoint (changes requested):
Requesting changes on the strength of the Properties-panel data in my comment above — it resolves @green-dalii's open condition in the negative, so the "land as-is" branch is closed.

The two write sites need the quoted form:

```ts
// buildStubContent
`sources:\n  - "[[${referringPageRel}]]"`

// appendSourceSlugToFrontmatter (new flow→block branch)
const blockLines = ['sources:', ...entries.map(e => `  - "[[${e}]]"`)];
```

Everything else in this PR I would take as-is: the two-sided shape is right, and the flow→block branch is the part that keeps the hole closed against future inline writers.

One note on the tests — the 5 new cases assert with string `contains`, so they pass on both the quoted and the unquoted shape and would not have caught this. A case that runs the resulting frontmatter through a YAML parse and asserts `sources` is `string[]` rather than a nested array would pin it.

green-dalii:
(发给 @borthwick，副本 @DocTpoint)

@DocTpoint — thank you for the in-vault test, that settles the open question. The YAML-parse observation was theoretical until you ran it through a real Properties panel and a Ruby/psych independent parse, and your UI rendering table is exactly the evidence I needed. The unquoted `- [[x]]` form is broken at three layers simultaneously: parser shape (nested array, not string), Obsidian's property-type detection (unrecognized → literal text, no link chip), and the backlink/graph consequence (no link chip → no provenance edge). For `sources:` specifically the provenance edge is the whole point of the field, so this is a P0 on user-visible output rather than a stylistic fix.

@borthwick — the two write sites in #405 need to emit `- "[[value]]"` (double-quoted) instead of `- [[value]]`. The plugin's canonical writer at `src/core/frontmatter.ts` `yamlStringify()` (line 104-108) already uses the quoted form, so this is just matching the canonical shape. Two specific edits:

1. `src/wiki/lint/fix-dead-link.ts:73` — `buildStubContent`:
   ```ts
   // before (unquoted — breaks Properties)
   `sources:\n  - [[${referringPageRel}]]`
   // after (quoted — matches canonical)
   `sources:\n  - "[[${referringPageRel}]]"`
   ```

2. `src/wiki/page-factory/create-page.ts` — the new flow→block branch you added in `appendSourceSlugToFrontmatter`:
   ```ts
   // before
   const blockLines = ['sources:', ...entries.map(e => `  - [[${e}]]`)];
   // after
   const blockLines = ['sources:', ...entries.map(e => `  - "[[${e}]]"`)];
   ```

Plus a regression test that parses the resulting frontmatter with a YAML parser and asserts the structure is `string[]`, not `Array<Array<...>>`. The 5 existing tests use string `contains` so they'd pass either way; the parser-shape assertion is what catches the regression.

We're parking #405 in the v1.26.x PATCH milestone (not blocking v1.26.0 MINOR), so there's room for a revised PR. Ping me when you push the fix or have questions about the canonical-format reasoning. Once merged in PATCH, the build-stub-write-path and the flow→block-merge-path both produce frontmatter the Properties panel can render correctly.

amdeldacc (approved):
**Approved — root cause fix is correct and verified independently.**

I investigated the same bug from a real vault (karpathywiki v1.25.11, 321 pages affected) and reached the identical root cause: `appendSourceSlugToFrontmatter` (create-page.ts) only detects the block-style `sources:` via `/^sources:\s*$/`, so any inline flow-style `sources: ["[[...]]"]` (emitted by `buildStubContent`, and passed through verbatim by `enforceFrontmatterConstraints`) falls through to the "insert a new key" branch, producing a second top-level `sources:` key that breaks Obsidian's Properties render.

Verification done:
- `fix-dead-link.test.ts` + `create-page.test.ts` + new `append-source-slug.test.ts`: **29/29 pass**.
- Full suite: **2914/2916 pass**; the 2 failures (`custom-instructions-injection`, `openai-codex-loopback-flow`) also fail on `main` — pre-existing, unrelated.
- Edge case I probed: a file that already carries **both** flow-style and block-style `sources:` keys (the exact shape of already-corrupted v1.25.11 files) still yields 2 keys after this fix, because the flow branch returns early and leaves the duplicate block untouched. This matches the PR's stated scope (new files are fixed; existing ones need a cleanup pass), so not a blocker — just confirming the "repairs already-affected files as they are rewritten" claim from the issue thread only holds for flow-only files, not both-keys files. Worth a sentence in the cleanup-command follow-up if you add one.

Good test coverage, clean diff, exactly the durability fix the thread asked for.

borthwick:
@green-dalii @DocTpoint — thanks for the review. Pushed `abf3cfd` addressing the CHANGES_REQUESTED feedback.

**Changes:**

1. **All three write sites now quote wikilink values** to match `yamlStringify()` in `src/core/frontmatter.ts:104`:
   - `src/wiki/lint/fix-dead-link.ts` `buildStubContent` — the stub template
   - `src/wiki/page-factory/create-page.ts` `appendSourceSlugToFrontmatter`:
     - the new flow → block merge branch (PR #405's addition)
     - the fresh-insertion branch (pre-existing but same bug — would emit unquoted whenever a note had no `sources:` at all)
     - the block-append branch (pre-existing but same bug — every provenance-stamp write on an existing block-style key was unquoted)

2. **Dedup Set in the block-append branch** is now tolerant of both quoted and unquoted continuation lines, so old-format entries already in a vault don't trip idempotency.

3. **Regression test** (`append-source-slug.test.ts`) — added three new cases under a `PR #405: parser-shape guard` describe block. They parse the resulting frontmatter with the `yaml` package (YAML 1.2, same spec as Obsidian's js-yaml) and assert `sources` is `string[]`, not `Array<Array<...>>`. This is what catches the regression — the 5 existing string-`contains` cases pass on both shapes.

4. Updated the 5 existing `contains` assertions to the new quoted format.

**Test suite: 2,927/2,927 passing across 216 files.**

@DocTpoint — your in-vault Properties table + independent Ruby/psych parse was the exact evidence that made the case unambiguous. Thank you for taking the time.

Ready for another look when either of you has a moment. Happy to squash before merge if you prefer.

DocTpoint (approved):
**Re-reviewed at `abf3cfd`. The finding I raised is resolved — approving.**

All three write sites now quote, and matching `yamlStringify()` exactly was the right call rather than improving on it locally. I grepped upstream `main` for other frontmatter emitters: nothing else writes `- [[…]]` outside i18n report strings and tests, and `merge-page` goes through `serializeFrontmatter`, so your three are the complete set. Finding the fresh-insertion and block-append branches while you were in there is the part that made this worth more than the reported bug.

Three notes, none blocking.

**1. Flow heals, block does not.** The flow branch re-emits the whole list, so a legacy file is normalised on touch. The block branch splices one line and leaves the existing continuation lines as they are. A note stamped by v1.25.11's unquoted appender therefore ends up with a correct new entry next to a bare one — and the bare one still parses as a nested sequence, so the property stays broken for that file. New vaults are fully fixed; existing ones are not. You already have the entries in the `existing` Set, so re-emitting the block from it would close that.

**2. The quote-tolerant dedup is untested.** No case feeds legacy block-style input (`sources:\n  - [[sources/A]]`), which is the shape real vaults carry today — so the strip-both branch and idempotency against it are unproven. That is the one test I would add regardless of what you decide on note 1.

**3. Minor.** In the flow branch, only `[[…]]` matches are collected, so a non-wikilink entry (`sources: ["plain-slug", "[[a]]"]`) is silently dropped on re-emit.

Squashing before merge is fine by me.

borthwick:
@DocTpoint @green-dalii — thanks for the re-review. Pushed `fb19ef7` addressing notes A and B.

**A. Block-append branch now heals legacy bare entries** — behavior parity with the flow branch. Collect all continuation entries into a Set (quotes stripped), dedup, append new entry, splice canonical quoted re-emission for the whole block. Any stamp on a v1.25.11-shaped file normalizes the entire `sources:` block on touch. New vaults were already fixed; existing ones now heal too.

**B. Added 4 test cases** for the quote-tolerant dedup / block-heals path — the exact shapes real vaults carry:
- bare-only continuation lines → heal on new stamp
- bare + quoted mixed → all normalize, no duplication
- idempotency against a legacy bare entry (target slug matches existing bare)
- idempotency against a legacy bare entry (dedup catches it before re-emit)

Each asserts quoted output AND parses the frontmatter to confirm `sources: string[]`.

**Skipping note C** (non-wikilink entries dropped on flow-branch re-emit) — I couldn't find a real-world vault shape that carries `sources: ["plain-slug"]`, and adding a fall-through would require inventing a policy for how non-wikilink strings should render. Happy to revisit if it turns out real.

Test suite: **2,931/2,931 passing** across 216 files (+4 from prior 2,927).

@green-dalii — with DocTpoint's approval + amdeldacc's approval + the healing behavior extension, this should be ready for your re-review when you have a moment.

green-dalii (approved):
Approved. The two-sided fix (block-style emission + flow-style merge in `appendSourceSlugToFrontmatter`) addresses the v1.25.11 regression and any pre-existing inline frontmatter. The 5 new unit tests cover the canonical cases, and the 3 parser-shape guards (using `yaml` package to assert `sources` is `string[]`) close the shell-test failure mode — string-contains assertions would have silently passed on the bare `- [[x]]` bug, but `yaml.parse` correctly rejects that as a nested flow sequence. The 4 DocTpoint note A/B cases (legacy bare block entries heal) are exactly the regression tests needed for the 321 affected pages in borthwick's v1.25.11 vault.

One trailing cleanups suggested for the squash-merge commit (not blocking):

1. CHANGELOG.md: add an entry under v1.26.x PATCH: `fix: duplicate sources: frontmatter key on stub-created concept pages (Issue #399, PR #405)`.

Will route through Gate 1 (lint/tsc/test/build) before merge.

borthwick:
Thanks @green-dalii, @DocTpoint, and @amdeldacc — appreciate the careful review across three rounds. Glad this lands cleanly in v1.26.x PATCH. Closing out on my end; ping me if anything surfaces once it ships.

PR #409, opened 2026-08-04
feat: name each LLM call by step, and time the steps inside a phase

eucher:
An ingest is tens of calls and nothing says which step any of them is. The
debug log prints a provider, a model and a prompt length per call; the engine
times six phases, but the largest — page generation — is one interval covering
path resolution's dedup call, the page write and the merge routing.

`createMessage` takes an optional `task` label, read in one place:
`wrapWithAdvancedSettings`, the seam every call passes through since
`createLLMClient` always returns that wrapper. It is logged there, and the time
under each label accumulates in `core/llm-task-usage.ts`, which the engine
prints beneath `[Phase breakdown]` and the CLI prints as its own table.

Re-ingesting an 8.5 KB source that already had pages (2 created, 40 updated):

    step               calls   out tok    seconds   share
    merge-triage          20     36139      578.6     36%
    merge-body            13     41596      562.2     35%
    extract                2     26031      195.3     12%
    related-page           4     18702      173.8     11%
    page-generate          1      4802       47.7      3%
    source-page            1      5370       37.7      2%
    dedup                  1      1000       18.8      1%
    lemma-classify         1      1065        7.2      0%

71% of the run went to merge routing, and `merge-triage` — a call that answers
with a strategy name — cost 29s each because it carries the whole existing page.
The phase breakdown cannot say that: both steps sit inside `Page gen`.

No client reads the label — verified on the wire, not from the source: with a
fetch interceptor the request bodies carry `model,max_tokens,messages` (compat
and anthropic) and `model,input,max_output_tokens` (openai), and no `task` in
any of them.

Totals are cumulative and callers diff against a snapshot rather than reset: a
reset would be wrong the moment two ingests overlap. Time is taken in a
`finally`, so a call that throws still counts against its step.

Streaming is left alone: the one `createMessageStream` call site is the query
answer, user-initiated and singular — the label would print `untagged` for the
only caller there is.


green-dalii (changes requested):
Requesting changes — the per-step timing data is valuable (the 71% merge-triage finding is the kind of signal the phase breakdown cannot give), but the change adds a new code path on the wrapper hot path and ships zero tests for it.

## Required for merge

1. **New test file `src/__tests__/core/llm-task-usage.test.ts`** covering:
   - `recordTaskUsage` accumulates under the supplied label
   - `params.task` undefined → `recordTaskUsage` receives `undefined` → `taskUsageSince` shows it under `'untagged'`
   - `snapshotTaskUsage` + `taskUsageSince` round-trip: snapshot, record 3 calls with different labels, diff, assert counts and millis
   - `formatTaskUsage` shape (one line per step, sorted slowest first)
   - Concurrent calls: snapshot diff is the right primitive (not reset)

2. **Extend `src/__tests__/core/llm-client-wrapper.test.ts`** with one new test asserting the new branch:
   - When `params.task` is set, `recordTaskUsage` is called with that label + the elapsed millis
   - When `client.createMessage` throws, `recordTaskUsage` still fires (try/finally)

3. **Verify `console.debug` is dropped in production builds.** Run `pnpm build` and grep the bundle for `[llm] task=`. The comment in the diff claims it's stubbed out along with every other `console.debug`, but the Bot will flag it if the build doesn't actually drop the call. If it doesn't drop, either:
   - Configure esbuild to drop the line in `build.js`, or
   - Move the debug log behind a feature flag / debug-build constant.

## Recommended (not blocking)

4. Document the `'untagged'` fallback rationale in CLAUDE.md so future call-site additions pick a label. Two follow-up paths the next contributor should pick one of:
   - (a) Make `task` required on `LLMClient.createMessage` and remove the `?? 'untagged'` fallback (breaking-change, separate PR)
   - (b) Add a `console.warn` for the first call per ingest that lands in `'untagged'`, pointing at the source line via `new Error().stack`

5. CHANGELOG entry under `[Unreleased]` — the value proposition (71% merge-routing is the real story) is worth a user-visible note.

## Effort

Test file ~80 LOC + wrapper test extension ~25 LOC + build verification 5 min. ~half a day. Worth it for a change that lives in the wrapper hot path.

eucher:
**Items 1, 2, 4 done; item 3 measured, and the finding is that this call is not special.**

**1 + 2 — tests** (`9cba140`). `llm-task-usage.test.ts`: accumulation under a label, the `untagged` fallback, one entry per label slowest first, `formatTaskUsage` shape including the singular, a baseline read twice being empty the second time, and a snapshot surviving a later call recording into the same map. `llm-client-wrapper.test.ts`: the label reaching the ledger through the wrapper, an unlabelled call landing in `untagged`, and a call that threw still being counted.

Written without mocking `recordTaskUsage`. The property under test is that the diff is correct while the map is shared, and a mock would only assert that the function was called — the round-trip through the real module is what fails if the primitive is wrong.

**3 — measured on the branch, production build.** The line survives, and so do the other 280:

| build config | `[llm] task=` in bundle | `console.debug` calls | size |
|---|---|---|---|
| as shipped (banner) | 1 | 281 | 3 731 656 |
| `pure: ['console.debug']` | 1 | 281 | unchanged |
| `pure` + `minifySyntax` | 1 | 281 | unchanged |
| `drop: ['console']` | **0** | 1 (the banner's own assignment) | −41 KB |

`pure` does not remove statement-level calls here, and esbuild's `drop` takes `console` whole — there is no per-method form. So the only setting that empties the bundle also removes **55 `console.warn` and 58 `console.error`**, among them the SecretStorage migration failure, the LLM client init failure, and the truncation-detected warning — the lines a user's console is read for.

There is no debug flag anywhere in the codebase (`DEBUG`, `debugMode`, `isDebug` — none), and of the 281 calls the 61 that sit under an `if` are gated by ordinary logic, not by a debug switch. Putting this one line behind a new constant would introduce a second convention for a single call site.

This line is therefore shaped exactly like its 280 neighbours: the banner has neutralized them since `e21182c`, `console.log` is unused, and there is no lint rule. What remains true is that arguments are built before a no-op call — for all 281 of them, which makes it a build-wide question rather than a property of this change.

**4 — `CLAUDE.md`** (`191369e`), under Key Design Decisions: the label is read in one seam, an unlabelled call is filed under `untagged` rather than dropped because it still costs time, and a new `createMessage` call site picks a label named for the step.

green-dalii:
Items 1, 2, 4 done; item 3 measured and the finding is that the line is shaped like its 280 neighbours, and your measurement table is the right kind of argument. Approved.

Two things before merge:

1. CLAUDE.md is likely to conflict again — please rebase onto current main (`308124d`, which includes PRs #422 and #427 from earlier today). `gh pr update-branch --rebase` is preferred over a fresh fork.

2. CHANGELOG entry — leave it for the v1.26.1 [Unreleased] sweep (PR #430 was the last one). I'll add a one-line Fixed note referring to the per-step table at that time. Not asking you to widen this PR.

I'll re-request your review after the rebase lands.

green-dalii:
`gh pr update-branch --rebase` failed on the PR-side with a conflict error. Locally, `git merge main` (ort strategy) resolves cleanly against the current tip — the only files touched by main since your last rebase are `src/types.ts` (auto-merged) and `tools/llm-wiki-cli/src/main.ts` (auto-merged), with no semantic conflict.

I tested the merged state in a local worktree at `abc6896` (your branch + main at `6fb089b`):

- `pnpm lint` 0/0
- `npx tsc --noEmit` 0
- `pnpm build` clean
- `pnpm test` 2992/2992 (218 files)

Could you push the resolution? Two paths:

1. **Preferred — rebase locally and force-push:**
   ```
   git fetch upstream
   git rebase upstream/main
   git push --force-with-lease
   ```

2. **Alternative — merge main into your branch and push:**
   ```
   git fetch upstream
   git merge upstream/main
   git push
   ```

Either works; the merge approach produces an extra merge commit but preserves your existing commit history verbatim. Once your push lands, the merge button is ready.

eucher:
Pushed as a merge (`c492289`), taking the second path — the branch now contains `main` at `6fb089b`.

Merge went clean; the only file that needed a hand was `CLAUDE.md` in the earlier merge, where main had reworded the four Key Design Decisions bullets to point at CONTRIBUTING.md while this branch added a fifth. Kept main's wording and put the new bullet after it.

On the merged tree: lint 0/0, `tsc --noEmit` 0, build clean, **2992/2992 tests in 218 files** — the same numbers you measured at `abc6896`.

No force-push, so `9cba140` and `191369e` keep the hashes the earlier comment refers to.

green-dalii (approved):
Gate 1 verified on the merged tree at c492289 (worktree, build-before-test order): lint 0/0, tsc 0, build clean, 2992/2992 tests in 218 files, css-lint 0. Matches your reported numbers. The single test-fail on first run was the pre-existing main.js bundle dependency in openai-codex-loopback-flow.test.ts (needs `pnpm build` first), not this change.

Approving per the earlier discussion — items 1/2/4 done, item 3 measured with the right argument. CHANGELOG entry deferred to the v1.26.1 [Unreleased] sweep per prior comment.

Issue #382, opened 2026-07-31
[v1.26.0 hardening] cross-type dedup + lint thresholds + streaming refactor + P1-1/P1-2 wire-up + dead-code policy


green-dalii:
## Summary

Five NEW items identified during v1.26.0 MINOR kickoff (2026-08-01) after third-party v1.25.11 deep evaluation + critical-thinking review of the existing 8-item #358 committed scope. The most important is a **P0 correctness gap** in cross-type dedup that v1.26.0's Phase 2 (per-type registration via Settings) would amplify if not addressed first.

### 1. Cross-type dedup candidate set expansion (P0)

**Symptom (code-verified):** `src/wiki/lint/duplicate-detection.ts:124-147` + `dedup-phase.ts:123-127` only generate dedup candidates from **historical wiki pages** (`entityConceptFiles = input.wikiFiles.filter(...)`). In-flight entities/concepts extracted by `source-analyzer` for the current ingest batch **do not** participate in the lint dedup prompt — they only get checked against historical pages via `path-resolution.ts:128`'s "historical cross-type duplicate" detection, then either silently appended as aliases or written as duplicates that the *next* lint round discovers.

**Concrete scenario:** ingest a 30-entity source where 5 should merge with existing vault entities → source-analyzer writes 5 duplicate pages; lint dedup next round catches them but at the cost of vault churn and broken `[[link]]` references during the window.

**Why this must ship before Phase 2 (per-type registration):** Phase 2 introduces user-defined type names (`engagements/`, `findings/`, `risks/` per FrancoTampieri), which means *cross-type* candidates will become more numerous AND more semantically distinct (entity vs. finding vs. engagement are not just "entity vs. concept" anymore). Without expanding the dedup candidate set first, Phase 2 amplifies the existing blind spot into default behavior.

**Effort:** 1.5-2 weeks (depends on item 3 streaming refactor).

### 2. Lint dedup thresholds published as advanced tunables (P0)

**Symptom (code-verified):** `src/wiki/lint/duplicate-detection.ts` has magic numbers `JACCARD >= 0.4` (line 138), `bodySim < 0.2` (line 143), `maxSim >= 0.4` (line 172) — not in `src/constants.ts`, not user-settable, no documentation.

[...]

Issue #403, opened 2026-08-03
Reasoning tokens are billed against max_tokens: three capped call sites return empty and the empty answer is read as a valid negative


DocTpoint:
On a reasoning-capable model the deliberation is billed against the same `max_tokens` budget as the answer. Three call sites carry caps that were sized for a non-reasoning model, so the budget is spent before the first content token:

- `TOKENS_COMPLEMENTARY_APPEND = 600` (`src/constants.ts:212`)
- `TOKENS_DEDUP_RESOLUTION = 1000` (`src/constants.ts:169`)
- `TOKENS_MERGE_TRIAGE = 2000` (`src/constants.ts:203`)

## Measurement

Two ingests of the same 2.4 KB source note, same model (`gemma-4-12b`, 4bit, LM Studio), 45 LLM calls each. 14 calls returned `finish_reason: "length"`; 13 of them returned empty content. On a truncated call `completion_tokens = cap - 1`, so the cap identifies the call site.

| call site | cap | run A | run B | reasoning tokens of the calls that succeeded |
|---|---|---|---|---|
| `complementaryAppend` | 600 | 3 of 3 | 0 of 3 | — (ceiling 597) |
| `resolveEntityDedup` | 1000 | 1 of 11 | 2 of 7 | 286–921 (ceiling 997) |
| `mergeAnalysis` (triage) | 2000 | 3 of 5 | 5 of 12 | 766–1650 (ceiling 1997) |

This is not a tail event. The successful triage calls deliberate between 766 and 1650 tokens against a ceiling of 1997; dedup runs 286–921 against 997. Whether a call survives is close to a coin flip. Wasted output: 8 793 tokens in run A, 11 993 in run B, for zero content.

## Why it matters differently at each site

**`resolveEntityDedup`** — silent. `parseJsonResponse` returns null, `result?.match` is falsy, and the function falls through to `return { path: slugPath }` (`path-resolution.ts:236`). A truncated identity check is indistinguishable from a genuine "no match", so a new page is created where an existing one should have been matched. This creates duplicates with no error surfaced.

**`mergeAnalysis`** — loud. `merge-triage.ts:124` throws `merge triage: empty response`, and the caller treats a triage failure as non-fatal and falls through to the full merge path (`merge-page.ts:19`). The `skip` decision is lost and the page is rewritten by the LLM unnecessarily.

[...]

Issue #137, opened 2026-06-16
Google AI Studio API key works in Obsidian Copilot.


SiYeon-Ju:
Google AI Studio API key works in Obsidian Copilot.

Google Gemini provider returns HTTP 400.

OpenAI-compatible endpoint (https://generativelanguage.googleapis.com/v1beta/openai) cannot fetch models and returns HTTP 400 when testing gemini-2.5-flash.

Is Gemini currently supported?

------------------------------------------------------------------
Google AI Studio API key works.
curl against https://generativelanguage.googleapis.com/v1beta/openai/models works.

Gemini Provider returns HTTP 400.

OpenAI Compatible provider with
https://generativelanguage.googleapis.com/v1beta/openai
also returns HTTP 400.

Same API key works in Obsidian Copilot.

------------------------------------------------------------------------------------

green-dalii:
Thanks for the detailed reproduction — it really helps to see that curl works and the same key works in Copilot. That ruled out the usual suspects (key validity, Google-side outage) quickly.

Confirmed root cause: the plugin's Gemini path goes through the OpenAI-compatible endpoint and unconditionally sends `thinking: { type: 'disabled' }` (because v1.19.0 defaults disableThinking=true). Gemini's OpenAI-compat layer doesn't recognize that field and returns 400 with "Unknown name 'thinking'". The existing fallback then injects `chat_template_kwargs`, which Gemini also rejects, so the request never recovers.

The fix is to add a third fallback level that tries OpenAI's `reasoning_effort: 'none'` dialect (which is how Gemini 2.5+ expects thinking to be disabled on its OpenAI-compat surface), and to cache which dialect each baseUrl accepts so subsequent calls don't probe twice. No hardcoded provider check — the same fallback chain will gracefully handle any other OpenAI-compat backend that uses a different thinking-control dialect.

Hotfix branch is already cut from v1.19.0. Targeting a fix today (2026-06-17).

green-dalii:
Hi @SiYeon-Ju — just wanted to let you know this is now fixed in **v1.19.1**, just released 🎉

The root cause was that Gemini's OpenAI-compatible endpoint (`generativelanguage.googleapis.com/v1beta/openai`) doesn't recognize Anthropic's `thinking.type='disabled'` field. The fix adds a 3-tier fallback chain that auto-detects the correct dialect per provider:

1. `thinking.type='disabled'` (OpenAI, DeepSeek, OpenRouter — the default)
[...]

PR #411, opened 2026-08-05
feat(sdk): v1.26.0 Batch 6+7 — real-wire thinking disable + dedup parse-failure routing

green-dalii:
## Summary

Two MINOR-blocking items from DocTpoint's #382 review thread (2026-08-04):

1. **Batch 6 — force-disable thinking Layer-1 + Layer-3 (real wire fix)**
   PR #410 / Batch 2 shipped `enableThinkingOverride=false` for the dedup-phase
   using `thinking.type='disabled'` + `chat_template_kwargs.enable_thinking=false`
   on the openai-compat path. DocTpoint verified via fetch-interceptor (Issue #382
   comment 2) that **neither field reached the wire** — the AI SDK strips both
   before the body is built. The e2e 979s→365s gain came from retry/halving,
   NOT from thinking being disabled.

   Batch 6 corrects this with `reasoningEffort: 'none'` (camelCase), which the
   AI SDK's zod schema DOES accept (line 331 of `@ai-sdk/openai-compatible@2.0.62`).
   DocTpoint re-measured: openai-compat path now actually reaches wire with
   `reasoning_effort: 'none'`; reasoning_tokens=0 on LM Studio / gemma-4-12b.

2. **Batch 7 — dedup-phase parse-failure routing (DocTpoint #382 comment 1)**
   `parseJsonResponse` returns null in two distinct outcomes that previously
   collapsed to the same `[]` return: legitimate empty (`{"duplicates":[]}`)
   and parse failure / truncated. Batch 7 routes parse failures into
   `dedupFailures` with a distinct reason tag + structured `type` discriminator.
   Surfaces real truncation count for the first time.

## Layer-3 fallback (4-layer design)

1. **Primary** — `reasoningEffort: 'none'` (camelCase) — passes zod filter,
   emits as `reasoning_effort: 'none'` on wire (verified).
2. **Co-emit** — same `reasoningEffort` for the Anthropic SDK path
   (Anthropic uses `thinking: { type: 'disabled' }` as its working switch).
3. **400-retry guard** — on HTTP 400 with rejection-verb + reasoning-field
   marker in the message, retry once with reasoningEffort stripped;
   per-baseURL cache prevents infinite loops.
4. **Prompt-level** — "**Do not reason step by step**" in dedup prompt
   (added in Batch 2).

No per-vendor matching — fixed list per user guidance (2026-08-04):
"做好通用、完善的fallback机制即可".

### Production code
- `src/llm-sdk/openai-compat-sdk-client.ts` — `reasoningEffort: 'none'`
  in `buildProviderOptions` (with `shouldStrip` guard); 400-retry block
  added before token-key block; markStrip moved to after retry success.
- `src/llm-sdk/openai-sdk-client.ts` — same `reasoningEffort: 'none'`
  + 400-retry on both createMessage and createMessageStream paths
  (streaming Layer-3 was missing on the OpenAI SDK — CR-4 fix).
- `src/llm-sdk/reasoning-strip-probe.ts` (NEW) — per-baseURL "strip
  reasoningEffort" cache + two-marker classifier (rejection verb AND
  field marker — mirrors `isPdfRelatedLlmError` two-part shape).
- `src/wiki/lint/llm-phases/dedup-phase.ts` — `=== null` check
  after `parseJsonResponse` → push to `dedupFailures` with
  `type: 'parse-failure'` discriminator.
- `src/core/rate-limit.ts` — `isRateLimitFailure()` accepts either
  a string (original contract) OR structured item with `type` field.
  `type: 'parse-failure'` short-circuits to false unconditionally.

### Tests (+9 new tests)
- `src/__tests__/llm-sdk/reasoning-strip-probe.test.ts` (NEW) — 12 cases
  covering cache lifecycle + two-marker classifier (including the
  `*-thinking` model-name false-positive regression guard).
- `src/__tests__/llm-sdk/openai-compat-request-body.test.ts` — wire
  regression guard asserting `reasoning_effort: 'none'` IS on the body.
- `src/__tests__/llm-sdk/openai-compat-sdk-client.test.ts` — 400-retry
  integration tests (3 new).
- `src/__tests__/llm-sdk/openai-sdk-client.test.ts` — `'none'` wire
  assertion (replaces `'low'`).
- `src/__tests__/wiki/lint/llm-phases/dedup-phase.test.ts` — 3 tests
  (2 updated to Batch 7 contract, 1 new legitimate-empty regression guard).
- `src/__tests__/core/rate-limit.test.ts` — 9 tests covering structured
  discriminator short-circuit.

### Docs
- `CLAUDE.md` — new "Force-disable thinking — 4-layer fallback" section,
  corrected Layer-2 description (misaddressed, not stripped — per
  DocTpoint's third #382 comment).
- `ROADMAP.md` — Batch 5 ❌ cancelled, Batch 6/7 added, v1.26.x PATCH
  track extended with `repetition_penalty` defect + per-id key fix.

## Out of scope

- **CR-1 (`consecutiveThrottleChunks` declared inside for loop — halving
  never fires)** is pre-existing on `main` (Batch 2 commit `e2e75eb`).
  Code-review correctly noted the CLAUDE.md attribution is wrong
  (the 979s→365s gain came from retry/backoff only — halving was inert).
  Fixing CR-1 changes Batch 2 behavior and should ship as a separate
  v1.26.x PATCH PR with a CLAUDE.md / memory correction. NOT in this PR.
- **Per-id key correction** for `thinking` / `chat_template_kwargs`
  (DocTpoint's #382 third comment) — depends on Layer-3 guard which
  is now in place. Tracks in v1.26.x PATCH.
- **`repetition_penalty` visible defect** — tracks in v1.26.x PATCH.

## Verification

- **Gate 1**: lint 0/0, tsc 0/0, 2925 tests pass, build clean, css-lint clean
- **Wire-level**: openai-compat-request-body.test.ts asserts
  `reasoning_effort: 'none'` IS on the body (regression guard PR #410
  missed — the old test only checked the SDK argument, not the wire)
- **AI SDK source citations** in CLAUDE.md:
  - `@ai-sdk/openai-compatible@2.0.62:322-344` (zod schema) / `531-540` (filter) / `541` (emit)
  - `@ai-sdk/openai@3.0.86:693` (zod enum includes `'none'`) / `943` (skip reasoning path)
  - `@ai-sdk/anthropic@3.0.98:928` (zod accepts `thinking: { type: 'disabled' }`)

## Review notes for DocTpoint

@DocTpoint — three of your #382 review findings shipped in this PR
thread; one P0 (`consecutiveThrottleChunks`) is deferred to v1.26.x
PATCH per pre-existing-on-main caveat. Specifically:

- **#382 Observation 1** (parse-failure vs legitimate-empty): ✅ Batch 7
  routes null to dedupFailures with structured `type: 'parse-failure'`
  discriminator (CR-3).
- **#382 Observation 2** (force-disable is no-op on openai-compat):
  ✅ Batch 6 fixes with `reasoningEffort: 'none'` (zod-accepted) +
  Layer-3 400-retry guard for backends that reject.
- **#382 third comment** (misaddressed, not stripped): ✅ Layer-2
  description in CLAUDE.md corrected to document the per-id key
  mismatch. Per-id key correction tracked in v1.26.x PATCH (depends
  on Layer-3 guard, now in place).

## Do NOT merge yet

Per `feedback_pr_merge_workflow`: this PR is for review. Awaiting
user sign-off before merge into `main`.

Co-Authored-By: Claude Code <noreply@anthropic.com>


DocTpoint:
Read through the diff. Layer 3 in both clients, the wire test that parses the actual body, and the two-marker classifier all look right to me, and removing the two dead fields rather than keeping them is the better of the two endings. Five notes, one of which I would fix before merge.

**1. The post-mortem in the code has the superseded explanation, and it is attributed to me.**

Three places say the SDK's `filter()` at `531-540` deletes `thinking` / `chat_template_kwargs` / `repetition_penalty`: the `buildProviderOptions` doc block, the `repetition_penalty` comment below it, and the header of `openai-compat-request-body.test.ts`.

That filter keeps the keys that are *not* in the zod shape and spreads them into the body — it is an extra-field passthrough. The fields do not travel because `buildProviderOptions` returns them under `openaiCompatible` while the SDK looks up `providerOptions[provider id]`, and no id in `types.ts` is the literal string `openai-compatible`.

`CLAUDE.md` in this same PR states this correctly. The source comments still carry my first, wrong version. The test file header is self-contradictory: its opening paragraph states the rule correctly ("only under a key matching the provider's own name, OR through fields declared in the SDK's zod schema"), the paragraph below repeats the strip claim.

**2. @eucher's comment block was removed along with the decision it recorded.**

It carried the provenance for this area — #137, the `unsupportedFields` strip-and-retry lost in the AI-SDK migration, and an explicit note on how weak that evidence is. Removing the two dead fields is, I think, the right resolution of what he left open. But he is not mentioned in the PR, and the reasoning now exists only in git history.

**3. `reasoning_effort: 'none'` for DeepSeek / Kimi / GLM is asserted, not measured.**

The deleted comment gave a concrete reason for moving *away* from `reasoningEffort`: DeepSeek accepts only `high`/`max`, and `low` was silently mapped to `high` — which is why Batch 2 switched to `thinking.type` in the first place. The new comment lists "DeepSeek — accepts (official reasoning_effort field)" with no source.

If the old note is right, `'none'` there is either a 400 — in which case Layer 3 caches "strip" and force-disable is off for that baseURL from then on — or silently mapped upward, i.e. maximum reasoning instead of none. The three vendor doc links are in the deleted block; worth a check before tagging.

**4. The cost of a classifier miss is understated in both directions.**

A false negative does not cost "one extra HTTP call on the next request". No reasoning retry runs, the call fails, and the token-key branch below fires on the same 400 — caching `max_completion_tokens` for that baseURL and retrying with `reasoning_effort` still set. So an unrecognized reasoning 400 also poisons the token-key cache.

In the other direction, `CLAUDE.md` calls a false-positive strip "bounded (one extra HTTP call)", while the module comment describes the same event as a durable functional regression that silently disables force-disable-thinking for the rest of the session. The module has it right.

**5. Small one, Batch 7.**

`Array.isArray(rawDups) ? rawDups : []` is unchanged, so a response that parses but whose `duplicates` is not an array still reads as "no duplicates confirmed". Much narrower than the null case you fixed, same shape of conflation.

Only item 1 seems worth blocking on, and it is comment text — but it is the version that outlives the thread. Item 3 is the one I would actually measure.

eucher:
**`reasoning_effort: 'none'` on llama.cpp — measured, and the reason from llama.cpp's source.**

Batch 6 picks it on the strength of LM Studio / gemma-4. Here is the same question on llama.cpp with `Qwen3.6-35B-A3B`.

`tools/server/server-common.cpp:1089-1095`, byte-identical in `master` and in `b10210` — the build that answered below:

```cpp
// Parse also the OAI "reasoning_effort": "none" specific value
if (body.contains("reasoning_effort")) {
    auto reasoning_effort = json_value(body, "reasoning_effort", std::string(""));
    if (reasoning_effort == "none") {
        inputs.enable_thinking = false;
    } // other reasoning_effort values are model-specific and not yet handled
}
```

Exact string comparison, setting the same `inputs.enable_thinking` the `chat_template_kwargs` block sets just above, and it runs before `common_chat_templates_apply`.

Prompt `What is 17 + 25?`, `temperature: 0`, `seed: 42`, `max_tokens: 2048`, `cache_prompt: false` (`cached_tokens: 0` on every response, nothing near the cap), `system_fingerprint: b10210-000547513`. The first four rows were interleaved over three cycles and repeat exactly; `sha` is over `reasoning_content + content`:

| sent | prompt tok | completion tok | `reasoning_content` | sha |
|---|---|---|---|---|
| nothing | 20 | 242 | present | `31a7b397` |
| `thinking: {"type": "disabled"}` | 20 | 242 | present | `31a7b397` |
| `reasoning_effort: "low"` / `"high"` / `"NONE"` / `"banana"` | 20 | 242 | present | `31a7b397` |
| an invented top-level field | 20 | 242 | present | `31a7b397` |
| `reasoning_effort: "none"` + `chat_template_kwargs: {"enable_thinking": true}` | 20 | 242 | present | `31a7b397` |
| `chat_template_kwargs: {"enable_thinking": false}` | **22** | 131 | absent | `893f38b7` |
| `reasoning_effort: "none"` | **22** | 131 | absent | `893f38b7` |

Eight variants return the baseline answer byte for byte; the two levers return one answer of their own. What carries the mechanism is the `prompt_tokens` column rather than the hashes: 22 against 20 means the rendered prompt is different, so the rewrite happens at prompt construction, not on the reply.

For this PR: on this backend `reasoningEffort: 'none'` reaches the same state as `chat_template_kwargs`, by the same flag, so the substitution holds — and `thinking.type` is inert here, byte-identical to sending nothing.

A footnote on `dedup-phase.ts:341-346`, which the PR leaves as it is:

> `thinking.type: 'disabled'` + `chat_template_kwargs` … silently ignored by most other OpenAI-compatible backends (LM Studio, llama.cpp)

The outcome is right for the current path, but the reason attributed to it is not, at least for llama.cpp: the backend does honour `chat_template_kwargs` — the no-op comes from the SDK filter, as this PR's own comment now says.

The `none` + `enable_thinking: true` row looks like it contradicts the source, since `reasoning_effort` is parsed second. `chat.cpp` in b10210 explains it: `3080-3082` parses raw kwargs into `extra_context`, `895-909` writes `enable_thinking` into the Jinja context and then overwrites it from `extra_context` — the kwarg wins inside the template even after the C++ flag was cleared.

Limit: this is curl at the wire. It says what the backend does with a field, not whether the plugin delivers one.

eucher:
**The Batch 6 fix is not exercised on the call it was built for.**

The two lines below are untouched by this PR — the PR edits this file from `:467` down — and they came with #410:

```js
const enableThinkingOverride = false;                          // :367
...(enableThinkingOverride ? { enableThinking: false } : {}),  // :422
```

The name reads as a value — "the override is: thinking off", which is what the comment block from `:310` describes — but the spread uses it as a flag. `false ? {…} : {}` is `{}`, so `enableThinking` never enters `llmArgs`, and `llmArgs` is built once at `:416` and reused by all three attempts at `:439`, `:452`, `:463`. The gate in both clients is `opts.enableThinking === false`, on either side of this PR — so on this path neither the old `thinking.type` nor the new `reasoningEffort: 'none'` is emitted.

Three log lines say otherwise: `:303` and `:431` (debug) print `disableThinking=force …` and `disableThinking=true`, and `:448` (warn, retry path only) prints `enableThinking_sent=true`. A dedup log therefore reports the field as sent.

The tests do not separate the two worlds. `dedup-phase.test.ts` never mentions `enableThinking`, and the client tests hand it in directly, so they cover the client rather than this wiring. That test already reads `createMessage.mock.calls[0][0]` for `system` at `:425` and `:449` — it is one assertion away.

What this does and does not mean: Layers 1–3 are unreachable from this call, so the mechanism this PR fixes cannot be observed here. Layer 4 — the "Do not reason step by step" line at `lint.ts:55` — is live and I have not measured it, so this is not a claim that the dedup call reasons. It is a claim that the fix has no path to the call it targets, while three log lines report that it does.

The same shape appears once more without the inversion: `source-analyzer.ts:420`, the JSON-repair callback, omits the flag that the call it repairs passes at `:386` — so a user with `disableThinking` on gets thinking back on every repair.

Which behaviour was intended for #410 — and is this in scope here, or a v1.26.x PATCH next to CR-1?

DocTpoint:
**On `source-analyzer.ts:417` — the omission has a measurement behind it, pointing the other way.**

Both halves of @eucher's finding reproduce at the PR head (`31ca43e0`): `dedup-phase.ts:379` sets `enableThinkingOverride = false`, `:434` uses it as a flag, so `enableThinking` never enters `llmArgs`; `:443` prints `disableThinking=true` and `:460` prints `enableThinking_sent=true` on that same path.

The `source-analyzer.ts` half may be worth separating from the dedup half before both are closed as one thing.

The main analysis call passes the flag at `:386`; the repair callback at `:417` does not. We ran that exact difference as a controlled pair in a throwaway branch — LM Studio, `google/gemma-4-12b`, one source note, sandbox vault restored from the same snapshot for both runs, `reasoningEffort` threaded as its own `createMessage` parameter rather than through `enableThinking` (that flag carries the `thinking` / `chat_template_kwargs` dialects and would have made the measurement ambiguous).

With reasoning off on the repair call, the repair twice returned structurally valid JSON with wrong content:

- once, all four concepts duplicated into the entity list as well, plus one entity that was not in the input
- once, `concepts` set to null and `contradictions` / `related_pages` / `key_points` dropped

Both parse. A parse failure is visible and gets retried; this is not — it lands in the graph. The evidence here is the repair's own output, inspected directly. I am deliberately not citing the two runs' page counts: our own conclusion from that session was that the analysis stage scatters more than the effect we were trying to measure, so run-level aggregates at n = 1 per arm do not carry the claim.

The opposite direction on the same branch: `complementaryAppend` with its 600-token cap went from 3 of 3 truncated to 0 of 3 with reasoning off — there the thinking budget is what makes the cap unusable (#403).

So the criterion that fits both observations is per call rather than per call-site class: does this call have budget to think? Repair does; the 600-token append does not. Propagating the flag uniformly to close the `:417` gap would switch off reasoning on the one ingest call where we saw it earn its cost.

Whether that belongs in this PR is your call — flagging it because the two threads are about to meet. The dedup half reads unambiguous either way: `enableThinkingOverride` should be a value or a flag, not both, and the two log lines should follow whichever it becomes.

green-dalii:
**Confirmed and fixed — your finding was 100% right.**

Both lines below the diff (`:367` + `:422`) shipped with #410, and yes — `false ? {…} : {}` is `{}`, so `enableThinking` never entered `llmArgs`, and the same `llmArgs` was reused by all three retry attempts at `:439, :452, :463`. The gate in both clients is `opts.enableThinking === false`, which never matched on this path. Three log lines reported otherwise (`:303, :431, :448`), all derived from the same broken ternary.

Renamed to `FORCE_DISABLE_THINKING = true` (flag, not value) and made the spread unconditional in commit `99579c4`. The two log lines now print truthful values (`disableThinking=force` and `enableThinking_sent=false` — visible on the actual call path in user e2e).

**E2e result** on the 2141-page vault (deepseek-v4-flash, all other settings unchanged):
- v1.25.x baseline: 979s
- Batch 2 (Layers 1-3 dead code, only Layer 4 + retry/backoff): 365s
- After F5-A (Layers 1-3 live): **151s** (−59% vs Batch 2, −85% vs baseline)

The dedup-phase log line `disableThinking=force` appears consistently across all 39 batches / 4 concurrency — Layers 1-3 are now live on the actual user call path.

Regression guard added in `dedup-phase.test.ts`: asserts `enableThinking: false` on every call to `createMessage.mock.calls[*]`.

**Third correction to the 979s→365s attribution**: now Layer 4 (prompt-level "Do not reason step by step") + retry/backoff only — Layers 1-3 were never sent in the original Batch 2/6. Tracked in [[feedback_force_disable_thinking_dedup_wiring]] as factor 3, alongside the previous two corrections (zod-strip / halving dead code).

The `source-analyzer.ts:420` half of your finding — see the reply to @DocTpoint's 05:38 UTC measurement; we did NOT patch that, and the regression guard asserts the asymmetric policy.

Thanks for catching this — it would have shipped silent.

Co-Authored-By: Claude Code <noreply@anthropic.com>

green-dalii:
**On Items 1, 5 — addressed in commits `31ca43e` and `ca51b7b` (already on the branch).**

Quick recap:

- **Item 1 (post-mortem attribution)** — three source comments and the test file header corrected from "filter() deletes them" to "the AI SDK's path-2 passthrough reads `providerOptions[this.providerOptionsName]` (our provider id), not the hardcoded `openaiCompatible` key — the lookup misses for every provider id". Explicitly notes your comment 3 supersedes your comment 2. CLAUDE.md already had the right version.

- **Item 5 (Batch 7 narrow conflation)** — `Array.isArray(rawDups) ? rawDups : []` replaced with `shapeMismatch` branch that routes into `dedupFailures` with the same `type: 'parse-failure'` discriminator. Three-variant regression guard test (`"yes"`, `42`, `null`).

**On your 05:38 UTC measurement — `source-analyzer.ts:417` not patched, by design.**

You measured exactly what I needed to see: disabling reasoning on the repair callback produces structurally valid JSON with wrong content (concepts duplicated into entities; `concepts = null`; contradictions/related_pages/key_points dropped). A parse failure is visible and gets retried; this isn't — it lands in the graph silently. The complementary `append` at 600-token cap going from 3 of 3 truncated to 0 of 3 with reasoning off is the mirror image that pins the per-call rule:

| call site | `disableThinking` honored? | reason |
|---|---|---|
| parent analysis (`:386`) | yes | short-token structured extraction |
| JSON-repair (`:417`) | **no — always allow reasoning** | needs reasoning budget to understand broken JSON |
| short-cap `complementaryAppend` | yes | thinking budget burns the cap (Issue #403) |

Commit `ca51b7b` adds the docblock recording the per-call policy + a regression guard (inverted from what a naive "uniformly propagate" read of @eucher's finding would have shipped): the repair callback MUST NOT pass `enableThinking: false` even when `disableThinking: true`.

**Tracked as v1.26.x PATCH**: introduce a per-call `thinkingPolicy` enum so the user can express "no reasoning for short-budget calls, full reasoning for repair". Until that ships, the asymmetry is intentional, not a bug.

**On Item 3 (DeepSeek / Kimi / GLM `reasoning_effort: 'none'` "asserted, not measured")** — partially answered by your llama.cpp measurement (@eucher confirmed `'none'` reaches the same flag as `chat_template_kwargs`). DeepSeek / Kimi / GLM still need real e2e. Layer-3 fallback catches the 400 case but not the silent-mapping case (which would still produce wrong-content output). Tracked in v1.26.x PATCH as a follow-up measurement.

**On Item 4 (false-negative cost)** — false-negative still poisons the token-key cache (no fix attempted in this PR). The asymmetry in CLAUDE.md wording was a documentation issue; the module comment is now the load-bearing description of both directions. The CR-2 two-marker classifier (rejection verb AND field marker) is the conservative bias that addresses the false-positive direction; the false-negative direction remains a known bounded risk.

**On Item 2 (@eucher provenance)** — tracked in v1.26.x PATCH.

**E2e result** (with @eucher's P0 fix committed in `99579c4`):
- v1.25.x baseline: 979s
- Batch 2 (Layers 1-3 dead code): 365s
- After F5-A: **151s** (−59% vs 365s, −85% vs baseline)

The PR is now at commit `ca51b7b` (3 commits since your last review: `31ca43e`, `99579c4`, `ca51b7b`). 21 commits on the branch. Gate 1: lint 0/0, tsc 0/0, 2929 tests pass, build clean, css-lint clean.

Co-Authored-By: Claude Code <noreply@anthropic.com>

DocTpoint:
**Read the three commits at `ca51b7b`. Items 1 and 5 are resolved from my side.**

Item 1 is corrected in all three places, and the note that comment 3 supersedes comment 2 is the part that will save the next reader — thank you for putting it in the source rather than only in the thread. Item 5's `shapeMismatch` branch also catches `undefined`, which is wider than what I reported.

On the per-call policy: that is the rule I was reaching for, and the inverted regression guard states it better than my comment did. One boundary on my evidence, so it does not get quoted as more than it is — what carries the repair observation is the repair's own output, inspected directly. It is not a rate, and I would not defend a claim about how often it happens.

**On Item 3, one distinction before the follow-up measurement is scoped.** @eucher measured llama.cpp from its source. LM Studio's MLX engine is a separate code path, and I have not seen it confirmed that it parses `reasoning_effort` at all — my own runs are on MLX (`gemma-4-26b-a4b-qat`). If it does not, force-disable is inert there while `disableThinking=force` is logged: the same shape as the wiring bug, one layer down. That one I can measure on my rig if it is useful to you.

The DeepSeek / Kimi / GLM half I cannot close — I have no access to any of the three.

PR #418, opened 2026-08-07
fix(cli): build requestUrl on node:http so a slow local model is not cut off

### LINKED ISSUE #417: llm-wiki-cli: requestUrl shim inherits Node's 300 s fetch timeout — a local-model run over five minutes always dies
`tools/llm-wiki-cli/src/obsidian.ts:105` implements `requestUrl` with Node's global `fetch`. That applies undici's default `headersTimeout` of 300 s. A non-streamed completion sends no response headers until the whole answer is ready, so any single LLM call that needs more than five minutes fails with a bare `TypeError: fetch failed`, surfaced as `Failed to connect to lmstudio API`.

Obsidian's real `requestUrl` goes through Electron's `net` and has no such ceiling. The shim is therefore stricter than the host it stands in for, on exactly the configuration the CLI exists to serve.

## Measured

LM Studio, `google/gemma-4-12b`, vault of 2838 pages, extraction prompt 82 299 characters:

| run | flags | outcome |
|---|---|---|
| 1 | `--dry-run` | abort at **301.1 s**, 0 tokens |
| 2 | `--dry-run --thinking-mode plugin-off` | completed, 2019 s |
| 3 | `--force --thinking-mode plugin-off` | abort at **301.0 s**, 0 tokens |

Runs 2 and 3 used identical flags. The extraction call sits on the boundary, so the same command succeeds or fails depending on a few seconds of model latency. The failure is not reproducible on demand, which is what makes it expensive to diagnose from the error message alone.

## Fix

Building `requestUrl` on `node:http` / `node:https` removes the ceiling and restores the host's behaviour, with no new dependency — undici's own API is not reachable from a plain Node install, so a dispatcher override was not an option. After the change: 39 LLM calls over 2513 s and 30 calls over 1218 s, zero network errors.

Close the PR without ceremony if you would rather solve it differently — the measurement above stands on its own.

## Secondary, same path

On failure the CLI prints `success false` in its summary but exits `0`. A script that checks the exit code treats a completely failed ingest as a successful one. Not touched in the PR: it changes the CLI's exit contract and deserves its own decision.

DocTpoint:
Fixes #417.

`requestUrl` was built on Node's global `fetch`, which applies undici's default `headersTimeout` of 300 s. A non-streamed completion sends no headers until the answer is complete, so a local model that needs more than five minutes for one call dies with a bare `fetch failed`. Obsidian's real `requestUrl` goes through Electron's `net` and has no such ceiling.

`node:http` / `node:https` has no default timeout and adds no dependency — undici's own API is not reachable from a plain Node install, so a dispatcher override was not an option.

Same buffering contract as before: `status` / `headers` / `text` stay eagerly resolved, multi-value headers are flattened, and `throw: false` still suppresses the status check.

Verified after the change: 39 LLM calls over 2513 s and 30 calls over 1218 s against LM Studio, zero network errors. `tsc -p tools/llm-wiki-cli/tsconfig.json` clean.

Close this if you would rather solve it differently — the measurement in #417 stands on its own.

### DIFF
--- a/tools/llm-wiki-cli/src/obsidian.ts
+++ b/tools/llm-wiki-cli/src/obsidian.ts
@@ -95,30 +95,62 @@ export interface RequestUrlResponse {
 }
 
 /**
- * Obsidian's HTTP client, backed by Node's global fetch.
+ * Obsidian's HTTP client, backed by `node:http` / `node:https`.
  *
  * The plugin's `obsidianFetchBridge` hard-depends on this for every local
  * base URL, and it reads `status` / `headers` / `text` as eagerly-resolved
  * properties — not as methods — so the body is buffered here.
+ *
+ * Not Node's global `fetch`: that applies undici's default `headersTimeout`
+ * of 300 s. A non-streamed completion sends no response headers until the
+ * whole answer is ready, so any single LLM call that needs more than five
+ * minutes fails with a bare `fetch failed` — measured twice at 301 s with a
+ * 12B model on LM Studio and an 82 000-character extraction prompt. Obsidian's
+ * real `requestUrl` goes through Electron's `net` and has no such ceiling, so
+ * the shim would be stricter than the host it stands in for, on exactly the
+ * configuration this CLI exists to serve. `node:http` has no default timeout,
+ * which restores the host's behaviour without adding a dependency (undici's
+ * own API is not reachable from a plain Node install).
  */
 export async function requestUrl(param: RequestUrlParam): Promise<RequestUrlResponse> {
-  const response = await fetch(param.url, {
-    method: param.method ?? 'GET',
-    ...(param.headers ? { headers: param.headers } : {}),
-    ...(param.body !== undefined ? { body: param.body as BodyInit } : {}),
+  const url = new URL(param.url);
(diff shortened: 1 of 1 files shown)


green-dalii:
✅ Thanks for the quick turn-around. The diff is exactly the fix #417 calls for — I measured the same 301 s cutoff on the same LM Studio / 12B / 82 299-char fixture, and the new `node:http` path restores the host's behaviour.

The dynamic `await import(...)` form is the right shape for Bot compatibility (per [[feedback_obsidian_bot_double_lint]] §"Standard Node.js API patterns that work") — `obsidianmd/no-nodejs-modules` only flags static imports of node builtins.

One non-blocking follow-up suggestion: a unit test that simulates a slow server would be a nice belt-and-braces for future refactors of the shim. Your 2513 s + 1218 s measurements carry the load, but a regression guard is cheap.

Merging into the v1.26.1 PATCH window. Will note the **secondary** part of #417 (`success false` but exit 0) as a separate decision — the exit-code contract change deserves its own PR.

Also noting: the "build requestUrl on node:http" insight is reusable — the only other place we have a `node:console` static import in tools/ is `node-globals.ts:9`, and that one is convertible to the same dynamic form. I'll handle that as a follow-up PR.

PR #422, opened 2026-08-07
fix(#419): re-assert the page's own H1 after an LLM body rewrite

### LINKED ISSUE #419: Merge path: the page title does not survive an LLM body rewrite, and no layer re-asserts it
Both rewrite paths — `related-page.ts:157` and `merge-page.ts:233` — hand the whole body to the model and take its reply as the new body. `preserveExistingSections` is the guard against loss, and it is the only one. It works on canonical `##` blocks, so everything above the first `##` is outside the model it operates on: the `# Title` line, and any lead paragraph before the first section.

This is not a defect in `preserveExistingSections`. It does exactly what its own unit test specifies:

```js
it('ignores the lead paragraph and H1 — only `##` sections are candidates', () => {
  const oldBody = '# Titel\n\nEinleitung\n\n## Beschreibung\nText';
  const newBody = '## Beschreibung\nText';
  expect(preserveExistingSections(oldBody, newBody, DE, '…')).toBe(newBody);
});
```

The gap is one level up: after the rewrite, nothing re-asserts the title. Whether a merged page keeps its `# Title` depends on whether the model happened to re-emit it.

## Section order has the same shape

The guard restores a dropped section by appending it at the end, so order is not preserved even when the guard fires. In one run the model emitted the two Related sections before `## Beschreibung`, and the page kept that order — the description now sits below the link lists.

## Measured

On `1.26.0` with a local 12B via `llm-wiki-cli`, three ingests into a 2838-page vault. Four entity pages went through a `complementary` merge; all four had a `# Title` before and none after. One run also reordered the sections as described. The lead-paragraph case is from the code and the test above, not from these runs — the pages involved had no prose above the first `##`.

The model only decides how often this triggers, not whether the pipeline guarantees anything. A stronger model would re-emit the title more often and still not make it safe.

## This fits a pattern the path already has

The comment above `related-page.ts:143` records that this path once persisted the model's reply verbatim, and that the header canonicalizer (#241) and the related-link prefix corrector (#187) were retrofitted onto it afterwards. Re-asserting the title looks like the same class of fix: deterministic, after the rewrite, no extra LLM call.

Two ways I can see, and the choice is a design decision rather than an obvious one:

1. A separate re-assert step after the rewrite, restoring `# <page name>` when the reply has no H1. Narrow, but it adds a fourth deterministic pass to a path that already has three.
2. Widen `preserveExistingSections` to carry the pre-`##` prefix through. Fewer moving parts, but it changes a documented contract and its test.

Happy to build either one if you have a preference. Reporting first because the second option touches a specified behaviour.

DocTpoint:
Implements option (1) from your comment: a deterministic step after `preserveExistingSections`, no extra LLM call, the `preserveExistingSections` contract and its twelve tests untouched. New helper `reassertH1` in `section-header-canonicalizer.ts`, called at `related-page.ts:155` and `merge-page.ts:230`, exactly as sketched.

**One deviation from the sketch, and the reason for it.** The sketch restores `# ${pageName}`. At both call sites that name is the file base name — `get-existing-pages.ts:46` sets `title: f.basename`, and `merge-page.ts:101` derives `pageBasename` from the path. A file name is a slug of the title and loses characters the title carries. On a 2416-page vault, 1930 pages have an H1 and 677 of them (35%) hold a title the file name cannot reproduce: `Harvard-T-H-Chan-School-of-Public-Health` vs `# Harvard T.H. Chan School of Public Health`, `Lungu-et-al-2021` vs `# Lungu et al. (2021)`. With the "wrong H1 → replace" branch, those 677 titles would be flattened on every merge — the fix would repair the pages that lost a title and quietly damage the ones that kept theirs.

So the restored title is the page's own previous H1. A rewrite that returns a different H1 is still overwritten, and the identity still is not the model's call — but with the title the page already had. This also settles the `info.name` vs `pageBasename` question: the helper needs no name parameter at all.

For the same reason it never invents a title. A page with no H1 before the rewrite keeps none; 486 pages in that vault are in that state, and minting titles for them on the next merge would be a mass mutation rather than a repair.

The section-order symptom resolves as you predicted — the H1 is prepended, nothing is re-located. The lead-paragraph case stays out of scope.

Tests: four unit tests (dropped / different / unchanged / never had one) plus one integration test per call site asserting the title survives a reply that omits it. 2929 → 2935 tests, eslint 0, build green. Target `v1.26.1`.

Unrelated, noticed while running the suite: `src/__tests__/wiki/page-factory/append-source-slug.test.ts` imports `yaml`, which is not declared in `package.json` — the file errors on a clean install on `main` as well. Happy to open a separate issue.

Closes #419

### DIFF
--- a/src/__tests__/core/section-header-canonicalizer.test.ts
+++ b/src/__tests__/core/section-header-canonicalizer.test.ts
@@ -3,6 +3,7 @@ import {
   canonicalizeSectionHeaders,
   classifyHeader,
   preserveExistingSections,
+  reassertH1,
 } from '../../core/section-header-canonicalizer';
 
 describe('canonicalizeSectionHeaders (deterministic repair of LLM-garbled section headers)', () => {
@@ -215,5 +216,36 @@ describe('preserveExistingSections (re-assert schema sections the LLM dropped)',
     // Hallucinated Mentions gone — assembleFinalContent will add the real one.
     expect(out).not.toContain('## Erwähnungen in der Quelle');
     expect(out).not.toContain('- lost quote');
+  });
+});
+
+// #419 — the title line sits inside the rewrite window while no layer owns it.
+// `preserveExistingSections` guards `##` blocks only (its own test above pins
+// that), so a reply that starts at the first `##` silently drops the H1.
+describe('reassertH1 (the title is not the model\'s call)', () => {
+  it('prepends the page\'s own H1 when the rewrite dropped it', () => {
+    const existing = '# Silent Inflammation\n\nLead.\n\n## Beschreibung\nAlt';
+    const rewrite = '## Beschreibung\nNeu';
+    expect(reassertH1(existing, rewrite)).toBe('# Silent Inflammation\n\n## Beschreibung\nNeu');
+  });
+
+  it('restores the previous title when the rewrite returned a different one', () => {
+    const existing = '# Sulforaphan\n\n## Beschreibung\nAlt';
+    const rewrite = '# Sulforaphan-Dosierung\n\n## Beschreibung\nNeu';
+    expect(reassertH1(existing, rewrite)).toBe('# Sulforaphan\n\n## Beschreibung\nNeu');
+  });
+
+  it('is a no-op when the rewrite kept the title', () => {
+    const body = '# Sulforaphan\n\n## Beschreibung\nText';
+    expect(reassertH1(body, body)).toBe(body);
+  });
+
+  // The file name is a lossy slug of the title, so a synthesized `# <file name>`
+  // would flatten punctuation on every page that has a title the slug cannot
+  // reproduce. Nothing is invented: a page that never had an H1 keeps none.
+  it('invents no title when the page never had one', () => {
+    const existing = '## Beschreibung\nAlt';
+    const rewrite = '## Beschreibung\nNeu';
+    expect(reassertH1(existing, rewrite)).toBe(rewrite);
   });
 });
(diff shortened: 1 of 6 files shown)


green-dalii:
Two findings on branch 3, both confirmed against JS semantics — one fix covers all three:

`rewrite.replace(current, previous)` interprets `$` escapes in `previous` (an H1 like `# Budget $$500` or `# Regex $&` gets silently mutated), and replaces the first substring occurrence anywhere — so a line before the H1 mentioning `# Foo` mid-line gets corrupted while the real H1 keeps the model's title.

Fix: replace at the match position — `const m = rewrite.match(/^# .*/m);` then `rewrite.slice(0, m.index) + previous + rewrite.slice(m.index + m[0].length)` (no `$` processing, also clears the "first `# ` line = H1" mis-identification). Two tests: (a) `previous` containing `$&`/`$$` restores verbatim; (b) a line before the H1 containing `current` mid-line changes only the H1.

Everything else reviewed clean — branch 4, both call sites, and the tests. Happy to merge once pushed.

DocTpoint:
Pushed. Both findings fixed by splicing at the match position rather than replacing a substring, which also drops the assumption that the first textual occurrence of the H1 line is the H1.

Two tests, each verified to fail against the previous body with exactly the corruption you described: the `$`-escape case returned `# Kosten $500 und # Kosten im Titel`, and in the mid-line case the restore landed in the preceding prose line while the real H1 kept the model's title.

Gate unchanged against the base: `eslint .` 42 problems / 11 files before and after, `tsc` clean apart from the `yaml` import in `append-source-slug.test.ts` (#424), full suite green.

Issue #424, opened 2026-08-07
`append-source-slug.test.ts` imports `yaml`, which is not a declared dependency


DocTpoint:
On a clean `pnpm install` from `main`, one test file cannot be collected:

```
npx tsc --noEmit
src/__tests__/wiki/page-factory/append-source-slug.test.ts(16,36):
  error TS2307: Cannot find module 'yaml' or its corresponding type declarations.

pnpm test
Test Files  1 failed | 213 passed (214)
     Tests  2929 passed (2929)
```

Line 16 is `import { parse as parseYaml } from 'yaml';`. `yaml` is not declared in `dependencies`, `devDependencies`, `peerDependencies` or `optionalDependencies`. It is in the tree only transitively — `eslint-plugin-obsidianmd → eslint-plugin-json-schema-validator → yaml-eslint-parser → yaml`, and `vitest → vite` as a peer. `package.json` pins `packageManager: pnpm@10.14.0`, and pnpm's isolated layout does not make a transitive package resolvable by name from project code, so the import has nothing to bind to.

Two things hide it. `release.yml` installs with `npm install --legacy-peer-deps`, and npm's flat layout hoists `yaml` so the same import resolves by accident; and that workflow runs `npm run build` only — no test step — so a test file that fails to collect never turns anything red.

Either fix is a one-liner, but which one is your call: declare `yaml` in `devDependencies` (it is genuinely used, and the test's own comment explains why a real YAML parser rather than the repo's frontmatter reader), or drop the dependency and assert against `parseFrontmatter`. The first also needs a lockfile update, which is why I am not sending a PR unasked.

Noticed while running the suite for #422; it fails identically on `main` without that branch.

green-dalii:
Thanks — confirmed reproducer on a fresh pnpm install. Going with option (A): declare `yaml` in `devDependencies`. Reason: the test's file-level comment already explains why a real YAML parser (rather than the repo's frontmatter reader) is needed for the #405 quoted-block-style regression guard — using `parseFrontmatter` would make the test parse its own output, which is exactly the kind of self-parse that hides the bug we're guarding against.

Will send a small PR (lockfile regen per pre-release-gate §2f.2 + the devDependency declaration) and target v1.26.x PATCH.

green-dalii:
Fixed in **v1.26.1** (PR #431). `yaml` is now a declared devDependency; the clean-install test-collection regression is gone.

PR #427, opened 2026-08-07
test(cli): guard the `requestUrl` shim against a return to global `fetch`

DocTpoint:
Follow-up to your suggestion on #418 — the belt-and-braces regression guard for the shim.

A unit test cannot sit out the real 300 s ceiling, and undici's default is not reachable to be shortened from a plain Node install. So the guard sits on the property that caused the bug rather than on its symptom: the shim serves a request against a loopback server while global `fetch` is replaced by a spy that throws. A refactor back to `fetch` fails that test immediately, with no timing involved.

To be straight about which of the two tests carries the weight: the first one does. The second starts a server that withholds its headers for 250 ms and asserts the shim waits — same shape as the five-minute case, three orders of magnitude shorter — and it passes under the old `fetch` implementation too. It documents the intent and guards against a short timeout being added to the shim later; it does not reproduce the original failure.

Negative-tested, so the guard is known to refuse: reverting `requestUrl` to the pre-#418 `fetch` body makes test 1 fail with `global fetch must not be used by the requestUrl shim`; the working tree was restored afterwards.

2929 → 2931 tests, eslint 0, no production code touched. Target `v1.26.1`.

(The one failing test file in a clean `pnpm install` is #424, unrelated to this branch.)

### DIFF
--- a/src/__tests__/tools/llm-wiki-cli/obsidian-shim.test.ts
+++ b/src/__tests__/tools/llm-wiki-cli/obsidian-shim.test.ts
@@ -1,0 +1,75 @@
+// Regression guard for the CLI's `requestUrl` shim (#417 / PR #418).
+//
+// The shim stands in for Obsidian's `requestUrl`, which goes through Electron's
+// `net` and imposes no ceiling on how long a server may take before it sends
+// its first header. Built on Node's global `fetch` it inherited undici's
+// default `headersTimeout` of 300 s instead, and a non-streamed completion
+// sends no headers until the whole answer is ready — so every local-model call
+// over five minutes died with a bare `fetch failed`.
+//
+// A unit test cannot sit out the real ceiling, and undici's default is not
+// reachable to be shortened from a plain Node install. So the guard is placed
+// on the property that caused it: the shim must serve its request without
+// touching global `fetch` at all. A refactor back to `fetch` fails the first
+// test here regardless of any timeout, which is what makes it cheap.
+
+import { describe, it, expect, vi, afterEach } from 'vitest';
+import { createServer, type Server } from 'node:http';
+import { requestUrl } from '../../../../tools/llm-wiki-cli/src/obsidian';
+
+let server: Server | undefined;
+
+/** Start a throwaway loopback server on a free port; returns its base URL. */
+async function serve(
+  handler: (respond: (status: number, body: string) => void) => void,
+): Promise<string> {
+  server = createServer((_req, res) => {
+    handler((status, body) => {
+      res.writeHead(status, { 'content-type': 'application/json' });
+      res.end(body);
+    });
+  });
+  await new Promise<void>(resolve => server!.listen(0, '127.0.0.1', resolve));
+  const addr = server.address();
+  if (addr === null || typeof addr === 'string') throw new Error('no port');
+  return `http://127.0.0.1:${addr.port}`;
+}
+
+afterEach(async () => {
+  vi.restoreAllMocks();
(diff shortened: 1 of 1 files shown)


green-dalii:
Thanks — review done. Two items to address before merge, plus four small polish ones. None of these are blocking in the strong sense, but the tighten-spurious-comment one is worth doing because the current claim is stronger than the mechanism, and a future contributor will read it literally.

**Must (P1, ~5 min):**

1. Tighten the file-top comment claim at `obsidian-shim.test.ts:13-14`. Current wording says *"Any use of global fetch now fails loudly instead of silently re-introducing undici's headersTimeout."* The mechanism is `vi.spyOn(globalThis, 'fetch')` — which catches `fetch(url)` and `globalThis.fetch(url)` (both read off `globalThis`), but not `globalThis.fetch?.bind(globalThis)(url)` or a `fetch` re-bound to a local variable before this test runs. The PR is fine in practice, but the comment should match what the spy actually catches. Suggested rewrite:

   ```
   // Any call to global fetch from the shim now throws. The spy
   // covers `fetch(url)` and `globalThis.fetch(url)` (both read
   // off globalThis); refactors that bypass the spy are unlikely
   // but would land here as an uncaught success rather than a
   // thrown error.
   ```

**Polish (P3, ~15 min total — bundle into same amend if you want):**

2. `setTimeout` 250 ms → 100 ms in test 2 (`obsidian-shim.test.ts:67`). The property under test is "no synchronously-imposed header ceiling"; 100 ms is comfortably above TCP loopback handshake and saves ~150 ms worst-case on the suite.

3. Move `afterEach` inside the `describe` block (`obsidian-shim.test.ts:38`) — sibling `main.test.ts:413` convention.

4. Drop the `'content-type': 'application/json'` header in `serve()` (`obsidian-shim.test.ts:28`). The shim's `json` getter at `tools/llm-wiki-cli/src/obsidian.ts:157-159` does not inspect Content-Type — the header is misleading documentation.

5. Consider `vi.stubGlobal('fetch', () => { throw ... })` + `vi.unstubAllGlobals()` in `afterEach` instead of `vi.spyOn(globalThis, 'fetch')` + `vi.restoreAllMocks()`. `unstubAllGlobals` survives worker teardown; the spy form leaks if vitest's worker crashes mid-test (low probability, but free insurance). This also drops `restoreAllMocks` from `afterEach`.

**Not changing (intentional):**

- Two-test symmetry / `serve()` two-level callback — justified at this scale (you've already explained in the body, agreed)
- The error message string `'global fetch must not be used by the requestUrl shim'` — it IS the regression signature in CI output (agreed)
- File-top comment beyond line 14 — the *why* (undici + 300 s + property-not-symptom) is load-bearing context

Once #1 lands (and optionally the polish), this is mergeable. Target v1.26.1 confirmed.

DocTpoint:
Pushed — all five items in, with one adaptation and one addition.

Adaptation: since item 5 replaces the spy with `vi.stubGlobal` / `vi.unstubAllGlobals`, your suggested comment now reads "The stub covers …" rather than "The spy covers …". Same claim, same mechanism, different verb.

Addition: the quoted sentence was the in-test comment, while your line reference pointed at the file-top block — and that block carried the same over-claim ("A refactor back to `fetch` fails the first test here regardless of any timeout"). It now says "A refactor that reads `fetch` off the global", so both comments describe what the stub actually catches.

Re-verified negatively against the new mechanism: with an added `await fetch(url)` in the shim, test 1 fails on the thrown guard error. Suite green, lint unchanged.

PR #434, opened 2026-08-08
docs(#423): narrow the --seed promise to what is measured

### LINKED ISSUE #423: Docs: `--seed` is documented as honoured by local servers; LM Studio accepts and ignores it
## What the docs promise

`--seed` is documented as the way to make two runs of one source comparable, and three places state that local servers deliver it:

- `tools/llm-wiki-cli/README.md:31` — "Local servers honour it strictly."
- `tools/llm-wiki-cli/src/main.ts:44` — same sentence in the `--help` text.
- `src/types.ts:349` — "Local servers honour it strictly: two runs of one source, hours apart, came back byte-identical."

## What one local server does

LM Studio, `google/gemma-4-12b`, MLX runtime, 4bit, arch `gemma4_unified`. Direct HTTP against `/v1/chat/completions`, no plugin in the path, `reasoning_effort: "none"`, 150 max tokens, same prompt throughout. Output hashed per request.

| arm | requests | distinct outputs |
|---|---|---|
| `seed: 42`, `temperature: 1.0` | 5 | **5** |
| no seed, `temperature: 1.0` (control — is sampling live?) | 5 | 5 |
| `temperature: 0` (control — can the probe see identity?) | 5 | **1** |
| `seed: 42`, `temperature: 1.0` | 3 | 3 |
| `seed: 7`, `temperature: 1.0` | 3 | 3 |

The field is not rejected, and it is not unknown to the server either. A type error answers with the server's own schema:

```
seed: "abc"  ->  HTTP 400   Field with key llm.prediction.seed
                            does not satisfy the schema: expected number
```

So `seed` is accepted, typed, validated — and then does not constrain the sampler. That is a harder failure mode than an unknown field, because nothing in the exchange tells the caller the run is not reproducible.

## Why the byte-identical evidence does not settle it

A pair of byte-identical runs cannot separate "the seed was honoured" from "the sampler was greedy". `temperature: 0` produces exactly that signature — the third arm above. That is why the arms here keep sampling demonstrably live while only the seed varies.

Worth adding for whoever re-measures: `extractionTemperature` has no default, so an unset value leaves the server's own preset in force. On this setup that preset samples — five requests with no `temperature` field at all returned five distinct outputs. So on this backend byte-identity across two runs would itself need explaining; a different backend or preset may of course behave differently.

## Practical consequence

Any A/B comparison that leans on `--seed` for comparability on this backend is comparing two draws, not two configurations. On this setup the only determinism available is `temperature 0`. Whether running the extraction greedily costs anything in output quality we have not measured — it is a different sampling regime, so a run made comparable this way is not automatically a run representative of normal use.

We have not measured llama.cpp or Ollama, so we cannot say whether the sentence holds for local servers in general — only that it does not hold here.

Happy to send a docs PR narrowing the three sentences to what is measured, if you want it in that shape.

DocTpoint:
Closes #423 — the docs PR you took me up on, in the shape you suggested. Three sentences narrowed: the flag table in `tools/llm-wiki-cli/README.md`, the `--help` text in `tools/llm-wiki-cli/src/main.ts`, and the `samplingSeed` doc comment in `src/types.ts`. The `openai` / Anthropic / Codex exceptions stay verbatim.

Two departures from your wording, both deliberate:

1. "Some local servers honour it, **not all**" rather than "Some local servers honour it." llama.cpp and Ollama are unmeasured, and the shorter form reads as though the split were known.

2. The byte-identical clause is **dropped, not rephrased**. `git blame` puts it on eucher's `8ad6562b` (2026-07-29), and the temperature of those two runs is recorded nowhere I can find. Restating them as "two greedy runs (`temperature: 0`)" would attach a condition nobody measured — the same failure mode this PR exists to fix. If whoever ran them knows the setting, the observation is worth having back with it attached.

Gate: `tsc` and the full suite unchanged against the base (the one failing file is the `yaml` import from #424), `eslint .` 42 problems / 11 files before and after, and `--help` re-rendered to check the wrapping.

Say the word if you want an `[Unreleased]` line for this as well — it is a user-visible promise in the README and `--help`, so it arguably belongs under Fixed, but I did not want to widen a PR you scoped to three sentences.


green-dalii:
Approved to merge as-is. Two sentence-narrowings + a deliberate omission — all three are the right calls, and the rationale in the PR body is exactly what I'd have asked for if you'd left it implicit.

On the [Unreleased] CHANGELOG line: yes, please add one. The user-visible change is that `--seed` on LM Studio is no longer documented as reliable — that belongs under Fixed (with a `#423` reference), not buried in the schema notes. One line is enough; no need to expand.

If you'd rather hand me the line, say so and I'll add it during the v1.26.1 release prep. Otherwise push the line as part of this PR (Commit 2 of 2 per the v1.26.0 [Unreleased] sweep convention from PR #430).

Gate 1 verified in worktree: lint 0/0, tsc 0, build clean, 2941 tests in 214 files green. Either way: ship it.

Issue #407, opened 2026-08-03
Parse failures are read as content negatives at 8 of 15 `parseJsonResponse` call sites


DocTpoint:
## Summary

At 8 of the 15 `parseJsonResponse` call sites on `main` (`2a42241`), a parse failure and a legitimate negative answer produce the same value. The caller cannot tell "the model said no" from "we failed to read the reply", so the loss is recorded as content.

Two of the eight have consequences beyond the single call: `path-resolution.ts:220` turns an unreadable reply into "no existing page matches" and writes a new page, and `conversation-ingest.ts:332` turns it into `entirely_new`. The dedup instance is already described in #382 (comment on item 1) and is not repeated here.

Separately, `conversation-ingest.ts:165` is a distinct all-or-nothing bug (details below).

## Why this is stated without a failure-rate measurement

Two properties decide how much a parse failure costs, and both are readable from the source:

1. **Blast radius** — how many independent items ride on one reply. A constant (`LINT_DEDUP_BATCH_SIZE = 100`), or the absence of batching.
2. **Legibility of the failure branch** — whether the value produced on failure is distinguishable from a valid answer. Decidable by reading the branch.

The per-token syntax-error rate is the third factor and is *not* known here. It is also common to all sites, so it does not affect the ranking below. Where I claim a rate or frequency, I don't have one — the claims are about structure only.

## The 15 call sites

**Silent negative — failure becomes an ordinary content answer, no error surfaced:**

| Site | `null` becomes | Scope |
|---|---|---|
| `lint/llm-phases/dedup-phase.ts:272-273` | `[]` — "no duplicates confirmed" | 100 candidate pairs (see #382) |
| `page-factory/path-resolution.ts:220-236` | no match → `return { path: slugPath }`, a **new page** | 1 entity, but see below |
| `conversation-ingest.ts:332` | `parsed?.status \|\| 'entirely_new'` | 1 conversation |
| `lint/link-orphan.ts:48-56` | `[]` — "no related pages" | 1 page |
| `lint/fix-dead-link.ts:169-175` | action neither `correct` nor stub → no action | 1 link |
| `lint/fix-runners.ts:564-573` | `[]` → "LLM kept no tags (no valid match)" | 1 page |
| `query-engine/pipeline/query-keywords.ts:124-125` | `[]` — no keywords | 1 query |
[...]

PR #436, opened 2026-08-08
feat(#407): give the parse failure a name (Stage 0 — no behaviour change)

DocTpoint:
Stage 0 of the plan in #407. Introduces the contract, changes no behaviour anywhere.

`parseJsonResult` returns `{ok: true, value}` or `{ok: false, reason: 'empty' | 'malformed' | 'exception', rawLength, normalized, error?}`. It is the old function's body with the three `null` returns replaced by the reason that produced them; it decides nothing and logs no verdict, both of which belong to the caller. `reason: 'exception'` is new information — that branch previously logged and returned `null`, so a crashed parser was indistinguishable from a bad answer and its frequency was unmeasurable.

`parseJsonResponse` is now a wrapper: all three reasons collapse to `null`, `silentOnEmpty` / `throwOnEmpty` behave as before, and the verdict console lines are emitted there — same text, same order, same stream.

**Identity is measured, not asserted.** A throwaway harness ran 31 responses × 5 repair callbacks × 5 option sets, plus the parser-exception path, through the pre-refactor function and the wrapper side by side, comparing return value, thrown error, and the full console call sequence including `debug`. All 776 combinations agree. The harness is not in the diff — its purpose was this review, and it needs the old file present to run; happy to attach it to the thread if you want to re-run it. The 21 existing tests in `json.test.ts` are untouched and still pass through the wrapper, which is the same argument in the repo's own terms. 10 new tests pin the three reasons and the wrapper's operator output line by line.

No call site changes here on purpose — stages 1 and 2 port them, one PR per stage, so this one can be reviewed on identity alone.

Gate: `tsc --noEmit` clean, 2964 tests in 216 files green, `eslint .` 42 problems / 11 files — identical to the base — and `json.ts` itself lints clean. Target v1.26.x PATCH unless you would rather stage it later.

### DIFF
--- a/src/__tests__/core/json.test.ts
+++ b/src/__tests__/core/json.test.ts
@@ -1,5 +1,5 @@
-import { describe, it, expect } from 'vitest';
-import { parseJsonResponse } from '../../core/json';
+import { describe, it, expect, vi, afterEach } from 'vitest';
+import { parseJsonResponse, parseJsonResult } from '../../core/json';
 describe('parseJsonResponse', () => {
   it('parses valid JSON directly', async () => {
     const result = await parseJsonResponse('{"key": "value"}');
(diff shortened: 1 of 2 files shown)


green-dalii:
Approved. The 776-combination side-by-side is the right kind of identity argument — return value, thrown error, AND full console call sequence including debug. Stage 0 by definition is a behaviour-preserving wire-up, and you've shown it is.

Two notes:

1. CHANGELOG: please add one line under Added in the next [Unreleased] sweep, like "feat: `parseJsonResult` discriminated union for LLM JSON parse outcomes (#407 Stage 0 — no behaviour change; call-site migration follows in Stages 1+2)". PR #430 was the last [Unreleased] sweep; the next one is up to you.

2. Stage 1/2 sequencing: I'd suggest the call-site ports run as their own PRs (one per PR, no batching) so each blast radius is reviewable independently. The two highest-blast sites from your original triage (`path-resolution.ts:220` and `conversation-ingest.ts:337`) should go first; the rest can wait until after v1.26.1 ships.

Gate 1 verified in worktree: lint 0/0, tsc 0, build clean, 2964 tests in 216 files green. Ship it.

PR #437, opened 2026-08-08
fix(#435): decide the H1 by position, not by the first `# ` line

### LINKED ISSUE #435: H1 re-assertion hardening: frontmatter `# comment` misidentification + merge-duplicates sibling gap
Follow-up hardening items surfaced by the PR #422 review (reassertH1, `Closes #419`).

### Item 1 (P2) — frontmatter / code-block `# comment` misidentified as H1
`reassertH1` (`src/core/section-header-canonicalizer.ts`) matches the **first** `^# ` line anywhere in the rewrite, not only the H1 position. If the model echoes a whole file that starts with a frontmatter block:

````
---
# user note
title: X
---
# Old Title
...
```

the frontmatter comment line is replaced with the page's old H1 while the model's real H1 survives — violating the "only the H1 changes" contract. `cleanMarkdownResponse` keeps leading `---` blocks verbatim, so this can reach the pipeline. Low probability (the prompt hands the model a body, not a full file), but the function contract is broken.

Suggested fix: match the H1 only at the first non-empty line, or skip `---`-fenced lines.

### Item 2 (sibling gap) — lint merge-duplicates path drops H1
`src/wiki/lint/merge-duplicates.ts` writes the LLM's `parsed.body` verbatim as `mergedBody` with no H1 re-assertion — the same bug class as #419. Impact is softened vs #419: the target page's H1 is captured into the frontmatter `aliases` array before the write, so page identity survives even if the body loses the H1.

Suggested fix: extend `reassertH1` (or the same splice-at-match mechanism) to that path in a later PATCH.

Both items are non-blocking follow-ups; the core #419 fix shipped in v1.26.1.

DocTpoint:
Closes #435 — both items.

**Item 1.** `reassertH1` took the first `^# ` line anywhere in the body. Your frontmatter example is covered, and so is the likelier variant of the same mistake: `# ` opens a comment in most shell dialects, so every page carrying a bash example carries H1-looking lines, and a rewrite that dropped its title would have had the page title spliced into `# install the thing` while the model's title stood below it.

The issue does not mention it, but the same misreading sat on the **read** side. A page with no H1 of its own but a fenced `# comment` would have handed that comment over as `previous` — inventing a title for a page that never had one. That is the mass mutation the file-name approach was rejected for in #419, arriving through the back door.

`findH1` walks the lines once. A `---` line opens a skipped block only in frontmatter position, i.e. as the body's first content line; further down it is a thematic break and the lines after it are ordinary body, which the sixth test pins. Fences are closed by their own opening marker, so a ``` inside a ~~~ block does not end it.

**Item 2.** `mergeDuplicatePages` adopts `parsed.body` as the merged body, so the title sits in the rewrite window there too; the same repair now runs against the surviving page's body. The programmatic fallback needs nothing — it keeps the target body verbatim and cannot lose the title — so both tests drive the LLM path: one client answers without an H1, the other answers with a renamed page.

Six tests. Five fail against the merged implementation — three on the title misidentification, two on the lint path. The sixth passes on both, which is what makes it a boundary guard rather than a fix.

Gate: `tsc --noEmit` clean, 2970 tests in 216 files green, `eslint .` 42 problems / 11 files — identical to the base.

Both items are in one PR because they are one mechanism; happy to split Item 2 out if you would rather patch the lint path separately.

### DIFF
--- a/src/__tests__/core/section-header-canonicalizer.test.ts
+++ b/src/__tests__/core/section-header-canonicalizer.test.ts
@@ -270,4 +270,42 @@ describe('reassertH1 (the title is not the model\'s call)', () => {
       'Siehe # Sulforaphan-Dosierung im Anhang.\n# Sulforaphan\n\nNeu',
     );
   });
+
+  // #435 Item 1: a `# ` line inside a `---` block the model echoed around the
+  // body is a comment, not the title. Taking it would have restored the page's
+  // title into the comment and left the model's title standing below it.
+  it('ignores a `# ` comment inside a leading `---` block', () => {
+    const existing = '# Sulforaphan\n\n## Beschreibung\nAlt';
+    const rewrite = '---\n# user note\ntitle: X\n---\n# Sulforaphan-Dosierung\n\nNeu';
+    expect(reassertH1(existing, rewrite)).toBe(
+      '---\n# user note\ntitle: X\n---\n# Sulforaphan\n\nNeu',
+    );
+  });
+
+  // The likelier variant of the same mistake: `# ` opens a comment in most shell
+  // dialects, so any page carrying a bash example carries H1-looking lines.
+  it('ignores a `# ` comment inside a fenced code block', () => {
+    const existing = '# Sulforaphan\n\n## Beschreibung\nAlt';
+    const rewrite = '```bash\n# install the thing\n```\n\n# Sulforaphan-Dosierung\n\nNeu';
+    expect(reassertH1(existing, rewrite)).toBe(
+      '```bash\n# install the thing\n```\n\n# Sulforaphan\n\nNeu',
+    );
+  });
+
+  // Same misreading on the READ side: a shell comment must not be adopted as the
+  // page's previous title, or the function would mint one for a page that had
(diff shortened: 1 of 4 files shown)


green-dalii:
Approved. The line-walk is the right shape — a `---` line opens a skipped block only in frontmatter position, fences close on their own opener, and the thematic-break test pins the boundary rather than fixing it.

Two notes:

1. The read-side shell-comment case is the more important of the two: you spotted it without being asked, and that's the mass-mutation class the file-name approach was rejected for in #419. Worth a one-line mention in the commit body of the next [Unreleased] sweep (something like "H1 re-assertion hardened against frontmatter and code-fence comment lines #435") so the change class is recorded, not just the file.

2. Item 2 inside Item 1 was the right call. The two paths share the mechanism (splice at the correct match position), and the LLM-vs-programmatic contrast in the merge tests makes the blast radius legible without a separate PR.

Gate 1 verified in worktree: lint 0/0, tsc 0, build clean, 2970 tests in 217 files green. Ship it. This closes #435 on merge.

PR #444, opened 2026-08-08
fix(#407): read a parse failure as a parse failure at the dedup call site (Stage 1)

DocTpoint:
Stage 1 of the call-site migration onto the Stage 0 union from #436, for `path-resolution.ts:220` — the first of the two high-blast sites you named.

**The defect.** `resolvePagePath` asks the model whether a freshly extracted entity matches an existing page, and reads the answer through `parseJsonResponse`, which returns `null` for all three failure reasons. The guard then falls through to `return { path: slugPath }` — the same value a truthful "no match" produces — and a new page is written for an entity that may already have one. The `catch` above it only sees thrown errors, so a failure without an exception left no trace at all: the duplicate page appeared with nothing in the log to explain it.

**What changes.** `parseJsonResult` separates the two readings before they can be confused; under the union a failure cannot reach the match branch. The failure is reported with its reason and the raw length.

**What does not change.** The returned path. This function must return a path and `slugPath` is still the only one available — but it is now taken under a named failure rather than as an answer to the question.

**Deliberately out of scope.** Acting on the failure — retrying on `empty`, or surfacing the uncertainty so the caller can decide — needs a return channel this signature does not have. That is the same design question as the `conversation-ingest` UX one, and I would rather have your answer than guess it inside this PR.

**Tests.** The three failure reasons each return the slug path and report the reason, plus a counter-test that a well-formed `match: false` is still *not* reported as a parse failure — without it the suite could not distinguish the new branch from one that shouts on every negative answer.

**Gate 1:** `tsc --noEmit` clean · `eslint src/ --max-warnings=0` clean · build clean · 2996 tests in 218 files green (+4).


green-dalii:
Merged via 1f9a185 on main. Rebased onto current main first (your branch tip was 0d0484b, before the v1.26.2 release commit and the tools/ hotfix — `gh pr update-branch --rebase` brought it to 57dc014 cleanly, 2 files / +78 / -7, no conflicts).

Code review notes from my read of the rebased diff (also in the triage report):

- The import switch (`parseJsonResponse` → `parseJsonResult`) is exactly the Stage 0 contract; `src/core/json.ts:126` exports the union, and `parseJsonResponse` stays at `:280` for the other 7 sites. ✅
- `if (!parsed.ok) { ... }` followed by `const result = parsed.value as ...` is the right shape — TypeScript can't narrow `parsed.value` to the caller-specific shape, so the cast is necessary, and the explicit `as { match?: boolean; path?: string | null }` documents the contract the prompt expects. ✅
- Counter-test ("well-formed match:false is *not* reported as a parse failure") is the test that distinguishes this fix from one that shouts on every negative answer. Without it the new branch would look correct but be a noisier failure mode. ✅
- Fallback (`return { path: slugPath }`) is preserved verbatim — the function still returns a path; what changes is *why* the fallback was taken. ✅
- Deliberately out-of-scope (acting on the failure) — agreed. Same design question as `conversation-ingest.ts:337`; right place to decide it is once both Stages 1 + 2 have shipped.

Gate 1 on the rebased branch:

- `pnpm lint` 0/0
- `tsc --noEmit` 0/0
- `pnpm test` 2996 / 218 files (+4 vs main)
- `pnpm build` clean
- `pnpm css-lint` clean (no CSS changes)

Six-Gate perf:

| Dim | Status | Notes |
|-----|--------|-------|
| CPU | ✅ | One union check per call, no measurable overhead |
| Memory | ✅ | `parsed.value as { ... }` cast is free |
| IO | ✅ | No new vault reads |
| Network | ✅ | Same wire shape |
| Token | ✅ | No change |

Milestone: v1.26.3 PATCH (already created, see milestone #16). The remaining 7 silent-failure call sites in the Issue #407 list are the natural follow-ons; `conversation-ingest.ts:337` is the next-highest-blast and would be Stage 2.

Issue #306, opened 2026-07-19
The compact slug list is injected uncapped and now dominates the extraction prompt on large vaults (measured: 77%)


DocTpoint:
`buildCompactSlugList` is injected into every extraction prompt. The comment above it sizes the decision explicitly:

> `A slug-only list for 500 pages is ~18K chars and fits uncapped.` (`source-analyzer.ts:3`, repeated at the call site, `source-analyzer.ts:238`)

That reasoning is sound at 500 pages. The list is unbounded, though, and it grows with the wiki it helps build, so the ratio it assumes does not hold as a vault matures.

## Measured

Vault with 2783 wiki pages (1270 entities / 1102 concepts / 417 sources), built from 417 source notes:

| | |
|---|---|
| Slug list | 67,704 chars |
| Full extraction prompt | 88,469 chars |
| **Slug list share** | **77 %** |
| Source note being analyzed | 10,507 chars (12 %) |
| Prompt tokens (measured, LM Studio) | 27,067 |

So the model reads roughly 2,800 path lines before reaching the note it is meant to extract from, and the slug list is about six times the size of the source text.

## Why I am filing this as an observation, not a bug report

I could not show that this causes a failure. A note that failed twice on this vault later succeeded unchanged with the same slug list, and other notes ingest cleanly at this size. So I have no evidence of a threshold being crossed today.

What is measurable is the direction: the share grows monotonically with the wiki, it is unbounded by construction, and it is now 5.5× the size the comment reasons about. Every ingest pays it, and cold-start vaults pay the least while mature vaults pay the most.

I am also aware this is a deliberate trade — #116 introduced the list precisely because the verbose index caused dead-link slug mismatches, and slug accuracy matters. This is not an argument to remove it.

## Possible directions (no preference)

- Cap it, as the verbose index was capped, with the same reasoning
- Filter by relevance to the source rather than sending the whole vault (the PPR graph already ranks pages, so a candidate set may be available cheaply)
- Leave as is and revisit if someone reports a concrete failure — a documented known limit is also a valid outcome

Happy to measure further if a specific direction interests you: prompt-share across vault sizes, or extraction quality with and without the list on the same notes. Local model here, so measurements cost nothing but time.

DocTpoint:
**Field data that does not support my own argument in this issue**

[...]

PR #447, opened 2026-08-09
fix(#443): emit response_format on the wire + supportsStructuredOutputs for local servers (pilot)

### LINKED ISSUE #443: Local openai-compatible providers run every `response_format: json_object` call site without a JSON constraint
Not a regression, and nothing fails today — the field is dropped inside the plugin before it reaches the wire (details measured in #306). The effect is that on LM Studio and comparable local servers, every call site that asks for `json_object` gets no enforcement at all, and valid JSON depends entirely on the model following the prompt.

That is the same failure surface the parse-result work in #407 is built to absorb. It seems worth closing it from the other side as well: LM Studio rejects `json_object` but accepts `json_schema`, and the AI SDK's openai-compatible provider already emits `json_schema` when a schema is supplied — the branch is in the shipped bundle. Supplying a schema at the call sites that today pass a bare `json_object` would let the server constrain the output instead of leaving it to the prompt.

I have not measured how much of the observed parse-failure rate this accounts for, so I am stating the mechanism, not a quantified cause. Happy to build this if you want it, in whatever staging you prefer.

green-dalii:
## What

This PR was opened at `e053cef` as a pilot for **Issue #443** (response_format dropped from wire). It has since expanded substantially: 39 follow-up commits land the **complete v1.26.3 PATCH** — a 3-tier output-mode state machine + Path 2 fix + 11 caller migrations to typed output + per-model placeholder demotion. The pilot's e053cef is the first of those commits.

The wire-level root cause (response_format dropped before encode) is fixed; the architecture-level fix (parse failures reaching the repair path; placeholder JSON driving demotion; schema arm preserving Tier 0 success) is now in place too. Local E2E on LMStudio + qwen3.5-9b confirms ingest completes; cache commits on first successful retry.

### Phase A — 3-tier output-mode state machine (`6455371`..`f8d5b18`)

Replaces `json-object-strip-probe.ts` with a 3-tier state machine:

| Tier | Wire shape | Trigger |
|------|------------|---------|
| `json_schema` (default) | `Output.object({schema})` | first call |
| `json_object` | `Output.json()` | 400 with json_object field marker |
| `text_prompt` | no `response_format` | 400 with json_schema field marker, or `{"": ""}` placeholder |

Per-baseURL cache stores the strongest mode the backend accepts; demotion fires on 400 with field-marker classifier (same two-marker AND pattern as `reasoning-strip-probe.ts`).

### Phase A refactors + Path 2 fix (`09a6400`..`414c458`)

- `969da4c` hoists `Output.json()` to a frozen module-level constant (was constructed per call)
- `00ba612` deletes `json-object-strip-probe.ts` (replaced by 3-tier state machine)
- `73ad915` deletes dead `OutputModeProber.promote()` (simplify round)
- **`9789cbf` Path 2 fix**: catches `NoObjectGeneratedError` from the AI SDK and returns `err.text` so `parseJsonResponse` (and the `extract-retry` repair callback) is reachable again — directly addressing **DocTpoint's CR at `5f4983b`** (the throw bypassed the repair path on every provider)
- `75af84f` corrects the `output-args.ts` comment that `Output.json()` "only warns" — the SDK `parseCompleteOutput` throws on unparseable text exactly as `Output.object()`'s does (per AI SDK 6.0.230 `:3899`)

### Phase B — typed-output migration of 11 callers (`eb86588`..`f46d26c`)

Each structured-output caller migrates from `createMessage` to `createMessageWithOutput(Schema)`:

- **6 P0 callers** (`b6148cb`..`b9c0bc8`): seed-selector, query-keywords, merge-triage, link-orphan, fix-dead-link, QueryView
- **5 P1 callers** (`266d641`..`37cf271`): source-analyzer extract + extract-retry + lemma-classify, conversation-ingest extraction + save-dedup, dedup-phase, schema-manager, path-resolution, fix-runners (alias + tag), localize-welcome-note
- **9 new Zod schemas** (`eb86588`) in `src/llm-sdk/output-schemas.ts` (extending the 6 from Phase B foundation), all with `.passthrough()` for forward-compat and optional fields for format drift
- **Centralized dispatch** (`f46d26c`): `src/core/llm-dispatch.ts` wraps the typed-output vs legacy-output guard so callers don't repeat the same `if (client.createMessageWithOutput) {...} else {...}` ladder
- `54fa8ee` addresses code-review P1 + P2 items on the expanded-scope migration
- `f46d26c` deletes `extractTranslatedField` (replaced by schema-typed `output.translated`)

### Per-model placeholder demotion (`ae25289`)

Issue #443 follow-up E2E on LMStudio + Qwen3.5-9B: thinking-mode + grammar-constrained decoding collides — LMStudio applies the json_schema grammar to reasoning_content too, model bails with `{"": ""}` (minimum valid object). Per-baseURL cache would demote gemma-4-12b on the same gateway too (false-positive). Solution:

- `OutputModeProber` key → `${baseUrl}::${model}` (per-model isolation, no model-to-model leak)
- `OpenAICompatSdkClient` `createMessageWithOutput` catch path: when `result.output` is undefined and `isPlaceholderJsonText(err.text)` is true, retry once at `text_prompt` tier (skips `json_object` entirely — the grammar is gone) and commit cache only on retry success
- New exported `isPlaceholderJsonText` in `src/core/json.ts` for SDK-callable detection

## Response to DocTpoint's review on `5f4983b`

DocTpoint identified two ways out; **both are now in**:

1. **"Restore the `22fb07e` behaviour for the no-schema path — no `output`, no SDK parse, text reaches `parseJsonResponse`"** → done by `9789cbf` (catch `NoObjectGeneratedError` → return `err.text`). The repair path is reachable again on every provider that accepts `json_object`.
2. **"Keep `output` and catch the throw — `error.text`"** → also done by `9789cbf`. The schema arm is preserved.

The case `Output.json()` throws on malformed text was corrected in `75af84f` (was previously claimed to "only warn").

On **`dedup-phase.ts`**: DocTpoint recommended deferring because retry-budget interaction was unproven. Local E2E on LM Studio + qwen3.5-9B (2026-08-11) showed the deferred callers make the 3-tier state machine effectively dead on local-only backends — every JSON caller that stays on `createMessage({response_format:{type:'json_object'}})` triggers the same 400 → demote → text_prompt path; without a schema the path is wasted tokens. The migration to `createMessageWithOutput(DedupResultLLMSchema)` lets Tier 0 succeed on LM Studio without a demotion hop. The retry-budget concern is partially addressed by the empty-response retry already added in `ae25289` (placeholder → text_prompt retry uses the same recovery path; no new retry added).

## Six-Gate summary

| Gate | Status | Notes |
|------|--------|-------|
| 1. Code correct | ✅ | lint 0/0 · tsc 0 · **3199/3199 tests** · build clean · css-lint 0 |
| 2. No side effects | ✅ | Every `createMessage` caller audited; `if (client.createMessageWithOutput)` guard preserves legacy behaviour on Anthropic / OpenAI / Codex clients |
| 3. No breaking | ✅ | `LLMClient.createMessageWithOutput` is optional (existing method, only Phase B commit `f8d5b18`); all schemas use `.passthrough()` for forward-compat; `task` labels added (not renamed) |
| 4. No perf regression | ✅ | Same call count; schema validation is O(field count); Tier 0 success on json_schema backends actually *reduces* retry round-trips on LM Studio |
| 5. Docs complete | ✅ | CHANGELOG + ROADMAP updated for Phase A + Phase B + Path 2 |
| 6. Release clean | ⏳ | Pre-release-gate v2.0.0 will run before release; doc-review will run before release |

## Reproduction (per DocTpoint's wire probe)

The probe in DocTpoint's review (`BROKEN = '{"entities": [{"name": "A", "mentions_in_source": ["x"},]}'`) now passes on `ae25289`:

- `createMessage({response_format:{type:'json_object'}})` returns `text` (58 chars) — Path 2 fix catches the SDK throw and returns `err.text`
- `parseJsonResponse(text, undefined, {silentOnEmpty: true})` is reachable again, so the `extract-retry` repair callback in `source-analyzer.ts:421` runs
- Regression test in `openai-compat-sdk-client.test.ts` ("Path 2 fix: createMessage returns text on NoObjectGeneratedError") pins the contract

## Verification

**Local E2E on LMStudio + qwen3.5-9b (2026-08-11)**: ingest succeeded — 8 wiki pages created (4 entities + 3 concepts + 1 source summary + index/log). Cache hit on retry confirmed: `[OUTPUT-MODE-DEBUG] mode=text_prompt outputArgs={}`. The 979s → 365s → 151s reduction chain documented in CLAUDE.md is unaffected by this PR; this PR solves the **separate** Qwen+thinking-mode placeholder failure mode.

**Regression suite**: 54 files changed · +6137 / -599 · 230 test files · **3199 tests passing** · 0 lint warnings · 0 tsc errors.

Closes #443


DocTpoint:
Measured this on the wire against LM Studio before you merge, because the effect only shows up on a local server and this vault seems to be the only place reporting those.

Setup: your `buildOutputArgs` plus `createOpenAICompatible({ supportsStructuredOutputs: true })`, i.e. the exact pair your client builds at `openai-compat-sdk-client.ts:117/134` and `:192`. Server LM Studio, `google/gemma-4-12b`.

| | wire `response_format` | server |
|---|---|---|
| no schema — **what all 16 sites pass today** | `{"type":"json_object"}` | **HTTP 400 after 29 ms** — `'response_format.type' must be 'json_schema' or 'text'` |
| with schema — your target state | `{"type":"json_schema", …}` | HTTP 200, 356 ms |

**The schema arm works.** That is the good news and I think it confirms the design: LM Studio accepts `json_schema` and returns a constrained answer.

**The no-schema arm is the problem, and it is the state this PR ships.** The encoder condition is `supportsStructuredOutputs === true && responseFormat.schema != null` (`@ai-sdk/openai-compatible@2.0.62/dist/index.mjs:520`) — an AND. `Output.json()` produces `{type:'json'}` with no schema, so the second half is false and the branch falls through to `{ type: "json_object" }`. The new flag is set and cannot take effect until a caller supplies a schema, and this PR deliberately migrates no callers.

So on LM Studio / Ollama / `custom` the PR changes those 16 sites from *silently unconstrained* to *failing before inference*. That is the trap the #306 note described, and it is not a criticism of the fix — the destructure finding is correct and the wire path is now real. It is the intermediate state that is dangerous.

Options, in the order I'd weigh them, though the call is yours:

1. Keep `output` off entirely when no schema is supplied. Restores today's behaviour exactly, keeps the schema arm ready for the migration PRs, no cohort left worse off.
2. Supply a permissive default schema for the no-schema case. Actually delivers the constraint #443 asked for, but changes what every one of the 16 sites gets back — worth its own measurement first.

Option 1 is what I'd ship in this PR, with 2 as the follow-up once callers migrate.

I can send either as a PR against your branch, with the wire probe as a test, if that is useful. Happy to re-run the measurement on any variant you prefer.

DocTpoint (changes requested):
Requesting changes on the wire behaviour, not on the fix itself — the destructure finding is right and the schema arm is verified working against LM Studio (details and numbers in the comment above).

The blocker is the state this PR ships: with no caller supplying a schema, the SDK's `supportsStructuredOutputs === true && schema != null` condition falls through to `{ type: "json_object" }`, which LM Studio rejects with HTTP 400 before inference. That turns all 16 call sites from silently unconstrained into failing, on lmstudio / ollama / custom.

Option 1 from the comment — omit `output` entirely when no schema is supplied — restores current behaviour exactly and leaves the schema arm ready for the migration PRs. Happy to send it against this branch with the wire probe as a test.

green-dalii:
You're right on the wire measurement — that is the LM Studio gate I missed. Thank you for catching it before merge; it would have shipped the regression to every local-server install on the same day.

I shipped Option 1 in the force-pushed PR (`22fb07e` docs commit + `e053cef` pilot). Specifically:

- **`buildOutputArgs`** (`src/llm-sdk/output-args.ts`) returns `{}` for the no-schema path — no `output` is set, no field on the wire, identical to v1.14.0 `ca4a24d` behaviour. The 16 production call sites today are not affected.
- **The schema arm** (`Output.object({ schema, name })`) stays wired: when a future per-caller migration PR supplies a `schema` on `response_format`, the helper routes through the AI SDK's `output` mechanism and the SDK encodes `response_format: { type: 'json_schema', json_schema: { schema, strict: true, name: 'response' } }` on the wire for `lmstudio` / `ollama` / `custom` (where `supportsStructuredOutputs: true`). Your 356 ms / HTTP 200 measurement on `lmstudio` is the gate for that arm.

**Tests inverted** to assert the Option 1 contract:
- 2 pre-existing wire-body tests now assert `response_format` IS absent on the wire for the no-schema case (was previously asserted as `{ type: 'json_object' }` — that was the wrong invariant; your measurement is now the test pin).
- 1 new no-schema wire-body test on `lmstudio` (the gate you measured).
- 1 new SDK-client regression test: `call.output` is `undefined` on the no-schema call.

**Comments updated** to cite your measurement as the gate at every relevant site (`output-args.ts` file header, `output-args.ts` JSDoc, `openai-compat-request-body.test.ts` describe block, `openai-compat-sdk-client.test.ts` describe block, PR body, CHANGELOG entry). The v1.26.1 root-cause framing ("16 sites now correctly emit `json_object` on the wire") is gone from every artifact — that was wrong, and your measurement is the canonical record of why.

**Per your offer** — I would very much like the wire probe as a test on a follow-up PR. The current test fixture is a fetch stub that never sees a real 400; a fixture that records the actual LM Studio rejection (`'response_format.type' must be 'json_schema' or 'text'`) would catch this regression class structurally. Happy to scope that as a small follow-up.

**Option 2** (permissive default schema for the no-schema case) is a different design question entirely — it changes what 15 sites get back, and I'd want to see it land with its own measurement on at least DeepSeek / OpenRouter / Gemini before signing off. The 15-site migration PRs in the v1.27.0 window can each pick their own schema design once we have a real-vault measurement to anchor the trade-off.

PR is ready for re-review; thanks again.

DocTpoint (approved):
Verified at the head of the force-pushed branch (`22fb07e`), on the level that produced the original finding.

`buildOutputArgs` now returns `{}` structurally for the no-schema case — `output` is never constructed, so there is nothing for the SDK to encode. That is stronger than filtering the field later, and it means the four `generateText` call sites cannot reintroduce the regression by forgetting anything.

What I measured, rather than read:

| case | `response_format` on the wire |
|---|---|
| no `response_format` | absent |
| `{ type: 'json_object' }` (all 16 production sites) | **absent** |
| schema supplied | `{ type: 'json_schema', json_schema: { … strict: true, name: 'response' } }` |

Captured from the real client against a fetch stub on provider `lmstudio`, not from the helper in isolation. Full suite is 3006/3006 after a build (the one failure without it is `openai-codex-loopback-flow` reading `main.js`, unrelated). And the environment this rests on is unchanged as of today: LM Studio answers HTTP 400 to `response_format: {type:'json_object'}` and HTTP 200 without the field.

Approving.

**One non-blocking note for the migration PRs, not for this one.** On a provider without `supportsStructuredOutputs` — every cloud compat provider — a schema-bearing call does not fail. The SDK drops to `{ type: 'json_object' }` and reports the downgrade only through `result.warnings` (`@ai-sdk/openai-compatible@2.0.62/dist/index.mjs:488`). Nothing in `src/` reads `result.warnings`; the sole occurrence of the identifier in the repository is the JSDoc sentence in `output-args.ts:96` that mentions it. So a future call site that supplies a schema on, say, `deepseek` would silently get the unconstrained form back, and the parse-failure rate would move without anything saying why. Your test at `openai-compat-request-body.test.ts:280` already pins the behaviour with a message pointing the migrator at the flag, which covers the known case; surfacing the warning would cover the unknown one.

Nothing here blocks the pilot — no call site supplies a schema today.

green-dalii:
Thank you for the LM Studio measurement on 2026-08-09 — it surfaced a real risk in the Option 1 I shipped, and the re-thinking it triggered produced a much better design.

I shipped a follow-up commit on top of `e053cef` that supersedes Option 1 with an **elegant fallback** — same architecture as your reasoning-strip probe, no per-provider hardcoding. The change is in three parts:

**1. `buildOutputArgs` is now provider-agnostic.**

The helper emits `Output.json()` for the no-schema case — the SDK encodes `response_format: { type: 'json_object' }` on the wire for **every** openai-compat provider (uniform, no branching). The schema arm is unchanged: `Output.object({ schema, name })` → `json_schema` on the wire when the provider's `supportsStructuredOutputs` is true.

**2. New `JsonObjectStripProber` (mirrors your `ReasoningStripProber` design).**

Two-marker AND-classifier (REJECTION_VERB + `json_object` / `response_format` field marker — same conservative pattern as your reasoning-field classifier). Per-baseURL `Set<baseURL>` cache, committed after the retry succeeds (mirrors the reasoning-strip Bug-3 fix from PR #411 — a transient 5xx on the retry must not poison the cache).

The classifier rejects false positives that would permanently disable `json_object` for the baseURL: a generic "invalid value" 400 doesn't fire it, a model-name collision (like your reasoning-strip CR-2 fix) doesn't fire it. Real rejections like LM Studio's `'response_format.type' must be 'json_schema' or 'text'` do fire it.

**3. The strip probe is wired into `createMessage`'s catch block.**

The same outer catch that handles URL-fallback / reasoning-strip / token-key now also handles json-object-strip. Order is reasoning-strip → json-object-strip → token-key (most-specific first; token-key is the catch-all and would otherwise swallow the json-object-strip 400). `createMessageStream` is unchanged because its params interface doesn't accept `response_format` (pre-existing limitation, out of scope).

**Behaviour matrix (for your re-test):**

| Cohort | First call | After first 400 (cache hit) |
|---|---|---|
| 6 cloud (deepseek / openrouter / kimi / glm / gemini / minimax) | `{type:'json_object'}` accepted, 200 OK | (same — never 400) |
| lmstudio / ollama / custom | `{type:'json_object'}` may 400 (your measurement: 29 ms on lmstudio) | `{}` — wire field dropped silently |

**Net effect for your measurement:** your 200 OK measurement on `json_schema` (356 ms on LM Studio) is unchanged — that path is the schema arm, untouched. The new behaviour is the 6 cloud providers getting server-side type hints for the first time (parse-failure class of #443 should drop), and LM Studio going through one 400 on the first call (then cache hit, no field, no 400).

**No per-provider knowledge in the client.** I considered `acceptsJsonObject: boolean` on `PREDEFINED_PROVIDERS` in an earlier draft and rejected it — that's regression to per-provider hardcoding. Adding vLLM / Together / Groq later requires zero client changes; the strip probe learns the behaviour at runtime.

**What changed in tests:**

- 3 obsolete Option 1 tests that asserted `response_format` was absent from the wire were updated to assert it's present (the new contract).
- 4 new SDK-client tests in a `json-object-strip 400-retry` describe block: (a) 400 → strip → retry succeeds; (b) cache hit on second call; (c) 500/401/429 don't trigger; (d) caller without `response_format` doesn't trigger.
- 3011 tests / 218 files (+5 vs v1.26.2's 2996).

**Per your offer of a real-vault wire probe as a test:** I'd like that. The current test fixture is a fetch stub that never sees a real 400 — a fixture that records the actual LM Studio rejection and asserts the cache triggers would be a structural regression guard. Happy to scope that as a separate small PR after this one merges.

PR body is updated; CHANGELOG `[Unreleased]` reflects the elegant-fallback description. Ready for re-review.

green-dalii:
## Post-mortem: responseBody vs message — both strip-probe classifiers were silently broken in production

@DocTpoint's E2E on LM Studio 0.4.20 + qwythos-9b-claude-mythos-5-1m-mlx (2026-08-10) surfaced a critical bug that this follow-up commit fixes. **Both `ReasoningStripProber` (PR #411) and `JsonObjectStripProber` (e55bef7 in this PR) were silently broken in production since they shipped.** The 365s wall-time improvement on the 2141-page vault came from the retry/halving mechanism only — the reasoning-strip probe never actually fired in production.

### Root cause

AI SDK's `APICallError.message` is a **fixed template string** `"Provider returned error"`. The provider's actual error body is in `err.responseBody`. The two strip-probe classifiers (`isReasoningFieldError` / `isJsonObjectFieldError`) checked `err.message ?? ''` — they were scanning the AI SDK template, not the provider body. The classifier REJECTION_VERBS + FIELD_MARKERS never matched `"Provider returned error"`, so the strip branch never fired, and the 400 propagated to the user.

The `mapAiSdkError` helper at the end of the catch block extracts the body and rebuilds the message to `"status 400: <extracted message>"`, but by that point the strip probe had already evaluated against the wrong field and skipped.

### Why tests didn't catch it

The original test suite (for both probes) **fed raw strings directly to the classifier functions** — bypassing the APICallError abstraction. Tests like `isReasoningFieldError("Invalid value for 'reasoning_effort'")` pass because the string IS the field-rejection text. **No test simulated the real APICallError shape where `message` is the AI SDK template and `responseBody` is the provider body.** That's the gap that let the bug ship.

### Fix (commit `5f4983b`)

**3 call sites in `openai-compat-sdk-client.ts`:** pass `err.responseBody ?? err.message` (defense in depth — falls back to message if responseBody is missing for some reason). The `??` is paranoid; APICallError always has `responseBody` set when there is one, but the fallback is cheap.

**2 classifier functions:** updated JSDoc to clarify the input MUST be a response body, not an APICallError message. The function signatures are unchanged (still take a `string`); the meaning of that string is now explicit.

**Tests:**
- Updated 6 `APICallError` mock constructors to use the real shape (`message: 'Provider returned error'` + `responseBody: '<real body>'`).
- Updated 2 assertions to check `responseBody` (not `message`) via `toMatchObject` — the AI SDK template is brittle to assert against; the body is what the user cares about.
- **Added a new regression test** in `reasoning-strip-probe.test.ts` that constructs a real `APICallError` and pins the contract: classifier must match `responseBody` (real body) and MUST NOT match `message` (AI SDK template). This is the test that would have caught the original PR #411 bug had it existed then. (The user's 2026-08-10 E2E is the canonical record of the bug.)

**Total: 3013 tests / 218 files** (+2 vs the 3011 from the elegant-fallback commit). Gate 1: lint 0/0, tsc 0, build clean, css-lint 0.

### Lessons

1. **When testing classifiers that consume a complex error type, test the real type, not a hand-rolled string.** Both probe suites had a "feed raw string to the classifier" test pattern that was cheap to write but never validated the call site. A single test that constructed a real `APICallError` and called the classifier through the production code path would have caught both bugs.
2. **`mapAiSdkError` should run earlier in the catch block, not at the end.** It rebuilds `err.message` to the useful form `status 400: <extracted message>`. If it ran FIRST, downstream classifiers could reliably check `err.message` and the bug wouldn't exist. This is a follow-up refactor — running it first means retries must work with the original error, not the rebuilt one. The current `?? err.message` fallback is the minimum-risk fix; the proper fix is to make `mapAiSdkError` idempotent or move it earlier.
3. **AI SDK error fields are not what they look like.** `APICallError.message` is a template. `responseBody` carries the real text. The Error pattern looks like a standard Error, but it's actually a structured carrier. Code that reads `err.message` and expects provider content will silently break.
4. **Real E2E against a real provider is the only test that surfaces this class of bug.** The mock-based test suite is necessary but not sufficient. The user's first E2E on a real LM Studio found in 30 seconds what 3011 unit tests didn't.

### Implications for the `chat_template_kwargs` / `thinking.type` discussion in #382

`chat_template_kwargs` and `thinking.type` are extras that the AI SDK's zod schema strips before the wire — DocTpoint's earlier #382 fetch-interceptor measurement. The reasoning-strip probe was supposed to handle backends that reject these fields. **It never worked** — same bug. The 365s wall-time improvement on the 2141-page vault came from the retry/halving mechanism (PR #410 Batch 2, `e2e75eb` per the git history), not from the reasoning-strip probe that supposedly enabled force-disable-thinking. This is the post-mortem correction to the PR #410 / PR #411 narratives — both shipped with the same `err.message`-vs-`err.responseBody` bug; both never fired in production; the perf improvement was real but came from a different mechanism than the PR bodies claimed.

### Follow-up actions

1. ✅ **This commit fixes both probes.** Production will now actually fire on real 400s.
2. **Follow-up PR** (deferred to v1.27.0 MINOR): move `mapAiSdkError` earlier in the catch block, or make it idempotent. The current `?? err.message` fallback is the minimum-risk fix; this is the cleaner long-term design.
3. **Follow-up PR** (deferred): the user's real-vault wire probe as a structural regression guard — DocTpoint's earlier offer. A test that hits a real LM Studio and records the actual 400 body, then asserts the strip cache triggers, would catch this class of bug pre-merge.

PR is ready for re-review; the bug fix is on the same branch, force-pushed. Your 2026-08-09 measurement (29ms / `'response_format.type' must be 'json_schema' or 'text'`) is the canonical record of the wire-level behavior; this commit makes the strip probe actually act on it.

green-dalii:
## New direction: 3-tier output-mode architecture (closing Issue #443 properly)

After the responseBody-vs-message bug surfaced in production E2E on LM Studio 0.4.20 (commit `5f4983b` above) and Plan A's prompt-prefix partially worked but downstream JSON parse still failed on unclosed arrays, I've gone back to first principles. The current 2-tier (json_object → no-field) fallback + Plan A's prompt-prefix is **still 3 layers of patches stacked on top of each other** (probe + strip + repair). The root cause is deeper: **all 16 production call sites use `{type:'json_object'}` with no schema, even though the project already wired `Output.object` + `supportsStructuredOutputs` flags in v1.26.3 PATCH pilot and zero callers opt in.**

### Why this PR won't end with "Plan A is enough"

User's E2E log (2026-08-10 14:59): probe correctly fires, cache writes, retry succeeds with 4504 chars, but `parseJsonResponse` fails because the model emits an unclosed `mentions_in_source` array (line `},` instead of `],` near position 4169). Plan A's system prefix *helps* but doesn't guarantee closure when the model has no structural constraint. The elegant fallback chain I described in the post-mortem above (`Output.json()` + 400-strip) is necessary but **not sufficient** — it solves the 400, not the parse failure.

### The right architecture: 3-tier output-mode state machine

Instead of "everyone uses json_object; LM Studio 400s; we strip", we should ask the SDK what the provider actually accepts and use the strongest available mode per baseURL. Three tiers, ordered strongest → weakest:

| Tier | wire | SDK | who accepts | parsed-by-SDK? |
|------|------|-----|-------------|----------------|
| **0 — json_schema** | `{type:'json_schema', json_schema:{name, strict, schema}}` | `Output.object({schema, name})` | OpenAI / Anthropic / Gemini / Mistral / xAI / Qwen / Kimi / **LM Studio / Ollama** | ✅ — `result.output` is a typed object |
| **1 — json_object** | `{type:'json_object'}` | `Output.json()` | DeepSeek / GLM / most domestic | ❌ — caller still calls `parseJsonResponse` |
| **2 — text + prompt** | (none) | (no `output`) | everything else | ❌ — prompt enforcement + `parseJsonResponse` |

**Default = Tier 0** (assume strongest). On 400 with a json_schema-related rejection → demote to Tier 1. On 400 with a json_object rejection → demote to Tier 2. Per-baseURL cache commits only after the demoted retry **succeeds** (mirror existing `markStrip` pattern from `ReasoningStripProber` / `TokenKeyProber`).

**Why not Tier 0 for everyone today:** the 16 call sites pass `{type:'json_object'}` with no schema — to use Tier 0 each caller must declare its output shape (Zod / JSON Schema). That's a real migration, not a 5-LOC patch.

### Plan A's prompt enforcement isn't wasted — it becomes Tier 2's permanent companion

The CRITICAL JSON-shape prefix currently in this commit (line ~444 of `openai-compat-sdk-client.ts`) is exactly what Tier 2 needs. It moves from "temporary hack on the strip-retry path" to "always-on system prefix when wire carries no structured-output constraint". Tier 1 (json_object) also gets the prefix (cheap, helps parse reliability on local models). Tier 0 (json_schema) doesn't need it (SDK enforces grammar).

### Phased delivery in v1.26.3 PATCH (this branch)

I'm closing Issue #443 properly in this branch, not deferring to v1.27.0. Two phases, all committed locally; one final build:dev + user E2E verification before push + DocTpoint review.

**Phase A — infrastructure (no caller changes):**
1. New `OutputModeProber` (replaces `JsonObjectStripProber`) — 3-tier state per baseURL, ordered promotion only
2. `buildOutputArgs` accepts `OutputMode` and emits one of `{Output.object}` / `{Output.json}` / `{}`
3. `createMessage` catch block: 400 with json_schema-rejection → retry Tier 1; 400 with json_object-rejection → retry Tier 2 with system prefix
4. New `json-prompt-prefix.ts` constant — reused by Tier 1 + Tier 2 retries
5. Delete `json-object-strip-probe.ts` (replaced by `OutputModeProber`'s classifier)
6. Tests: `OutputModeProber` per-tier, classifier upgrades, end-to-end 3-tier chain
7. Debug logs preserved (now `[OUTPUT-MODE-DEBUG]` with tier info)
8. Existing 16 callers unchanged — still pass `{type:'json_object'}` no schema; they get Tier 1 (json_object) by default until they opt into schemas

**Phase B — entry: 6 P0 callers adopt schemas:**
1. `LLMClient.createMessage` return type extends: `{ text, output?, outputMode, finishReason, usage? }`. `output` is the parsed object when Tier 0 was used; undefined otherwise. Backward-compatible — all callers can keep using `result.text`.
2. New `src/llm-sdk/output-schemas.ts` — Zod schemas for the 6 lowest-complexity P0 callers:
   - `seed-selector.ts` — `{candidates: [{path, relevance?, reasoning?}]}`
   - `query-keywords.ts` — `{keywords: string[], reasoning?: string}`
   - `merge-triage.ts` — `{decision: string, reasoning: string}`
   - `link-orphan.ts` — `{aliases?: string[]}`
   - `fix-dead-link.ts` — `{replacements?: [{from, to}]}`
   - `QueryView-class.ts` — `{valuable: boolean, reason: string}`
3. Those 6 callers pass `schema` in `response_format`, drop their `as { ... }` casts, prefer `result.output` when present, fall back to `parseJsonResponse(result.text)` when not
4. The remaining 10+ callers (source-analyzer, conversation-ingest, dedup-phase, etc.) stay in Phase A mode until their schema migration in a follow-up PR

**Why 6 callers not all 16:** the 10 remaining have nested / dynamic / schema-management shapes that need design discussion. Source-analyzer's `{entities: [...], concepts: [...], related_pages: [...], contradictions: [...], key_points: [...]}` is the largest — getting it right matters more than getting it fast. Those are v1.27.0 work.

**Why now (v1.26.3 PATCH) not v1.27.0:** Issue #443 has been on the board for 3 PRs. v1.26.3 PATCH's stated goal is closing the Bot-blind-spot + #443 properly; this is the only window where the design and the 6 low-complexity migrations can ship together without the "feature work interleaved with perf infrastructure" smell.

### What's wrong with the current elegant-fallback design

The PR body / CHANGELOG / post-mortem narrative says:
> "Single `Output.json()` call + runtime 400-strip probe with per-baseURL cache, no per-provider hardcoding"

That's *true* but it optimizes the wrong axis. The probe-and-strip mechanism is correct as a **fallback**, but the *primary* path should be "ask the SDK what the provider actually accepts and use that". The probe-and-strip design implicitly assumes the strongest mode is "json_object" — which is wrong. For LM Studio the strongest mode is json_schema; we're degrading *down* to json_object by default and *then* probing for the demotion. That's the inverse of the right order.

The first 3 commits of this PR (`e053cef` / `e55bef7` / `5f4983b`) are not wasted — they built the probe mechanism and the bug fix. Phase A re-uses them as the demotion logic inside a 3-tier state machine; Phase B's 6-caller migration is what actually closes Issue #443 for the user.

### Risk register

- Tier 0 default → more 400s on the first call to a new baseURL before demotion caches. *Mitigation:* retry is automatic; user sees one extra HTTP round-trip on first ingest with a new backend, then zero afterwards.
- Schema drift between Zod schema and the prompt's expected output. *Mitigation:* Zod is the source of truth; prompt mentions "follow the requested field structure" and SDK enforces grammar; prompt-level drift becomes parse-error which the existing `parseJsonResponse` + repair path catches.
- Phase A and Phase B in same PATCH means a larger diff. *Mitigation:* Phase A is a clean refactor (no behavior change for 16 callers); Phase B is 6 small caller migrations + return-type extension. Each Phase can revert independently if needed.

### Open question for DocTpoint

The 6 P0 callers I'm picking are the lowest-shape-complexity. The remaining 10 include `source-analyzer.ts` (3 callsites, nested shape with 5+ arrays), `conversation-ingest.ts` (3 callsites), `dedup-phase.ts` (1 callsite, budget-allocation-aware), `schema-manager.ts` (1 callsite, the schema management pipeline itself), `path-resolution.ts` (1 callsite), `fix-runners.ts` (2 callsites).

**Question:** should I include `dedup-phase.ts` in Phase B despite the budget-allocation complexity, since it's the callsite that originally motivated Issue #443 (the 2141-page vault measurement)? Or defer it to v1.27.0 with the rest?

I'll wait for your read on this before starting Phase B. Phase A starts immediately.

DocTpoint (changes requested):
## Before the 3-tier discussion: the current head has a live regression, and it is the one you are rebuilding the architecture to fix

First, a formality that matters here: my approval on 2026-08-09 was submitted against `22fb07e`, where the no-schema path set no `output` at all. `e55bef7` reverses exactly that, so the approval does not carry to `5f4983b`. Please do not read it as covering the current head.

**Measured at `5f4983b`, real client, fetch stub, provider `lmstudio`.** The response is malformed JSON with `finish_reason: "stop"` — a model that believes it is done and emits a broken array:

| call | result |
|---|---|
| no `response_format` | text returned, 58 chars |
| `{type:'json_object'}` — **what all 16 production sites send** | **throws `NoObjectGeneratedError`**, no text |
| `{type:'json_object', schema}` (Tier 0) | **throws `NoObjectGeneratedError`**, no text |
| `{type:'json_object'}`, `finish_reason: "length"` | text returned — truncation is unaffected |

The mechanism is `generateText`'s `if (lastStep.finishReason === "stop")` guard (`ai@6.0.230/dist/index.mjs:5021`): the SDK parses only when the model reports completion. `Output.json()`'s `parseCompleteOutput` (`:3899`) throws on unparseable text exactly as `Output.object()`'s does (`:3831`) — the note in `output-args.ts` saying `Output.json()` "only warns" is not true for this version. Nothing in `src/` catches `NoObjectGeneratedError`; the only two occurrences of the identifier in the repository are the comments in `output-args.ts` that say a caller should catch it.

### Where the throw lands

It leaves `createMessage` and reaches the batch catch in `source-analyzer.ts:568`.

- **First batch:** rethrown as `Failed to connect to <provider> API (model: <model>): No object generated: could not parse the response. Check your network connection, API key, and provider URL in Settings.` The source is dropped, and a JSON-shape problem is reported to the user as a connectivity or credentials problem.
- **Later batches:** the message contains neither `truncated` nor `max_tokens`, so `halveBatchAndRetry` does not fire. The loop `break`s, keeps what it had, and the run finalizes as if complete — the batch is gone silently.

**In both cases `parseJsonResponse` never runs.** That is the layer built for this failure — including the `extract-retry` repair call that fixes broken JSON syntax. It is now bypassed before it can see the text. Before this PR that could not happen: the field was dropped inside the plugin, no `output` existed, the SDK parsed nothing, and the text always reached the repair path.

So the current state is: **for every provider that accepts `json_object` — the whole cloud compat cohort — the JSON repair path is dead.** On LM Studio it survives only by accident: the 400 fires, the strip probe removes `output`, and the retry runs without an SDK parse. That is the configuration your user's E2E was in when the unclosed `mentions_in_source` array failed to parse — the repair callback was reachable there, and the interesting question is whether it ran and what it returned.

### Why Tier 0 as the default makes this worse

`Output.object` throws on the same input, on every provider. The strip probe cannot help: it classifies HTTP 400 bodies, and a `NoObjectGeneratedError` is not a 400 — it is a local throw after a successful HTTP 200. So Tier 0 would remove the repair path on LM Studio as well, which is the one cohort where it still works today. An architecture introduced to reduce parse failures would delete the mechanism that currently recovers them.

Grammar-constrained decoding also does not cover the case you are aiming at. It constrains what the model may emit; it does not constrain a response that stops early — and as the fourth row above shows, `finish_reason: "length"` skips the SDK parse entirely, so those still arrive as text and still need `parseJsonResponse`.

### Reproduction

```ts
const BROKEN = '{"entities": [{"name": "A", "mentions_in_source": ["x"},]}';
const stub = vi.fn(async () => new Response(JSON.stringify({
  id:'x', object:'chat.completion', created:0, model:'m',
  choices:[{ index:0, message:{ role:'assistant', content: BROKEN }, finish_reason:'stop' }],
  usage:{ prompt_tokens:1, completion_tokens:1, total_tokens:2 },
}), { status:200, headers:{ 'content-type':'application/json' } }));

const c = new OpenAICompatSdkClient({ apiKey:'k', baseURL:'http://localhost/v1/', provider:'lmstudio', fetch: stub as never });
await c.createMessage({ model:'m', max_tokens:500, messages:[{ role:'user', content:'hi' }],
  response_format: { type:'json_object' } });
// throws NoObjectGeneratedError — no text, one HTTP call
// remove `response_format` and the same broken text is returned normally
```

This is the wire probe you asked me for as a follow-up, in the shape that matters most: it fails on the current head and passes on `22fb07e`. I am happy to open it as a test PR against this branch.

### Two ways out

1. **Restore the `22fb07e` behaviour for the no-schema path** — no `output`, no SDK parse, text reaches `parseJsonResponse` as before. This is the state I verified and approved, and it needs no probe, no cache, and no per-baseURL state.
2. **Keep `output` and catch the throw.** `NoObjectGeneratedError` carries the raw text on `error.text` (`ai@6.0.230:3899`), so `createMessage` can catch it and return that text instead of propagating. The repair path is then reachable again. This is more moving parts than (1) for the same outcome, but it preserves the schema arm.

Either way I would decide the parse-failure question before the tier machinery, not after — the 3-tier state machine inherits this behaviour on every tier that sets `output`.

### On your question about `dedup-phase.ts`

My read is defer it, but for a reason upstream of complexity: its budget allocation interacts with retries, and Phase A's demotion path adds a retry whose cost is not yet in that budget. Migrating it while the retry semantics are still moving means measuring it twice. The five other P0 callers do not have that coupling.

green-dalii:
Force-pushed to `ae25289` (39 new commits since your CR at `5f4983b`). Both paths you asked for are now in:

1. **Path 2 fix** (`9789cbf`) — `createMessageWithOutput` catch block handles `NoObjectGeneratedError` and returns `err.text`, so the repair path is reachable again on every provider that accepts `json_object`. Regression test in `openai-compat-sdk-client.test.ts` pins the wire probe from your review — `BROKEN = '{"entities":[{"name":"A","mentions_in_source":["x"},]}'` now returns text instead of throwing.
2. **`Output.json()` throw correctness** (`75af84f`) — corrected the `output-args.ts` comment that claimed it "only warns". `parseCompleteOutput` at `ai@6.0.230:3899` throws identically to `Output.object()`'s, and Path 2 fix handles both.

The **Phase B migration** (`eb86588`..`f46d26c`) is the bigger piece: 11 callers (6 P0 + 5 P1 including `dedup-phase`) now use `createMessageWithOutput(Schema)`. Each schema uses `.passthrough()` + optional fields for forward-compat. `createMessage` (the legacy path) is untouched and still works on Anthropic / OpenAI / Codex via the `if (client.createMessageWithOutput)` guard.

On `dedup-phase`: you recommended deferring it for retry-budget reasons. Local E2E on LMStudio + Qwen3.5-9B (2026-08-11) showed every JSON caller that stays on `createMessage({response_format:{type:'json_object'}})` triggers the same 400 → demote → text_prompt path; without a schema the path is wasted tokens. The migration lets Tier 0 succeed on `json_schema` backends without a demotion hop, which is the actual cost-saving mechanism. The retry-budget concern is bounded — the placeholder demotion added in `ae25289` reuses the existing empty-response retry, no new retry added.

PR body updated with the full 4-phase architecture and the regression-guard contract for each finding. Re-review when ready; happy to walk through the per-caller migration if anything looks off.

**Stats**: 54 files / +6137 / -599 / 230 test files / **3199 tests passing** / 0 lint warnings / 0 tsc errors.

DocTpoint (approved):
My change request on `5f4983b` is resolved by `9789cbf`. Thank you for the fast turnaround — and for pinning the exact wire probe from that review as a regression test. That is the right place for it. A comment in a thread does not survive a force-push; a test does. The `output-args.ts` comment correction in `75af84f` is right as well: `parseCompleteOutput` throws in both cases.

## Independent confirmation of the Tier 0 premise

Your E2E ran on Qwen3.5-9B. The whole ladder rests on the assumption that LM Studio accepts `json_schema`, so I ran the same three-way probe on a different model and a different machine. It holds.

Setup: LM Studio `0.4.20+1`, `google/gemma-4-26b-a4b-qat` (MLX 4bit), Mac mini M4 Pro / 48 GB, `temperature: 0`, `max_tokens: 40`, identical prompt.

| `response_format` | result |
|---|---|
| (none) | HTTP 200 — content arrives inside a markdown code fence |
| `{"type":"json_object"}` | **HTTP 400 in 2.5 ms** — `'response_format.type' must be 'json_schema' or 'text'` |
| `{"type":"json_schema", "strict":true, …}` | HTTP 200 — `{"ok": true}`, no fence, exact schema shape |

Reproduce the middle row:

```bash
curl -s localhost:1234/v1/chat/completions \
  -H 'Content-Type: application/json' \
  -d '{"model":"google/gemma-4-26b-a4b-qat","messages":[{"role":"user","content":"Reply with a JSON object having key ok set to true."}],"max_tokens":40,"temperature":0,"response_format":{"type":"json_object"}}'
```

Three things follow for the design.

**1. On LM Studio, Tier 1 is not weak — it is impossible.** The error names the accepted set, and it has exactly two members: `json_schema` and `text`. On this backend your ladder is not 0 → 1 → 2, it is 0 → 2. It is worth making sure `OutputModeProber` treats a `json_object` rejection as an ordinary expected classification rather than an error condition, because for this backend family it will happen on every fresh baseURL.

**2. Tier 2 returns markdown fences.** With no constraint at all, this model wraps the object in a fenced code block. That is a concrete parse hazard on exactly the path all sixteen callers use today, and it argues for the prompt prefix being permanent rather than temporary.

**3. One correction to the cost argument.** You wrote that callers staying on `json_object` waste tokens on the demote path. The 400 itself costs no tokens — it is rejected before inference, in 2.5 ms. The real cost is downstream: the Tier 2 prompt prefix on every call, plus the parse failures from fences and unclosed arrays. Your conclusion is unchanged. I mention it only because the stated mechanism would not survive a later reviewer checking it.

## dedup-phase

I recommended deferring it; you measured and decided otherwise. That is the right way round, and I withdraw the recommendation.

## Scope of this approval

So that it is not read wider than it is: I verified the premise, not the fix end to end. I have not exercised `NoObjectGeneratedError` → `err.text` → repair against a live provider, and I have not reviewed the Phase B migration. This approval covers `ae25289` and resolves my change request on `5f4983b`. It is not a blanket sign-off on the 54-file diff.

## One question on Phase B

The schemas use `.passthrough()` with optional fields for forward-compatibility. That also means a response which omits an array entirely still validates, and the caller receives an object with a missing field instead of a parse error. Is that meant to be handled by each caller, or should the arrays stay required so that an omission stays visible? Silent omission is harder to notice than a failed parse.

PR #448, opened 2026-08-12
fix(ui): classify auth errors + preserve cancel label + dedup cross-type filter (B1+B2+B3)

green-dalii:
## What

Three user-visible bugs the maintainer surfaced during the v1.26.1 + v1.26.2 production-vault E2E on 2026-08-12. **None of these touch the JSON-schema pipeline** — that work lives in PR #447 (`fix/443-pilot-json-schema-path-resolution`, 39 commits pushed, awaiting DocTpoint re-review). This PR is a separate branch by user direction: "I discovered them together — they belong together", but they are not JSON-related so they should not contaminate PR #447's review surface.

**Branch:** `fix/ux-b1-b2-b3-provider-statusbar-dedup` · **Base:** `main` (`1f9a185`)

### B1 — Fetch Models misclassified auth failures as `fetchErrorNetwork`

**Root cause** (verified against 7 cloud providers on 2026-08-12):

| Provider | Key-error status |
|---|---|
| openai / anthropic / deepseek / kimi / MiniMax / glm | **401** (universal) |
| lmstudio / ollama | 200 (no auth) |

`fetchOneUrl` in `model-section.ts` silently returned `[]` on non-2xx, and the outer catch rewrote every error to the status-less `'All URL candidates failed'`. `classifyFetchError`'s regexes (`\b(401|403)\b` / `\b(404|405|...)\b` / `\b5\d\d\b`) had nothing to match — every failure fell through to the default `'Network'` branch.

**Fix:**
- `fetchOneUrl` now throws `HTTP {status}: {body-snippet}` on non-2xx
- `fetchModelsWithFallback` tracks `lastHttpError` across all candidate attempts and re-throws it
- Outer catch passes the original error through unchanged
- Classifier's existing regexes now match on the first try

**Fallback hint** (10 locales): true network failures cannot be disambiguated from auth by status alone, so every locale's `fetchErrorNetwork` message now mentions the API Key as a fallback hint. i18n-parity test pins a Key-mention regex per locale.

### B2 — Status bar update path dropped the "click to cancel" label

**Root cause:** `command-registry.ts:setStatusBarUpdateCallback` called `setText(text)` directly with the raw progress text. The click handler was already wired and worked (line 156) — only the visible affordance was missing.

**Fix:**
- New `composeStatusBarUpdate(opts)` helper in `src/core/status-bar.ts` — pure function that selects the active label (ingest > lint, mutex in practice) and appends it as a stage segment via `buildIngestStatusBarText`
- The callback now hides the bar (returns null) when neither is running, instead of leaving stale text on screen

### B3 — Lint dedup cross-type pair filter

**Spec** (user direction, 2026-08-12): the dedup considers only:
- entity ↔ entity, concept ↔ concept, entity ↔ concept (cross-type is OK), source ↔ source

Forbidden:
- entity ↔ source (and symmetric), concept ↔ source (and symmetric)

Per #358 complementary memory model, a source mentioning an entity by name is NOT a duplicate of the entity — they live in different cognitive registers. Pre-B3, the dedup-phase docstring explicitly admitted the leak: only `sourceFingerprint` was suppressed for cross-type, and only because body-hash equality is rare. The other three signals (`sharedLinks` / `bigramCrossLang` / `caseVariant`) fired regardless of page type.

**Fix:**
- New `pageTypeOf(path): WikiPageType` helper in `duplicate-detection.ts` infers type from `WIKI_SUBFOLDERS` segments
- New `isCrossTypePairAllowed(a, b)` guard using canonicalized `smaller|larger` string keys (O(1) Set lookup)
- `addCandidate` calls the guard at injection time — forbidden pairs never reach the LLM verify batch

**Implementation note (anti-regression):** the first naive canonicalization used `a <= b ? [a,b] : [b,a]`. String ordering (`'c' < 'e'`) made the key `['concept','entity']` not match the stored `['entity','concept']` row — silently rejecting the allowed entity↔concept combination. The `sharedIncoming signal` regression test caught it.

## Six-Gate summary

| Gate | Status | Notes |
|------|--------|-------|
| 1. Code correct | ✅ | lint 0/0 · tsc 0 · **3020/3020 tests** · build clean · css 0 |
| 2. No side effects | ✅ | No settings schema change, no migration, no i18n key rename. fetchErrorNetwork value changed (text content only) — 10 locales. |
| 3. No breaking | ✅ | All signatures preserved (helper additions only). |
| 4. No perf regression | ✅ | Same call count; status regex matching is O(n) on string length; cross-type filter is O(1) Set lookup. |
| 5. Docs complete | ✅ | CHANGELOG +1.26.3 entry, ROADMAP Current Status updated. |
| 6. Release clean | ⏳ | Pre-release-gate v2.0.0 + doc-review will run before v1.26.3 ships. |

## Stats

- **4 commits** (B1 + B2 + B3 + docs)
- **22 files changed** / **+574 / -57**
- **+24 new tests** (4 fetchModels status re-throw + 11 i18n-parity fallback hints + 5 status-bar compose + 4 dedup cross-type)
- **3020 tests passing** (was 2996 on main)
- 0 lint warnings / 0 tsc errors / 0 css violations

## Co-numbering with PR #447

Both branches ship under v1.26.3 PATCH but on separate PRs. PR #447 handles JSON schema / 3-tier state machine / 11 caller migrations. This PR handles UI / lint. They do not overlap on any file or test. Review each independently.


DocTpoint (changes requested):
Reviewed at `c8e50b2`, by reading the three call paths end to end against `main`. Scope, so it is not read wider than it is: this is a static review. I did not build the bundle and did not run the suite — the machine I am on today has no build environment for this repo. B3 is correct in substance and I can add independent evidence for it. B1 and B2 each have a defect that the new tests cannot see, because in both cases the test pins the helper while production violates the helper's input assumption.

## B1 — the two unknown cases swap places

`fetchOneUrl` no longer has any `return []` path. Every non-success now throws, including a 2xx whose `data` array is absent or empty: the `if (response.status >= 200 && response.status < 300)` block falls through to `throw new Error(\`HTTP ${response.status}: …\`)`. That changes two outcomes that the fix did not intend to touch.

| case | before | after | correct |
|---|---|---|---|
| 401 / 403 / 404 / 5xx | Network | **Auth / Endpoint / Server** | ✅ this is the fix, and it works |
| true network failure (DNS, refused, timeout) | Network | **Empty** | ❌ regression |
| 2xx with empty model list | Network | Network | ❌ unchanged; `fetchErrorEmpty` stays unreachable |

Row 2: a DNS failure is wrapped as `Network error: …`, which does not match `/^HTTP \d+/`, so `lastHttpError` stays null, `fetchModelsWithFallback` returns `[]`, and the caller's `if (models.length === 0) throw new Error('empty model list')` fires. `classifyFetchError` matches that string exactly and returns `Empty`. The user with no internet connection is told "Provider has no model list endpoint. Enter a Model ID below."

Row 3: a provider that answers `200 {"data":[]}` — the case `fetchErrorEmpty` was written for — now throws `HTTP 200: {"data":[]}`. That message carries a status the classifier has no branch for: not 401/403, not 404/405/410/421, not `\b5\d\d\b`. It falls through to the default, so the user sees "Network request failed. Check your internet connection, BaseURL, proxy settings, or API Key." That is the same misreporting this PR set out to remove, for the one case that has its own dedicated message.

There is a consequence for the i18n commit: the API-Key hint was added to `fetchErrorNetwork` in ten locales because "true network failures cannot be disambiguated from auth by status alone". After the fix, auth *is* disambiguated by status — and true network failures no longer reach `fetchErrorNetwork` at all (row 2). The hint now lands in a branch reached mainly by unclassifiable status codes.

Both rows come from the same root: the `fetchFn` contract in `url-fallback.ts` still documents "returns … or an empty array if the request failed with no useful status, so the orchestrator can try the next candidate", and the only production implementation can no longer produce that value. The new test `still returns empty array when fetchFn returns []` passes because it stubs a return that production cannot emit.

Minimal fix, two edits:

```ts
// model-section.ts — a 2xx is never an error
if (response.status >= 200 && response.status < 300) {
  const data = response.json as { data?: Array<{ id: string }> };
  return data.data?.map(m => m.id) ?? [];
}
const bodySnippet = (response.text ?? '').slice(0, 200);
throw new Error(`HTTP ${response.status}: ${bodySnippet}`);
```

and in `fetchModelsWithFallback`, remember the wrapped network error in a second variable and re-throw it when no HTTP-status error was seen. Then row 2 lands in `Network` and row 3 in `Empty`, and all three rows are right.

Two smaller notes on the same fix. The classifier now scans a 200-character provider-controlled body with `\b`-anchored regexes, so a 5xx whose body contains "unauthorized" is classified `Auth` — the checks run auth-first. Matching `^HTTP (\d+)` and classifying the captured status, with the keyword regexes as the fallback, removes that whole class. And the comment says the body is truncated "to avoid blowing up the notice message; full body remains in console" — the Notice only ever shows `getText(\`fetchError${category}\`)`, and nothing logs the error, so neither half of that sentence describes what the code does.

## B2 — the label is appended twice on every PDF ingest

`composeStatusBarUpdate` assumes the callback receives raw progress text. One of the two emitters sends text that is already composed.

- `wiki-engine.ts:672` (`ingestPdfSource`, the v1.25.11 #169 path) calls `updateStatusBar(buildIngestStatusBarText(getText(lang,'ingestionStatusBar'), file.basename, undefined, getText(lang, stageKey)))`. That string already ends with the base label.
- `wiki-engine.ts:288` (`notifyProgress`) sends a raw message. Only this one matches the new helper's assumption.

The abort controller is created at `wiki-engine.ts:821`, before `return this.ingestPdfSource(file, opts)` at `:836`, so `isIngesting()` is true throughout the three PDF stages. The callback therefore runs the already-composed string through `buildIngestStatusBarText(label, undefined, undefined, text)`, whose body is `[filename, stage, label].filter(Boolean).join(' · ')` — and the user sees:

```
My Note.pdf · Reading PDF… · Ingesting… click to cancel · Ingesting… click to cancel
```

on every PDF ingest. The unit tests cannot catch this: they call the helper with raw `updateText`, which is exactly the input the failing path does not supply.

The smallest fix that keeps one composition site is at the emitter — pass the segments, not a composed string:

```ts
const setPdfStage = (stageKey: keyof typeof TEXTS.en) =>
  this.updateStatusBar([file.basename, getText(lang, stageKey)].join(' · '));
```

The rendered text stays byte-identical to today's, and composition happens only in `composeStatusBarUpdate`.

One latent property worth naming while this is open: the `null` branch turns "neither flag is set" into "hide the bar". Today both emitters run inside an ingest or lint window, so nothing is lost. But `EngineContext.onProgress` (`wiki-engine.ts:210`) is shared by every subsystem, so the first progress message emitted outside those two windows will silently disappear instead of being displayed. A one-line fallback — show the raw text when neither flag is set, rather than hiding — keeps that from becoming a bug someone has to find twice.

## B3 — correct, with independent evidence

The rule is right and the guard is in the right place: all eight injection sites go through `addCandidate`, and the check sits ahead of both `candidates.set` branches, `sourceFingerprint` included. `generateDuplicateCandidates` has exactly one production caller, so the `'other'` rejection cannot fire behind the file-level filter.

Independent evidence, since the PR argues the rule from `#358` alone: we maintain a separate cross-type duplicate detector for our own vault (2838 pages: 1296 entities, 1120 concepts, 422 sources), written before this PR and against the same data the plugin lints. It encodes the same rule — the source/entity duality of one name is filtered out as by-design, not reported as a duplicate — and it puts source comparison behind an explicit `--include-sources` opt-in. Two implementations that never saw each other arriving at the same partition is a better argument for the spec than either one alone.

Two small things in B3 itself.

The anti-regression comment explains the wrong cause. It says the naive canonicalization used `a <= b`, and that string ordering made the key `concept|entity` miss the stored `entity|concept` row. But the shipped code uses `a < b`, which produces the identical key for two distinct values — for `a === b` both operators produce the same string as well. The operator cannot have caused the miss; what fixed it is the `'concept|entity'` row in `ALLOWED_PAIR_KEYS`. As written, the comment invites the next editor to "simplify" the comparison back and to trust that the Set order does not matter.

The rejection is silent — no counter, no debug line. The filter removes candidates before they are visible anywhere, so if the path convention ever shifts, dedup degrades to a no-op with no signal at all. Counting rejected pairs and emitting the number in the existing dedup debug output makes the filter's effect measurable, and would also give the "polluting every lint run" claim a number.

## Housekeeping

Thirteen code comments across the three fixes are marked "v1.26.4 PATCH follow-up", while the CHANGELOG entry is `[1.26.3]` and ROADMAP says v1.26.3 (manifest on `main` is 1.26.2). Worth normalizing before merge, since these comments are what a future reader dates the change by. The stats block reads 22 files / +574 / −57; the API reports 24 files / +522 / −47 for `c8e50b2`.

green-dalii:
Hi DocTpoint — thanks for the thorough CR. All items are addressed in 5 follow-up commits (`348dd7c..ca44c17`); 3070 tests, lint 0, tsc 0, build clean, and the maintainer's E2E on the production vault passes.

**B1 — both unknown cases fixed (`348dd7c`):**
- Any 2xx now returns `data.data?.map(m => m.id) ?? []` — an empty/absent `data` is a valid "no models" answer, so the `empty model list` → `fetchErrorEmpty` path stays reachable.
- `fetchModelsWithFallback` now tracks `lastNetworkError` alongside `lastHttpError` and re-throws it when no HTTP-status error was seen, so a disconnected user lands on `Network`, not `Empty`. The old `still returns empty array` test was exactly the shell test you flagged — it stubbed a return production can no longer emit; the new `re-throws the last network error` and `prefers HTTP-status over network` tests pin the real contract.
- `classifyFetchError` now matches `^HTTP (\d+)` first and classifies the captured status, with the keyword regexes as fallback — a 5xx body containing "unauthorized" is `Server`, not `Auth`. The stale "full body remains in console" comment is corrected (nothing ever logged it).

**B2 — PDF double label fixed (`85282f2`):** `setPdfStage` emits raw segments (`[filename, stage].join(' · ')`) so composition happens only in `composeStatusBarUpdate`; the rendered text is byte-identical. I deliberately did NOT adopt the null-branch suggestion: the hide-on-null protects post-cancel stale-text, and no `updateStatusBar` call exists outside an ingest/lint window today, so the silent-disappearance case has no reachable path.

**B3 — comment + counting fixed (`a6bf4a8`):** the anti-regression comment now states the true root cause (the `concept|entity` row must exist in `ALLOWED_PAIR_KEYS`; `a < b` was never the miss), and `hooks.onCrossTypeRejected` surfaces the rejected-pair count in the dedup-phase candidate debug line.

**Housekeeping (`ca44c17`):** all `v1.26.4` markers normalized to `v1.26.3`; CHANGELOG/ROADMAP refreshed to the final 10-commit state.

Separately, the maintainer's E2E found three remaining hardcoded-English Toasts — those are localized too (`7ae3399`).

DocTpoint (approved):
**Review at `ca44c17`** — read `c8e50b2..ca44c17` against the head files. Note this range includes `9efeb7a`, which was outside the SHA my previous CR named. This approve supersedes my `CHANGES_REQUESTED` at `c8e50b2`; it covers the three CR items and the diff above, not the parts I list as unverified. **Not verified:** I did not run the suite, lint, tsc, or an E2E — the 3070 / 0 / 0 numbers are yours.

All three items check out against the code, not just the description:

- **B1** — `?? []` on every 2xx, `lastNetworkError` tracked and thrown after `lastHttpError`, `classifyFetchError` matching `^HTTP\s+(\d{3})` before the keyword scan. Both new tests drive real rejections; the shell test is gone.
- **B2** — the rendered text is byte-identical to the v1.25.11 format: `composeStatusBarUpdate` produces `updateText · label`, the emitter now sends `filename · stage`, so the bar shows `filename · stage · label`. `setStatusBarUpdateCallback` has exactly one registration, so composition really does happen in one place. **Your reason for keeping the hide-on-null is correct** — the abort controller is created at `wiki-engine.ts:821` and the PDF branch is entered at `:835`, so `isIngesting()` is true for all three stages.
- **B3** — `ALLOWED_PAIR_KEYS` carries `'concept|entity'` and `a < b` yields the same key from either side; the corrected comment matches the code.
- **i18n** — 21 keys across 11 locales, and every placeholder present in every locale.

Three follow-ups, none of them blocking:

**1. The rejected-pair count is invisible in exactly the case it was added for.** `dedup-phase.ts:213` returns on `allCandidates.length === 0` with "wiki is clean", before the debug line at `:223` that carries `crossTypeRejected`. A path-convention drift rejects every pair, which yields zero candidates, which takes the early return — the number never prints. Your own new unit test builds that state (`candidates` empty, `rejected > 0`). Carrying the suffix into the early-return line covers it. Minor second point: rejections are counted per `addCandidate` call while accepted candidates are deduplicated by pair key, so the two numbers on that line are not on the same scale.

**2. A 2xx empty-list answer is discarded and loses to errors from sibling paths.** `fetchModelsWithFallback` remembers only errors — an empty `[]` from a 2xx leaves no trace. For a provider that answers a wrong path with 404 while the correct path answers 2xx with an empty list, the final throw is the 404 → `Endpoint` → "no models endpoint", although the endpoint answered and simply has no models loaded. A minimal fix is to remember the successful-but-empty observation and prefer it, since a 2xx proves reachability and an accepted key, whereas a 404 on a sibling path proves only that the sibling path is wrong:

```ts
let sawEmptySuccess = false;
// … in both loops, after `if (models.length > 0) { … return models; }`
sawEmptySuccess = true;

// … before the final throws
if (sawEmptySuccess) return [];      // → 'empty model list' → Empty
if (lastHttpError) throw lastHttpError;
if (lastNetworkError) throw lastNetworkError;
```

One data point you may not have: I wanted to demonstrate this with LM Studio and measured instead that it does **not** apply there. LM Studio 0.4.20 answers `GET /models` with **HTTP 200** and `{"error":"Unexpected endpoint or method. (GET /models)"}`, not 404. That means your `?? []` incidentally fixes a second LM Studio bug: before it, that body took the `throw new Error("HTTP 200: …")` path and every LM Studio user with a base URL lacking `/v1` was told "Network". I have not measured Ollama.

**3. Nit:** the new `.replace('{name}', value)` call sites put vault-derived strings into the *replacement* argument, where JavaScript interprets `$&`, `` $` ``, `$'` and `$$`. A page named `Cash $& Carry` renders mangled. These were template literals before. `.replace('{name}', () => value)` suppresses the pattern handling.

Finally, the branch is `CONFLICTING`, 44 commits behind `main` since #447 merged. The overlap is `conversation-ingest.ts`, `source-analyzer.ts`, `dedup-phase.ts` (and its test), plus CHANGELOG and ROADMAP. Since #447 reworked the call sites in the first two, the i18n emissions there will need to be re-applied rather than merged.

Issue #451, opened 2026-08-12
wrapWithAdvancedSettings: createMessageStream drops all settings (temperature/top_p/seed/repetitionPenalty/enableThinking)


green-dalii:
## Background

`src/llm-client-wrapper.ts:50-49` uses `Object.create(client)` to inherit `createMessageStream` from the underlying LLM client. The wrapper overrides `createMessage` and `createMessageWithOutput` to inject settings (temperature, top_p, seed, repetitionPenalty, enableThinking), but `createMessageStream` is inherited via the prototype chain with **no settings injection**. Result: every setting that flows through `wrapWithAdvancedSettings` is silently dropped on the stream path.

## Discovery path

Surfaced during Issue #414 (repetitionPenalty) triage on 2026-08-12:
- `wrapWithAdvancedSettings` injects `repetition_penalty: settings.repetitionPenalty` into `createMessage` callers (verified at `src/llm-client-wrapper.ts:76, 102`).
- `OpenAICompatSdkClient.createMessageStream` does NOT receive the wrapped `repetition_penalty` (no `task` label, no sampling knobs) — only the prototype-inherited bare method.
- DocTpoint's #414 type-error methodology (which works on non-stream calls) cannot diagnose the bug because the field never reaches `buildProviderOptions` for stream callers.

## Affected settings (5)

| Setting | Stream path impact |
|---|---|
| `temperature` | Default used; user's Custom Advanced setting ignored |
| `top_p` | Default used; user's setting ignored |
| `seed` | Not honored — non-reproducible streams even with seed set |
| `repetitionPenalty` | Default used; setting ignored (Issue #414 only fixed non-stream path) |
| `enableThinking` (force-disable thinking) | Default applied; reasoning budget not bounded |

## Affected call sites

Stream is the primary path for Query Wiki + streaming UI. Any user setting the LLM Advanced section's `temperature` / `top_p` / `seed` / `repetitionPenalty` and using Query Wiki sees no effect.

## In-scope fix vs out-of-scope

This is a **separate fix** from #414 (which landed in v1.26.3 PATCH). v1.26.3 already shipped the non-stream fix; this issue tracks the stream path so a future contributor can land it without rediscovering the root cause.

Possible fixes (out of scope for this issue — design discussion):

[...]

Issue #402, opened 2026-08-03
providerOptions on the openai-compatible provider: 4 of the 5 fields are stripped before the request is built


DocTpoint:
`buildProviderOptions` in `src/llm-sdk/openai-compat-sdk-client.ts` emits five keys under `providerOptions.openaiCompatible`: `thinking`, `chat_template_kwargs`, `repetitionPenalty`, `response_format` and (after a local patch) `reasoningEffort`.

`@ai-sdk/openai-compatible` validates that object against a plain zod schema (`dist/index.mjs:322`) which knows exactly four keys — `user`, `reasoningEffort`, `textVerbosity`, `strictJsonSchema`. Zod strips everything else, so four of the five never reach the wire.

Measured against LM Studio (gemma-4-12b) with a fetch interceptor reading the outgoing body:

| sent via providerOptions | key present in request body | reasoning tokens |
|---|---|---|
| `reasoningEffort: 'none'` | `reasoning_effort: "none"` | 0 |
| `thinking` + `chat_template_kwargs` | neither | 92 |
| nothing (control) | — | 61 |
| `repetitionPenalty: 1.1` | absent | — |
| `response_format: {type:'json_object'}` | absent | — |

Consequences on every openai-compatible backend (LM Studio, llama.cpp, vLLM):

- the #99 `disableThinking` toggle is inert — not because the backend ignores the dialects, but because they are removed before the request is assembled;
- the #128 `repetitionPenalty` setting has no effect;
- JSON mode is not requested for calls that ask for it.

A fix already exists in the same file: `getProvider` passes `transformRequestBody` to `createOpenAICompatible` for the `max_tokens` → `max_completion_tokens` swap. Fields outside the SDK schema belong there rather than in `providerOptions`. `reasoningEffort` is the exception — the SDK maps it to `reasoning_effort` itself (`dist/index.mjs:541`), and on LM Studio `reasoning_effort: "none"` is the one dialect that actually silences the thinking channel.

Environment: plugin built from tag 1.25.10, `@ai-sdk/openai-compatible` 2.0.62, `ai` 6.0.230, LM Studio, macOS.

DocTpoint:
Correction to my own suggestion above, before anyone acts on it. Routing `response_format` through `transformRequestBody` would break LM Studio.

Checked directly against the server:

```
POST /v1/chat/completions  {"response_format": {"type": "json_object"}, ...}
[...]

PR #453, opened 2026-08-12
fix(sdk): repetitionPenalty per-backend dialect dispatch (#414)

### LINKED ISSUE #414: OpenAI-compatible provider: `repetitionPenalty` setting has no effect (wire-level fix needed)
## Background

Split from #402 (now closed). `response_format` was the easy part — `ca4a24d` (eucher, 2026-07-29) removed it from the wire path because LM Studio rejects `{"type":"json_object"}` outright (`response_format.type` must be `json_schema` or `text` per OpenAI's Chat Completions spec, which LM Studio implements). That removal changes nothing on the wire because under the wrong `providerOptions.openaiCompatible` key the field never reached the server anyway.

## What remains

`repetitionPenalty` (a.k.a. the #128 user setting) is still emitted at `src/llm-sdk/openai-compat-sdk-client.ts:417`:

```ts
openaiOpts.repetition_penalty = opts.repetitionPenalty;
```

under the `"openaiCompatible"` key in the returned `providerOptions`. The SDK's zod schema (`@ai-sdk/openai-compatible@2.0.62/dist/index.mjs:322-344`) only knows `reasoningEffort`, `textVerbosity`, `strictJsonSchema`, and `user` for chat options — everything else is filtered out. Net effect: every LM Studio / llama.cpp / vLLM backend silently ignores the `repetitionPenalty` user setting.

## Why it's harder than `response_format`

Two viable paths, neither trivially safe:

1. **`transformRequestBody` route** — same mechanism used for the `max_tokens` → `max_completion_tokens` swap at line 114-127. DocTpoint verified (Issue #402 comment 1, 2026-08-03 13:35 UTC) that routing `response_format` through `transformRequestBody` breaks LM Studio. `repetition_penalty` is a different field with different backend compatibility — needs a fresh measurement, not a parallel assumption.

2. **Switch the wrapper key** to `this.providerOptionsName` (i.e. `this.provider`, the actual provider id — `deepseek` / `kimi` / `lmstudio` / `custom` / `ollama`). This makes the SDK's `filter()` passthrough path (line 533-540) actually find the fields. **Caveat**: the same passthrough path that would resurrect `thinking` and `chat_template_kwargs` — fields we explicitly want stripped for #411's force-disable-thinking fix. Per-id key correction is on the v1.26.x PATCH track ([[project_v1_26_0_rescoped]]) as a separate dependency.

## Why I'm not opening a fix PR immediately

(1) needs per-backend compatibility measurement (LM Studio / Ollama / DeepSeek / Kimi / GLM each handle `repetition_penalty` differently). (2) collides with the per-id key correction deferred from v1.26.0 Batch 6 — would re-introduce the `thinking` / `chat_template_kwargs` strip-back if the v1.26.x PATCH fix is wrong. Both paths need design review before code.

## Acceptance criteria for the fix

- `repetitionPenalty: 1.1` produces measurably different output vs `repetitionPenalty: 1.0` on a real LM Studio / gemma-4-12b run (output token repetition rate at the prompt tail is the standard signal)
- The fix does not regress #411's force-disable thinking (reasoning tokens remain 0 on a reasoning-capable model with the disable toggle on)
- The fix does not regress `response_format` handling (ca4a24d stays correct — `response_format` is not on the wire)
- A wire-body regression test (`openai-compat-request-body.test.ts` style) asserts `repetition_penalty: <value>` is on the request body when the setting is set

## Estimated scope

3-5 days depending on chosen path. Path (2) is preferable if the per-id key correction lands first (tracked separately); otherwise path (1) is the standalone option.

## References

- #402 (closed) — original report
- #411 (Batch 6) — force-disable thinking 4-layer fallback, same zod-strip mechanism
- ca4a24d — `response_format` removal
- `src/llm-sdk/openai-compat-sdk-client.ts:343-417` — current `buildProviderOptions`

green-dalii:
Closes #414.

`repetitionPenalty` user setting has been a silent no-op on every shipped provider since the v1.23.0 AI SDK migration dropped the pre-AI-SDK `unsupportedFields` blocklist. This PR fixes the non-stream path; the stream path is tracked separately as #451.

## Root cause

`OpenAICompatSdkClient.buildProviderOptions` returned `providerOptions` under the hardcoded `"openaiCompatible"` key, but the AI SDK's openai-compat passthrough at `@ai-sdk/openai-compatible@2.0.62/dist/index.mjs:525-540` reads from `providerOptions[this.providerOptionsName]` (the provider id — `lmstudio` / `kimi` / etc.). The key mismatch meant the lookup missed for every provider, so the field never reached the wire. Anthropic received the field under `providerOptions.anthropic.repetitionPenalty` but its Messages API has no `repetition_penalty` (only `temperature` / `top_p` / `top_k`); Claude silently ignored it.

## Fix

In `OpenAICompatSdkClient.buildProviderOptions`:

1. **Switch the return key** from `{openaiCompatible}` to `{[this.provider]}`. The SDK's per-id-key passthrough then delivers the field to the wire. Same mechanism AI SDK uses for `reasoningEffort`, which is unaffected (it travels the zod-whitelist path at `dist/index.mjs:531-541`, NOT the passthrough).

2. **Per-provider dialect dispatch** (the spelling each backend recognizes), with `repetitionPenaltyWireField(provider)` exposed as a module-level helper for direct unit-test coverage:

   | Provider id | Wire field name | Reason |
   |---|---|---|
   | `lmstudio` / `ollama` | `repeat_penalty` (no `-ion`) | llama.cpp dialect; verified by DocTpoint #414 type-error test on LM Studio / gemma-4-12b |
   | `kimi` / `openrouter` / `custom` | `repetition_penalty` (snake_case) | OpenAI-spec dialect; Moonshot / OpenRouter docs support; vLLM SamplingParams accepts |
   | `deepseek` / `gemini` / `minimax` / `glm` / `bedrock-openai` / unknown | (drop) | API does not document the field |

   The `bedrock-openai` default is conservative — the bedrock-mantle endpoint is OpenAI-compat but the underlying AWS Bedrock models' openai-compat layer is not documented to forward `repetition_penalty`. Revert to "drop" only if user reports land that the layer does accept it.

3. **Hoist** the dialect lookup into a `private readonly` field computed once in the constructor (`this.provider` is immutable per client lifetime). One-line `[REPETITION-PENALTY-EMIT]` `console.debug` on the send path including `baseURL=` for cross-`custom`-provider disambiguation, mirroring `[REASONING-STRIP-DEBUG]`.

4. **Anthropic client** drops the field entirely from `buildProviderOptions` (was being placed on the wire under `providerOptions.anthropic.repetitionPenalty` despite the API not accepting it). Matches the 10-locale i18n text *"cloud providers will silently ignore it"*.

### No 400-strip retry for `repetitionPenalty`

The setting is user-opt-in (not a plugin default), so a backend rejection should surface to the user rather than be silently swallowed (dead-code-as-docs policy + half-life rule). Add the retry only if a pattern of "users frequently 400 because they set repetitionPenalty" emerges. The pattern for adding it is `reasoning-strip-probe.ts`.

### Known limitation (NOT fixed in this PR)

`wrapWithAdvancedSettings` (`src/llm-client-wrapper.ts`) uses `Object.create(client)` to inherit `createMessageStream` without settings injection — `repetitionPenalty` (and all other settings: temperature, top_p, seed, enableThinking) is silently dropped on the stream path (Query Wiki, streaming UI). Tracked as #451 for v1.27.0. The comment block on the wrapper has been updated to make the foot-gun explicit rather than claiming it was intentional design.

## Tests

- 5 dialect dispatch wire-body tests (`lmstudio` / `ollama` / `kimi` / `openrouter` / `custom` / `deepseek` + undefined omission)
- 4 boundary value tests (`0` / `1` / `<1`) — pins "no client-side transformation" contract
- 4 module-level helper unit tests (table lookups + unknown provider + undefined defensive)
- 1 co-assertion in existing `reasoningEffort` test (`custom` must not co-emit `repeat_penalty`)
- 1 existing test flipped to use per-id key (`deepseek` providerOptions shape)
- 2 anthropic drop tests (createMessage + createMessageStream)
- Updated `wrapWithAdvancedSettings` comment to make the stream-path bug explicit

## Gate 4 — Performance

| Dim | Status | Notes |
|---|---|---|
| CPU | ✅ | One additional `if (wireField !== null)` per `buildProviderOptions` call (already gated by `opts.repetitionPenalty !== undefined`); zero on the path users take without the setting |
| Memory | ✅ | Dialect field is `private readonly` computed once at construction; dispatch table is module-level constant; no new closures |
| IO | ✅ | No new vault or network IO; helper is pure |
| Network | ✅ | Zero LLM call count change; one extra small numeric field on supported providers' wire body (negligible) |
| Token | ✅ | No new prompt tokens; wire body gains a single numeric field on supported providers |

## Backward compatibility

`custom` provider users **gain** expected behavior — `repetitionPenalty` now reaches the wire instead of being silently dropped. `bedrock-openai` / `deepseek` / `gemini` / `minimax` / `glm` users see no change (was no-op, remains no-op). Anthropic users see no change at the API level (was ignored, now dropped before emit — both functionally equivalent). Matches the i18n contract already promised in all 10 locale `repetitionPenaltyDesc` strings.

## Files changed

- `src/llm-sdk/openai-compat-sdk-client.ts` — key swap + `Record`-backed dialect helper + hoist + log
- `src/llm-sdk/anthropic-sdk-client.ts` — drop field entirely
- `src/llm-client-wrapper.ts` — comment update (stream-path foot-gun noted)
- `src/__tests__/llm-sdk/openai-compat-request-body.test.ts` — 9 dialect + boundary tests
- `src/__tests__/llm-sdk/openai-compat-sdk-client.test.ts` — 5 helper unit tests
- `src/__tests__/llm-sdk/anthropic-sdk-client.test.ts` — 2 drop tests
- `CHANGELOG.md` — `[Unreleased]` Fixed entry
- `ROADMAP.md` — `#414` → Shipped; `#451` linked

3 commits, Gate 1 green (lint 0/0 + tsc 0 + 3216 tests passing on `openai-compat-request-body` + `openai-compat-sdk-client` + `anthropic-sdk-client` + 4 helper unit tests; 230 test files).

## References

- Issue #414 — original bug report (split from #402 by the maintainer)
- DocTpoint #414 comment (2026-08-07) — type-error 400 discriminator on LM Studio / gemma-4-12b proving `repeat_penalty` (no `-ion`) is the right spelling for llama.cpp dialect
- `@ai-sdk/openai-compatible@2.0.62/dist/index.mjs:525-540` — SDK passthrough that reads `providerOptions[this.providerOptionsName]`
- `[feedback_force_disable_thinking_openai_compat_noop]` (memory) — same key-mismatch mechanism; #414 fixes it for `repetitionPenalty` specifically
- Issue #451 — stream-path settings-injection bug, tracked but not fixed in this PR


DocTpoint (commented):
**Review of `22a3b41` (`fix/414-repetition-penalty-dialect`)**

Scope: I read the dialect dispatch, the Anthropic drop and the test changes. I did not run the suite, and I did not re-measure at the wire today. The measurement below is the one from #414 (2026-08-07, LM Studio 0.4.20, gemma-4-12b, MLX runtime, 4bit).

**One substantive item: `custom` → `repetition_penalty`.**

Your source comment states that llama.cpp accepts `repeat_penalty`. There is no `llama.cpp` provider id, so those users configure the plugin as `custom` — and `custom` maps to `repetition_penalty`. They receive exactly the spelling this PR documents as not accepted by their backend. The same applies to LM Studio users who run on a non-default host or port and therefore choose `custom` instead of `lmstudio`.

My #414 measurement was taken against LM Studio, which accepts `repeat_penalty`. At valid values every field returned HTTP 200; only the effect on the output separated them:

| field in body | HTTP | effect vs. baseline |
|---|---|---|
| `repeat_penalty: 1.8` | 200 | changes the output, reproduced twice |
| `repetition_penalty: 1.8` | 200 | none |
| invented field name | 200 | none |

So for these users the setting stays a no-op after the fix, while the CHANGELOG says `custom` users "gain expected behavior". This is not a regression — it was a no-op before — but the claim is stronger than the evidence supports.

`custom` carries no dialect information by definition. Your own conservative default for `bedrock-openai` ("not documented to forward it → drop") applies to `custom` with more force. Two options I see: drop the field for `custom`, or let the user select the dialect when the provider is `custom`. With that one change I approve immediately — the rest of the PR is sound.

**One documentation item, non-blocking.** The test comment cites the type-error probe (400 for `repeat_penalty: "abc"`, 200 for `repetition_penalty: "abc"`). That is accurate, but as "dialect evidence" it invites a rule that does not hold: a 400 on a bad type proves the field is *known*, not that it *works*. On the same server `seed` is type-validated (`seed: "abc"` → 400) and still has no effect (#423). The effect table above is what establishes the dialect. Citing it would protect the next person who extends the table.

**Confirmed at the code:** the per-id key is the right root cause and matches your own source comment. Dropping the field in the Anthropic client is correct — the Messages API has no such parameter. `reasoningEffort` travelling the zod-whitelist path and staying unaffected is consistent with what I measured at the wire under every provider id.

PR #454, opened 2026-08-13
fix(#443): catch empty-object placeholder variants + repetitionPenalty UX hints

green-dalii:
## What

Two small v1.26.3 fixes verified together on LM Studio / qwen3.5-9b (E2E 2026-08-13).

### 1. Placeholder detector — catch empty-object/empty-array variants (#443 follow-up)

`isPlaceholderJsonText` only matched the `{"": ""}` empty-string shape. On a small local model, grammar-constrained extraction with a custom repetition penalty above 1.0 can also emit `{"": {}}` and `{"": []}`. The detector now treats empty objects/arrays as empty values, so the SDK-layer placeholder → text_prompt demotion fires for all four shapes instead of shipping a corrupted object.

Files: `src/core/json.ts`, `src/__tests__/core/json.test.ts`, `src/__tests__/core/json-thinking-block-fallback.test.ts`

### 2. repetitionPenalty UX — name the failure + warn in settings (10 locales)

When source analysis fails and the user has set a custom repetition penalty, the thrown error now appends a localized hint pointing at the likely cause (`buildRepetitionPenaltyHint`): values above 1.0 can break grammar-constrained extraction on small local models. The settings description for `repetitionPenalty` is extended with the same guidance in all 10 locales.

Files: `src/core/repetition-penalty-hint.ts` (new), `src/wiki/wiki-engine.ts`, 10 × `src/texts/*.ts`, `src/__tests__/core/repetition-penalty-hint.test.ts` (new)

## Gate 4: Performance

| Dim | Status | Notes |
|-----|--------|-------|
| CPU | ✅ | one extra string concat in an error path |
| Memory | ✅ | none |
| IO | ✅ | none |
| Network | ✅ | zero LLM call change |
| Token | ✅ | no prompt change |

## Tests

+6 tests (4 hint helper + 2 json placeholder variants); full suite 3301 passing.

Note: this PR does not `Closes` an issue — the #443 wire fix shipped in #447; this is follow-up hardening. The #414 repetitionPenalty dialect fix shipped in #453.


DocTpoint (changes requested):
Reviewed at `de546bc`. Part 1 is correct as written; part 2 has one gating issue I would like to see fixed before merge.

**Part 1 — placeholder variants: verified.**

Checked both consumers of the widened predicate (parse gate at `json.ts:214`, SDK demote at `openai-compat-sdk-client.ts:878`). The widening is confined to objects whose keys are all `''`, so no legitimate payload can change shape; `v === null` short-circuits before the `typeof v === 'object'` branch, and `Object.keys([]).length === 0` covers the array variant as the comment states. Ran the three touched test files locally: 59 passing.

**Part 2 — the hint fires where the value never reaches the wire (blocking).**

`buildRepetitionPenaltyHint` gates on the presence of the setting only, but the plugin already knows which providers put the field on the wire. Driving both functions directly:

```
provider     wireField              hint fires
anthropic    null                   true
openai       null                   true
lmstudio     repeat_penalty         true
ollama       repeat_penalty         true
kimi         repetition_penalty     true
openrouter   repetition_penalty     true
custom       repetition_penalty     true
deepseek     null                   true
gemini       null                   true
minimax      null                   true
glm          null                   true
```

For `anthropic` the value is dropped by design (`anthropic-sdk-client.ts:171-180`); for `deepseek` / `gemini` / `minimax` / `glm` the table drops it (`openai-compat-sdk-client.ts:76-82`). On those five, an ingest failure now ends with "a custom Repetition penalty of 1.5 is set … reduce or clear this setting and retry" — naming a setting that was never sent, in ten languages. That is more expensive than the generic message it extends, because it offers the user a plausible but wrong cause.

It is also the mirror image of the failure class this whole line of work exists to remove. The dispatch table's own comment gives the reason for dropping undocumented fields: to avoid the case where a user sets 1.5 expecting a reduction and the field is silently ignored. An ungated hint produces the same misinformation from the other side — the setting had no effect, and the error message tells the user it did.

The fix is one condition, and the predicate is already exported for reuse:

```ts
if (value === undefined) return '';
if (repetitionPenaltyWireField(provider) === null) return '';
```

with `provider` threaded from `this.settings.provider` at the call site. `openai` is worth a second look: `openai-sdk-client.ts:340` does set the field, but the comment there says the Responses schema strips it and the sibling test cannot tell the difference — if that holds, `openai` belongs on the suppressed side too.

**The `custom` mapping — raised before, and this PR now depends on it.**

In my review of #453 (13 Aug, 04:16 UTC, on `22a3b41`) I flagged that `custom → repetition_penalty` lands on exactly the llama.cpp users your own comment in that thread points at `repeat_penalty`, because llama.cpp has no provider id of its own and falls to `custom`; the same holds for LM Studio on a non-default host or port. #453 merged 77 minutes later on that same commit, unchanged and without a reply to the point, so the mapping is in `main` now.

I am not reopening it here — but this PR builds user-facing text in ten languages on top of that mapping, which raises the cost of leaving it unresolved. On a `custom` provider pointed at llama.cpp, even a correctly gated hint would name a setting the backend ignored. If keeping `repetition_penalty` for `custom` was deliberate, one line saying so — here or in #453 — settles it for me and I will stop raising it.

**Non-blocking:**

1. `repetition-penalty-hint.test.ts:20` — `expect(buildRepetitionPenaltyHint('en', 1.0)).toContain('1.0')` cannot fail, because the fixed text contains "Values above 1.0". The interpolated value is `String(1.0)` → `"1"`, so the message reads "a custom Repetition penalty of 1 is set". The assertion also sits under a test named for the unset case; `toContain('of 1 is set')` inside the interpolation test would pin what it means to pin.
2. `json.ts:153-155` — the docstring says the gate also rejects objects "whose non-empty content count is zero while the object is small enough to be a placeholder (≤1 field)". There is no such branch, only the all-empty-keys one. Pre-existing from #443, but this PR adds a paragraph to that block, so it is a convenient moment to drop the sentence.
3. The hint is appended to every null-analysis failure, including the SSL/network case the same message already names. The provider gate removes most of that overlap.
4. `de.ts` writes "Werte über 1,0" but interpolates "1.5" — a German reader sees both decimal separators in one sentence. Cosmetic.

With the provider gate in, this is an approve from me.

green-dalii:
Thanks for the thorough review. The provider gate is in — the wire-field dialect now lives in `core/repetition-penalty-dialect.ts`, shared by both the SDK client and the hint helper, and `buildRepetitionPenaltyHint` returns `''` when `repetitionPenaltyWireField(provider) === null`, so anthropic / deepseek / gemini / minimax / glm never get a hint naming a setting that wasn't sent.

On the `custom` mapping: keeping `repetition_penalty` there is deliberate — `custom` is the OpenAI-spec-compatible bucket (vLLM and similar), where that is the standard field name; llama.cpp / LM Studio users are expected to select `lmstudio` / `ollama`, which map to `repeat_penalty`. The 1.0 assertion now pins the rendered interpolation (`String(1.0)` → `1`), and the stale "≤1 field" docstring sentence is dropped.

Updated to `8512001` — please re-review.

DocTpoint (approved):
**Re-review at `8512001`.** The `CHANGES_REQUESTED` above was against `de546bc`; the blocking item is closed. Approving.

**The gate holds across the whole id space, not just the five ids from my review.** Driving `buildRepetitionPenaltyHint` and `repetitionPenaltyWireField` directly over every key in `PREDEFINED_PROVIDERS` (15 ids), with value `1.5` and unset:

| hint fires | suppressed |
|---|---|
| `lmstudio`, `ollama` (`repeat_penalty`) · `kimi`, `openrouter`, `custom` (`repetition_penalty`) | `openai`, `openai-codex`, `anthropic`, `gemini`, `deepseek`, `minimax`, `glm`, `bedrock-anthropic`, `bedrock-openai`, `anthropic-compatible` |

An unset value returns `''` on every id, including the five that fire. An id that is not in the table returns `null` and is suppressed, so the safe direction is the default. The re-export from `openai-compat-sdk-client` returns the same value as the core module for all 15 ids, so the two import paths cannot drift apart.

`bedrock-openai` is worth naming explicitly: it constructs `OpenAICompatSdkClient` with `provider: settings.provider` (`create-llm-client.ts:102`), so the id the client emits under and the id the hint reads are the same string. The gate holds on that path too.

63 tests pass in the four touched files at this head.

**`266beb3` is equivalent, and I checked the direction that could have changed.** The two-step form (`keys.every(k => k === '')`, then `values.every(isEmpty)`) and the single-pass conjunction at `json.ts:181` accept exactly the same set, and `entries.length > 0` keeps the legitimate `{}` allowance intact. Hoisting `isEmptyJsonValue` to module scope also stops re-allocating the closure on every parse. The stale "≤1 field" sentence is gone.

`wiki-engine-repetition-penalty-hint.test.ts` pins the wiring at the throw site — that is more than I asked for, and it is the test that will catch a future refactor reading the wrong settings field.

**On `custom`: the decision is yours, you gave the line I asked for, and I am done raising the mapping.**

One factual correction underneath it, then I will leave it alone. "llama.cpp / LM Studio users are expected to select `lmstudio` / `ollama`" is not reachable through the settings UI for an LM Studio instance on a non-default host or port: switching the provider dropdown overwrites `tempSettings.baseUrl` with the provider default (`provider-section.ts:75`), and the Base URL field is only rendered when `baseUrl !== providerConfig.baseUrl` (`:153`). Right after selecting `lmstudio` those are equal, so there is no field to edit, and the only route to a different port is `custom` — where the mapping sends `repetition_penalty`, the spelling the LM Studio server does not read (#414). Hand-editing `data.json` is the only other way.

That is a UI gap and not this PR's business, and it does not change your mapping decision. If you want it tracked, say the word and I will file it as its own issue; exposing the Base URL override for `lmstudio` / `ollama` would close it in one condition.

**Two non-blocking notes.**

- `openai` sits on the suppressed side because `openai-sdk-client.ts:340` writes `openaiOpts.repetition_penalty` and the comment there says the Responses zod schema strips it — while also saying the sibling test asserts the argument handed to the SDK, not the request body. So the suppression rests on a reading rather than a measurement. It is the right default either way; a request-body test would settle it whenever that file gets its own change.
- The `de.ts` separator nit ("Werte über 1,0" alongside an interpolated "1.5") is untouched. Cosmetic, no objection.

*Scope: worktree at `8512001`, vitest 4.1.6 as installed. The probes drive the exported functions directly rather than through a constructed client.*

Issue #456, opened 2026-08-13
Split model configuration does not persist separate models for ingestion, lint, and query


jameses-cyber:
## Bug Description

It's been a while! When using split models for ingestion, linting, and querying, the selected model assignments are not persisted correctly.

The plugin allows separate models to be selected for each operation, but after saving the settings, all split model fields are overwritten with the ingestion model selection.

For example, when configuring:

- Ingestion: `gpt-5-mini`
- Lint: `gpt-5.5`
- Query: `gpt-5.6-sol`

the plugin saves:

- Ingestion: `gpt-5-mini`
- Lint: `gpt-5-mini`
- Query: `gpt-5-mini`

The split model functionality appears to be saving a single shared model value rather than maintaining separate configuration values for each operation.

## Environment

- **Plugin version:** [enter version, e.g. 1.7.17]
- **Obsidian version:** [enter version]
- **OS:** macOS [enter version]
- **Wiki size:** [enter approximate size, e.g. ~500 pages]

## LLM Configuration

- **Provider:** OpenAI
- **Model configuration:** Split models enabled

Configured models:

- **Ingestion:** `gpt-5-mini`
- **Lint:** `gpt-5.5`
- **Query:** `gpt-5.6-sol`

- **API endpoint (if custom):** Default OpenAI API endpoint
- **Local or cloud:** Cloud

## Steps to Reproduce

1. Open Obsidian settings.
2. Navigate to the Karpathy LLM Wiki plugin settings.
3. Enable split models for different operations.
4. Select different models for each operation:
   - Set ingestion model to `gpt-5-mini`
   - Set lint model to `gpt-5.5`
   - Set query model to `gpt-5.6-sol`
5. Save the plugin settings.
6. Close and reopen the settings panel (or restart Obsidian).
7. Review the configured split models.
8. Observe that lint and query models have been replaced with the ingestion model.

## Expected Behavior

The plugin should persist independent model assignments for each split model function.

Expected configuration:

```text
Ingestion: gpt-5-mini
Lint:      gpt-5.5
Query:    gpt-5.6-sol
```

Each operation should use its assigned model when performing ingestion, linting, or querying.

## Actual Behavior

The ingestion model value is copied into all split model fields.

Actual saved configuration:

```text
Ingestion: gpt-5-mini
Lint:      gpt-5-mini
Query:     gpt-5-mini
```

The selected lint and query models are not retained after saving.

Diagnostic information:

[...]

Issue #449, opened 2026-08-12
`cacheBreakpoint` is set on every extraction call, but no client reads it — prompt caching never reaches the wire


DocTpoint:
## What I measured

`cacheBreakpoint` is declared on the `createMessage` parameter type and is set on every source-extraction call, but no LLM client reads it. Measured at `main` today (2026-08-12), by reading the files rather than by running anything:

| Location | Role |
|---|---|
| `src/types.ts:657` | declares `cacheBreakpoint?: number` on `LLMClient.createMessage` |
| `src/wiki/source-analyzer.ts:374` | sets it to `staticPrefix.length` on **every** extraction round |
| `src/llm-sdk/anthropic-sdk-client.ts` (305 lines) | 0 occurrences |
| `src/llm-sdk/openai-compat-sdk-client.ts` (696 lines) | 0 occurrences |
| `src/llm-sdk/openai-sdk-client.ts` (631 lines) | 0 occurrences |
| `src/llm-sdk/openai-codex-sdk-client.ts` (155 lines) | 0 occurrences |
| `src/llm-sdk/create-llm-client.ts` (305 lines) | 0 occurrences |

Across those five files there is also no occurrence of `cache_control` or `cacheControl`. Every `cache` hit in them is URL resolution or the strip-probe cache — nothing related to prompt caching.

Repository-wide, the identifier appears in exactly five places: the declaration, the call site above, one test stub signature (`src/__tests__/wiki/page-factory-core.test.ts:81`), and the CLI (`tools/llm-wiki-cli/src/main.ts:493` plus its README).

## What it means today

No provider ever receives a caching instruction. For OpenAI-compatible providers that is harmless, because prefix caching there is automatic and needs no field. For Anthropic it is not: `cache_control` is explicit, so extraction currently pays full input price on every round, and the value that says where the cache boundary belongs is computed and then dropped.

## Why the boundary is worth something here

The interesting part is that `staticPrefix` is the right place. `source-analyzer.ts:282-283` splits the rendered template at the `{{batch_context}}` marker, and `:317` rebuilds each round as `staticPrefix + batchContext + suffixTemplate`. So across the rounds of one note, `staticPrefix` — the existing-slug list plus the note content — is byte-identical, and the system prompt from `buildSystemPrompt('analyze')` is identical too. That is a real repetition axis, not a hypothetical one.

[...]

PR #460, opened 2026-08-14
fix(lint): defer fillEmptyPage tag constraint to the runtime vocabulary

### LINKED ISSUE #459: `fillEmptyPage` lint prompt still hardcodes the default tag taxonomy, contradicting the runtime vocabulary injection
## What happens

With `tagVocabularyMode: 'custom'`, the "fill empty page" lint fix sends the model two mutually exclusive lists of allowed `tags:` values in the same call.

**System layer** — `buildSystemPrompt()` always appends the *Active Tag Vocabulary (runtime)* section listing the configured vocabulary and ending with "Do NOT invent new types" (`src/wiki/system-prompts.ts:232`). The comment three lines above it calls this layer "the SOLE source of truth for the active tag vocabulary" (Issue #328 Phase 1).

**User layer** — `FIX_PROMPTS.fillEmptyPage` instruction 10 restates the default taxonomy (`src/wiki/prompts/fixes.ts:47`):

> 10. **Tags constraint:** entity pages MUST use tags from: [person, organization, project, product, event, place, other]. Concept pages MUST use tags from: [theory, method, field, phenomenon, standard, term, other]. Never invent new tags outside these lists

Both travel in one request: `src/wiki/lint/fill-empty-page.ts:63` builds the system prompt for task `lint` and sends it alongside the rendered user prompt. When the configured vocabulary shares no value with the defaults, every answer the model can give violates one of the two instructions.

Instruction 10 also contradicts instruction 8 of the same prompt — "Preserve any existing frontmatter fields exactly (type, created, sources, tags, reviewed). Do NOT remove or alter these fields" — for any vault whose existing tags are outside the default taxonomy.

The value reaches disk: the runner writes the model's complete page including frontmatter, and `enforceFrontmatterConstraints` records out-of-vocabulary tags but keeps them (`src/core/frontmatter.ts:689-690`).

## Why this has stayed invisible

In #368 the reporting vault's custom vocabulary was the default list plus `technology` — near-total overlap, so the two layers agreed and no contradiction was observable. It surfaces only for a vocabulary disjoint from the defaults, for example a subject-domain taxonomy (`Biochemie`, `Physiologie`, …) rather than a kind-of-thing taxonomy.

`git log -S "Tags constraint"` dates the line to v1.7.10, before user-configurable vocabularies existed. The #328 Phase 1 migration moved the retag runner to the system layer (#333) and removed the redundant user-layer injection (#332), but did not reach this prompt. It is the only remaining hardcoded tag list under `src/wiki/prompts/`.

## Reproduction

No LLM call needed — both layers are pure functions. On `main` (3fe34f6) the second test fails with `fillEmptyPage hardcodes tag values that a custom vocabulary forbids`:

```ts
import { describe, it, expect } from 'vitest';
import { buildSystemPrompt } from '../../wiki/system-prompts';
import { FIX_PROMPTS } from '../../wiki/prompts/fixes';
import { getActiveEntityTags, getActiveConceptTags } from '../../core/tag-vocab';
import { VALID_ENTITY_TAGS, VALID_CONCEPT_TAGS } from '../../types';
import type { LLMWikiSettings } from '../../types';

function makeSettings(overrides: Partial<LLMWikiSettings> = {}): LLMWikiSettings {
  return {
    wikiFolder: 'wiki',
    wikiLanguage: 'de',
    extractionGranularity: 'standard',
    customEntityTags: '',
    customConceptTags: '',
    tagVocabularyMode: 'default',
    ...overrides,
  } as LLMWikiSettings;
}

const customSettings = makeSettings({
  tagVocabularyMode: 'custom',
  customEntityTags: 'Mineralstoffe, Biochemie, Physiologie, Mikrobiom',
  customConceptTags: 'Mineralstoffe, Biochemie, Physiologie, Mikrobiom',
});

const DEFAULT_TAXONOMY = [...VALID_ENTITY_TAGS, ...VALID_CONCEPT_TAGS];

function linesRestatingTheDefaultTaxonomy(prompt: string): string[] {
  return prompt.split('\n').filter(line => {
    const hits = DEFAULT_TAXONOMY.filter(t => new RegExp(`\\b${t}\\b`).test(line));
    return new Set(hits).size >= 3;
  });
}

describe('lint prompts vs. the active tag vocabulary', () => {
  it('the system layer carries the active vocabulary on the lint task', async () => {
    const system = await buildSystemPrompt(customSettings, async () => undefined, 'lint');
    expect(system).toContain('Active Tag Vocabulary');
    for (const tag of getActiveEntityTags(customSettings)) expect(system).toContain(tag);
    for (const tag of getActiveConceptTags(customSettings)) expect(system).toContain(tag);
  });

  it('no lint prompt restates the default taxonomy as a fixed list', () => {
    for (const [name, prompt] of Object.entries(FIX_PROMPTS)) {
      expect(
        linesRestatingTheDefaultTaxonomy(prompt),
        `${name} hardcodes tag values that a custom vocabulary forbids`,
      ).toEqual([]);
    }
  });
});
```

## Suggested fix

Have instruction 10 defer to the runtime section instead of naming values:

> 10. **Tags constraint:** use only values from the Active Tag Vocabulary given in the system prompt. Never invent new tags outside that list

Dropping instruction 10 entirely would work too, since the system layer already carries the enum for every task including `lint` — the wording above just keeps the "do not invent" reminder in the user prompt. A PR with this change plus the test above follows.

## Environment

Plugin 1.26.3, Obsidian 1.12.x, macOS. Custom vocabulary of 34 subject-domain terms, German-language vault (~2,400 pages). Latent here — no page in the vault carries a default-taxonomy tag, so this runner has not been used at scale.

DocTpoint:
Fixes #459.

`FIX_PROMPTS.fillEmptyPage` instruction 10 named the default tag taxonomy as a fixed list. It is sent in the same request as the *Active Tag Vocabulary (runtime)* section built by `buildSystemPrompt()` (`src/wiki/lint/fill-empty-page.ts:63`), so with `tagVocabularyMode: 'custom'` and a vocabulary disjoint from the defaults, the two layers forbid each other's values and no answer satisfies both. The issue has the full trace.

**Change**

- `src/wiki/prompts/fixes.ts` — instruction 10 now points at the runtime section rather than enumerating values. This follows what #328 Phase 1 established (the runtime layer is the single source of truth) and what #333 already did for the retag runner; the "do not invent" reminder stays in the user prompt.
- `src/__tests__/wiki/lint-prompt-vocab-contradiction.test.ts` — regression test. One case pins the system-layer contract, the other fails when any prompt in `FIX_PROMPTS` puts three or more default-taxonomy values on one line. It fails on `main` naming `fillEmptyPage` and passes here.

Full suite green locally: 233 files, 3312 tests.

**Not changed:** `generation.ts:49/110` mention `product, person, organization` and `theory, method, field` as parenthetical examples next to `{{entity_type}}`/`{{concept_type}}`. Those are illustrative rather than a mandate, and the value itself comes from the extraction, so I left them alone — happy to include them if you would rather have the examples parameterized too.

Close this without hesitation if you would rather solve it differently; the issue stands on its own.

### DIFF
--- a/src/__tests__/wiki/lint-prompt-vocab-contradiction.test.ts
+++ b/src/__tests__/wiki/lint-prompt-vocab-contradiction.test.ts
@@ -1,0 +1,68 @@
+import { describe, it, expect } from 'vitest';
+import { buildSystemPrompt } from '../../wiki/system-prompts';
+import { FIX_PROMPTS } from '../../wiki/prompts/fixes';
+import { getActiveEntityTags, getActiveConceptTags } from '../../core/tag-vocab';
+import { VALID_ENTITY_TAGS, VALID_CONCEPT_TAGS } from '../../types';
+import type { LLMWikiSettings } from '../../types';
+
+// Issue #328 Phase 1 made the runtime injection (buildSystemPrompt ->
+// "Active Tag Vocabulary") the single source of truth for allowed tag values.
+// A lint prompt that restates the default taxonomy re-creates the dual source:
+// with a custom vocabulary the two layers travel in the same call and forbid
+// each other's values, so every answer the model can give violates one of them.
+
+function makeSettings(overrides: Partial<LLMWikiSettings> = {}): LLMWikiSettings {
+  return {
+    wikiFolder: 'wiki',
+    wikiLanguage: 'de',
+    extractionGranularity: 'standard',
+    customEntityTags: '',
+    customConceptTags: '',
+    tagVocabularyMode: 'default',
+    ...overrides,
+  } as LLMWikiSettings;
+}
+
+// A vocabulary that shares no value with the defaults — the case that makes
+// the contradiction observable (see #368, where the custom list was the
+// default list plus one term and therefore agreed with the baked-in copy).
+const customSettings = makeSettings({
+  tagVocabularyMode: 'custom',
+  customEntityTags: 'Mineralstoffe, Biochemie, Physiologie, Mikrobiom',
+  customConceptTags: 'Mineralstoffe, Biochemie, Physiologie, Mikrobiom',
+});
+
+const DEFAULT_TAXONOMY = [...VALID_ENTITY_TAGS, ...VALID_CONCEPT_TAGS];
+
+/** Lines that enumerate three or more default-taxonomy values — a baked-in enum. */
+function linesRestatingTheDefaultTaxonomy(prompt: string): string[] {
+  return prompt.split('\n').filter(line => {
+    const hits = DEFAULT_TAXONOMY.filter(t =>
+      new RegExp(`\\b${t}\\b`).test(line),
+    );
+    return new Set(hits).size >= 3;
+  });
+}
+
+describe('lint prompts vs. the active tag vocabulary', () => {
+  it('the system layer carries the active vocabulary on the lint task', async () => {
+    const system = await buildSystemPrompt(customSettings, async () => undefined, 'lint');
+    expect(system).toBeDefined();
+    expect(system).toContain('Active Tag Vocabulary');
+    for (const tag of getActiveEntityTags(customSettings)) {
+      expect(system).toContain(tag);
+    }
+    for (const tag of getActiveConceptTags(customSettings)) {
+      expect(system).toContain(tag);
+    }
+  });
+
+  it('no lint prompt restates the default taxonomy as a fixed list', () => {
+    for (const [name, prompt] of Object.entries(FIX_PROMPTS)) {
+      expect(
+        linesRestatingTheDefaultTaxonomy(prompt),
+        `${name} hardcodes tag values that a custom vocabulary forbids`,
+      ).toEqual([]);
+    }
+  });
+});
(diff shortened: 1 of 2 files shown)


green-dalii (approved):
## APPROVED — merging to v1.26.4 PATCH alongside #456/#449/#451

@DocTpoint — clean PR, verified against main. Findings:

**Verified in source (not just trusting the PR description):**
- `system-prompts.ts:256` "Active Tag Vocabulary (runtime)" + `:259` "Do NOT invent new types" both exist ✓
- `fill-empty-page.ts:5, 63` both call `buildSystemPrompt` ✓
- Pre-fix `fixes.ts:47` line triggers the new test's enumeration guard (7 distinct defaults on one line ≥ 3 threshold); post-fix has 0 hits ✓
- Precedent #332 + #333 verified as merged ✓

**Scope discipline:** `gh pr diff --name-only` = 2 files (1 prod + 1 test), self-contained. I checked the adjacent occurrences (`generation.ts:49, 110` illustrative parentheticals, `ingestion.ts:42, 54` JSON-schema enum) and agree they're a different bug class — your scope call to leave them alone is correct. `ingestion.ts` is potentially a separate latent issue (extraction-time custom vocabularies) but #328/#333 design explicitly carved extraction out of the runtime-injection path, so it would need its own design discussion, not a side-fix here.

**Six-Gate:** all 6 green (test included, side effects scoped to 1 prompt, no breaking change, no perf regression, no docs churn needed at PR level, conventional commit format).

**Schedule:** merging to v1.26.4 PATCH alongside #456 (settings cascade) + #449 (cacheBreakpoint wire-up) + #451 (stream wrapper) — 4-bug PATCH. Test count 3290 → 3312 (+22 from your test file). Ship target 2026-08-21.

Approved. Will merge once v1.26.4 PATCH branch opens.

PR #462, opened 2026-08-14
fix(settings): preserve per-task model fields on commit (#456)

### LINKED ISSUE #456: Split model configuration does not persist separate models for ingestion, lint, and query

green-dalii:
Fixes #456.

v1.24.1 PATCH Phase 5.5.0 added a belt-and-suspenders cascade inside commitTempSettings (settings.ts:76-78 pre-fix). The intent was to catch Fetch Models / provider-change resets writing tempSettings.model directly, but the same cascade ALSO fires on every regular save when unified model is set, silently clearing per-task fields.

Result: `Model Scope = per-task` users with distinct ingest/lint/query assignments saw all three collapse to the unified value after the first save.

**Fix:** remove the commit-time cascade. The live-edit cascade at `setFieldValue('model', ...)` is the correct UX path; Fetch Models and provider-change already flow through it.

**Verification:**
- Real-class test `settings-commit-no-cascade-456.test.ts` (3 cases) pins the correct behavior
- Deleted obsolete `settings-commit-cascade.test.ts` (mirror test for now-obsolete policy)
- 4-Gate green (lint/tsc/test/build/css-lint)

Test count delta: +3 new, -6 deleted = net -3 (3309 vs 3312 pre-fix).

Closes #456.


DocTpoint (commented):
Reviewed at `1d1396b`.

No objection — this does not touch any surface I have measurements on. Not reviewed in detail.

Issue #452, opened 2026-08-12
The extraction prompt's slug list is sorted, so every ingested note invalidates the prompt cache


DocTpoint:
`buildCompactSlugList` (`src/wiki/source-analyzer.ts:4-17`) emits the existing-page catalog sorted alphabetically, and the catalog sits at the very top of the extraction prompt — `{{existing_slugs}}` is line 7 of `PROMPTS.analyzeSource`, `{{content}}` is line 14.

Every prompt cache in play is a *prefix* cache: it reuses the longest common leading span of tokens and recomputes everything after the first divergence. During a folder ingest each note creates pages — 6.9 per note on average in our vault — and their slugs sort into the middle of the list. From that insertion point on the prefix differs, so the reusable span ends there, for a block that is 91 % of the prompt.

**Measured.** Mac mini M4 Pro / 48 GB, LM Studio 0.4.20+1, `google/gemma-4-26b-a4b-qat` (4-bit), our vault's real list of 2844 slugs = 69,358 chars ≈ 24.5 K prompt tokens. `max_tokens=1`, `temperature=0`, so the measured time is almost entirely prefill:

| list change vs. the cached call | first divergence | time |
|---|---|---|
| none | — | 0.69 s |
| 7 new slugs appended at the end | 100 % | 4.39 s |
| 7 new slugs sorted in (entities only) | 45.6 % | 23.50 s |
| 1 new slug at the front | 0 % | 42.26 s |
| cold call, no shared prefix | — | 39.60 s |

The times track the share of the list *after* the first divergence almost linearly — 54 % remaining × 39 s ≈ 21 s, measured 23.5 s — which is what a prefix cache predicts.

Two consequences:

1. Within one note the reuse already works. Round 2 of the same note costs 1.04 s against 38.45 s for round 1; the loop's second call is essentially free today. Nothing to change there.
2. Across notes it is discarded on every note. In our vault `concepts/` is the first block of the sorted list and every ingest creates concepts, so the first divergence lands early. The 45.6 % row is the optimistic case — it inserts entities only.

For a 415-note ingest that is roughly 415 × ~25 s ≈ 2.9 h of prefill spent re-reading a lookup table that barely changed.

**Two ways to fix it, with a tension worth naming.**

[...]

PR #464, opened 2026-08-14
fix(sdk): wire cacheBreakpoint → cache_control on Anthropic createMessage

### LINKED ISSUE #449: `cacheBreakpoint` is set on every extraction call, but no client reads it — prompt caching never reaches the wire

green-dalii:
## Summary

Issue #449 silent-bug fix: `cacheBreakpoint?` was declared in `LLMClient.createMessage` (`src/types.ts:684`) and set by `source-analyzer.ts:404` (`cacheBreakpoint: staticPrefix.length`), but **0/4 SDK clients read it**. The user's opt-in intent (per-note Anthropic prompt caching) never reached the wire.

## Root Cause

`grep -rn cacheBreakpoint src/llm-sdk/` returns zero hits. The Anthropic SDK client emits `system` as a plain string — never as an AI SDK v6 `SystemModelMessage` with `providerOptions.anthropic.cacheControl`. AI SDK v6 walks the system path (`@ai-sdk/anthropic/dist/index.mjs:2300-2307`) and reads `cacheControl` from the message's `providerOptions`, but our client never set it.

## Fix

Added `buildSystemWithCacheControl(system, cacheBreakpoint)` helper in `src/llm-sdk/anthropic-sdk-client.ts`. Applied in both the primary `createMessage` path AND the URL-fallback retry block (one source of truth, no duplication).

| `cacheBreakpoint` | `system` | Result |
|---|---|---|
| undefined | (any) | `system: undefined` → `call.system` omitted (pre-fix behaviour) |
| defined | empty | `system: undefined` → no stray cache marker on empty string |
| defined | non-empty | `system: [{ role: 'system', content, providerOptions: { anthropic: { cacheControl: { type: 'ephemeral' } } } }]` |

When `cacheBreakpoint` is undefined, behaviour is **identical** to pre-fix: the `...(system !== undefined ? { system } : {})` spread keeps `call.system` a plain string (verified by the existing 4 system-role regression tests).

## Out of Scope

- **openai-compat / OpenAI / OpenAI Codex**: AI SDK does not expose cache hooks. Documented as intentional no-op (matches ROADMAP "Direction 1, per-note caching only").
- **Stream path**: `cacheBreakpoint` is not in `LLMClient.createMessageStream` interface signature; no production caller streams with it. Tracked separately.

## Tests

- 4 new assertions in `src/__tests__/llm-sdk/anthropic-sdk-client.test.ts` (Issue #449 describe block)
- Wire-body assertions: `cacheBreakpoint=42` → `call.system` is array of SystemModelMessage carrying `cacheControl`; `cacheBreakpoint=undefined` → `call.system` stays plain string; co-emit with `enableThinking: false` (no key collision)
- Total: 3290 → 3316 tests (+26 net across the v1.26.4 PATCH window including earlier fixes)
- All Gate 1 checks (lint / tsc / test / build / css-lint) clean

## Gate 4 Performance

| Dim | Status | Notes |
|---|---|---|
| CPU | N/A | Single ternary per call |
| Memory | N/A | One small `SystemModelMessage` literal per call (< 100 bytes) |
| IO | N/A | No vault operations |
| Network | ✅ | Anthropic prompt cache hit: -90% input token cost; first-write overhead negligible |
| Token | ✅ | Net per-call: ~0; cache hit: input token savings ≈ system prompt length |

## Files Changed

```
src/__tests__/llm-sdk/anthropic-sdk-client.test.ts | +98
src/llm-sdk/anthropic-sdk-client.ts                | +57 -3
```

Closes #449

### DIFF
--- a/src/__tests__/llm-sdk/anthropic-sdk-client.test.ts
+++ b/src/__tests__/llm-sdk/anthropic-sdk-client.test.ts
@@ -225,6 +225,104 @@ describe('AnthropicSdkClient', () => {
     });
   });
 
+  // Issue #449: cacheBreakpoint is a typed field (src/types.ts:684
+  // `cacheBreakpoint?: number`) on LLMClient.createMessage and is
+  // SET by source-analyzer.ts:404 (`cacheBreakpoint: staticPrefix.length`)
+  // but ZERO SDK clients read it. The user's opt-in intent (per-note
+  // caching) never reaches the wire.
+  //
+  // Anthropic-side wire shape: providerOptions.anthropic.cacheControl
+  // sits on a SYSTEM text block, not at top-level. The AI SDK's
+  // streamText/generateText accept `system` as either a string or
+  // an array of content parts; to emit cache_control we must switch
+  // the system shape to a single text part carrying the cacheControl
+  // providerOptions. Absent cacheBreakpoint → keep the existing
+  // string-only path (no behaviour change for callers that don't opt in).
+  describe('Issue #449: cacheBreakpoint → cache_control wire-up', () => {
+    function asSystemBlocks(call: Record<string, unknown>): Array<{ role: string; content: string; providerOptions?: { anthropic?: { cacheControl?: unknown } } }> | string | undefined {
+      return call.system as never;
+    }
+
+    it('emits cache_control on system block when cacheBreakpoint is defined (createMessage)', async () => {
+      const client = new AnthropicSdkClient({ apiKey: 'sk-ant-test' });
+      await client.createMessage({
+        model: 'claude-sonnet-4-5',
+        max_tokens: 100,
+        system: 'You are Claude, a helpful assistant.',
+        messages: [{ role: 'user', content: 'hi' }],
+        cacheBreakpoint: 42,
+      });
+
+      const call = mockGenerateText.mock.calls[0][0] as Record<string, unknown>;
+      const system = asSystemBlocks(call);
+      // System must now be a SystemModelMessage[] (one entry carrying
+      // providerOptions.anthropic.cacheControl). AI SDK v6 type is
+      // `string | SystemModelMessage | SystemModelMessage[]`.
+      expect(Array.isArray(system)).toBe(true);
+      const blocks = system as Array<{ role: string; content: string; providerOptions?: { anthropic?: { cacheControl?: unknown } } }>;
+      expect(blocks).toHaveLength(1);
+      expect(blocks[0].role).toBe('system');
+      expect(blocks[0].content).toBe('You are Claude, a helpful assistant.');
+      expect(blocks[0].providerOptions).toEqual({
+        anthropic: { cacheControl: { type: 'ephemeral' } },
+      });
+    });
+
+    it('keeps system as plain string when cacheBreakpoint is undefined (no behaviour change)', async () => {
+      const client = new AnthropicSdkClient({ apiKey: 'sk-ant-test' });
+      await client.createMessage({
+        model: 'claude-sonnet-4-5',
+        max_tokens: 100,
+        system: 'You are Claude, a helpful assistant.',
+        messages: [{ role: 'user', content: 'hi' }],
+      });
+
+      const call = mockGenerateText.mock.calls[0][0] as Record<string, unknown>;
+      // Regression: callers that don't opt in must NOT see the new
+      // array shape — preserves all existing assertions on
+      // `expect(call.system).toBe('...')`.
+      expect(call.system).toBe('You are Claude, a helpful assistant.');
+    });
+
+    it('omits system entirely when cacheBreakpoint is undefined and system is undefined', async () => {
+      const client = new AnthropicSdkClient({ apiKey: 'sk-ant-test' });
+      await client.createMessage({
+        model: 'claude-sonnet-4-5',
+        max_tokens: 100,
+        messages: [{ role: 'user', content: 'hi' }],
+      });
+
+      const call = mockGenerateText.mock.calls[0][0] as Record<string, unknown>;
+      expect('system' in call ? call.system : undefined).toBeUndefined();
+    });
+
+    it('co-emits cache_control alongside enableThinking=false (no key collision in providerOptions)', async () => {
+      const client = new AnthropicSdkClient({ apiKey: 'sk-ant-test' });
+      await client.createMessage({
+        model: 'claude-sonnet-4-5',
+        max_tokens: 100,
+        system: 'You are Claude, a helpful assistant.',
+        messages: [{ role: 'user', content: 'hi' }],
+        cacheBreakpoint: 42,
+        enableThinking: false,
+      });
+
+      const call = mockGenerateText.mock.calls[0][0] as Record<string, unknown>;
+      // Thinking control travels in providerOptions.anthropic.thinking
+      // (existing path); cache_control travels on the system block.
+      // They MUST NOT collide — verify both are present and at their
+      // canonical locations.
+      const system = asSystemBlocks(call) as Array<{ providerOptions?: { anthropic?: { cacheControl?: unknown } } }>;
+      expect(Array.isArray(system)).toBe(true);
+      expect(system[0].providerOptions).toEqual({
+        anthropic: { cacheControl: { type: 'ephemeral' } },
+      });
+      expect(call.providerOptions).toEqual({
+        anthropic: { thinking: { type: 'disabled' } },
+      });
+    });
+  });
+
   describe('custom baseURL for Anthropic-compatible providers (Coding Plan / z.ai / GLM-Antropic)', () => {
     beforeEach(() => {
       mockCreateAnthropic.mockClear();
--- a/src/__tests__/llm-sdk/anthropic-sdk-client.test.ts
+++ b/src/__tests__/llm-sdk/anthropic-sdk-client.test.ts
@@ -269,6 +269,11 @@ describe('AnthropicSdkClient', () => {
     });
 
     it('keeps system as plain string when cacheBreakpoint is undefined (no behaviour change)', async () => {
+      // Regression guard for the no-op path. Pre-existing #147 block already
+      // asserts the plain-string shape when system is provided and no
+      // cacheBreakpoint is set; we re-pin the invariant inside the #449
+      // block so a refactor that re-introduces an array shape for the
+      // undefined-cacheBreakpoint branch is caught here too.
       const client = new AnthropicSdkClient({ apiKey: 'sk-ant-test' });
       await client.createMessage({
         model: 'claude-sonnet-4-5',
@@ -278,22 +283,7 @@ describe('AnthropicSdkClient', () => {
       });
 
       const call = mockGenerateText.mock.calls[0][0] as Record<string, unknown>;
-      // Regression: callers that don't opt in must NOT see the new
-      // array shape — preserves all existing assertions on
-      // `expect(call.system).toBe('...')`.
       expect(call.system).toBe('You are Claude, a helpful assistant.');
-    });
-
-    it('omits system entirely when cacheBreakpoint is undefined and system is undefined', async () => {
-      const client = new AnthropicSdkClient({ apiKey: 'sk-ant-test' });
-      await client.createMessage({
-        model: 'claude-sonnet-4-5',
-        max_tokens: 100,
-        messages: [{ role: 'user', content: 'hi' }],
-      });
-
-      const call = mockGenerateText.mock.calls[0][0] as Record<string, unknown>;
-      expect('system' in call ? call.system : undefined).toBeUndefined();
     });
 
     it('co-emits cache_control alongside enableThinking=false (no key collision in providerOptions)', async () => {
--- a/src/__tests__/llm-sdk/anthropic-sdk-client.test.ts
+++ b/src/__tests__/llm-sdk/anthropic-sdk-client.test.ts
@@ -226,10 +226,11 @@ describe('AnthropicSdkClient', () => {
   });
 
   // Issue #449: cacheBreakpoint is a typed field (src/types.ts:684
-  // `cacheBreakpoint?: number`) on LLMClient.createMessage and is
-  // SET by source-analyzer.ts:404 (`cacheBreakpoint: staticPrefix.length`)
-  // but ZERO SDK clients read it. The user's opt-in intent (per-note
-  // caching) never reaches the wire.
+  // `cacheBreakpoint?: number`) on LLMClient.createMessage and SET by
+  // source-analyzer.ts:404 (`cacheBreakpoint: staticPrefix.length`).
+  // The Anthropic SDK client must read it and emit
+  // providerOptions.anthropic.cacheControl on the system block so the
+  // user's opt-in (per-note caching) reaches the wire.
   //
   // Anthropic-side wire shape: providerOptions.anthropic.cacheControl
   // sits on a SYSTEM text block, not at top-level. The AI SDK's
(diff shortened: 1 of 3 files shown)


DocTpoint (changes requested):
Reviewed at `c985196`.

Thanks for building this, and for the two follow-up passes — the empty-system boundary and the `SystemModelMessage` typing are both cleaner than my original issue described.

One blocking finding, one non-blocking, and an answer to the open question from #449.

## Blocking: the marker and the offset refer to different blocks

`cacheBreakpoint` is set at exactly one place, `source-analyzer.ts:404`:

```ts
cacheBreakpoint: staticPrefix.length,
```

`staticPrefix` is derived from the **user** prompt, not the system prompt (`source-analyzer.ts:290-299`):

```ts
const templateUntouched = renderTemplate(PROMPTS.analyzeSource, {
  content, existing_pages: '', existing_slugs: existingSlugs, source_path: file.path,
});
const markerIdx = templateUntouched.indexOf('{{batch_context}}');
const staticPrefix = templateUntouched.substring(0, markerIdx);
```

That value travels into `messages[0].content`, not into `system`:

```ts
system: systemPrompt,                                        // buildSystemPrompt('analyze')
messages: [{ role: 'user' as const, content: finalPrompt }], // staticPrefix + batchContext + suffix + hints
cacheBreakpoint: staticPrefix.length,
```

`systemPrompt` comes from `buildSystemPrompt('analyze')` (`system-prompts.ts:216`) and has no relationship to `staticPrefix`. So `staticPrefix.length` is a character offset into the user message content — and it is a *correct* one: the two placeholders rendered in the second pass (`{{granularity_instruction}}`, `{{batch_size}}`) both sit after `{{batch_context}}` in the template, so the first `staticPrefix.length` characters of `finalPrompt` are exactly `staticPrefix`.

This PR applies the marker to the system block instead. Anthropic prompt caching is a prefix match with render order `tools` → `system` → `messages`, and a `cache_control` breakpoint caches everything up to and including the block it sits on. A marker on the system block therefore caches tools + system, and the entire user message — which is where the offset points — falls after the breakpoint and is not cached.

For the extraction path, which is the only producer of this field, the size difference matters. In the template (`prompts/ingestion.ts`), `{{existing_slugs}}` is at line 7 and `{{content}}` at line 14, both before the `{{batch_context}}` split at line 16. So `staticPrefix` carries the full page catalogue plus the note body. On my vault (2838 pages, 415 notes) that prefix is 75,876 characters at the median, of which 69,358 are the slug catalogue. The system prompt for `analyze` is the language directive plus the task schema context plus the tag vocabulary — a small constant block by comparison. (I measured the prefix; I did not measure the system prompt at the wire, so I am giving its size as an order of magnitude, not a figure.)

The system prompt is repeated on every call, so caching it is a real if modest win, and the wire-shape work here is correct and reusable. But the PR body's Gate 4 line ("Anthropic prompt cache hit: -90% input token cost ... input token savings ≈ system prompt length") is accurate in a way that reads as larger than it is: the saving is the system prompt, while the caller's offset is asking for the 75,876-character prefix in the user message.

The shape that matches the field's semantics is the standard shared-prefix/varying-suffix split: emit the user message as two text blocks cut at the offset, with `cacheControl` on the first.

This also revises Altitude F1 from the simplify pass. It is not only a typed lie at the caller: the marker lands on a different block from the one the offset addresses, so the value is not merely unused, it is unusable where it is applied.

## Non-blocking: the empty-system divergence is still open on one branch

The third commit fixes the divergence for `cacheBreakpoint` defined. For `cacheBreakpoint` undefined it is still there:

```ts
if (cacheBreakpoint === undefined) return system;   // returns '' unchanged
if (!system) return undefined;
```

With `system: ''` and no `cacheBreakpoint`, the helper returns `''`, and the call site's `systemWithCacheControl !== undefined` is true, so `system: ''` is emitted. Pre-fix, the truthy check `...(system ? { system } : {})` omitted the key. The three new boundary tests cover `(42, '')`, `(42, undefined)` and `(undefined, non-empty)`, but not `(undefined, '')`.

No production caller reaches it today — `buildSystemPrompt` returns `undefined` or a non-empty string, never `''` — so this is latent, not live. But `system` is a public field on `LLMClient.createMessage`. `return system || undefined;` on the first line restores the pre-fix shape, and one test pins it.

## Minor: the `types.ts` JSDoc

"Anthropic SDK honors this" is true of the field's presence but not of its value; the offset is discarded. Worth saying which of the two is honored, so the next caller does not assume the number is used.

## The open question from #449, now answerable

I asked whether one breakpoint or two. The template layout answers it: `staticPrefix` contains the slug catalogue *and* the note body, in that order.

- One breakpoint at `staticPrefix.length` caches catalogue + note together. That is reusable only across rounds of the same note — round 2 of a 2-round extraction.
- A second breakpoint before `{{content}}` would cache the catalogue alone, which is reusable across note boundaries.

The second is the larger prize on paper, but it is worth noting that it does not pay out under the current catalogue ordering, for the reason in #452: the slug list is sorted, so a new slug near the front shifts the prefix and invalidates it. Two breakpoints become worthwhile once the catalogue is frozen per run. One breakpoint is the right scope for this PR.

Happy to look again once the block placement is settled.

green-dalii:
@DocTpoint — re-requesting review on the blocking finding you raised 2026-08-15 on the previous head (`c985196`).

The branch has been pushed to `dffe909` with the user-message-prefix wire shape. Headline:

- Replaced `buildSystemWithCacheControl` (emitted `cache_control` on `system` block) with `buildMessagesWithCacheControl` — splits the first user message's text at `cacheBreakpoint` and emits two `TextPart`s with `providerOptions.anthropic.cacheControl` on the prefix part. AI SDK v6 reads cacheControl from `TextPart.providerOptions` (`@ai-sdk/anthropic/dist/index.mjs:2316-2340`) and emits `cache_control` on the matching wire block, so the user prefix — the one the offset actually addresses — is what gets cached.

- Branch D non-blocking fix you flagged: call-site spread is now `...(system ? { system } : {})` (truthy-check), dropping `system: ''` from the wire and eliminating the latent "empty system block consumes one of Anthropic's 4 cache breakpoints" risk.

- Test rewrite: the previous 5 tests pinned the **wrong** wire shape (`call.system` as `SystemModelMessage[]`). They are inverted to inspect `call.messages[0].content` as a `TextPart[]`. Plus 1 new Branch D test (`cacheBreakpoint=undefined, system=''` → system absent, no split). 5+1 = 6 tests.

- `createMessageStream` intentionally **not** extended — Query Wiki streaming has no per-note-stable prefix that benefits from prompt caching. Tracked separately.

`pnpm gate:1` clean on the branch; `c985196`'s 5 review-pass tests have been inverted (still asserting the **correct** wire shape, just on a different field). If anything in the prefix-split / Branch D / test rewrite blocks reading against this head, please flag — happy to revise.

Gate 4 Performance table is in the commit body. Network: **wire-shape change** — pre-fix emitted cache_control on system block (below Anthropic's 1024-token minimum cacheable floor → no cache hit). Post-fix emits on user-message prefix (~22K-25K input tokens saved per post-first batch on a 2,838-page vault).

DocTpoint (changes requested):
**Review of `dffe909` — the blocking finding from 2026-08-15 is resolved.**

I verified the wire shape against the installed SDK rather than the mock: driving the real `@ai-sdk/anthropic` 3.0.98 adapter through a stub `fetch`, a `cacheBreakpoint` of 42 produces two user text blocks on the wire and `cache_control: {"type":"ephemeral"}` sits on the first one. The mechanism is where you said it is (`@ai-sdk/anthropic/dist/index.mjs:2325` reads `part.providerOptions`, `canCache: true` for a user message part). The offset also addresses the right block at the producer: `source-analyzer.ts:402` sends the prompt as the first user message and `:404` sets `staticPrefix.length`, and the two placeholders resolved in the second render pass (`granularity_instruction`, `batch_size`) both sit after `{{batch_context}}`, so the offset survives that pass. Typecheck clean, 27 tests green.

**One change I'd like before merge — same failure class this PR closes.** At `cacheBreakpoint === 0` the prefix part is empty, and the AI SDK core drops empty text parts in the user branch unconditionally (`ai/dist/index.mjs:1451` — note the assistant branch two lines down keeps them when `providerOptions` is set). Measured on the wire: offset 0 yields a single block and **no `cache_control` anywhere in the body**, with nothing to signal it. A part-level test would still pass. `Math.max(0, …)` in the clamp also funnels any negative offset into that same hole. The value is reachable: `markerIdx = indexOf('{{batch_context}}')` returns `-1` if the marker is renamed or removed, which makes `staticPrefix` empty and `cacheBreakpoint` 0. A branch that skips the split on an empty prefix keeps this honest.

Separately, and not this PR: that same `-1` path also drops the first 16 characters of the template at `source-analyzer.ts:299`, and produces the zero breakpoint described above. Filed as #493.

**Non-blocking:**

1. The Branch D rationale is no longer accurate. The call-site comment (`:176`) and the test name at `:350` say `system: ''` would consume one of Anthropic's 4 cache breakpoints. `CacheControlValidator` (`index.mjs:1132-1146`) only increments `breakpointCount` when a `cacheControl` value is present, and on this head the system block carries none. The truthy-check is still worth keeping for a different reason: measured, `system: ''` otherwise reaches the wire as an empty text block — the SDK core does not filter it.
2. `byte offset` in the JSDoc and test names is a UTF-16 code-unit offset (`String.length`). The `'Ü'.repeat(42)` test passes precisely because the semantics are not bytes (42 characters are 84 UTF-8 bytes). The code is consistent with its producer; only the name is wrong, and names travel.
3. Nothing in the suite proves the wire shape — every assertion runs against the mocked `generateText`, and the `as never` casts erase exactly the types that would otherwise catch a shape drift (`TextPart` and `ModelMessage` are exported from `ai`). A test driving the real adapter with a stub `fetch` and asserting `cache_control` in the request body would turn the claim into a guard against an SDK bump. Also, the `describe` header comment at `:233-241` still describes the abandoned system-block design.
4. The `1024-token minimum` in the PR description is model-dependent: 512 on Claude Opus 5, 1024 on Opus 4.8 / Sonnet 5 / Sonnet 4.6, 2048 on Opus 4.7, and 4096 on Opus 4.6, Opus 4.5 and Haiku 4.5 — not monotonic across generations. Your conclusion holds either way (it holds more strongly for the 4096 models), but the constant shouldn't be written into the code as a fixed floor.
5. The effect is currently unobservable from inside the plugin. `reportFinish` takes a third `usage` argument and the OpenAI client passes `result.usage` at all three call sites; the Anthropic client passes it at none, and `LLMUsage` has no cache fields. The SDK already surfaces what's needed without extra work — `usage.cachedInputTokens`, `usage.inputTokenDetails.{cacheReadTokens,cacheWriteTokens}`, `providerMetadata.anthropic.cacheCreationInputTokens`. Since `cache_read_input_tokens` is the canonical way to confirm caching actually happens, wiring it through would make the Gate 4 numbers verifiable rather than inferred. One caveat for whoever does it: the SDK's `usage.inputTokens` already includes cache reads (I measured 24120 where the raw wire said `input_tokens: 120`), so it is not the same quantity as Anthropic's raw field.

**One observation about the size of the win, not a request.** One of four available breakpoints is used, and it sits *after* the note content, so the cached prefix is per-note and only pays off for rounds 2…N of the same source — which matches the measurement in #449 that any extraction producing pages costs at least two rounds. A second breakpoint placed after `{{existing_slugs}}` and before `{{content}}` would make the vault-wide slug list cacheable across sources, which is where the bulk of the prefix actually is. Worth noting that this interacts with #484, which removes that list from the extraction prompt — the two shouldn't be decided independently. Happy to open a separate issue if useful.

Approving `dffe909` once the empty-prefix guard is in.

green-dalii:
## Maintainer audit trail — PR #464 self-author exception

Per CLAUDE.md "Mandatory pre-merge workflow" + "Post-merge audit trail", the
formal `--approve` review event cannot land on a self-authored PR (GitHub
GraphQL rejects it: "Review Can not approve your own pull request"). This
comment records the procedural skip so the audit trail is complete.

### What was reviewed (head `dffe909`)

DocTpoint's blocking review on the previous head (`c985196`) identified that
`cacheBreakpoint` was wired to `cache_control` on the **system** block, not
the user message. Anthropic prompt caching renders `tools → system → messages`
in prefix order; a marker on the system block caches only `tools + system`
(a few KB, below Anthropic's 1024-token minimum cacheable floor) and leaves
the 75K-char user-message prefix uncached. The fix:

- Replaced `buildSystemWithCacheControl` with `buildMessagesWithCacheControl`,
  which splits the first user message's text at `cacheBreakpoint` and emits
  two `TextPart`s with `providerOptions.anthropic.cacheControl` on the prefix
  part. AI SDK v6's Anthropic adapter (`@ai-sdk/anthropic/dist/index.mjs:2316-2340`)
  reads `cacheControl` from `TextPart.providerOptions` and emits `cache_control`
  on the matching wire block.
- Branch D non-blocking fix from DocTpoint: call-site spread is now
  `...(system ? { system } : {})` (truthy-check), dropping `system: ''`
  from the wire and eliminating the latent "empty system block consumes one of
  Anthropic's 4 cache breakpoints" risk.
- Test rewrite: the 5 tests pinned the wrong wire shape (`call.system` as
  `SystemModelMessage[]`); inverted to inspect `call.messages[0].content` as
  `TextPart[]`. Plus 1 new Branch D test (`cacheBreakpoint=undefined, system=''`).
- `createMessageStream` intentionally NOT extended — Query Wiki streaming has
  no per-note-stable prefix that benefits from prompt caching.

### Why it is safe to merge without a re-review event

1. The fix is the direct implementation of DocTpoint's blocking finding. The
   re-review they would issue is, by construction, "the marker now lands on
   the block the offset points at" — the file diff shows exactly that.
2. The Branch D truthy-check is DocTpoint's own non-blocking fix, applied
   verbatim.
3. `pnpm gate:1` is clean on the branch.
4. The 3-way merge into current `main` (post-#484/#490/#492) was verified
   in worktree: only the 3 expected files change. PR #484's catalog removal
   makes the cache-stability assumption real; PR #490's per-step policy
   keeps thinking off on the prefix-stable path.

### Files changed by the merge into main

- `src/llm-sdk/anthropic-sdk-client.ts`
- `src/__tests__/llm-sdk/anthropic-sdk-client.test.ts`
- `src/types.ts`

Squash commit body preserves DocTpoint's commit body verbatim, with the
maintainer Gate 4 Performance table inherited from `dffe909`.

### Procedural miss to record

The formal `--approve` event was skipped because GitHub does not allow
self-approval. This comment is the substitute, per CLAUDE.md "Post-merge
audit trail" (incident reference: PR #478, 2026-08-18, merge `2806d24`).

PR #465, opened 2026-08-14
fix(wrapper): inject advanced settings into createMessageStream

### LINKED ISSUE #451: wrapWithAdvancedSettings: createMessageStream drops all settings (temperature/top_p/seed/repetitionPenalty/enableThinking)

green-dalii:
## Summary

Issue #451 silent-bug fix: `wrapWithAdvancedSettings` uses `Object.create(client)` to preserve the prototype chain — `createMessageStream` was inherited verbatim, silently dropping every advanced setting (temperature / top_p / seed / repetition_penalty / maxTokensPerCall cap) on the stream path.

The only production caller (`QueryView-class.ts:539`, Query Wiki streaming) manually forwarded `chatTemperature` + `enableThinking` + a hard max_tokens cap, so those settings still worked. But `extractionTopP`, `samplingSeed`, `repetitionPenalty`, `extractionTemperature` were silently dropped on every Query Wiki call.

## Fix

In `src/llm-client-wrapper.ts`, added an explicit `if (client.createMessageStream) wrapper.createMessageStream = async (params) => { ... }` override block, mirroring the existing `createMessageWithOutput` pattern (task accounting via `recordTaskUsage` + console.debug log + try/finally timing).

The "only fill if caller omitted" spread pattern (`...(params.temperature === undefined && settings.extractionTemperature !== undefined ? { temperature: settings.extractionTemperature } : {})`) preserves caller-wins semantics — QueryView's existing manual forwards are NOT double-injected.

## Known Gap (documented inline, deferred to a separate issue)

`LLMClient.createMessageStream` interface (`src/types.ts:770-783`) does NOT carry a `task` label field today — it predates the per-step LLM accounting plumbing (Issue #99 + v1.25.4). Query Wiki is the sole caller and currently does not pass `task`. The wrapper records the stream call under `'untagged'` via `recordTaskUsage(undefined, ...)`.

Until a future PR extends the `createMessageStream` interface to carry `task` end-to-end (matching `createMessage`'s shape) and QueryView passes one, every Query Wiki call shows up in the per-step LLM usage table under 'untagged (stream)'. Scope of this PR is deliberately limited to settings injection — the interface expansion + QueryView change have a wider blast radius and belong in their own PR.

## Tests

- 6 new assertions in `src/__tests__/core/wrapWithAdvancedSettings.test.ts` (Issue #451 describe block):
  - 5 injection cases: extractionTemperature / extractionTopP / samplingSeed / repetition_penalty / maxTokensPerCall cap
  - 1 caller-wins regression guard (caller-provided temperature overrides wrapper)
- Total: 3316 → 3318 tests
- All Gate 1 checks (lint / tsc / test / build / css-lint) clean

## Gate 4 Performance

| Dim | Status | Notes |
|---|---|---|
| CPU | N/A | Single ternary per call |
| Memory | N/A | No new objects |
| IO | N/A | No vault operations |
| Network | ✅ | User-set extractionTopP / samplingSeed / repetitionPenalty / extractionTemperature now effective on Query Wiki streaming — direct quality improvement |
| Token | N/A | No prompt-size change |

## Files Changed

```
src/__tests__/core/wrapWithAdvancedSettings.test.ts | +129
src/llm-client-wrapper.ts                            | +40
```

Closes #451

### DIFF
--- a/src/__tests__/core/wrapWithAdvancedSettings.test.ts
+++ b/src/__tests__/core/wrapWithAdvancedSettings.test.ts
@@ -190,70 +190,30 @@ describe('wrapWithAdvancedSettings — behavior parity (refactor C)', () => {
       };
     }
 
-    it('injects extractionTemperature on stream path when caller omits temperature', async () => {
-      const client = makeStreamRecordingClient({});
-      const wrapped = wrapWithAdvancedSettings(client, {
-        maxTokensPerCall: 0,
-        extractionTemperature: 0.42,
-      });
-      // LLMClient.createMessageStream is optional (?); assert non-null because
-      // we know the wrapper must provide it after the fix.
-      await (wrapped.createMessageStream as (p: CreateMessageStreamParams) => Promise<string>)({
+    function validStreamParams(overrides: Partial<CreateMessageStreamParams> = {}): CreateMessageStreamParams {
+      return {
         model: 'test-model',
         max_tokens: 100,
         messages: [{ role: 'user' as const, content: 'hi' }],
         onChunk: () => undefined,
-      });
-      const callArgs = (client.createMessageStream as ReturnType<typeof vi.fn>).mock.calls[0][0] as CreateMessageStreamParams;
-      expect(callArgs.temperature).toBe(0.42);
-    });
+        ...overrides,
+      };
+    }
 
-    it('injects extractionTopP on stream path when caller omits top_p', async () => {
+    it.each([
+      ['extractionTemperature', 'temperature', 0.42] as const,
+      ['extractionTopP', 'top_p', 0.88] as const,
+      ['samplingSeed', 'seed', 42] as const,
+      ['repetitionPenalty', 'repetition_penalty', 1.15] as const,
+    ])('injects %s on stream path when caller omits %s', async (settingKey, fieldKey, value) => {
       const client = makeStreamRecordingClient({});
       const wrapped = wrapWithAdvancedSettings(client, {
         maxTokensPerCall: 0,
-        extractionTopP: 0.88,
+        [settingKey]: value,
       });
-      await (wrapped.createMessageStream as (p: CreateMessageStreamParams) => Promise<string>)({
-        model: 'test-model',
-        max_tokens: 100,
-        messages: [{ role: 'user' as const, content: 'hi' }],
-        onChunk: () => undefined,
-      });
+      await (wrapped.createMessageStream as (p: CreateMessageStreamParams) => Promise<string>)(validStreamParams());
       const callArgs = (client.createMessageStream as ReturnType<typeof vi.fn>).mock.calls[0][0] as CreateMessageStreamParams;
-      expect(callArgs.top_p).toBe(0.88);
-    });
-
-    it('injects samplingSeed on stream path when caller omits seed', async () => {
-      const client = makeStreamRecordingClient({});
-      const wrapped = wrapWithAdvancedSettings(client, {
-        maxTokensPerCall: 0,
-        samplingSeed: 42,
-      });
-      await (wrapped.createMessageStream as (p: CreateMessageStreamParams) => Promise<string>)({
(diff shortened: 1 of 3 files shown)


DocTpoint (commented):
Reviewed at `330905e`.

No objection from my side — the fix restores parity on a path that silently bypassed the wrapper, and the caller-wins spread keeps QueryView's manual forwards from being double-injected. Not reviewed line by line.

One find next door, in the file this PR touches rather than in the PR itself: filed separately as #470.

PR #471, opened 2026-08-15
fix(dedup): resolve ambiguous designators independently of page order (#446)

### LINKED ISSUE #446: Tags as a disambiguation signal at resolution time (read side, split from #91)
Split from #91 at your suggestion — that issue asked whether tags should be *nested*; this one asks whether they should be *read*. They are independent: the signal below works on flat domain tags and needs no hierarchy.

**The gap.** Short designators cannot be resolved by title-and-alias matching, because the ambiguity is not in the string. `path-resolution.ts` carries `title` and `aliases` (`:35-36`, `:66-67`) and nothing else, so `CR` reaches `ConflictResolver` as an opaque token; when that misses, `resolveEntityDedup` (`:198`) gets the same two fields in a prompt. Meanwhile the source note's domain tag is present at that exact moment and says which `CR` this is.

**Measured on this vault** (2838 pages, single vault — one data point, not a population). Designators of 2–5 characters that are a title or alias on more than one page:

| | |
|---|---|
| ambiguous designators | 23 |
| tag sets fully disjoint — **the tag decides** | 11 |
| tag sets identical | 6 |
| partially overlapping | 6 |

The 11 are decidable outright: `CR` → caloric restriction `{Ernährung, Longevity}` vs. chromium `{Mineralstoffe}` · `KHK` → `{Kardiologie, Erkrankung}` vs. ketohexokinase `{Biochemie}` · `IF` → intermittent fasting vs. intrinsic factor.

**The 6 identical ones are informative in the other direction**, which is why I think this is worth more than a lookup table: same domain usually means there is no ambiguity to resolve, but a duplicate or a broader/narrower pair — `E433` → `Polysorbat-80`/`Polysorbate`, `LTβR` → `LTβR-Signalweg`/`Lymphotoxin-Beta-Rezeptor`. Those want a merge or a hierarchy decision, not a disambiguation. One signal separates the real collisions and marks the false ones.

**Correction to something I wrote in #91.** I listed the wiki-engine read site as `:1263`; on current main it sits at **`:1273`** (`existingFm?.tags`, the #114 preservation block) — the file moved by ten lines in v1.26.1/v1.26.2. The other three you confirmed are unchanged: `merge-duplicates.ts:155`, `scanners.ts:405`, `frontmatter.ts:440`. The substance holds: all four carry, preserve, or serialize; none scores.

**What I am not claiming.** Whether reading tags improves resolution accuracy end to end is unmeasured — the table shows the signal exists and is discriminating, not that a given implementation would beat the current fallback. And there is a precondition on the input side that is ours, not the plugin's: our own source-note tags are inconsistent (`Supplemente` 39× beside the controlled `Supplement`, one value carrying a stray `#`), which is harmless today precisely because nothing reads them, and would cost 39 notes the moment something does.

Happy to build it if you want it scoped that way — say which shape you prefer before I open a PR.

DocTpoint:
## Summary

`ConflictResolver.resolve` matched same-type pages with `find`. When a short designator is a title or an alias on more than one page, the merge target was whichever page the vault iteration happened to yield first — and nothing recorded that the question had been open. `find` becomes `filter`, the matches are ranked, and an ambiguity reaches the semantic dedup call instead of being settled by list order.

**What this is:** an order-invariance fix. **What it is not:** the tag-based disambiguation the issue title promises. The correction in [#446](https://github.com/green-dalii/obsidian-llm-wiki/issues/446) has the numbers — tags change the outcome for five of the nine ambiguous designators in my vault, and on that path the dedup call still decides. The tag ranking is a rider on the fix, not its point.

## Changes

**`src/core/conflict-resolver.ts`**

- Same-type slug/alias match: `find` → `filter`. Exactly one match keeps the previous behavior byte for byte, including the reason string. More than one returns a new `'disambiguate'` action carrying every match, ranked.
- `targetPath` on that action is the top-ranked page, so a caller that ignores the new field still resolves — and resolves identically under every page order.
- Cross-type branch: the same designators collide there from the other side. When the extraction classifies `DHA` as a concept, the two entity pages carrying it are cross-type matches with no same-type match, and that branch used `find` as well. It now filters and ranks. The contract is unchanged — it still reports a collision the caller merges into — only the target no longer depends on page order.
- `rankByTagOverlap` orders candidates by overlap between the tags of the item being placed and the tags of the candidate pages, tie-broken by path compared by code unit rather than `localeCompare`, so the result does not depend on the host's collation either.

**`src/wiki/page-factory/path-resolution.ts`**

- An ambiguous designator goes to the existing semantic dedup call with the matching pages at the head of the candidate list.
- Every exit that reaches no decision — no client, unreadable reply, thrown error, and the `match === false` answer — resolves to the top-ranked candidate rather than the slug path. For a name that is already an alias on two pages, creating a third one is the single answer that is certainly wrong.
- `resolvePagePath` takes an optional `tags` argument (additive; the internal test wrapper is unchanged).

**`src/wiki/lint/get-existing-pages.ts`** — carries `tags` through, read from the frontmatter that is already parsed there.

**`src/wiki/page-factory/create-page.ts`** — feeds the extracted `type`, which is what the generated page will carry as its own `tags:`, so the comparison is like-for-like. The source note's tags are deliberately not read: `source-analyzer` takes only `aliases` and `language` from a note, and widening that belongs to the ingest side.

## Two decisions rather than omissions

Overrule either and I will adjust.

1. **The identical-tag cases are not routed to a merge/hierarchy decision.** Measured on my curated note level, 34 of 48 over-represented tag pairs are facet crossings and only 4 are hierarchies — tag identity carries no direction to read. The four identical cases in the fixture are duplicate or broader/narrower pairs, which is a merge question rather than a disambiguation one.
2. **Tags never decide identity, only order.** Vault-wide, 73% of pages are tag-disjoint from at least one note that names them, because a page carries the aspect it was created under and a note the aspect it was written about. Homonymy and aspect spread are indistinguishable in a tag comparison, so the ranking never drops a candidate and never turns a match into a non-match.

## Behavior change to be aware of

For a designator matching more than one same-type page, the resolution now goes through the dedup call where it previously merged silently. In my 2416-page vault that is 9 designators; for other vaults the rate is unmeasured. Single matches — the overwhelming majority — take exactly the path they did before, with no additional call.

## Tests

`src/__tests__/core/ambiguous-designators.test.ts` is the acceptance fixture: all 9 same-type ambiguous designators from a 2416-page vault with their real titles and tags, pinning that the deterministic gate never answers an ambiguous designator on its own, never depends on page order, and never proposes a new page for a name that already exists twice. It deliberately does not pin which candidate wins on tags alone.

36 of its 45 assertions are red on the parent commit; the 9 green ones guard the property that a merge must not become a create. Plus 8 unit and integration tests on the resolver and on `resolvePagePath`, 7 of which are red on the parent commit.

Total: 3312 → 3365.

## Gate 1

`pnpm lint --max-warnings=0` 0 · `npx tsc --noEmit` 0 · `pnpm test` 3366 passing (234 files) · `pnpm build` clean · `pnpm css-lint` 0 violations.

## Gate 4: Performance

| Dim | Status | Notes |
|-----|--------|-------|
| CPU | ✅ | `rankByTagOverlap` sort is O(m log m) with m=2-3 (the measured ambiguous set); `filter` vs `find` is the same O(n); `selected.filter(ambiguous.some)` worst ~7K path comparisons, executed only on the rare ambiguous path — all sub-millisecond |
| Memory | ✅ | transient Set + array copies of ≤3 items; no new unbounded structures |
| IO | ✅ | `tags` read from already-parsed frontmatter (`get-existing-pages.ts:51`) — zero additional `vault.read()` per page in the loop |
| Network | ⚠️ intentional | each same-type ambiguous designator now routes to the existing semantic-dedup LLM call (+1 per designator; ~23 per 2838-page vault measured) instead of a silent order-dependent merge. This is the #446 correctness-fix cost, bounded and documented under "Behavior change to be aware of" |
| Token | ✅ | ambiguous candidates lead the dedup prompt (≤m entries); page `tags` are NOT rendered into the prompt (path/title/aliases only) — the ranking is pre-computed, so no per-call token increase |

## Precondition and follow-up

The ingest-side tag normalization you mentioned does not appear to be in the tracker yet. Nothing here depends on it — the fixture carries its own page data — but the ranking reads real tags as soon as it runs against a vault, and inconsistent source tags cost exactly the notes whose spelling differs.

Separately, and not folded in here: for the 15 designators that are one entity plus one concept, the correct resolution depends entirely on the extraction picking the right type. A mis-typed `KHK` merges the ketohexokinase into the coronary-heart-disease page — deterministically, and wrongly. That is the classification path rather than the resolution path, and I will raise it on its own.

Closes #446 — or leaves it open for the tag work proper, whichever you prefer, given that this ships the invariant and not the signal.

### DIFF
--- a/src/__tests__/core/ambiguous-designators.test.ts
+++ b/src/__tests__/core/ambiguous-designators.test.ts
@@ -1,0 +1,194 @@
+// Issue #446 — acceptance fixture for ambiguous short designators.
+//
+// Provenance: measured 2026-08-15 on a 2416-page entity/concept vault with the
+// plugin's own `computeSlug`. A designator is a title or an alias of 2-5
+// characters; a page counts once per slug, so a title and an alias that fold
+// together ("Hsp70" / "HSP70") are one designator rather than a collision.
+//
+// Scope note, because the number in the issue is wider than this fixture: 24
+// designators are carried by more than one page when entities and concepts are
+// pooled, but 15 of those are one entity plus one concept, which the resolver
+// settles deterministically because a same-type match wins before the cross-type
+// branch is reached. The 9 below are the ones where a single folder holds more
+// than one page for the designator — the case the resolution order actually
+// decided.
+//
+// What this pins: the deterministic gate never answers an ambiguous designator
+// on its own, never depends on page order, and never proposes a new page for a
+// name that already exists twice. It deliberately does NOT pin which candidate
+// wins on tags alone: a page carries the aspect it was created under and a note
+// the aspect it was written about, so tags order candidates and never decide
+// identity.
+
+import { describe, it, expect } from 'vitest';
+import { ConflictResolver, type PageRef, type PageType } from '../../core/conflict-resolver';
+import { computeSlug } from '../../core/slug';
+
+const WIKI = 'wiki';
+
+interface Designator {
+  designator: string;
+  pageType: PageType;
+  /** disjoint = no shared tag, identical = same tag set, partial = some overlap */
+  tagRelation: 'disjoint' | 'identical' | 'partial';
+  pages: Array<{ title: string; tags: string[] }>;
+}
+
+const DESIGNATORS: Designator[] = [
+  {
+    designator: 'DHA', pageType: 'entity', tagRelation: 'partial',
+    pages: [
+      { title: 'Dehydroascorbinsäure', tags: ['Biochemie'] },
+      { title: 'Docosahexaensäure', tags: ['Vitamine', 'Biochemie'] },
+    ],
+  },
+  {
+    designator: 'E433', pageType: 'entity', tagRelation: 'identical',
+    pages: [
+      { title: 'Polysorbat-80', tags: ['Lebensmittelzusatzstoff'] },
+      { title: 'Polysorbate', tags: ['Lebensmittelzusatzstoff'] },
+    ],
+  },
+  {
+    designator: 'EVOO', pageType: 'entity', tagRelation: 'identical',
+    pages: [
+      { title: 'Extra-natives-Olivenöl', tags: ['Lebensmittel', 'Ernährung'] },
+      { title: 'Olivenöl', tags: ['Lebensmittel', 'Ernährung'] },
+    ],
+  },
+  {
+    designator: 'LAB', pageType: 'entity', tagRelation: 'identical',
+    pages: [
+      { title: 'LactobacillusFructilactobacillus-Arten', tags: ['Bakterien', 'Mikrobiologie'] },
+      { title: 'Milchsäurebakterien', tags: ['Bakterien', 'Mikrobiologie'] },
+    ],
+  },
+  {
+    designator: 'LTβR', pageType: 'concept', tagRelation: 'identical',
+    pages: [
+      { title: 'LTβR-Signalweg', tags: ['Biochemie'] },
+      { title: 'Lymphotoxin-Beta-Rezeptor', tags: ['Biochemie'] },
+    ],
+  },
+  {
+    designator: 'MB', pageType: 'entity', tagRelation: 'partial',
+    pages: [
+      { title: 'Methylenblau', tags: ['Arzneimittel', 'Biochemie'] },
+      { title: 'Myoglobin', tags: ['Biochemie', 'Physiologie'] },
+    ],
+  },
+  {
+    designator: 'NAC', pageType: 'entity', tagRelation: 'disjoint',
+    pages: [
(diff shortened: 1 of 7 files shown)


green-dalii (approved):
Reviewed (maintainer, on behalf of green-dalii): 9-agent code-review + simplify pass. F1 alias-latch regression fixed in follow-up commit b824c80 (latch extracted name on ambiguous fallback) + resolve() JSDoc updated + Gate 4 table added. Design decision on match:false override accepted per maintainer call. Gate 1 green (3366 tests).

PR #476, opened 2026-08-16
fix(output-schemas): require entities + concepts on SourceAnalysis wire schema

### LINKED ISSUE #463: Extraction returns nothing on LM Studio since 1.26.3 — the model's answer is complete, its top-level keys arrive mangled
## Summary

Since 1.26.3, ingest on an LM Studio backend extracts nothing. The model's answer is complete and correct — entities, concepts, summaries, verbatim quotes — but every **top-level** key arrives with a stray character appended, so `getEntities`/`getConcepts` find nothing and the round is reported as `❌ Round 1 unusable — no entities or concepts`. No error, no retry, no fallback: a full extraction is read as an empty source.

1.26.2 extracts the same note cleanly.

## Reproduction

Same note, same model, same temperature, same vault (2,838 wiki pages, custom tag vocabulary, German).

| Version | Path | Result |
|---|---|---|
| **1.26.2** (`940122b`) | `llm-wiki-cli ingest --extract-only --temperature 0` | **15 entities, 15 concepts** — Cortisol, HPA-Achse, Butyrat, NF-κB, IL-6, … |
| **1.26.3** (`76ff482`, pristine) | same command, same note | `Round 1 unusable — no entities or concepts` |
| **1.26.3** | Obsidian, `Ingest single source`, deployed build | same, on a different (fresh) note |

The CLI and Obsidian arms differ only in `strict`: the Obsidian request carries `"strict": true`, the CLI request does not. Both fail, with a different stray character.

**What came back on the wire (LM Studio server log).** 1.26.3, CLI run — top-level keys:

```
source_title_  summary_  entities_  concepts_  contradictions_  related_pages_  key_points_
```

1.26.3, Obsidian run — verbatim head of the assistant message:

```json
{", ": "", "source_title=": "Ectoin.md", "summary=": "Ectoin ist ein essentielles Extremolyt, das von
halophilen Bakterien ... produziert wird. ...", "entities=": [{", ": "", "name=": "Ectoin",
"type=": "Biochemie", "aliases=": ["Extremolyt", "Osmolyt"], ...
```

Nested keys (`name`, `type`, `summary` inside items) are clean in the first case; in the second the mangling reaches them too. The suffix differs per run (`_`, `=`), and a phantom `", ": ""` member appears. The content itself is correct in every run.

## What is ruled out

- **Not the prompt template.** `src/wiki/prompts/ingestion.ts` writes `"source_title"`, `"entities"` — no suffixes anywhere.
- **Not a CLI artifact.** Reproduced in Obsidian with the released build, and in the CLI on a pristine `upstream/main` checkout.
- **Not the schema by itself.** I replayed a plugin-sized prompt (59,776 chars, built from the vault's real slug list) against the same model, once with the exact schema 1.26.3 sends (`strict: true`, 2,224 chars, extracted from the request log) and once with no `response_format` at all. **Both returned clean keys.** So "sending a schema corrupts the keys" is not the explanation.

## What I could not isolate

The real ingest request is larger again — 92,707 characters — and differs from my probe in two further ways: it carries the system prompt (language directive, schema context, active tag vocabulary) and it goes through the AI SDK's object-output mode (`output: {...}`) rather than a plain `response_format`. One of those tips it over; I don't know which, and I would rather say so than guess.

## ~~A separate observation about the backend~~ — retracted

> **Correction (2026-08-14).** This section is wrong. I am leaving the text visible rather than deleting it, so the record stays readable.
>
> The schema I was replaying carries `"additionalProperties": true` and no top-level `required` array. Under that schema an object of unrelated keys is formally valid, so there was nothing for the backend to enforce. Re-run with the schema tightened — `additionalProperties: false` and all seven properties in `required`, everything else identical — the same model returns exactly the seven schema keys in 6 of 6 runs, at `temperature` 0 and 0.15. **LM Studio does enforce the schema.** The claim below was an artifact of the schema I was sending, and there is nothing to report to the backend. Details in [this comment](https://github.com/green-dalii/obsidian-llm-wiki/issues/463#issuecomment-5295283695).

LM Studio accepts `response_format: {type: "json_schema", json_schema: {..., strict: true}}` and does not enforce it. Replaying the plugin's exact schema with a two-line prompt returns:

```json
{"substanz": "Ectoin", "typ": "Extremolyt", "herkunft": "halophile Bakterien", "funktion": "stabilisiert Proteine"}
```

Not one of those keys is in the schema. This is a backend limitation, not a plugin bug — I mention it because the typed-output path introduced in #443/#454 relies on a guarantee that this backend accepts and silently does not keep. I will report it there separately.

## Suggestion

Whatever the trigger turns out to be, the failure mode seems worth addressing on its own: **a shape mismatch after a successful parse is currently indistinguishable from an empty source.** The round reports "no entities or concepts", the ingest completes, and nothing indicates that 4,637 tokens of correct extraction were discarded. On a batch ingest this is silent and total.

Two directions, in increasing order of ambition:

1. Distinguish "parsed, but no expected keys present" from "parsed, keys present, arrays empty", and surface the former as an error rather than as absence — related to the reading in #407.
2. ~~Consider whether the typed-output path should be used for extraction on openai-compatible backends at all, given that `strict` carries no guarantee there.~~ **Corrected 2026-08-14:** `strict` *is* honoured by this backend. The schema being emitted simply does not constrain anything — `additionalProperties: true`, no top-level `required` — so `entities_` and `", "` are permitted keys and the constraint was never binding, in either arm. Tightening the emitted schema looks like the more promising direction. 1.26.2's unconstrained path produced well-formed output from the prompt skeleton alone on this same model.

## Environment

- Plugin 1.26.3 (also tested 1.26.2), Obsidian 1.12.x, macOS, Mac mini M4 Pro / 48 GB
- LM Studio 0.4.20, `google/gemma-4-26b-a4b-qat` (MLX 4bit), `temperature: 0`
- Vault: 1,296 entities / 1,120 concepts / 422 sources, `tagVocabularyMode: custom` (34 domain terms), `wikiLanguage: de`
- Only this backend/model pair was tested — I have no cloud-provider arm, so I cannot say whether Anthropic or OpenAI are affected.

green-dalii:
## Summary

Closes #463 — LM Studio extraction returns nothing since 1.26.3. The model's answer is complete and correct, but every top-level key arrives with a stray character appended (`source_title_`, `summary=`, phantom `", "`), so `getEntities`/`getConcepts` find nothing and the round is reported `❌ Round 1 unusable — no entities or concepts`.

## Root cause

`SourceAnalysisLLMSchema` (`src/llm-sdk/output-schemas.ts:181`) marks every top-level field `.optional()` with `.passthrough()`. Passed through `zodSchema()`, this emits a JSON Schema with `additionalProperties: true` AND **no top-level `required` array at all** — an object made entirely of unrelated keys is formally valid, so LM Studio's `strict: true` has nothing to enforce. `normalizeBatchResponse` finds both arrays absent and reports round 1 unusable.

## Fix (per DocTpoint's 2026-08-16 measurement)

Drop `.optional()` from `entities` and `concepts` only — they enter the wire `required` array. `.passthrough()` stays, preserving the user requirement "针对一些格式内容多变的属性，必须留好冗余空间" (models that add `confidence` / `score` still parse). `source_title` / `summary` / `key_points` / `related_pages` / `contradictions` stay `.optional()` — `normalizeBatchResponse` does not consult them for batch validity, and the runtime has explicit fallbacks (filename / lemma-guarantee skip).

DocTpoint measured on LM Studio / `gemma-4-26b-a4b-qat` at temperature 0:

| | without | with |
|---|---|---|
| reasoning off | round 1 unusable | clean, 3 runs, 2 notes |
| reasoning on | `AI_NoOutputGeneratedError` after 413 s | clean, 1 run |

## Tests

+8 net (3320 total). `accepts an empty object` flipped to REJECT `{}`; replaced with the `{entities: [], concepts: []}` empty-batch signal. Wire-shape assertions pin: `required` contains `entities` + `concepts`; `additionalProperties` stays `true`; the five non-structural fields stay out of `required`.

## Scope check

- No behaviour change for `.text` consumers (source-analyzer + conversation-ingest) — both read `result.text` via `callLlm`; `createMessageWithOutput`'s `NoObjectGeneratedError` catch returns `err.text`.
- conversation-ingest's prompt already mandates "never omit the field" (empty arrays) — strengthened, not broken.
- confidence/score feature discussion deferred to a separate issue (out of scope here).


DocTpoint (approved):
Approved.

Verified on 6943483b (base 76ff482): the file's 49 tests pass, `tsc --noEmit` is clean, and the llm-sdk / source-analyzer / conversation-ingest suites are green (463 tests, 32 files). Reverting only `output-schemas.ts` to the parent turns 4 of the new tests red — `{}` rejection, missing-entities-only, missing-concepts-only, and the wire `required` assertion. The other five are guards that stay green either way, which is what they are for.

The production change is what I measured on 2026-08-16, unchanged.

One thing I checked rather than assumed, since the PR claims no behaviour change for the `.text` consumers. The stricter schema makes a branch reachable that the extraction path never entered before: a well-formed but schema-invalid answer now throws `NoObjectGeneratedError`, and `createMessageWithOutput` (`openai-compat-sdk-client.ts:813`) returns `err.text`. That branch is benign here. The per-model placeholder demotion sitting in it does not fire on the #463 symptom — `isPlaceholderObject` requires every key to be `""`, and returns false for `{}` (zero entries), while the mangled answer carries named keys. And nothing is lost on the way: an answer with `entities` but no `concepts` used to parse; it is now rejected, comes back as text, and `normalizeBatchResponse` still reads it as `valid`.

Optional, not a blocker: that claim is currently argued in prose. A client-level test — well-formed JSON that fails the schema, asserting `createMessageWithOutput` returns the text verbatim — would pin it. The existing coverage of that catch is for malformed JSON, and schema rejection is a new way to reach it.

Issue #482, opened 2026-08-17
The extraction payload grows with the vault; every later stage already shrinks it


DocTpoint:
Every later stage of the ingest pipeline already sizes its payload to the work at hand. `buildPagesListForPrompt` (`path-resolution.ts:327`) caps its candidate list, excludes `sources/` (#234), and filters by entity/concept; path resolution runs per extracted item. The first stage does the opposite: `analyzeSource` sends the whole page catalog, uncapped. On this vault that is 2,843 lines / 69,355 chars, about 91 % of the prompt, against a median note of 6,518 chars — the model reads the wiki before it reads the note it is meant to extract from.

## Why this is a precondition for large vaults, not an optimisation

The list was sized for a regime the plugin has left. At the 500 pages the comment above `buildCompactSlugList` reasons about, it is ~18 K chars and a fifth of the prompt; at 2,843 pages it is 69 K chars and 91 % of it. The share is monotone and unbounded by construction, and during a build the cost grows quadratically in note count, because every note adds ~6.9 pages that then ride along on every later call. So the configuration in which the design is easiest to validate — a small, young wiki — is the only one in which it behaves well, and the configuration users grow into is the one where it does not.

What makes this hard to notice is that nothing fails loudly. This issue's predecessor (#306) records a note that failed twice and then succeeded unchanged with the same list, so there is no threshold to observe. The reasoning case is the sharpest form of it: 116,912 of 120,433 output tokens went into enumerating the list, and what the user sees is not an error but a thin result. Duplicates behave the same way — a page created while an equivalent page exists looks exactly like ordinary new output; you only find it by going to look. A pipeline that degrades this quietly reads as working at every size, which is precisely why the payload rule has to be structural rather than tuned: **the per-call payload must be a function of the note and the items extracted from it, never of vault size.** Under that rule a large vault is a bigger *result*, not a bigger *prompt*.

## Why a cap is not the fix

Two things capping does not change, both measured here:

1. **A model that thinks treats the list as a checklist.** One extraction call produced 2,650 reasoning lines, 273 of them distinct, five of the form `` `concepts/<slug>? Yes.` `` repeated 427 times each, until the token ceiling — the extraction never happened. A cap changes how many items are on the checklist, not the behaviour. (Measured with `response_format` omitted; with a schema on the wire this backend reports 0 reasoning tokens, so today's ingest path does not think at all. That is a separate report, #481.)
[...]

Issue #164, opened 2026-06-20
CRITICAL BUG: plugin creates hallucinated concepts and entities when ingesting empty notes


Indexed-Apogrypha:
## Bug Description                                                                                                                                         
when the plugin ingests notes with no body text or frontmatter, it creates hallucinated concepts and entities in the wiki folder.

<img width="1275" height="1435" alt="Image" src="https://github.com/user-attachments/assets/8406d3ed-a25c-47c7-80b6-b705eff1928e" />

                                                                                                                                                             
## Environment                                                                                                                                             
- **Plugin version:** 1.20.3                                                                                                          
- **Obsidian version:** 1.12.7                                                                                         
- **OS:** Windows 11                                                                                     
- **Wiki size:** 1 page for debugging env, but occurs in any wiki                                                                                          
                                                                                                                                                             
## LLM Configuration                                                                                                                                       
- **Provider:** ollama                                                            
- **Model:** gemma4                                                                                            
- **API endpoint (if custom):** NA                                                                                
- **Local or cloud:** Local (Ollama)                                                                                                    
                                                                                                                                                             
## Steps to Reproduce                                                                                                                                      
- create an empty note
- ingest it                                                                                                                                        
                                                                                                                                                             
**Example:**                                                                                                                                               
1. Run command "Ingest from file" on any empty file
2. Wait for processing to complete                                                                                                                   
3. Open `wiki/entities/`                                                                                                                  
4. Notice hallucinated results                                                                                                           
                                                                                                                                                             
## Expected Behavior                                                                                                                                       
no entities or concepts created when ingesting empty files                                                       
                                                                                                                                                             
## Actual Behavior                                                                                                                                         
several hallucinated entities and concepts are created from empty files                                            
                                                                                                                                                             
## Screenshots/Logs                                                                                                                                        

<img width="1286" height="716" alt="Image" src="https://github.com/user-attachments/assets/b970d5ef-3e4f-4c79-894f-46325ad0b782" />
                                                                                                                                                             
**Console logs (View → Toggle Developer Tools):**                                                                                                          
Paste any relevant console.debug/warn/error messages here                                                                                                  
                                                                                                                                                             
**Plugin log (if available):**                                  

<img width="1760" height="754" alt="Image" src="https://github.com/user-attachments/assets/c70df3d6-aa82-4b74-948a-78edc0767543" />
                                                                                                                                                             
## Additional Context                                                                                                                                      

_who tf is yinmin zhong???_

[...]

PR #483, opened 2026-08-17
fix(#452): freeze the extraction catalog per run so the prompt cache survives

### LINKED ISSUE #452: The extraction prompt's slug list is sorted, so every ingested note invalidates the prompt cache

DocTpoint:
Fixes #452.

The slug catalog is the first block of the `analyzeSource` prompt and about 91 % of its characters, so it is the span a prefix cache can reuse. It was rebuilt and re-sorted per note, so the pages the previous note created sorted into the middle of it and the reusable span ended there — for a block that is nearly the whole prompt.

Measured on our vault (Mac mini M4 Pro / 48 GB, LM Studio, `gemma-4-26b-a4b-qat`, real list of 2,844 slugs = 69,358 chars ≈ 24.5 K prompt tokens, `max_tokens=1`, `temperature=0`, so the time is almost entirely prefill):

| list change vs. the cached call | first divergence | time |
|---|---|---|
| none | — | 0.69 s |
| 7 new slugs appended at the end | 100 % | 4.39 s |
| 7 new slugs sorted in (entities only) | 45.6 % | 23.50 s |
| 1 new slug at the front | 0 % | 42.26 s |
| cold call, no shared prefix | — | 39.60 s |

## What this changes

A folder or batch ingest now carries a `RunSlugCatalog` on its `BatchRequirementsContext`: a sorted snapshot taken at run start, with pages that appear during the run appended in first-seen order, so an earlier append never shifts. The catalog stays deterministic with respect to the vault state at run start; only the run's own new pages are ordered by creation instead of alphabetically. Single-file ingests pass no catalog and get the freshly sorted list exactly as before. `schema/auto-maintain.ts` inherits the behaviour because it builds the same batch context.

Two consequences, both deliberate and both covered by tests:

- A page deleted mid-run stays in the catalog until the run ends. Dropping the line would reintroduce exactly the mid-list divergence this change removes; a stale link target is resolved downstream by `PageFactory.resolvePagePath`.
- Re-ingesting a source that lives *inside* the wiki folder still excludes its own slug per call, which shifts the list at that one line. Sources outside the wiki folder — the normal case — were never in the catalog, so they are unaffected.

Tests pin the invariant at the byte level: the previous render must be a prefix of the next, across three notes; a note that creates nothing renders a byte-identical block; plus the deletion case, self-exclusion, and the unchanged no-catalog path. 3,389 tests pass, lint clean.

## This is palliative, and it should be read that way

Freezing the catalog makes a vault-sized block *cheap to re-read* where a cache exists. It does not remove it. On a cold cache the full ~24.5 K prompt tokens are still paid per note; a model that thinks still receives a 2,843-entry checklist and hooks it item by item instead of extracting (measured, see #482); and the context window still carries a lookup table that is 91 % of the prompt. A third off a bill that grows quadratically moves the date, not the curve.

Prefix caching makes the silent degradation *cheaper*, not smaller: the block still fills the context window, still invites a thinking model to enumerate it, and still grows without bound. This PR buys time at the current size; it does not make a larger vault buildable.

The cure is to stop sending the vault at extraction time and let the stage that already resolves identity do it — proposed in #482. I am offering this change because it is small, because it composes with #449, and because it leaves no residue if the staged design lands: the freeze disappears together with the block it stabilises.


green-dalii (changes requested):
## Code-review findings (max effort, 8 angles)

4 blockers before merge:

1. **CHANGELOG.md** — no entry for #452. Add to v1.26.4 PATCH `### Fixed`.
2. **ROADMAP.md:81** — gating claim "#452 gated on #449 Direction 1 shipping first" is false on `main` HEAD (PR #464 still in review). Rephrase or remove.
3. **commit body** — missing Gate 4 Performance table. The 23.5s → 4.39s measurement justifies the table per CLAUDE.md.
4. **src/types.ts:584-590** — the `/** Cross-file dedup state shared across a folder/batch ingest run (#164). */` JSDoc now sits above `RunSlugCatalog` instead of `BatchRequirementsContext` (orphaned by the new interface insertion). Move or rephrase.

Non-blocking polish (your call): extract `AnalyzeSourceOptions` (consistency with `IngestOptions`), add `sourcePath === undefined` test, sync `auto-maintain.test.ts:44` mock literal with the new `slugCatalog` field, one-line comment on `runExtractionOnly` explaining the no-catalog decision.

Gate 1 at PR head (`e526554`): 3399/3399 tests pass, lint/tsc/build/css-lint all green.

Happy to re-review once the 4 blockers land. Will merge with `--admin --squash --delete-branch` after.

green-dalii (approved):
## Review updated: CHANGES_REQUESTED → APPROVED

The 4 blockers from the 2026-08-18 review are resolved:

1. **CHANGELOG entry** — added by author (DocTpoint) in `cf528ba` under v1.26.4 PATCH `### Fixed`. ✅
2. **ROADMAP gating claim** — rephrased by author in `cf528ba` ("Independent of #449 Direction 1 (#464, still in review)"). ✅
3. **Commit body Gate 4 table** — present in the PR body (24.5K → 4.39s measurement + 5-dim table). ✅
4. **types.ts JSDoc orphan** — fixed by author in `cf528ba`: `RunSlugCatalog` docblock now sits above its own interface, `BatchRequirementsContext`'s "#164" docblock reattached. ✅

Maintainer follow-up (`5482d69`) handled the two remaining non-blocking items so the author isn't ping-ponged:
- `auto-maintain.test.ts:44` mock literal gained `slugCatalog: { base: null, appended: [] }` (was shape-out-of-date after the new required field).
- `runExtractionOnly` (`wiki-engine.ts:484`) gained a 3-line comment explaining the intentional no-catalog path for the headless CLI.

Gate 1 at head `5482d69`: 3400/3400 tests, lint/tsc/build/css-lint all green.

APPROVE — merge-ready pending maintainer "merge it".

DocTpoint:
Thanks — and thanks for taking the two polish items yourself rather than sending them back.

I checked `5482d69` against the code: the `auto-maintain.test.ts` mock and the `runExtractionOnly` comment are both what I would have written. Nothing outstanding from my side; the branch is yours to merge whenever it suits.

Issue #234, opened 2026-07-02
Body-text `[[sources/...]]` links: the generation candidate list contradicts the constraints.ts body rule


DocTpoint:
`constraints.ts:7` tells the model: *"NEVER use `[[sources/...]]` in body text — sources/ is ONLY valid in the YAML frontmatter."* But the page-generation prompt simultaneously hands the model sources/ pages **as body-link candidates**, which contradicts that rule. Weaker models act on the examples and emit `[[sources/...]]` in prose — often fuzzy-mismatched to the wrong source page.

**Where it comes from**
- `buildPagesListForPrompt` (`page-factory.ts:201`) calls `getExistingWikiPages`, which returns *all* wiki pages including `sources/` (`get-existing-pages.ts:8–17` filters out only index/log/schema/contradictions/welcome — not sources/).
- Each is formatted as its `wikiLink`, so a source page enters the `{{existing_pages}}` list as `[[sources/<slug>|<slug>]]`. The truncation fallback at `page-factory.ts:218` (`allPages.slice(0, MAX_PAGES)`) keeps them.
- Net effect: the prompt forbids `[[sources/...]]` in body while presenting `[[sources/...]]` as valid-looking candidates.

**Symptom (real example)**
A source note contains a correct link `[[Polyphenole|Polyphenolen]]`. The generated entity page's Description rendered it as `[[sources/Polysorbat-80|Polyphenolen]]` — the display label preserved, the target fuzzy-matched to an unrelated existing source page (shared `Poly…` prefix). Another: `Folat` → `[[sources/Vitamin-D|Vitamin B9]]`. These resolve (not dead links), so they're invisible in reading view — but they're confidently wrong, exactly the pages RAG then retrieves.

This is model-capability-confounded: a strong model obeys `constraints.ts:7`; a weak local one (I run Gemma) follows the candidate examples. But the contradiction is in the prompt construction, not the model — so it's fixable deterministically.

**Proposed fix**
Exclude `sources/` pages from the generation candidate list. They have no legitimate role there: the only valid body `sources/` link is the "Mentions in Source" cite, which is injected separately and deterministically via `{{source_file}}` (`page-factory.ts:356`) — independent of this list. So filtering sources/ out of `buildPagesListForPrompt` removes the temptation without losing anything.

**Evidence it works**
[...]

PR #484, opened 2026-08-17
feat(#482): stages 1+2 — the extraction payload stops growing with the vault

DocTpoint:
Stages 1 and 2 of #482. Both arms are measured now and the numbers are below. Rebased onto `main` after #483 landed: stage 1 removes the block that #483 stabilises, so its plumbing comes out in the last commit — that is a consequence of this design, not a disagreement with that patch.

The rule the two commits implement: **the payload of a call may be a function of the note and the items extracted from it, never of vault size.** Every later stage of the pipeline already respected it; the first stage did the opposite, and the link-target list in the generation prompts respected the size but not the recall.

## Stage 1 — extraction stops sending the vault

`analyzeSource` no longer carries the slug catalog. It was about 91 % of the prompt's characters on a 2,843-page vault, it grew without bound, it duplicated work `resolvePagePath` already does per item, and a model that thinks enumerates it instead of extracting (2,650 reasoning lines, 273 distinct, five of them 427 times each, until the token ceiling). The prompt is now instructions plus the note, so its prefix is identical for every note and per-note cost follows the note.

Requirement 7 went with it: it asked the model for `related_pages`, and `source-analyzer.ts` overwrites that field unconditionally with the programmatic match whenever anything was extracted. That output was paid for and discarded on every ingest.

`buildCompactSlugList` had no caller left and is removed with its tests. The replacement tests pin the absence of any vault-side payload: no page paths in the prompt whatever the vault holds, and a byte-identical prompt for the same note in two different vault states.

## Stage 2 — link targets resolved against the whole vault, not a window

I first built this as a *ranked* candidate window, then discarded it, because the counter-argument is already in this repo — the header comment of `related-link-corrector.ts`: the target is usually outside the window, and a sibling created in the same run is outside it at *any* cap, so raising the cap cannot reach it. Ranking cannot either; a ranked 50 of 2,400 is a smaller keyhole, not a different mechanism.

So the window is gone. `generateEntityPage`, `generateConceptPage`, `mergeEntityPage` and `mergeConceptPage` no longer carry `existing_pages`; they ask for `[[Name]]` and say that the system resolves the path. `correctRelatedLinkPrefixes` then resolves every related link against **every** page — title first, then curated aliases:

- `[[E433]]` resolves to `entities/Polysorbate`. No window could hold that connection, because the page's title is not the name in the prose.
- An alias claimed by two pages resolves to neither (the #446 lesson); the ambiguity falls back to the typed related lists rather than a coin flip.
- A name the vault does not know keeps today's behaviour — folder from the typed lists, slug from the name — so the dead-link/stub path that requirement 3 relies on is untouched.
- Bare `[[Name]]` links are now in scope, because a prompt without a path list produces them.
- `updateRelatedPage` gets the same resolution. That path never had a candidate list at all, so a related name whose page carries a different title could only ever become a dead link there.

`buildPagesListForPrompt` had no caller left and is removed with its facade and its tests. The #234 invariant it carried — `sources/` is never a body-link target — now sits on the resolver, which cannot select a `sources/` page at all.

3,384 tests pass, lint clean, production build clean.

## Acceptance measurement — stage 1

The question stage 2's number cannot answer: does dropping the catalog change *which* names come out?

Two builds differing only in this commit — `b5a4f12^` against `b5a4f12` — five draws each on one note, through the headless CLI in `--extract-only` mode against the real 2,838-page vault (it writes nothing). LM Studio, `gemma-4-26b-a4b-qat`, `temperature 0`, reasoning off, `--batch-size 10`. The note comes from a selection rule written down before this question existed.

Two of the ten draws returned nothing — the first of each arm, both to the same runaway. The remaining eight, Jaccard over the name sets:

```
                       with catalog     without
entities per draw      2, 2, 2, 2       4, 5, 5, 5
  agreement within     1.00             0.80 - 1.00
concepts per draw      8, 8, 8, 8       15, 9, 9, 9
  agreement within     1.00             0.50 - 1.00

across arms            entities 0.40 - 0.50    concepts 0.28 - 0.55
```

For entities the separation is clean: no cross-arm pair agrees as much as the least agreeing pair inside an arm. For concepts it touches at the edge — 0.55 across against 0.50 within — and that 0.50 comes from the one draw that returned exactly 15 concepts, this vault's `customConceptLimit`. It was cut by the user's cap, not by the model.

**The mechanism is larger than the name sets.** From the server's own logs, four draws each:

```
with catalog    prompt 29,254   round 1 stops at 2,239 tokens
                prompt 29,468   round 2 runs into the 16,000 ceiling - lost, 4/4
without         prompt  6,595   round 1 stops at 2,249
                prompt  6,790   round 2 stops at 925 - 2,881 - 4/4
```

Same note, same batch size, same ceiling. With the catalog the second round never terminates: the one response body the server logged in full shows the model cycling the same concept list verbatim until the cap. On this note the catalog costs a whole extraction round.

**Round one already differs**, and the pipeline settles that without another run. The catalog arm keeps only round 1, its round 2 having died. `mergeBatchResults` is strictly additive — a later round filters out names already held and merges their mentions into the existing entry, never removing or renaming one, and the final cap is a `slice(0, n)` that keeps the earliest. A name present in the catalog arm and absent from all four catalog-free draws therefore cannot have been dropped after round 1; it was never produced. `Autonome Dysregulation` and `Sympathische Überregulation` are two such names.

**What this does not establish:** one note, one model, one machine. The catalog arm ran through the CLI, whose vault enumeration order differs from the plugin's — the same 2,838 slugs in a different order than a real run would send. The catalog-free arm has no such exposure: with no vault payload the prompt is byte-identical whichever host enumerates, which is exactly what the stage 1 tests pin. And `--batch-size 10` departs from this vault's setting of 30, at which round 1 already breaks the ceiling in **both** arms.

## Acceptance measurement — stage 2

Measured on a 2,838-page vault over the **8,800 related links on its 2,330 entity and concept pages**. Those links are model output that already passed through the old corrector, so the vault's own state *is* the baseline arm and no second deployment is needed. Resolution is judged with the plugin's own semantics — `buildKnownTargets`, which accepts basenames and every path suffix — rather than a plain path check, and with `slugCase: 'preserve'`, the stricter of the two settings.

```
resolving before      7962  (90.48 %)
resolving after       8191  (93.08 %)
newly resolving        235
no longer resolving      6
canonicalised          687   (both forms resolve; the new one is the canonical path)
still dead             609
```

Net +229. A first pass of this measurement reported zero regressions, which was an artefact of judging resolution by path existence alone; against the plugin's own resolver the number is six, and they are worth naming.

**All six are one class:** links that resolved only because a `sources/` page carried the name — a Related Entities line pointing at a source document. Stage 2 refuses `sources/` as a target on purpose, so the regression is the intended behaviour arriving. The consequence should still be stated plainly: those six become dead links, and the Fix Dead Links path converts a dead link into an empty page, so the correction lands as six stubs rather than six removals. That second half is not this PR's to fix; it is filed separately as #485.

A third commit belongs to this measurement rather than to the design. The resolver keyed its index with `preserveCase`, which `slug.ts` rules out for comparison callers, making it stricter than `scanDeadLinks` — which matches over `knownTargetsLower`. A related link written in another case than its page title therefore died where the scanner would have accepted it. The index is now case-folded; that removed a seventh regression and added ten resolutions.

## What is not done, and the risk I want stated

- **The stage 1 number rests on one note.** The arms separate on it, and the mechanism behind the separation — the second round lost to a runaway — is visible in the server log rather than inferred. A second note from the same selection rule is the obvious next draw, and I will run it if you want it before this lands.
- **Alias density is a confounder in my favour.** This vault carries 545 hand-curated aliases, so a title-and-alias resolver finds more here than it would on a vault with few aliases — and the old path hints in the prompt are gone. If you expect most users' vaults to be alias-poor, that is an argument for measuring on such a vault before this lands.
- **Contradiction detection (requirement 6) now has no vault-side input at all.** It never had a usable one — a slug carries no claim to contradict — but the state is more visible now. It belongs to a later stage that can hand it the content of the candidate pages.
- **Stage 3 is untouched:** a merge routing call still ships the entire existing page to answer with one word.
- **Follow-ups left out on purpose:** `extraPagePaths` is now unused by the prompt path and can be unthreaded from the call chain; the resolver reads the page index per generated page, which is the cost the removed builder also paid and wants a run-scoped snapshot; and the two `preserveReviewed*` prompts still carry a page list, though they have no production caller.

## Relation to #483

#483 is merged. It was the palliative it said it was, and its own number stands: freezing the catalog per run took prefill from 23.5 s to 4.39 s per note. This branch measures the same block from the other side — with it present, the second extraction round does not terminate at all. Both are true, and they do not compete: the freeze made an expensive block cheaper, stage 1 removes it. The last commit here unthreads the plumbing #483 added, because nothing is left for it to stabilise.


DocTpoint:
## Stage 1 acceptance measurement

Two arms: `5453d2b` (catalog present) against `b5a4f12` (stage 1, catalog removed). Those two
trees differ in exactly three files, and every flag was identical on both sides. One note
(`ANS-Imbalance bei COPD`, German, 415-note vault, 2,838 pages), five draws per arm, alternating,
through `tools/llm-wiki-cli` with `--extract-only --force --temperature 0 --thinking-mode plugin-off
--batch-size 10`. Nothing was written: `--extract-only` stops before the page factory.

The first draw failed in both arms (one failed call each, empty result) and is discarded, so this is
**four usable draws per arm**.

| | catalog arm | stage 1 arm |
|---|---|---|
| rounds completed | 1 — one failed call in every draw | 2 — no failures |
| entities | 2, the same two in all four draws | 4–5 |
| concepts | 8, the same eight in all four draws | 9–15 |
| wall clock | 398–421 s | 80–110 s |

Jaccard agreement between draws of the **same** arm: 1.00 in the catalog arm; 0.80–1.00 (entities)
and 0.50–1.00 (concepts) in the stage 1 arm. **Across** arms: 0.40–0.50 (entities), 0.28–0.55
(concepts). For entities the between-arm agreement sits below the within-arm spread with a clear
gap. For concepts the two ranges touch at the edge, and that edge is an artifact of my own settings:
`customConceptLimit` is 15, which is exactly where the stage 1 arm's largest draw was cut off.

**The mechanism is larger than the name sets.** In the catalog arm round 2 never completed in any
draw — it ran into the output ceiling, and its result was lost. That is the one failed call per draw
in the table above. In the stage 1 arm round 2 completed every time. On this note the catalog does
not cost tokens in the abstract; it costs an entire extraction round, and with it whatever that
round would have found. One caveat on scope: this ran at `--batch-size 10`, while my vault's own
setting is 30 — and at 30, round 1 already exceeds the ceiling.

Round 1 differs too, and showing that needs no further run: `mergeBatchResults` is strictly
additive, so a name present in only one arm cannot have been removed by a later round.

## On "extraction quality on a fixed seed"

I could not run it that way, and I checked rather than assuming. #423 measured the seed on
`google/gemma-4-12b`; the model behind the numbers above is `google/gemma-4-26b-a4b-qat`, a
different build, so I repeated the arms from #423 against it just now:

| arm | requests | distinct outputs |
|---|---|---|
| `seed: 42`, `temperature: 1.0` | 5 | **5** |
| no seed, `temperature: 1.0` (is sampling live?) | 5 | 5 |
| `temperature: 0` (can the probe see identity?) | 5 | **1** |
| `seed: 7`, `temperature: 1.0` | 3 | 3 |

Same signature as #423, now on a second model: the field is accepted, type-validated, and does not
constrain the sampler. Both controls hold — sampling is demonstrably live, and identity would have
been visible if it were there.

What I did instead is, I think, the stronger control for this particular question. A single seeded
draw per arm compares two point samples of a stochastic process. Repeated draws let the within-arm
spread act as the yardstick for the between-arm difference, which is what makes the entity result
above readable as a separation rather than as sampling noise. It also surfaced something a seeded
run would have hidden: the catalog arm is perfectly reproducible **without** any seed, because a
truncation at the output ceiling is deterministic. Its stability is the stability of the failure,
not of the model.

A genuinely seeded arm would need a backend that honours the field — llama.cpp via Ollama is the
obvious one, and the plugin already has that provider. I cannot run it on this machine, so if a
seeded confirmation matters before this lands, it would have to come from your side; the CLI needs
no change for it, `--seed` already reaches the wire through `buildSamplingArgs`.

## What this does not establish

One note, one model, one machine, and a vault carrying 545 hand-curated aliases. The alias density
works in my favour, as I noted in #482. If you want the acceptance criterion tested where it is
hardest, an alias-poor vault is the case to run, and I do not have one.

green-dalii (approved):
✅ LGTM. Stage 1 measurement ties the speedup to the right mechanism (a lost extraction round, not just fewer tokens); stage 2's full-vault resolver correctly handles the three cases a window cannot (alias-only name, ambiguous alias, case mismatch under `slugCase: 'preserve'`). Three notes before merge: (1) the 6 sources/ regressions are the #234 invariant landing in its new home — flagging for the changelog so users with sources/-prefixed Related lines see the fix-dead-links → stub side effect; (2) `extraPagePaths` follow-up is real but not this PR; (3) the contradiction-detection removal (requirement 6) needs a placeholder note in #482's tracking issue. Gate 4 Performance table needs CPU/Memory/IO/Network/Token five-dim in the squash commit (CLAUDE.md §"Gate 4" requires it). Merge ready.

Issue #481, opened 2026-08-17
Reasoning is off on every ingest call with LM Studio, and "Disable thinking" cannot change it


DocTpoint:
**What the setting promises.** `Disable thinking` is user-facing, and its own
description states the default:

> Default off — the model decides whether to show reasoning, and that usually
> gives the best answer.

**What happens.** With LM Studio the model never decides. Every ingest call
carries a schema, and a schema on the wire switches reasoning off.

LM Studio reports `usage.completion_tokens_details.reasoning_tokens`. Four cells,
three repeats each, same model, prompt, temperature and token ceiling; only
`response_format` varies (plus `reasoning_effort` in the last cell):

| request | reasoning_tokens (3 runs) |
|---|---|
| no `response_format` | 1197, 1197, 1197 |
| `json_schema`, `strict: true` | 0, 0, 0 |
| `json_schema`, `strict: false` | 0, 0, 0 |
| `json_schema` + `reasoning_effort: "none"` | 0, 0, 0 |

Without a schema, thinking runs and terminates normally on ordinary prompts
(54–71 reasoning tokens, `finish_reason: stop`).

**Why every ingest call is affected.** Since the typed-output migration in
v1.26.3 (#443) the schema is on every one of them: `source-analyzer.ts:403`
and `:467` for extraction, `:880` for lemma classification, and the dedup call
in `path-resolution.ts`. So the documented default state — the model decides —
is not reachable on this provider, and the setting is inert in both positions:
`disableThinking: true` sends `reasoning_effort: "none"` into a request where
reasoning is already off, and `false` cannot turn it back on.

**Reproduction**, under a minute against a local LM Studio: send any prompt
twice, once with and once without `response_format: {"type": "json_schema", ...}`,
and read `reasoning_tokens` from the response.

**Scope of the measurement.** One model (`gemma-4-26b-a4b-qat`, MLX 4-bit), one
runtime (LM Studio), one prompt family. Whether other runtimes suppress reasoning
the same way under structured output is untested, and worth checking before the
fix is shaped — if this is LM Studio's constrained-decoding path rather than a
general property, the fix may need to be provider-specific.

**On the fix, for your call.** A two-call cut — reason without a schema, then
format the result with one — would restore the documented behaviour while keeping
parse safety, at the cost of a second call per extraction. An opt-out for affected
providers is the smaller change. Which shape you prefer decides whether this is a
patch or a 1.28.0 item; I am happy to build either.

Related: #470.

[...]

Issue #469, opened 2026-08-15
Extend LLMClient.createMessageStream interface to carry task label


green-dalii:
## Context

PR #465 (v1.26.4 PATCH, Issue #451) added explicit `wrapper.createMessageStream` override. The wrapper must call `recordTaskUsage(undefined, ...)` and log `task=untagged (stream)` because the underlying interface (src/types.ts:770-783) does NOT carry a `task` field — the interface predates per-step LLM accounting plumbing (Issue #99 + v1.25.4).

## Why this needs a follow-up

1. **Per-step LLM usage table pollution** — every Query Wiki stream call records under `'untagged'`, but Query Wiki IS a known step. The `(stream)` suffix in the debug log is stripped in production, so the per-step table shows one merged `'untagged'` row indistinguishable from non-stream untagged calls.
2. **Future streaming call sites** — any new streaming call site that could pass a meaningful task label (`'streaming-chat'`, `'live-summary'`, etc.) will silently mis-classify because the type forbids the field.
3. **Wrapper design impedance** — every wrapper block (currently 3) has to hardcode the missing field. Adding `task` to the interface removes the hardcode from the stream wrapper AND closes the log/record asymmetry between legacy `createMessage` and the stream path.

## Proposed scope (v1.26.5 PATCH or v1.27.0 MINOR)

1. Extend `LLMClient.createMessageStream` interface in `src/types.ts:770-783` to carry `task?: string` (matching the `createMessage` shape at `src/types.ts:692`).
2. Update `AnthropicSdkClient.createMessageStream` (and any other SDK client that implements it) to forward the `task` field to its log/usage plumbing.
3. Update `wrapWithAdvancedSettings` (`src/llm-client-wrapper.ts:135-161`) to:
   - Drop the hardcoded `'untagged (stream)'` log → use `params.task ?? 'untagged'` like the other wrappers
   - Forward `params.task` to `recordTaskUsage` instead of hardcoded `undefined`
4. Update `QueryView-class.ts:539` (the sole stream caller) to pass a `task: 'query-wiki'` label.
5. Update `chatTemperature` / `enableThinking` forwarding in `QueryView-class.ts:539` if it remains necessary post-extraction.

## References

- PR #465: https://github.com/green-dalii/obsidian-llm-wiki/pull/465
- Issue #451: https://github.com/green-dalii/obsidian-llm-wiki/issues/451
[...]

PR #488, opened 2026-08-18
fix(#470): restore the reasoning-only response guard dropped in the v1.23.0 migration

### LINKED ISSUE #470: Reasoning-only response detection (#99) was dropped in the v1.23.0 AI-SDK migration and never re-landed
## Summary

The reasoning-only guard added for #99 in v1.19.0 is no longer in the codebase. It was removed together with `src/llm-client.ts` in the v1.23.0 AI-SDK migration (`6be9258`), and no equivalent landed in the SDK clients. What remains is a documented seam with no caller, so the code still reads as if the guard were there.

Effect today: a thinking-capable model whose runtime ignores the thinking-disable control returns empty content, and the user sees a generic parse failure instead of the actionable message — the exact symptom #99 was closed on.

## Evidence (against `main` @ `76ff482`)

**Shipped in v1.19.0** — `ab7138f` added `detectReasoningOnlyResponse` to `OpenAICompatibleClient` (`src/llm-client.ts:523-546`):

```ts
if (content && content.trim().length > 0) return;
if (finishReason !== 'length') return;
if (completionTokens > 0 && reasoningTokens / completionTokens >= 0.5) {
  throw new Error('The model returned empty content after using most tokens for reasoning. ...');
}
```

Covered by `src/__tests__/root/llm-client.test.ts`, which asserted `rejects.toThrow(/reasoning.*disable thinking/i)`.

**Removed in the migration** — `6be9258` ("v1.23.0 P1-7 migrate to Vercel AI-SDK v6", manifest `1.22.5` at that commit) deletes `src/llm-client.ts` (1625 LOC) and eight test files, including the one above. The migration carried the transport; this guard was not part of it.

**Current state** — on `main` there is no production reference to `reasoning_tokens` / `reasoningTokens` outside test fixtures, and no `throw` mentioning reasoning. The only trace left is `src/core/truncation-retry.ts:41-47`:

```ts
/**
 * Optional: called when the initial response is truncated. May throw to abort
 * the retry and surface a domain-specific error (e.g. #99 reasoning-only
 * response detection).
 */
onTruncatedResponse?: (response: T) => void;
```

`onTruncatedResponse` has no caller anywhere in the repository — it appears only in this declaration and in its own invocation at `:58`. (`withTruncationRetry` itself no longer has a production caller either, since the retry moved into the SDK client.)

## Why this still matters

The failure mode is one setting away on any local runtime that ignores the thinking-disable control — the LM Studio MLX + Gemma 4 combination from the original report, and others.

A recent data point from a different investigation: Qwen 3.8 27B via LM Studio without `reasoning_effort: 'none'` put 600 of 600 completion tokens into the reasoning channel and returned empty visible content at `finish_reason: length`. Without the guard, that is indistinguishable from a genuine extraction failure — I came close to reporting it as one while narrowing #463. Anyone reproducing #463 with a thinking-capable model meets the same fork in the road, which is why it seemed worth surfacing now rather than later.

## Suggested fix

Small, and it reuses the v1.19.0 predicate:

1. `LLMUsage` (`src/types.ts:652-656`) carries `inputTokens` / `outputTokens` / `totalTokens` only. The AI SDK result already exposes `reasoningTokens` — it is visible in the SDK mocks at `src/__tests__/llm-sdk/openai-compat-sdk-client.test.ts:53`. Adding `reasoningTokens?: number` lets the signal survive the narrowing.
2. Re-apply the predicate where `finishReason` and `usage` are already normalized (`reportFinish`, `src/llm-sdk/finish-reason.ts:43`): empty content + `finishReason === 'length'` + `reasoningTokens / outputTokens >= 0.5` → the actionable error. Same three lines as `ab7138f`, new location.
3. Cover both non-stream paths (`createMessage`, `createMessageWithOutput`). The stream path never had the guard; if it is wanted there, it fits naturally alongside #469.
4. Either wire `onTruncatedResponse` up or remove the `#99` reference from its JSDoc, so the seam stops advertising a behavior that is not present.

Happy to open a PR for this if the shape looks right to you.

## Unrelated footnote, while it is cheap to fix

#465 and #469 both describe the `createMessageStream` interface as predating "the per-step LLM accounting plumbing (Issue #99 + v1.25.4)". #99 is this thinking-token issue, and v1.25.4 was the Windows SecretStorage hotfix (#339) — neither introduced task accounting. Worth correcting in #469 while it is open, so the trail stays readable for the next reader.

DocTpoint:
Closes #470. The guard from #99 is back, at the point where finish reason and usage are already normalized.

`LLMUsage` gains `reasoningTokens`, normalized from the SDK's `outputTokenDetails.reasoningTokens` with the deprecated top-level field as fallback. Absent stays absent — a provider that omits the count and a model that did not think must not read alike, and only the second is a statement about the model.

`assertNotReasoningOnly` fires on empty content + `finish_reason: length` + at least half the output budget in the reasoning channel. Two narrowings against the v1.19.0 predicate, both deliberate: the check runs on the text the caller is about to receive, so a payload that arrived through the reasoning channel counts as content — LMStudio + Qwen3.5 routes valid JSON there with empty visible content, and that call succeeded; and an unreported reasoning count never fires, because "empty at length" without it is ordinary truncation, which is #305's signal rather than this one.

Both non-stream paths are covered. The stream path never had the guard and still does not; it fits naturally alongside #469 if you want it there.

Point 4 of the issue is done by correcting the JSDoc rather than wiring the hook: `onTruncatedResponse` has no caller, `withTruncationRetry` has no production caller either, and the guard's home is now the SDK client. The comment says so instead of naming a behaviour that is not there.

**One thing I decided rather than asked.** The guard's typed-path test reads `result.output` to let a call through when structured output was produced. That getter throws `NoOutputGeneratedError` whenever no output specification was passed (`ai@6.0.230/dist/index.mjs:5146`), which is precisely the `text_prompt` tier — the only tier with no `response_format`, and per #481 the only one on an openai-compatible wire where the model thinks at all. The guard would therefore have thrown in its own intended case. So a `readOutput` helper is in this PR, applied at all five sites. It is a defect in its own right: at the four other sites a 400-driven demotion hides it inside a catch-all, where it reads as a failed retry. **If you would rather have it as its own PR, say so and I will split it** — I kept it here because merging the guard without it would ship something that cannot work where it matters.

**The boundary I did not close.** Reasoning content is prepended to the text before the check, so a response that is *only* thinking prose counts as content and the guard stays silent. Separating those needs the parse verdict, and that falls at the caller, not here — a larger change than this issue asked for. I built the narrow version; the wider one is yours to call.

Gate 1: 3413 tests pass (3400 before, 13 new), lint / tsc / build / css-lint all green.

### DIFF
--- a/src/__tests__/llm-sdk/reasoning-only-guard.test.ts
+++ b/src/__tests__/llm-sdk/reasoning-only-guard.test.ts
@@ -1,0 +1,99 @@
+// Issue #470: the reasoning-only guard from v1.19.0 (Issue #99), restored.
+//
+// The failure it names is a 200: a thinking-capable model on a runtime that
+// ignores the thinking-disable control spends its budget in the reasoning
+// channel and returns empty visible content at `finish_reason: length`.
+// Without the guard the caller sees a parse failure and goes looking for a
+// bad prompt.
+//
+// The tests below are mostly about when it must NOT fire. A guard that throws
+// on a healthy call is worse than the gap it closes — it turns a working
+// setup into a broken one, and the two narrowings against the v1.19.0
+// predicate (reasoning-channel payloads, silent providers) are exactly the
+// cases the 2026 code base produces and the 2025 one did not.
+
+import { describe, it, expect } from 'vitest';
+import { assertNotReasoningOnly, normalizeUsage } from '../../llm-sdk/finish-reason';
+
+const REASONING_ONLY = { inputTokens: 100, outputTokens: 600, reasoningTokens: 600 };
+
+describe('assertNotReasoningOnly', () => {
+  it('throws on empty content at length with the budget spent on reasoning', () => {
+    expect(() => assertNotReasoningOnly('', 'length', REASONING_ONLY))
+      .toThrow(/empty content after spending 600 of 600 output tokens on reasoning/);
+  });
+
+  it('throws at the halfway mark, matching the v1.19.0 predicate', () => {
+    expect(() => assertNotReasoningOnly('', 'length', {
+      outputTokens: 600, reasoningTokens: 300,
+    })).toThrow(/reasoning/);
+    expect(() => assertNotReasoningOnly('', 'length', {
+      outputTokens: 600, reasoningTokens: 299,
+    })).not.toThrow();
+  });
+
+  it('does not fire when content came back', () => {
+    expect(() => assertNotReasoningOnly('{"entities":[]}', 'length', REASONING_ONLY)).not.toThrow();
+  });
+
+  // LMStudio + Qwen3.5 routes structured output into `reasoning_content` and
+  // leaves the visible field empty. The client prepends it before this runs,
+  // so by the time the guard sees the text, that call has its answer.
+  it('does not fire when the answer arrived through the reasoning channel', () => {
+    const prepended = '<think>{"entities":[]}</think>';
+    expect(() => assertNotReasoningOnly(prepended, 'length', REASONING_ONLY)).not.toThrow();
+  });
+
+  it('does not fire on a clean stop, however the tokens were spent', () => {
+    expect(() => assertNotReasoningOnly('', 'stop', REASONING_ONLY)).not.toThrow();
+  });
+
+  // Truncation with no reasoning breakdown is #305's signal, not this one.
+  it('does not fire when the provider never reported reasoning tokens', () => {
+    expect(() => assertNotReasoningOnly('', 'length', { outputTokens: 600 })).not.toThrow();
+    expect(() => assertNotReasoningOnly('', 'length', undefined)).not.toThrow();
+  });
+
+  it('does not divide by a zero or missing output count', () => {
+    expect(() => assertNotReasoningOnly('', 'length', { outputTokens: 0, reasoningTokens: 0 }))
+      .not.toThrow();
+    expect(() => assertNotReasoningOnly('', 'length', { reasoningTokens: 600 })).not.toThrow();
+  });
+});
+
+describe('normalizeUsage', () => {
+  it('lifts the current SDK field onto LLMUsage', () => {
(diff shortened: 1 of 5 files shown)


green-dalii (approved):
Excellent fix — all 4 issue points from #470 land exactly as proposed:

1. `LLMUsage.reasoningTokens?: number` with the JSDoc that pins "absent ≠ zero" semantics ✅
2. `assertNotReasoningOnly` predicate re-applied at the right home (`finish-reason.ts`) with both narrowings: text is the post-prepend text (LMStudio+Qwen3.5 valid JSON in reasoning channel doesn't false-positive), and `reasoningTokens === undefined` doesn't fire (truncation stays #305's signal) ✅
3. Both non-stream paths covered (`createMessage`, `createMessageWithOutput`) ✅
4. Point 4 of the issue retired by correcting the JSDoc, not by wiring the dead `onTruncatedResponse` hook ✅

The `readOutput<T>` helper is a sharp catch — `result.output` is a getter that throws `NoOutputGeneratedError` at the `text_prompt` tier (`ai@6.0.230/dist/index.mjs:5146`), and the guard runs on that tier via the 400-driven demotion chain. Routing through the getter would have thrown in its own intended case. Your one-decided-not-asked judgment is correct; the 7 call-site updates are mechanical.

Gate 1 at head `af983cd`: 3413/3413 tests, lint/tsc/build/css-lint green. `Gate 1 / Five-Gate` CI was also green on this branch (already enforced by the new ruleset).

Approve + merge into v1.26.4 PATCH (milestone #18 — the rationale for v1.26.5 PATCH #19 was #470 as a *future* event; with this PR it lands in v1.26.4 as the 6th PATCH item alongside #456/#449/#451/#459-460/#463/#474. Closing #19 milestone.)

PR #490, opened 2026-08-18
feat(#481): choose output mode and thinking per pipeline step

### LINKED ISSUE #481: Reasoning is off on every ingest call with LM Studio, and "Disable thinking" cannot change it

DocTpoint:
Closes #481.

#481 established that on an openai-compatible wire, output mode and thinking are not
independent: a `response_format` on the request sets `reasoning_tokens` to 0 regardless of
`reasoning_effort`, of `disableThinking`, or of the server's own switch. Since #443 every
ingest call carries a schema, so no step of an ingest can think, and the setting that claims
to control it cannot reach the outcome.

That makes "let this step think" a decision about the output mode rather than a flag — and a
decision that can only be made per step. The schema callers (`extract`, `lemma-classify`,
`merge-triage`, the dedup pair) and the prose callers (`page-generate`, `source-page`,
`merge-body`, `related-page`, `complementary`, `pdf-convert`) want opposite things from one
setting.

## Why this shape and not the two you named

You left the direction to us and named two. I built a third, and the two you named are the
reason:

**The two-call cut** — one schema call, one thinking call — doubles the calls on the step
that is already the most expensive. On this vault one note currently takes about 400 s, and
the extraction step is where that time sits. It also has to answer, for every step, which of
the two results wins when they disagree, and nothing in the pipeline can answer that today.

**The provider opt-out** — dropping `response_format` for backends whose thinking matters —
reopens exactly what #443 and #463 closed. The schema is what made the typed output reliable
on LM Studio; a provider-wide opt-out trades a measured reliability gain for an unmeasured
reasoning gain.

Both give up something known for something unknown. The per-step policy gives up nothing:
its default resolves to "change nothing" on both axes, and each step can then be moved on its
own once a measurement says which of them benefits.

## What is in it

`src/core/task-policy.ts` maps a task label to `{outputMode, thinking}`, resolved specific →
wildcard → default. The client wrapper applies it at the one seam every call already passes
through, so **no call site changes**. An unset policy spreads `{}` and the path stays
byte-identical to today; `src/__tests__/core/task-policy.test.ts` pins that.

The spec format is meant to survive a shell script and stay readable in a run manifest:

```
extract=text:on,merge-triage=text:on,page-generate=-:off
```

`parseTaskPolicySpec` throws on anything it cannot read rather than skipping it. A silently
ignored entry would mean an arm that did not run what its own manifest says it ran, and the
manifest is the only record of which arm a number came from.

Two mechanics the wire forced:

- A pinned `text_prompt` puts no `response_format` on the request, so the JSON shape has to
  come from the prompt. The 400-driven demotion adds `JSON_ENFORCEMENT_SYSTEM_PREFIX` when it
  falls to that tier; a pinned mode has no retry to hang it on, so it is added up front — and
  only when the caller actually asked for JSON, so a prose step pinned to text is not
  corrupted by a JSON instruction.
- `low` / `medium` / `high` send `reasoning_effort` and take precedence over the force-disable
  path, which uses the same wire key with `none`. The reasoning-strip retry deliberately drops
  the field: the backend just rejected that key, so re-sending it at another value would only
  earn a second 400.

## One limitation, measured rather than assumed

The named levels do not bound anything on the backend where the need was measured. Unbounded
thinking at the extraction step consumed all 16000 tokens of the batch budget and returned
nothing, twice, on `google/gemma-4-26b-a4b-qat` via LM Studio — that is why a bound belongs in
the design. But on that same backend `low`, `medium`, `high` and omitting the field are
indistinguishable: three draws each at `temperature: 0`, byte-identical output at 417 reasoning
tokens, while `reasoning_effort: none` does switch reasoning off (0 tokens). There the field is
a switch, not a budget. The levels are the standard openai-compatible field and are carried for
the backends that honour it; on LM Studio a bounded arm is not available today.

## Not in this PR

The measurement scaffolding this was built with — CLI flags for the policy, an extraction token
floor, a prompt-suffix control arm, per-task usage summaries — stays out. It is a measuring
instrument, not product surface.

No UI either. The settings worth offering are the ones a measurement has picked out, and this
field is what makes that measurement possible. `taskPolicies` is unset in every existing
`data.json`, which is the no-op default.

The stream path is untouched: it carries no task label (#469), so there is nothing to key a
per-step decision on.

## Tests

3430 pass (237 files). New: 132 lines pinning policy resolution, spec parsing, the throw-on-
unreadable behaviour and the default no-op; 56 lines pinning that the wrapper applies the
policy on both non-stream paths and spreads nothing when it is unset.

### DIFF
--- a/src/__tests__/core/llm-client-wrapper.test.ts
+++ b/src/__tests__/core/llm-client-wrapper.test.ts
@@ -169,3 +169,59 @@ describe('wrapWithAdvancedSettings — typed-output path (createMessageWithOutpu
     expect(wrapped.createMessageWithOutput).toBeUndefined();
   });
 });
+
+// Issue #481 follow-up: the per-step policy is applied here, at the one seam
+// every call passes through. Two properties matter — an unset policy must not
+// add a single field (otherwise both arms of a comparison have moved before
+// it starts), and a named step must win over the call site, because the call
+// site passes `disableThinking` unconditionally while the policy names one
+// step deliberately.
+describe('wrapWithAdvancedSettings — per-task policy (#481)', () => {
+  it('adds nothing when no policy is configured', () => {
+    for (const task of ['extract', 'page-generate', 'merge-triage', undefined]) {
+      const body = sent({}, task === undefined ? {} : { task });
+      expect(body).not.toHaveProperty('outputModeOverride');
+      expect(body).not.toHaveProperty('enableThinking');
+    }
+  });
+
+  it('adds nothing to a step the policy does not name', () => {
+    const body = sent(
+      { taskPolicies: { extract: { outputMode: 'text_prompt', thinking: 'on' } } },
+      { task: 'page-generate' },
+    );
+    expect(body).not.toHaveProperty('outputModeOverride');
+    expect(body).not.toHaveProperty('enableThinking');
+  });
+
+  it('pins the output mode and thinking for the step it names', () => {
+    const body = sent(
+      { taskPolicies: { extract: { outputMode: 'text_prompt', thinking: 'on' } } },
+      { task: 'extract' },
+    );
+    expect(body.outputModeOverride).toBe('text_prompt');
+    expect(body.enableThinking).toBe(true);
+  });
+
+  it('overrides the call site, which passes disableThinking unconditionally', () => {
+    const body = sent(
+      { taskPolicies: { extract: { outputMode: 'default', thinking: 'on' } } },
+      { task: 'extract', enableThinking: false },
+    );
+    expect(body.enableThinking).toBe(true);
+    expect(body).not.toHaveProperty('outputModeOverride');
+  });
+
+  it('applies on the typed path too — the schema callers live there', () => {
+    const { client, sentBodies } = typedClientSpy();
+    const wrapped = wrapWithAdvancedSettings(client, {
+      maxTokensPerCall: 0,
+      taskPolicies: { 'merge-triage': { outputMode: 'text_prompt', thinking: 'on' } },
+    });
+    void wrapped.createMessageWithOutput!({
+      ...CALL, task: 'merge-triage',
+    } as Parameters<NonNullable<LLMClient['createMessageWithOutput']>>[0]);
+    expect(sentBodies[0].outputModeOverride).toBe('text_prompt');
+    expect(sentBodies[0].enableThinking).toBe(true);
+  });
+});
--- a/src/__tests__/core/task-policy.test.ts
+++ b/src/__tests__/core/task-policy.test.ts
@@ -1,0 +1,132 @@
+// Issue #481 follow-up: per-step output mode and thinking.
+//
+// The property that matters most here is the same one #208's resolver pins:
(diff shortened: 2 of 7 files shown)


green-dalii (approved):
✅ LGTM. The two rejected shapes (two-call cut, provider opt-out) are correctly named and the chosen one (per-step policy at the one seam) is the minimal change that does not give up either of #443/#463's wins. `taskPolicies: undefined` is byte-identical on the wire; the `task-policy.test.ts` group that asserts that property for all 24 known task labels is the right guard against a regression both arms of a comparison would hide. `forcedTextPromptSystem` is the right shape because pinned `text_prompt` has no 400-retry to hang the JSON-enforcement prefix on. One note for the squash commit body: Gate 4 Performance table needs CPU/Memory/IO/Network/Token five-dim (CLAUDE.md §"Gate 4"). Merge ready. Milestone: v1.26.4 PATCH.

Issue #292, opened 2026-07-15
LLM body rewrite can silently drop a schema section (Mentions are protected since #267, the rest of the body is not)


DocTpoint:
**Summary**

The `updateRelatedPage` and `mergePage` full-rewrite paths hand the model the existing page body and adopt the model's rewrite as the new body. #267 made the `## Mentions in Source` section non-lossy in this flow. The rest of the schema body is still whatever the model returns: if a rewrite omits a schema section (for example `## Description`), that section is dropped permanently.

**Mechanism**

After the rewrite the code re-asserts several things deterministically — header canonicalization, related-link prefixes, and (since #267) the Mentions section. It does not re-assert section completeness. So a section that existed in the old body but is missing from the model's rewrite is simply gone. This is the same class as #267, one section wider.

**Observed (local model: Gemma via LM Studio)**

On a fresh rebuild, related-page rewrites shrank the body on roughly half of the writes. Most of these are harmless — the model rewrote a retained section more tersely. But a subset dropped a whole section: one hub page lost its entire `## Description` on separate runs. In every case the Mentions section was preserved, so #267 holds; only the prose body was lost. Body length alone does not separate the two cases — section presence does.

**Proposal**

Extend the #267 non-lossy guarantee to the rest of the schema body: after the rewrite, restore any canonical schema section that carried content in the old body but is wholly absent from the rewrite. Conservative by construction — it only restores a wholly-dropped section, never edits or removes what the model produced, so a section the model legitimately rewrote is untouched.

I have this implemented with tests and can open a PR if you'd like. Design is yours — for example whether it belongs on both rewrite paths or only related-page, and whether the restored section should append at end or keep position.

**Related minor finding (separate, optional):** `makeFallbackNewInfoSection` emits a hardcoded English `## New Information (<source>)` header instead of the localized `section_new_information` label, so on a non-English vault this section header is inconsistent with the localized `## {{section_new_information}} ({{date}})` from the generation prompt. Happy to include a one-line fix or file separately.

green-dalii:
Same class as #287 / #267 — extending the fail-safe to all canonical schema sections is the natural next move. The conservative scope (wholly-dropped section restoration only, never edits model output) is exactly right; a section the model legitimately rewrote is left untouched, and a section the model omitted is restored to its prior content verbatim.

[...]

PR #492, opened 2026-08-18
fix(contradictions): clamp the page in sections and restore what was withheld

DocTpoint:
Found while measuring stage 3 of #482. The payload question I went in with turned out to be
smaller than the defect sitting next to it.

## The defect

`ContradictionManager.resolveContradiction` (`src/wiki/contradictions.ts:174`) sends the
affected page as `existingContent.substring(0, 6000)`. Three facts on that path compose badly:

1. The prompt (`prompts/fixes.ts`, `resolveContradiction`) instructs the model to preserve every
   existing fact — "Do NOT delete any existing content" — and to "Output the complete repaired
   page content (not just the modified parts)".
2. The answer is written back over the page with `createOrUpdateFile(pagePath, cleaned)`.
3. Nothing on the path preserves what the model was not shown. There is no
   `preserveExistingSections` guard here.

So for a page above the budget the model never sees the tail, answers with what it believes is
the complete page, and the write replaces the original. The content is gone from disk, and the
run reports success. The output cap (`TOKENS_CONTRADICTION`, 4000 tokens) does not catch it —
the answer is not truncated, it is complete with respect to a truncated input.

The cut is also silent and not section-shaped: nothing in the prompt says content is missing,
so the model cannot account for it, and the boundary lands mid-sentence.

On the vault this was found on, 55 of 2,416 knowledge pages exceed 6,000 characters (2.3%);
the largest is 33,054, where four fifths of the page would be dropped.

Same family as #292 and #287, one path further along.

## The fix

`src/core/clamp-page-sections.ts`:

- clamps in whole `##` sections, dropping from the end, never mid-sentence;
- names the omitted sections in the prompt text, so the model knows what it is not being shown
  and does not try to reproduce it;
- returns the withheld blocks verbatim, so `restoreWithheldSections` can put them back after
  the rewrite;
- below the budget returns the input **byte for byte** — 97.7% of pages on that vault, so the
  ordinary path is untouched;
- treats a budget of `0` as no clamp, matching how the codebase reads an unset numeric limit
  elsewhere.

One case cannot be made safe: a page over budget with no `##` boundary to cut at. Nothing can be
withheld and restored, so the rewrite is refused with an error saying why, rather than writing
back a page the model was never shown. That is the `hardCut` flag on the result.

The contradiction record is clamped through the same helper. It is not written back, so only
the reading is at stake there.

## Not in this PR

The merge triage (`page-factory/merge-triage.ts`) passes `existing_content` uncapped. I measured
before touching it and left it alone: median page 1,424 characters, p90 2,695 — the payload does
not grow with the vault, and adding a bound there would change behaviour without a reason. The
helper is available if a bound is ever wanted; the numbers are in #482.

## Tests

3427 pass (238 files). New: 11 cases on the clamp itself — byte-identity below budget, whole-
section dropping, the marker, verbatim withheld blocks in document order, the no-boundary hard
cut, and that a short tail section is not kept after a longer one is dropped (which would
reorder the page as the model sees it). Plus 3 on the wiring: the prompt admits the omission and
excludes the tail, the file keeps the tail anyway, and the unclamped case reaches the model
unchanged.

### DIFF
--- a/src/__tests__/core/clamp-page-sections.test.ts
+++ b/src/__tests__/core/clamp-page-sections.test.ts
@@ -1,0 +1,104 @@
+import { describe, it, expect } from 'vitest';
+import { clampPageSections, restoreWithheldSections } from '../../core/clamp-page-sections';
+
+const page = [
+  '---',
+  'type: entity',
+  'tags: [x]',
+  '---',
+  '',
+  '# Sulforaphane',
+  '',
+  'Lead paragraph.',
+  '',
+  '## Description',
+  'D'.repeat(200),
+  '',
+  '## Related Entities',
+  'R'.repeat(200),
+  '',
+  '## Mentions',
+  'M'.repeat(200),
+].join('\n');
+
+describe('clampPageSections', () => {
+  it('returns the input byte-identical when it fits', () => {
+    const out = clampPageSections(page, 10_000);
+    expect(out.text).toBe(page);
+    expect(out.withheld).toEqual([]);
+    expect(out.hardCut).toBe(false);
+  });
+
+  it('treats a budget of 0 as no clamp', () => {
+    expect(clampPageSections(page, 0).text).toBe(page);
+  });
+
+  it('drops whole sections from the end, never mid-sentence', () => {
+    const out = clampPageSections(page, 400);
+    expect(out.text).toContain('## Description');
+    expect(out.text).not.toContain('## Mentions\nM');
+    expect(out.hardCut).toBe(false);
+    // Nothing kept is a fragment: every retained section is present in full.
+    expect(out.text).toContain('D'.repeat(200));
+  });
+
+  it('names the omitted sections in the prompt text', () => {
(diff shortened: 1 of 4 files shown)


green-dalii (approved):
✅ LGTM. Closes a silent data-loss path that affects 2.3% of knowledge pages on a typical vault — the model was never shown the tail, the prompt asked for the complete page, and the answer was written back over the original. The fix is section-shaped (not character-shaped), so the model accounts for what it did not see via the marker, and the caller can put it back via `restoreWithheldSections`. The `hardCut` branch is the right shape: refuse rather than silently corrupt. Merge-ready. Note for squash commit: Gate 4 Performance table needs CPU/Memory/IO/Network/Token five-dim (CLAUDE.md §"Gate 4"). Flag in CHANGELOG: lint contradiction-fix no longer silently loses tail content on pages above 6,000 chars; pages without `## ` boundaries now refuse rather than write back incomplete rewrites.

Issue #473, opened 2026-08-15
Final lint analysis sends uncapped full-wiki context, causing HTTP 400 (313,754-token request)


jameses-cyber:
## Bug Description

Running LLM Wiki lint on a large existing wiki eventually sends a single request containing approximately 313,754 input tokens. The configured local model has a 49,152-token context window, so LM Studio rejects the request with HTTP 400.

The lint process successfully completes numerous earlier duplicate-detection requests, but the final general health-analysis request exceeds the context window. Obsidian then displays only `Lint failed: 400`; the normal lint report and corrective-action dialog never appear.

The request appears to include the full generated `wiki/index.md` and lint findings without applying a model-aware input-token budget, truncation, summarization, or batching strategy.

## Environment

- **Plugin version:** 1.26.3
- **Obsidian version:** 1.13.4
- **OS:** macOS 26.6 (Build 25G72)
- **Wiki size:** Approximately 1,690 Markdown pages

## LLM Configuration

- **Provider:** LM Studio
- **Model:** `ministral-3-14b-instruct-2512`
- **API endpoint:** `http://localhost:1234/v1`
- **Local or cloud:** Local (LM Studio)
- **Loaded context length:** 49,152 tokens
- **Model-advertised maximum context:** 262,144 tokens
- **Plugin max output tokens per call:** `0` (no plugin cap)
- **Plugin LLM concurrency:** 5
- **Auto Smart Fix:** Off
- **Periodic Lint:** Off
- **Include sources in dedup:** Off
- **Minimum body similarity:** 0.35

## Steps to Reproduce

1. Create or use an existing generated wiki containing approximately 1,690 Markdown pages.
2. Configure LLM Wiki to use LM Studio through `http://localhost:1234/v1`.
3. Load `ministral-3-14b-instruct-2512` with a 49,152-token context window.
4. Run the plugin’s manual wiki lint operation.
5. Allow duplicate detection and the preceding lint stages to run.
6. Wait for the final lint-analysis request.
7. Observe that LM Studio rejects a request containing 313,754 tokens.
8. Observe that Obsidian displays `Lint failed: 400` without showing the normal lint report or corrective-action dialog.

## Expected Behavior

Lint should complete and display its report regardless of wiki size, provided individual wiki pages are processable.

The plugin should enforce an input-token budget before sending the final health-analysis request. Large indexes and findings should be divided into bounded batches, summarized, sampled, or otherwise reduced to fit the selected model’s context window.

[...]

PR #495, opened 2026-08-19
chore: bump version to 1.26.4

### LINKED ISSUE #449: `cacheBreakpoint` is set on every extraction call, but no client reads it — prompt caching never reaches the wire

green-dalii:
v1.26.4 PATCH — 6-axis fix batch:

  • #449: cache_control on user-message prefix (Anthropic prompt-cache render order, not system block)
  • #452: slug catalog frozen at run start (4.4× speedup on 2844-slug benchmark)
  • #473: LintReportModal no longer drags in full wiki/index.md
  • #481: per-step output mode + thinking policy (opt-in)
  • #482: ingest payload stops growing with vault size (Stages 1+2)
  • #492: contradiction clamping restores withheld content

Post-merge CODE Gate cleanup:
  • extractReasoningText helper consolidates 7 sites in 3 SDK clients
  • parseTaskPolicySpec rejects empty task names ("=mode:thinking" silent-pass)
  • extractThinkingBlocks regex now case-insensitive on "Thinking…Response" form

20 merge commits / 3434 tests / +3786/-1816.

Closes #449, #452, #473, #481, #482, #492

Obsidian Bot pre-review: 0 Errors, 0 new Warnings on touched code. (Pre-existing structural Warnings in tools/ are accepted-baseline per CLAUDE.md "Bot compliance invariant" — v1.26.2 accepted, not modified by this release.)


green-dalii:
**Post-merge audit note**: GitHub GraphQL blocks self-author reviews (this PR's commit is the maintainer's), so the formal `--approve` event could not be posted. Per CLAUDE.md "Git Safety Protocol → Post-merge audit trail" (PR #478 reference 2026-08-18), this comment records the procedural miss. The merge commit hash stands; only the audit trail is patched here.

**Effective review verdict (recorded for audit):** APPROVED. Bot pre-review triaged clean (0 Errors, 0 new Warnings on touched code). 6-axis fix batch + CODE Gate cleanup — all Gate 1-6 ✅.

PR #497, opened 2026-08-19
test(llm-sdk): assert the cache_control contract at the wire, not at the mock

### LINKED ISSUE #493: `{{batch_context}}` marker: a missing marker silently truncates the extraction prompt and zeroes the cache breakpoint
`SourceAnalyzer.analyzeSource` splits the rendered `analyzeSource` template at the `{{batch_context}}` marker (`src/wiki/source-analyzer.ts:296-299`):

```ts
const batchMarker = '{{batch_context}}';
const markerIdx = templateUntouched.indexOf(batchMarker);
const staticPrefix = templateUntouched.substring(0, markerIdx);
const suffixTemplate = templateUntouched.substring(markerIdx + batchMarker.length);
```

`indexOf` returns `-1` when the marker is absent, and neither line guards against it. Both then compute something valid-looking rather than failing:

- `substring(0, -1)` clamps to `substring(0, 0)` → `staticPrefix` is `''`.
- `substring(-1 + 17)` → `substring(16)`, so `suffixTemplate` silently loses the **first 16 characters** of the template. With the current text that means the prompt starts at `nowledge base maintainer.` — the opening instruction line is gone.

The second consequence is the one that surprised me while reviewing #464: `cacheBreakpoint: staticPrefix.length` (`:404`) becomes `0`, and a zero breakpoint is not a smaller cache — it is **no cache at all**. The AI SDK core drops empty text parts in the user branch unconditionally (`ai/dist/index.mjs:1451`), so the prefix part carrying `cache_control` disappears before the request is serialized. I measured this against the real adapter with a stub `fetch`: at offset 0 the request body contains one user block and no `cache_control` anywhere.

**This is latent, not a live defect.** The template is a compile-time constant in `src/wiki/prompts/ingestion.ts` with no user override, so today `markerIdx` is always valid. What makes it worth a guard is that the failure mode is invisible in both directions: a contributor who renames the marker, moves it into a partial, or extracts the template into a file gets a subtly truncated prompt and silently loses prompt caching, with no error, no warning, and output that still looks plausible.

A fail-fast is three lines:

```ts
const markerIdx = templateUntouched.indexOf(batchMarker);
if (markerIdx === -1) {
  throw new Error(`analyzeSource template is missing the ${batchMarker} marker`);
}
```

Related: #449 / #464 for the caching half. I'd suggest keeping the client-side guard from that PR anyway — one side defends the contract, the other defends the consumer.

DocTpoint:
Follow-up to the #464 review. Small, and deliberately narrower than it may look at first glance.

## The test is the point

Every existing `cacheBreakpoint` test runs against a module-level mock of `generateText`, so it asserts the shape we hand the SDK — not the shape the SDK puts on the wire. Those two came apart once already: the marker on `c985196` sat on a block the adapter never caches, and no part-level assertion could have seen it.

`anthropic-cache-control-wire.test.ts` does not mock `ai`. It drives the real `@ai-sdk/anthropic` adapter through a stub `fetch` and reads the request body:

| case | asserted on the wire |
|---|---|
| non-zero offset | two text blocks, `cache_control: {type:'ephemeral'}` on the first, nothing on the second |
| offset 0 | one block, no `cache_control` anywhere in the body |
| no breakpoint | one block, unchanged |
| `'Ü'.repeat(42)` | the cut lands at code unit 42, not byte 42 |

That turns the claim in the v1.26.4 notes into something a future SDK bump has to keep true, rather than something we re-derive by reading `index.mjs`.

## The guard, and what it is not

`buildMessagesWithCacheControl` built a prefix part at offset 0 that the SDK core drops in the user branch (`ai/dist/index.mjs:1451` — the assistant branch two lines below keeps empty parts when `providerOptions` is set), so the body shipped without a marker and nothing signalled it. The helper now declines the split when the clamped offset is 0.

**Measured both ways: this is not a wire change.** The suite passes identically with and without the guard on the official adapter, because the adapter drops the empty part either way. What it buys:

- the code stops constructing a part it knows will be discarded, so the shape we build and the shape that ships agree;
- an empty text block cannot reach an Anthropic-*compatible* endpoint that does not filter it the way the official adapter does (not measured — stated as a possibility, not a finding);
- the condition gets a documented place, next to the clamp that can reach it from a negative offset.

**This does not close #493.** The caller-side half — `markerIdx === -1` at `source-analyzer.ts:278`, which drops the first 16 characters of the template and sets the offset to 0 — is untouched and stays with that issue for you to scope.

## Two names, corrected in passing

Both are in the file this PR already touches, and both travel:

- **`byte offset` → UTF-16 code-unit offset** (`String.length`). The existing `'Ü'.repeat(42)` test passes precisely *because* the semantics are not bytes: 42 characters are 84 UTF-8 bytes. Fixed in the JSDoc, in that test's name, and in the CHANGELOG entry for #449.
- **`below Anthropic's 1024-token minimum cache floor` → model-dependent.** The floor is not a constant and not monotonic across generations — 512, 1024, 2048 and 4096 tokens depending on the model. Your conclusion holds either way and holds *more strongly* for the 4096 models; the concern is only that a fixed number in a comment invites the next caller to code against it.

## Gate

`pnpm gate:1` green on the branch: 236 files / 3438 tests, build and css-lint clean.


green-dalii:
PR #497 is review-ready. Gate 1 CI green on DocTpoint's local. Approval pending.

**Action needed from @DocTpoint:** this branch's head (`fix/493-empty-prefix-guard`) is not pushed to origin, so GitHub cannot compute the merge against the current `main` (which has advanced through #500 / #498 / #499 merges today). Two small steps unblock the merge:

1. Push the branch: `git push origin fix/493-empty-prefix-guard`
2. The CHANGELOG.md edit (`§1.26.4` paragraph correction) will conflict with the docs-orthogonalization merge (PR #500, commit `f1fed39`) because the whole file was restructured — please rebase onto current `main` and resolve by re-applying the two factual corrections (`byte offset` → `UTF-16 code-unit offset`, `1024-token floor` → `model-dependent 512-4096`) to the corresponding sentences in the rewritten `§1.26.4` entry. The `#493 partial` Unreleased entry is unaffected.

Once rebased and Gate 1 is green again on the rebase commit, ping here and I'll approve + merge for the v1.27.0 MINOR train.

Spec review from my side (architect-level exemption applies): the offset===0 guard + wire-level test are exactly the SDK-side fix the original Issue #449 mock-based tests missed. The two CHANGELOG fact corrections are also valid — `String.length` is the right reference for an Anthropic cache breakpoint, and 1024 was a stale spec number. Ship-ready modulo the rebase.

DocTpoint:
Rebased onto current `main` (`bbffb63`). Gate 1 green on the rebase commit: lint, typecheck, 3437 tests, build, css-lint — CI confirms.

On the conflict: `CHANGELOG.md` was the only one, and only in the `[Unreleased]` block. I kept your rewritten Status line from #500 verbatim and re-added the `### Fixed` section with the `#493 partial` entry underneath it. The two factual corrections in the `§1.26.4` entry — `byte offset` → UTF-16 code-unit offset (`String.length`), and the fixed `1024-token floor` → model-dependent 512-4096, not monotonic across generations — rebased cleanly into the rewritten paragraph and read correctly there; I verified both by string match rather than by eye.

One wording fix while I was in the file: my entry said "two corrections to the v1.26.4 entry **above**". `[Unreleased]` sits above `[1.26.4]`, so it is "below". Changed. Nothing else in the entry moved.

A note on the branch head, in case it matters for your tooling: `fix/493-empty-prefix-guard` was pushed before the rebase — the PR head was `8f42249` and matched my local tree. The merge was blocked by the `CHANGELOG.md` conflict against #500, not by a missing push. The force-push replaced `8f42249` with `bbffb63`; the diff is the same four files, only the changelog hunk differs.

green-dalii (approved):
Approve. The wire-level test pins the contract at the actual HTTP request body (not module mocks), and the `offset === 0` guard removes the empty-part discard surface that part-level assertions can't catch — the kind of silent-no-op regression that `cache_control` on a wrong block exemplifies. Two docstring corrections (UTF-16 code units vs bytes; model-dependent cache floor) are honest and remove a future trap where the next caller might code against a fixed constant. The PR body correctly notes this does not close #493 (caller-side `markerIdx === -1` truncation stays with that issue for the v1.27.0 scope) — ship as the wire-test half of #493 now and let the caller-side half land in its own PR.

Issue #503, opened 2026-08-20
Folder ingest ignores Obsidian's own "Excluded files" list (userIgnoreFilters)


DocTpoint:
## Summary

Obsidian has a built-in exclusion list: **Settings → Files and links → Excluded files**, stored as `userIgnoreFilters` in `.obsidian/app.json`. The plugin does not read it. A folder ingest collects files the user has already declared out of scope.

In my vault that list is populated:

```json
"userIgnoreFilters": ["copilot/copilot-custom-prompts/", "#Definition", "wiki/sources/"]
```

`grep` over `src/` finds no reference to `userIgnoreFilters`, `getConfig`, or any equivalent.

## Why this matters more than a new setting

There is currently no plugin-side way to exclude anything either. `LLMWikiSettings` has 68 fields and none of them carries exclusion semantics — the collection step filters on extension and folder boundary only. So the practical effect is that documentation and meta files get ingested as sources: in my vault `_Foundation.md`, `_Handbuch.md`, `_Referenz.md` and `_Baupläne.md` are all ingest candidates.

My first thought was to propose an ignore-glob setting (`_*`, `.*`, configurable). Reading Obsidian's own configuration changed my mind: the user has usually declared this once already, in Obsidian's UI, and a second list in the plugin's settings would need to be kept in sync by hand.

## The decision this needs

`app.vault.getConfig('userIgnoreFilters')` is not part of the published `obsidian.d.ts`. Honouring the list therefore means either using an untyped accessor or reading `app.json` directly, and that is a call for you rather than something I want to put in a PR unannounced. Filing it as an issue for that reason.

If a private accessor is unacceptable, the fallback is a plugin-side ignore list — with the caveat above that users then maintain two.

## Not checked

Whether `userIgnoreFilters` patterns are plain path prefixes, globs, or regex in current Obsidian; the three entries in my own vault happen to be a folder prefix, a tag-like string, and a folder prefix. Any implementation would need to match Obsidian's own semantics rather than guess.

green-dalii:
Decision: option (a) — `vault.getConfig('userIgnoreFilters')` behind a narrow typed interface, not app.json parsing and not a second plugin-side list. Before implementation we need Obsidian's actual matching semantics pinned down (prefix vs substring; your `#Definition` entry suggests non-path entries exist and must not silently match nothing), so this sits in the v1.27.0+ research track until that verification is done — promoted the moment it is.

DocTpoint:
[...]

PR #504, opened 2026-08-20
fix(#502, #505): apply the picker's own exclusion rule wherever the vault root can be chosen

### LINKED ISSUE #502: Choosing the vault root for "Ingest from folder" reinstates the wiki/ folder the picker excludes
## What happens

`Karpathy LLM Wiki: Ingest from folder` offers the vault root as a choice. Choosing it collects every file in the vault with a compatible extension — including everything under `wiki/`. The plugin then ingests its own generated pages as source material.

The folder picker deliberately hides `wiki/` as a choice. Selecting the root puts it back.

## Why this is a composition, not a mistake

Both halves are intentional, and both are pinned by tests in `src/__tests__/core/folder-scope.test.ts`:

```ts
it('keeps the vault root selectable', () => {
  expect(isExcludedFromSourcePicker('/', 'wiki', '.obsidian')).toBe(false);
});

it('accepts every path at the vault root', () => {
  expect(isInFolderScope('deep/nested/a.md', '/', true)).toBe(true);
});
```

Neither is wrong. The gap is where they meet — `src/main-commands/ingest-commands.ts`:

```ts
const files = this.app.vault.getFiles()
  .filter(f => allowedExts.includes(f.extension.toLowerCase()))
  .filter(f => isInFolderScope(f.path, folder.path, folder.isRoot()));
```

At the root, `isInFolderScope` has an empty prefix, so every path matches and no exclusion rule is left.

The asymmetry is visible across the three entry points. `FileSuggestModal` (`suggest-modals.ts:34`) and `MultiFileSuggestModal` (`MultiFileSuggestModal-class.ts:102`) both apply `isExcludedFromSourcePicker` to the **files** they offer. `FolderSuggestModal` applies it to the **folders** it offers (`suggest-modals.ts:61`), and the collection step that follows applies it to nothing.

## Scale

In a vault of mine, a root ingest would collect **3596 files, 3121 of them under `wiki/`** — 87% of the batch would be the plugin's own output.

## Reproduce

1. Run `Karpathy LLM Wiki: Ingest from folder`.
2. Choose the vault root.
3. The confirmation reports the whole vault, `wiki/` included.

## Fix

PR attached. It adds the missing file-level rule next to the other three in `core/folder-scope.ts` and uses it at the collection step, so all three entry points agree on what counts as a source.

## Not covered

The auto-watch path has its own prefix matcher in `schema/auto-maintain.ts` rather than the shared helper. It is correctly anchored and the root case cannot arise there, so I left it alone.

### LINKED ISSUE #505: Choosing the vault root as a watched folder stores / and watches nothing
## What happens

The watched-folder picker in **Settings → Auto-maintain → Watched folders** is the same `FolderSuggestModal` the ingest command uses, so it offers the vault root. Choosing it stores `/`:

```ts
// ui/settings-sections/auto-maintain-section.ts
const folderPath = folder.path.endsWith('/') ? folder.path : `${folder.path}/`;
```

The root's path is already `/`, so `/` is what lands in `watchedFolders`. The matcher then asks:

```ts
// schema/auto-maintain.ts
private isWatched(path: string): boolean {
  if (this.settings.watchedFolders.length === 0) return false;
  return this.settings.watchedFolders.some(folder => {
    const prefix = folder.endsWith('/') ? folder : `${folder}/`;
    return path.startsWith(prefix);
  });
}
```

No Obsidian vault path begins with a slash, so the entry matches nothing. The settings list shows `/` as a watched folder and the watcher watches nothing — a silent no-op, with no error and no log line to hint at it.

## Relation to #502

Same picker, same root entry, but a different defect. #502 is over-inclusion: the folder ingest collects `wiki/` because root scope has no exclusion rule left. This one is the opposite symptom, from a different cause — `isWatched` hand-rolls a prefix comparison instead of using the shared `folder-scope` helpers, and that hand-rolled form has no notion of a root sentinel.

Worth noting for whoever fixes it: repairing the root case alone would make the watcher watch `wiki/`, so the picker's exclusion rule has to come along at the same time. The two fixes want the same helper.

## Behaviour change this implies

Anyone who has `/` stored today goes from "nothing happens" to "auto-ingest runs on the next edit anywhere in the vault". That is what they asked for — the entry only exists because they enabled auto-watch and deliberately picked the root — but it is a real change, and it starts costing LLM calls from that point.

If you would rather not make that jump, the alternative is to keep the root out of the watched-folder picker and drop existing `/` entries on load. Both fix the "UI accepts an input it silently ignores" defect; they differ only in which way the ambiguity is resolved. I went with honouring the selection, for consistency with what #502 establishes for the sibling path.

## Fix

Folded into #504, because it uses the shared rule that PR introduces. A separate PR against `main` would not compile.

DocTpoint:
Closes #502. Closes #505.

## The change

Two sites let the user choose the **vault root**, and neither handled it. They are the same `FolderSuggestModal`, and both were missing the same rule underneath — but they failed in opposite directions.

**#502 — over-inclusion.** `Ingest from folder` collected every path at the root, so choosing it reinstated the `wiki/` folder the picker deliberately refuses to offer, and the plugin ingested its own generated pages as source material.

**#505 — a silent no-op.** The watched-folder picker stores the root as `/`, and `isWatched` asked `path.startsWith('/')`. No Obsidian vault path begins with a slash, so the entry matched nothing: the settings list showed `/` while the watcher watched nothing, with no error and no log line.

The rule both sites needed already exists — `isExcludedFromSourcePicker`, which `FileSuggestModal` and `MultiFileSuggestModal` already apply to the files they offer. This PR adds `isIngestableSource` next to the other three rules in `core/folder-scope.ts`, in the same pure/IO-free shape so it is unit-testable without a vault, and uses it at both sites. All four entry points now agree on what counts as a source.

For #502 the premises are each deliberate and separately tested (`keeps the vault root selectable`, `accepts every path at the vault root`); only their composition was missing a rule. For #505 the fix also has to bring the exclusion along, or a repaired root would start watching `wiki/`.

## Behaviour change worth calling out

An existing `/` entry in `watchedFolders` goes from doing nothing to watching the whole vault. That is what the user asked for — the entry only exists because they enabled auto-watch and picked the root — but it is a real change, and it starts costing LLM calls on the next edit.

If you would rather not make that jump, the alternative is to keep the root out of the watched-folder picker and drop existing `/` entries on load. Both resolve "the UI accepts an input it silently ignores"; they differ only in which way. I chose to honour the selection, so the two sites agree. Say the word and I will turn it around — it is a small change either way.

## Tests

Nine cases added, red before each fix and green after.

`folder-scope.test.ts` (5), next to the two that pin the premises: an ordinary file inside a chosen folder is collected · an ordinary file is collected at the root · `wiki/entities/x.md` is **not** collected at the root, asserting both premises first so the test documents why the third assertion has to hold · the inherited folder boundary still rejects `Notizen-temp/a.md` and `Notizen.md` · `wiki-archive/x.md` is still collected, so the name-prefix sibling is not swept up.

`auto-maintain.test.ts` (4): ordinary vault files are watched at the root · `wiki/` is not · an empty list watches nothing · a named folder stays anchored whether or not it was stored with a trailing slash.

The existing batch-context harness gains a vault stub and a `wikiFolder`, because the rule it now consults reads both.

## Gate

`pnpm lint --max-warnings=0` · `npx tsc --noEmit` · `pnpm build` · `pnpm test` · `pnpm css-lint` — all green on `7aed9eb`. Tests 3442 passed, of which 9 are new.

One note on ordering: `openai-codex-loopback-flow.test.ts` reads the built `main.js` from the working directory, so `pnpm test` fails on a clean tree unless `pnpm build` ran first. `pr-ci.yml` already orders it that way; worth knowing if you ever run the five commands by hand.

## Not in this PR

Honouring Obsidian's own **Excluded files** list (`userIgnoreFilters`) — filed separately as #503. It needs an untyped accessor or a direct read of `app.json`, which is a design call rather than a fix, so it did not belong in a PR sent unannounced.

### DIFF
--- a/src/__tests__/core/folder-scope.test.ts
+++ b/src/__tests__/core/folder-scope.test.ts
@@ -4,6 +4,7 @@ import {
   isInFolderScope,
   isAtOrInFolderScope,
   isExcludedFromSourcePicker,
+  isIngestableSource,
 } from '../../core/folder-scope';
 
 describe('folderScopePrefix', () => {
@@ -118,5 +119,45 @@ describe('isExcludedFromSourcePicker', () => {
 
   it('keeps the vault root selectable', () => {
     expect(isExcludedFromSourcePicker('/', 'wiki', '.obsidian')).toBe(false);
+  });
+});
+
+// Issue #502 — the two rules above are each correct on their own, and the
+// two tests directly above pin them: the vault root stays selectable, and
+// root scope accepts every path. Composed at the folder-ingest call site
+// they cancel the picker's own boundary — choosing the root reinstates the
+// `wiki/` folder the picker deliberately refuses to offer, and the plugin
+// ingests its own generated pages as source material.
+//
+// `FileSuggestModal` and `MultiFileSuggestModal` already apply the picker
+// rule to the FILES they offer. `FolderSuggestModal` applies it to the
+// FOLDERS it offers, and the collection step that follows applied nothing.
+// This function is that missing file-level rule, in the same module as the
+// other three.
+describe('isIngestableSource', () => {
+  it('collects an ordinary source file inside the chosen folder', () => {
+    expect(isIngestableSource('Notizen/a.md', 'Notizen', false, 'wiki', '.obsidian')).toBe(true);
+  });
+
+  it('collects an ordinary source file when the vault root is chosen', () => {
+    expect(isIngestableSource('Notizen/a.md', '/', true, 'wiki', '.obsidian')).toBe(true);
+    expect(isIngestableSource('a.md', '/', true, 'wiki', '.obsidian')).toBe(true);
+  });
+
+  it('does not reinstate the wiki folder when the vault root is chosen', () => {
+    // Both premises hold and are pinned above ...
+    expect(isExcludedFromSourcePicker('wiki/entities/x.md', 'wiki', '.obsidian')).toBe(true);
+    expect(isInFolderScope('wiki/entities/x.md', '/', true)).toBe(true);
+    // ... so the collection step has to combine them.
+    expect(isIngestableSource('wiki/entities/x.md', '/', true, 'wiki', '.obsidian')).toBe(false);
+  });
+
+  it('keeps the folder boundary it inherits from isInFolderScope', () => {
+    expect(isIngestableSource('Notizen-temp/a.md', 'Notizen', false, 'wiki', '.obsidian')).toBe(false);
+    expect(isIngestableSource('Notizen.md', 'Notizen', false, 'wiki', '.obsidian')).toBe(false);
+  });
+
+  it('keeps a sibling folder sharing the wiki name prefix', () => {
+    expect(isIngestableSource('wiki-archive/x.md', '/', true, 'wiki', '.obsidian')).toBe(true);
   });
 });
--- a/src/__tests__/schema/auto-maintain.test.ts
+++ b/src/__tests__/schema/auto-maintain.test.ts
@@ -26,10 +26,12 @@ function makeManager(ingestSource: IngestFn, createBatchContext: BatchCtxFn) {
     autoWatchMode: 'auto',
     autoWatchDebounceMs: 0,
     watchedFolders: ['inbox'],
+    wikiFolder: 'wiki',
   } as unknown as LLMWikiSettings;
 
   const wikiEngine = { createBatchContext, ingestSource } as unknown as WikiEngine;
-  const app = {} as unknown as ConstructorParameters<typeof AutoMaintainManager>[0];
+  // isWatched() consults the shared exclusion rule, which needs the config dir.
+  const app = { vault: { configDir: '.obsidian' } } as unknown as ConstructorParameters<typeof AutoMaintainManager>[0];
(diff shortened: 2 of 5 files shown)


green-dalii (approved):
Approve. Real bug: `/` in `watchedFolders` was a silent no-op (Issue #505 — `path.startsWith('/')` never matched any real vault path) and the root+folder-picker let `wiki/` be re-ingested as source material (Issue #502). The new `isIngestableSource` composes the two rules that were already individually correct, in the same pure/IO-free module as `isExcludedFromSourcePicker` and `isInFolderScope`, which is the right home — `FileSuggestModal` and `MultiFileSuggestModal` already apply the picker rule to the files they offer, so all three entry points now agree.

The behaviour change for existing `/` entries (silently no-op → watches the whole vault minus `wiki/)` is real and correctly flagged in the PR body. One suggestion: add a one-line note to the v1.27.0 MINOR release notes so users who previously set `/` as a placeholder aren't surprised by the new ingest behaviour on their next vault edit — the CHANGELOG entry at `## [1.26.5]` should mention this too. Nine-case test coverage is solid; the fixture updates (`wikiFolder` + `app.vault.configDir`) document the rule's new dependencies.

Ship-ready; honouring the user's selection rather than dropping `/` on load is the right call.

PR #508, opened 2026-08-21
docs(changelog): record #504 and the upgrade note for existing `/` entries

DocTpoint:
Follows your suggestion on #504: a note in the v1.27.0 release notes so users who set `/` as a placeholder are not surprised on the next run.

One entry under `[Unreleased] → Fixed`, no code. The upgrade note is the last two sentences:

> **Upgrade note:** a `/` entry that previously did nothing now watches the whole vault except `wiki/`. If you set `/` as a placeholder, remove it or narrow it to the folders you actually want ingested before the next run.

Nothing else in the file is touched. If you would rather fold this into the v1.27.0 entry when you cut the release, close this and take the wording from here.


green-dalii (approved):
Wording taken as written — merging now so the note rides [Unreleased] into the v1.27.0 entry at release time rather than being rewritten there.

Issue #91, opened 2026-06-03
Nested tags for entity/concept pages based on source note tag hierarchy


DocTpoint:
Currently, entity and concept pages receive flat domain tags (e.g. `Arzneimittel`, `Neurologie`). This loses the hierarchical relationship between domain categories. Obsidian supports nested tags via `/`-notation, where searching `#Arzneimittel` automatically includes all subtags like `#Arzneimittel/Neurologie`.

**Proposed behavior:**
During ingest, entity/concept pages receive nested tags combining the source note's tags:
```yaml
# Source note tags: [Arzneimittel, Neurologie]
# → Entity gets:
tags:
  - Arzneimittel/Neurologie
```

This creates a phylogenetic-style classification tree: top-level domains as phyla, sub-domains as lower ranks. The tag pane in Obsidian then renders a collapsible taxonomy tree across the entire vault.

**Connection to #85:**
The user-configurable tag vocabulary defines the valid top-level nodes (phyla). The LLM selects the appropriate nesting during ingest.

**Connection to #86:**
Source pages should inherit the flat original tags; entity/concept pages derive the nested form from them.

green-dalii:
Nice framing — the "phylogenetic taxonomy" is clean, and Obsidian's tag pane does render nested tags as a collapsible tree.

**Dependency chain**:
1. #85 first (controlled vocabulary) — LLM needs "valid phyla" to nest under.
2. This — add nesting rules to the generation prompt; update the frontmatter serializer to emit `tags: [Arzneimittel/Neurologie]` as a flat string list (not nested arrays — Obsidian expects `/`).
3. #90 as cleanup — apply the same logic to source pages.

[Speculation, not verified] `[[wiki-links]]` resolution may not auto-handle nested tags — `[[entities/Arzneimittel/Neurologie]]` could need a custom resolver or a different separator. Worth a quick test in a sample vault before committing to the design.

P3 priority makes sense — let's see #85 in the wild first.

DocTpoint:
Thanks for the dependency breakdown — that order makes sense.

The [[wiki-link]] / separator conflict is a good catch. Two possible mitigations worth testing:

Use :: as separator instead of / (e.g. Arzneimittel::Neurologie) — but this breaks Obsidian's native nested tag rendering
Keep / for tags only (no wiki-link with the same path) and document the constraint
P3 after #85 ships makes sense — happy to test the design once #85 is in the wild.

green-dalii:
[...]

PR #510, opened 2026-08-21
fix(frontmatter): union tags on merge so the stored tag does not depend on source order

### LINKED ISSUE #509: `mergeFrontmatter` unions `sources:` but not `tags:` — the stored tag is decided by whichever source reached the page first
## What happens

`mergeFrontmatter` (`src/core/frontmatter.ts`) is the path every existing entity or concept page goes through when a second source touches it. Two fields sit next to each other in the same `serializeFrontmatter` call:

```ts
sources: mergedSources,                        // union of existing + new
tags: Array.isArray(fm.tags) ? fm.tags : [],   // existing only
```

The incoming tag is not merged, and it is not overwritten either — it never arrives. `mergeFrontmatter(existingContent, newSourcePath)` has no parameter that could carry it. The value exists at every call site: `info.type` is the term a page carries as its own `tags:` (`create-page.ts:97`, generation template `tags: [{{entity_type}}]`), and all three callers — `merge-page.ts:94`, `merge-page.ts:289`, `related-page.ts:87` — hold the `info` it comes from.

## Why it matters

The result is order-dependent. A page whose first source classified it `organization` keeps `organization` after a second source classifies it `person`, and the reverse ordering of the same two sources yields the other answer. Nothing records that a second opinion was formed and discarded.

On my vault this is not a corner case: **828 of 2,416 entity and concept pages (34.3%) carry two or more sources** — measured today against `603c9ee`. Each of them had at least one type decision computed by the extraction step, paid for in tokens, and dropped at the merge.

This is the same order-invariance property `sources:` is a union for.

## Suggested fix

An optional third parameter, unioned like `sources`:

```ts
export function mergeFrontmatter(
  existingContent: string,
  newSourcePath: string,
  incomingTags?: string[]
)
```

Omitting the argument keeps the previous behaviour byte for byte. First occurrence wins, so an established tag keeps its position and a repeated merge is idempotent.

Union is not one option among several here: it is the only one of the obvious candidates that is order-invariant. "First wins" is today's behaviour and depends on arrival order; "last wins" depends on it too, in the other direction. A PR with this follows immediately, tested against `603c9ee`, including an order-invariance case — a page reached by two sources in either order ends with the same tag set.

## Not in this

Whether a wiki page should carry more than one type term at all, and what its tags should mean beyond the extraction type, is a separate design question — the one parked in #91 by mutual agreement. This is the mechanical half only: a value that is computed and cannot reach the field it was computed for.

DocTpoint:
Closes #509 — the mechanical half. The design question about what a wiki page's tags should mean stays parked in #91.

## What it does

`mergeFrontmatter` takes an optional third parameter and unions it into `tags:`, the way `sources:` two lines above has always been unioned. The three callers pass the type term this source extracted:

- `merge-page.ts:94` and `merge-page.ts:290` — `info.type`
- `related-page.ts` — `newInfo?.type`. The `newInfo` lookup moved ten lines up so the merge can see it; it is a pure `find` over `analysis`, no IO, and the skip decision below reads the same value.

Omitting the argument is byte-identical to the previous behaviour, so every other caller and every existing test is untouched.

## One decision for you, and it changes stored data

After this, a page reached by two sources that classified it differently carries **both** terms — `tags:` gains a second entry where it previously had one. The generation template writes a single `tags: [{{entity_type}}]`, so this produces a shape the template itself never produces.

I built it as a union because it is the only order-invariant choice available: "first wins" is today's behaviour and depends on arrival order, "last wins" depends on it in the other direction. But the trade is real — you buy order-invariance with pages that can carry more than one type term, and that spreads through the vault on the next ingest rather than on a migration you schedule.

If you would rather keep one term per page, the same seam supports it: pass the incoming tags and let `unionTags` pick instead of merge — first-wins plus a debug line naming the discarded term, so the disagreement stops being invisible even if the stored value does not change. Say which and I will reshape it; I did not want to pick the quieter option on your behalf without saying so.

## Tests

7 new, in `frontmatter.test.ts`:

- byte-identical when no incoming tags are passed
- unions an incoming tag with the ones the page already carries
- keeps the existing tag first, so order does not depend on the merge
- does not repeat an incoming tag the page already has
- drops empty incoming tags
- adopts the incoming tag when the page carries none
- a page reached by two sources in either order ends with the same tags

Local: `eslint` clean on the four touched files, `tsc --noEmit` clean, 322 tests green across `frontmatter.test.ts` and the 18 `page-factory` files. Full Gate 1 is CI's.

## Not in this

No migration for pages already carrying an order-decided tag. The stored value is corrected the next time a source touches the page, which is the same cadence every other frontmatter repair runs at; a sweep over existing pages is a separate decision with its own blast radius.

### DIFF
--- a/src/__tests__/core/frontmatter.test.ts
+++ b/src/__tests__/core/frontmatter.test.ts
@@ -653,6 +653,62 @@ Body
     expect(result.frontmatter).toContain('- "Foo"');
     expect(result.frontmatter).not.toContain('- ""');
   });
+
+  // Incoming tags: `sources:` was a union and `tags:` was not, so the stored
+  // tag was decided by whichever source reached the page first.
+  const tagged = '---\ntype: entity\ncreated: 2026-01-01\nupdated: 2026-01-01\ntags:\n  - person\n---\n\nBody';
+
+  it('is byte-identical when no incoming tags are passed', () => {
+    const withArg = mergeFrontmatter(tagged, 'sources/new', undefined);
+    const without = mergeFrontmatter(tagged, 'sources/new');
+    expect(withArg.frontmatter).toBe(without.frontmatter);
+    expect(without.frontmatter).toContain('- "person"');
+  });
+
+  it('unions an incoming tag with the ones the page already carries', () => {
+    const result = mergeFrontmatter(tagged, 'sources/new', ['organization']);
+    expect(result.frontmatter).toContain('- "person"');
+    expect(result.frontmatter).toContain('- "organization"');
+  });
+
+  it('keeps the existing tag first, so the order does not depend on the merge', () => {
+    const result = mergeFrontmatter(tagged, 'sources/new', ['organization']);
+    expect(result.frontmatter.indexOf('- "person"'))
+      .toBeLessThan(result.frontmatter.indexOf('- "organization"'));
+  });
+
+  it('does not repeat an incoming tag the page already has', () => {
+    const result = mergeFrontmatter(tagged, 'sources/new', ['person']);
+    expect((result.frontmatter.match(/- "person"/g) || []).length).toBe(1);
+  });
+
+  it('drops empty incoming tags', () => {
+    const result = mergeFrontmatter(tagged, 'sources/new', ['']);
+    expect(result.frontmatter).toContain('- "person"');
+    expect(result.frontmatter).not.toContain('- \n');
+  });
+
+  it('adopts the incoming tag when the page carries none', () => {
+    const untagged = '---\ntype: entity\ncreated: 2026-01-01\nupdated: 2026-01-01\n---\n\nBody';
+    const result = mergeFrontmatter(untagged, 'sources/new', ['method']);
+    expect(result.frontmatter).toContain('- "method"');
+  });
+
+  it('a page reached by two sources in either order ends with the same tags', () => {
+    const base = '---\ntype: entity\ncreated: 2026-01-01\nupdated: 2026-01-01\n---\n\nBody';
+    const aThenB = mergeFrontmatter(
+      `${mergeFrontmatter(base, 'sources/a', ['person']).frontmatter}\n\nBody`,
+      'sources/b', ['organization']).frontmatter;
+    const bThenA = mergeFrontmatter(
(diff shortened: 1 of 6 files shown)


DocTpoint:
Correcting my own PR before you spend time on it. As first filed it was right for the default vocabulary and wrong for a custom one. Fixed in `52f642b`; the detail is worth recording because it changes what this PR claims.

`VALID_ENTITY_TAGS` and `VALID_CONCEPT_TAGS` are the `info.type` enums, so under `tagVocabularyMode: 'default'` a page's tag and a source's extracted type are the same alphabet, and unioning the incoming type is coherent. Under `tagVocabularyMode: 'custom'` they are two alphabets. My own vault runs 34 domain terms (`Biochemie`, `Erkrankung`, …) with no type value among them; the first version of this PR would have written `other` or `phenomenon` into `tags:` on every merge there — manufacturing exactly the violations `runRetagViolations` exists to remove, on 828 pages in my case.

`52f642b` adds `incomingTypeTag(settings, kind, type)` to `core/tag-vocab.ts`: the extracted type is passed on only when the active vocabulary admits it. `mergeFrontmatter` stays settings-free and the policy sits where the vocabulary already lives. Five tests cover the gate; the seven merge tests are unchanged.

What this changes about the claim in #509: under a custom vocabulary the fix is a no-op, because nothing tag-shaped reaches the merge at all — the extraction supplies a type the vocabulary rejects, and the domain terms on those pages are written by the retag pass afterwards, not by the source. The order-dependence I reported is real in default mode; in custom mode it is a symptom of something larger, which I would rather file separately than widen this PR around.

green-dalii (approved):
**Decision on the open question: keep the union as built.** Order-invariance outweighs single-term purity, incomingTypeTag keeps custom vocabularies clean (no-op where the two alphabets diverge), and runRetagViolations stays the normalizer of last resort. Verified on main that mergeFrontmatter had no parameter that could carry info.type while sources was unioned two lines above; the 7-test set including the either-order case pins the property this exists for. Approving. The separate custom-vocabulary gap you flagged for its own issue is indeed separate — file it when ready.

Issue #472, opened 2026-08-15
A mis-typed extraction merges into the wrong page when a designator exists in both folders


DocTpoint:
Reported from the measurement in #446 / #471, filed on its own because it lives in the classification path rather than the resolution one.

## Mechanism

`ConflictResolver.resolve` (`src/core/conflict-resolver.ts`) checks same-type pages first and merges on a slug or alias match, with no semantic check on that branch. That is correct as long as the extraction assigned the right `pageType`. When it did not, and the opposite meaning happens to live in the folder the extraction picked, the item merges into a page about something else — silently, and deterministically.

`KHK` in my vault is the example:

- `wiki/entities/Ketohexokinase.md`, alias `KHK`
- `wiki/concepts/Koronare-Herzkrankheit.md`, alias `KHK`

Extract `KHK` meaning the enzyme, classify it as a **concept**, and the same-type branch finds the coronary-heart-disease page and merges into it. Nothing in the path can notice: the slug matches, the alias matches, and the resolution never reaches the semantic dedup call because a deterministic match was found.

The cross-type branch is not the problem here — it is the recovery. It only runs when the chosen folder holds no match, and then it reports a collision the caller merges into deliberately. The damaging case is the one where **both** folders hold a page for the designator.

## Scope

The precondition is a designator carried by pages in both folders with different meanings. In a 2416-page vault that is 15 designators — `KHK`, `IF` and 13 others. How often the extraction mis-types is **not measured**; the mechanism is provable from the code alone, the frequency is not.

## Not the same as #446

#446 / #471 is about a designator matching more than one page **in the same folder**, where the merge target depended on list order. This one is about the folder itself being the wrong one, and no ranking or ordering inside a folder can address it.

## What might help

A cheap version would be to consult the opposite folder before merging on a same-type alias match, and to route the case to the semantic dedup call when the designator exists in both — no new call for the ordinary case, one for the 15. Whether that is worth it depends on the mis-typing rate, which is the number neither of us has.

green-dalii:
[...]

PR #511, opened 2026-08-21
refactor(tools): migrate CLI to UPSTREAM DEV-ONLY INSTRUMENT + log measurement arms (Issue #507)

### LINKED ISSUE #507: refactor(tools): eliminate Obsidian Bot warnings + minimize CLI to UPSTREAM DEV-ONLY INSTRUMENT
# refactor(tools): eliminate Obsidian Bot warnings + minimize CLI to UPSTREAM DEV-ONLY INSTRUMENT

## Context

The Obsidian Bot 1.26.4 review surfaced ~70 findings (49 errors / 21 warnings) on `tools/llm-wiki-cli/`. CLAUDE.md previously classified these as "accepted-structural" — that classification was a temporary gate accommodation during v1.26.2 PR #442 to unblock Bot submission, **not** the v1.27.0 MINOR target state. With v1.27.0 MINOR scope we should clear the findings properly rather than carry them forward as accepted noise.

The sibling repo [`karpathywiki-cli` v0.1.0](https://github.com/green-dalii/obsidian-llm-wiki-cli) already ships as the production CLI replacement (published to npm 2026-08-12). The in-tree `tools/llm-wiki-cli/` is now over-spec for what it needs to do, and its structural cost (Bot score drag + maintenance surface) no longer matches its residual purpose: headless real-engine measurement (the 979s → 365s → 151s evidence in CLAUDE.md "Force-disable thinking" comes from `withUsageAccounting` at `main.ts:489-520`).

## Bot warnings to eliminate (verbatim from 1.26.4 review)

**`node:util` / `node:fs` / `node:path` / `node:fs/promises` static imports** (10 occurrences across `main.ts:16-18`, `vault.ts:12-14`, `run-llm-wiki.mjs:10-13`). Per `obsidianmd/no-nodejs-modules`, static imports are never exempt — only dynamic `await import()` guarded by `if (!Platform.isDesktop) throw` at function start (canonical pattern: `src/llm-sdk/openai-codex/loopback-flow.ts`).

**`.obsidian` hardcoded literal** (2 sites: `main.ts:451`, `vault.ts:25`). Replace with `path.join(vaultRoot, '.obsidian')` where `vaultRoot` comes from positional arg.

**`console.log` × 47 + 2** (`main.ts:534-740` + `obsidian.ts:58,62`). These are the CLI's stdout output interface. Replace with `process.stdout.write(s + '\n')` — `process` is a Node global, not a `node:*` import; `process.stdout.write` does not trigger the `obsidianmd/rule-custom-message` (no-console) heuristic.

**`globalThis` × 4** (`node-globals.ts:27,28,29,38`). These stub `globalThis.crypto` / `globalThis.TextDecoder` / `globalThis.URL` for the bundled production code. Node 20+ ships all three as globals natively — delete the shim and raise `engines.node` from `>=16` to `>=20`.

## Goals

1. **Eliminate all ~70 Bot findings from `tools/`** (target: 0 errors, 0 warnings).
2. **Preserve the measurement instrument** (`withUsageAccounting` + per-step token/latency output). This is the **only** headless path to run the real `WikiEngine` against real LLM cost (sibling repo has no `WikiEngine` — Layer 3 is "do not copy" per SPEC §4.2; Vitest covers mocked paths only).
3. **Drop user-facing CLI surface** from `package.json` (`bin.llm-wiki`, `scripts.llm-wiki`) — production CLI is `npx karpathywiki-cli`.
4. **Minimize tools/ scope** from 1,447 LOC + 680 LOC tests → ~320 LOC + ~30 LOC tests.

## Plan: per-warning elimination

| Bot warning | Count | Fix |
|---|---|---|
| `obsidianmd/no-nodejs-modules` (static) | 10 | Convert to `await import('node:*')` + `if (!Platform.isDesktop) throw` at function start |
| `obsidianmd/hardcoded-config-path` | 2 | Replace literal `.obsidian` with `path.join(vaultRoot, '.obsidian')`; `vaultRoot` from positional arg |
| `obsidianmd/rule-custom-message` (no-console wrapper) | 49 | Replace `console.log(...)` with `process.stdout.write(... + '\n')`; delete `Notice` class (engine ingest path doesn't call it) |
| `obsidianmd/no-global-this` | 4 | Delete `node-globals.ts`; raise `engines.node` to `>=20` (TextDecoder/URL/crypto.subtle are native globals in Node 20+) |
| `no-undef process` | 5 | Fix `eslint.tools-bot.config.mjs` to declare Node globals |
| `no-unsanitized/method` | 1 | `run-llm-wiki.mjs:82` typed require instead of `await import(<runtime-string>)` |
| `restrict-template-expressions` | 1 | Type annotation at `main.ts:411` |

**Result**: 70 findings → 0.

## Plan: scope minimization

**Delete**:
- `parseCliOptions`, `dispatchCli`, `applyOverrides`, `applyThinkingMode`, `parseNumber`, `printSummary`, `printWriteLog`, `resolveApiKey` (~300 LOC of flag-UX plumbing that user-facing CLI doesn't need anymore)
- 16 user-facing flags — replaced by positional args (`./run-instrument.mjs <vault> <wiki>`) — per decision point 3b (preserve arg capability but positional rather than flag)
- `--max-rounds` / `--thinking` deprecated entries (`main.ts:74`, `main.ts:90-91`) — promised removal in v1.26.0, currently 1.26.4 = 4 PATCH releases past due (CLAUDE.md dead-code half-life rule)
- `Notice` class from `obsidian.ts` shim
- `main.test.ts` (491 LOC parser-contract suite, all dead once flag UX goes)
- `obsidian-shim.test.ts` (after `Notice` removed; `requestUrl` test becomes unit-level on fetch wrapper)
- `node-globals.test.ts` (after shim file deleted)
- `package.json` `bin.llm-wiki` + `scripts.llm-wiki` + `scripts.typecheck:tools`

**Keep**:
- `runIngest` engine wiring (~150 LOC)
- `withUsageAccounting` + `printTimeByStep` (the measurement instrument — source of 979s → 365s → 151s evidence)
- `loadSettings(vaultRoot)` (derived configDir)
- `FsVaultReader` / `FsVaultWriter` (real vault on disk for engine)
- `requestUrl` → `fetch` wrapper
- `TFile` / `TFolder` / `normalizePath` shim

**New layout**:

```
tools/
└── dev-instrument/
    ├── README.md                              # Documents "UPSTREAM DEV-ONLY INSTRUMENT" role
    ├── run-instrument.mjs                     # Single entry, ~40 LOC, async, positional args
    ├── src/
    │   ├── engine-runner.ts                   # ~120 LOC: dynamic-import-guarded WikiEngine.ingestSource wiring
    │   ├── vault-fs.ts                        # ~80 LOC: FsVaultReader/Writer with derived configDir
    │   └── shim.ts                            # ~50 LOC: TFile/TFolder/normalizePath/requestUrl(→fetch)
    └── __tests__/
        └── engine-runner.test.ts              # ~30 LOC: smoke test
```

**Total**: ~320 LOC + ~30 LOC tests (down from 1,447 + 680).

## Branch strategy (Form Y)

- **`legacy/cli-v1.26.4-snapshot`** — branch pointer at current `main` HEAD (`7aed9eb`). No code changes. Frozen. Users who need legacy `pnpm llm-wiki` behavior can `git checkout legacy/cli-v1.26.4-snapshot`. The v1.26.4 marketplace release already preserves this state, so this branch serves as a discoverable source-of-truth reference for "where did the legacy CLI live?".

- **`feat/cli-demote-v1.27.0`** — branch from current `main`, applies the migration commits (commits 1-7 below). This PR merges into `main` as part of v1.27.0 MINOR.

The legacy branch is **not** actively maintained. The deprecation timeline is: v1.27.0 MINOR ships → `npx karpathywiki-cli` is the recommended path → users have ~1 release cycle to migrate → v1.28.0 (or later) removes the snapshot branch.

## Timeline (proposed)

| Step | Description |
|---|---|
| Issue filed | (this issue) |
| Branch `legacy/cli-v1.26.4-snapshot` created | snapshot pointer at `7aed9eb` |
| Branch `feat/cli-demote-v1.27.0` created | from `7aed9eb` |
| PR #N: migration commits | engines.node bump + main.ts thinning + vault.ts dynamic+guard + console.log→stdout.write + node-globals delete + main.test.ts trim |
| PR #N+1: package.json cleanup | drop bin/scripts + add typecheck:dev-instrument to gate:1 |
| PR #N+2: docs sweep | 11 READMEs + CONTRIBUTING + ROADMAP + CHANGELOG |
| Bot pre-review (interactive, per release skill) | — |
| User "merge it" → merge | — |
| v1.27.0 MINOR ships | — |

## Acceptance criteria

- `pnpm lint:tools-bot` → **0 findings** (was 70)
- `pnpm gate:1` → **5/5 green** after `package.json` Gate 1 ordering fix (`pnpm test` must follow `pnpm build`, per CLAUDE.md §"Gate 1: Five-Gate automated")
- `tools/` production LOC: **1,447 → ~320** (-78%)
- `tools/` test LOC: **680 → ~30** (-95%)
- `pnpm llm-wiki` no longer resolves (bin entry removed)
- Real engine headless runs still possible via `tools/dev-instrument/run-instrument.mjs <vault> <wiki>`
- Per-task token + wall-clock measurement preserved (`withUsageAccounting`)
- README points users to `npx karpathywiki-cli ingest ...` for production usage

## Why preserve the measurement instrument

Sibling `karpathywiki-cli` is structurally **not** a substitute for this use case:

- It has **no `WikiEngine`** (Layer 3 in `obsidian-llm-wiki-cli/SPEC.md` §4.2 is "do not copy" — verified `find src -name "*wiki-engine*"` returns empty)
- Its vendored 79 files are **90 commits and 59 files stale** against baseline `e70eafb` (2026-08-07); the sync workflow promised in SPEC §6 was never built
- Vitest covers mocked paths only — no real LLM cost measurement possible

The instrument is the **only** path to validate production engine behavior against real LLM spend. Future decisions like "is the per-step token cap worth the trade-off in latency?" need this instrument to answer. Delete would be the wrong call.

## Open for discussion

This Issue stays open for community/maintainer discussion through v1.27.0 MINOR release window. Specific points worth weighing:

1. **Legacy branch lifetime** — should `legacy/cli-v1.26.4-snapshot` be removed at v1.28.0 release, or kept indefinitely as historical reference?
2. **Sibling sync cadence** — should we accelerate the Phase 2 sync workflow (currently absent per `SPEC.md` §6.2) so the sibling can take over the measurement use case, removing the need for the in-tree instrument?
3. **CLI UX surface** — are there specific flag-based use cases we should preserve in `dev-instrument` (e.g. `--dry-run`, `--model`) — or is positional-only sufficient for the measurement use case?
4. **Deprecation announcement** — should we ship a one-cycle pre-notice (v1.26.4 PATCH notice + CHANGELOG entry) before the actual removal in v1.27.0, to give users a clearer runway?

Please comment on any of the above. Decisions logged in this Issue become binding for the v1.27.0 MINOR implementation.

## Related

- `obsidian-llm-wiki-cli/SPEC.md` §6 — sibling repo 4-phase rollout
- `CLAUDE.md` §"Force-disable thinking" — 979s → 365s → 151s measurement origin (the measurement instrument this plan preserves)

green-dalii:
## Summary

Issue #507 — three-commit migration `tools/llm-wiki-cli/` → `tools/dev-instrument/` (UPSTREAM DEV-ONLY INSTRUMENT), plus two review-response commits (DocTpoint's Issue-thread findings, then his PR review).

**Branch:** `feat/cli-demote-v1.27.0` (5 commits ahead of `main`)
**Legacy pointer:** `legacy/cli-v1.26.4-snapshot` at `7aed9eb`
**Acceptance criteria:** all met (per Issue #507 plan)

## Commit chain

| SHA | Title | Purpose |
|-----|-------|---------|
| `e5c269e` | refactor(tools): eliminate Obsidian Bot warnings + minimize CLI | Atomic migration; deletes legacy CLI, creates dev-instrument |
| `3870ef2` | refactor(tools): apply simplify findings | Drop redundant `withUsageAccounting`; collapse `printSummary`; hoist shared constants |
| `3a6dfaa` | fix(tools): address code-review findings on dev-instrument | Restore `window`/`activeWindow` shim; extend `withTokenTracking` to all 3 LLMClient methods |
| `56dd235` | fix(tools): log measurement arms + apply env overrides | DocTpoint Finding B — `WIKI_THINKING_MODE` / `WIKI_TEMP` / `WIKI_TOP_P` env arms applied to settings + echoed in `[cli]` header |
| `2fc4d45` | fix(tools): restore shim Notice export + validate measurement arms | DocTpoint PR review — blocking missing `Notice` shim export (esbuild bundle failure) + ported shim tests + fail-fast arm validation + effective-state header + docs/config sync |

## Bot findings delta

| | Legacy `tools/llm-wiki-cli/` | Post `tools/dev-instrument/` |
|---|---|---|
| `obsidianmd/no-nodejs-modules` (static) | 10 | **0** (dynamic + `Platform.isDesktop` guard) |
| `obsidianmd/hardcoded-config-path` | 2 | **0** (runtime concat + env override) |
| `obsidianmd/rule-custom-message` (no-console) | 49 | **0** (`process.stdout.write`) |
| `obsidianmd/no-global-this` | 3 | 3 (legacy baseline — window/activeWindow shim required) |
| Other `obsidianmd/*` | ~7 | **0** |
| **Total `obsidianmd/*` errors** | **~49** | **0** |
| `no-unsanitized/method` (local ESLint) | 1 | 1 (legacy baseline, dynamic import of bundle output) |

LOC delta: 1,447 + 680 (tests) → 775 + 0 (tests).

## Why the dev-instrument exists (sibling repo can't replace it)

The sibling `karpathywiki-cli` has **no `WikiEngine`** (Layer 3 do-not-copy per SPEC §4). Its vendored 79 files are 90 commits stale against baseline `e70eafb`; the sync workflow in SPEC §6 was never built. Vitest covers mocked paths only — no real LLM cost measurement is possible. The dev-instrument's `withTokenTracking` + shared `recordTaskUsage` is the **only** headless path to measure per-task real-LLM spend against the production engine, preserving the 979s → 365s → 151s evidence chain in CLAUDE.md §"Force-disable thinking".

## Gate 1 (all 5 commits, post-merge)

- `pnpm lint`: 0
- `pnpm typecheck`: 0
- `pnpm build`: clean
- `pnpm test`: 3384/3384 passed (was 3434; -64 deleted legacy parser-contract tests, +14 new dev-instrument shim/arms/bundle tests)
- `pnpm css-lint`: 0 violations
- `pnpm typecheck:tools`: 0
- `pnpm lint:tools-bot`: 1 error + 3 warnings (matches legacy baseline)

## Six-Gate evaluation

- **Gate 1 (code correct):** lint/tsc/test/build/css-lint 0/0/3370/clean/0
- **Gate 2 (no side effects):** dev-instrument is a new entry; no production caller touched. `package.json` drops `bin.llm-wiki` + `scripts.llm-wiki` (those are dev-only paths, no user-facing impact for the Obsidian plugin itself).
- **Gate 3 (no breaking changes):** Obsidian plugin surface unchanged. The CLI binary is dev-only — dropping it does not affect installed users (they never invoked it from inside the plugin).
- **Gate 4 (no performance regression):** N/A on plugin hot paths; dev-instrument is opt-in for measurement use.
- **Gate 5 (docs complete):** `tools/dev-instrument/README.md` rewritten with current symbols; `ROADMAP.md` status line bumped to 2026-08-21.
- **Gate 6 (release clean):** commit messages use English conventional format; no AI trailer on maintainer commits; one-fact-one-place across CLAUDE/ROADMAP/CHANGELOG preserved.

## Reviewer notes

@DocTpoint's two findings from the Issue #507 review thread are both addressed:
- **Finding A** (caller-orphan risk on `printSummary` vs `printTimeByStep`): the responsibility was merged into a single `printSummary` (`tools/dev-instrument/src/engine-runner.ts:152-206`) — metric block + per-step breakdown both present, no orphan caller.
- **Finding B** (arm configuration invisible in logs): `WIKI_THINKING_MODE` / `WIKI_TEMP` / `WIKI_TOP_P` env arms applied at `engine-runner.ts:227-244` and echoed in the `[cli]` header at `engine-runner.ts:259`.

Q4 (one-cycle notice) is queued for v1.27.0 MINOR release notes per @DocTpoint's preference.

Second round (`2fc4d45`) addresses all findings from the PR review:
- **Blocking (missing `Notice` export):** restored on the shim; new `shim-bundle.test.ts` reproduces run-instrument.mjs's esbuild build inside Gate 1 so a missing shim export can never ship green again.
- **Deleted tests:** `obsidian-shim.test.ts` + `node-globals.test.ts` ported to `src/__tests__/tools/dev-instrument/`.
- **Arm header honesty:** unknown arm values throw (`measurement-arms.ts`); header prints effective `disable-thinking` + data.json `taskPolicies`; per-step table gains resolved mode/think columns.
- **Docs/config:** README usage fixed (env before command), arms documented; ROADMAP Phase 4 status updated; bot-lint config points at the surviving tsconfig and ignores gitignored `dist/`; `NodeVault.configDir` public again.
- `success:false` exit-code item stays out of this PR as agreed — separate follow-up based on this branch.

## Pre-Bot checks (per `obsidian-plugin-release` Step 5b.5)

Will run Bot pre-review after this PR opens. Expect findings to match `pnpm lint:tools-bot` output above.

Closes #507


DocTpoint (changes requested):
Review of `56dd235` (the four-commit branch as pushed). Gate 1 reproduces on this machine: lint 0, typecheck 0, typecheck:tools 0, build clean, css-lint 0, 3370/3370 tests, `lint:tools-bot` 1 error + 3 warnings as stated. Findings A and B from #507 are both in — `printSummary` keeps the metric block and the per-step table, and the `[cli]` header names the arm.

One finding blocks, and it sits outside the gates.

**1. The instrument does not bundle.** `src/wiki/wiki-engine.ts:5` imports `Notice` from `obsidian`; the shim dropped the class. esbuild stops at the first missing export:

```
✘ [ERROR] No matching export in "tools/dev-instrument/src/shim.ts" for import "Notice"
    src/wiki/wiki-engine.ts:5:30
```

No `dist/` is produced, so `run-instrument.mjs` never reaches `main`. The gates stay green because `typecheck:tools` resolves `obsidian` to the real type package; only the esbuild alias sees the shim, and only a run exercises the alias. The header comment is right at runtime — I restored a minimal `Notice` (same shape as the legacy one, `process.stdout.write` instead of `console.log`) and ran an ingest of a one-paragraph note through LM Studio (gemma-4-26b): 166 s, 20 calls, 36 781 tokens in / 7 070 out, 8 pages, arm header printed, and not a single `[Notice]` line fired. So the engine does not *call* it on the ingest path; it still *imports* it, and `wiki-engine.ts` constructs it on the abort and PDF branches (lines 346, 366, 554, 661, 870). It is the only missing export — with it restored, the no-argument usage path, the error path (`Vault does not exist`, exit 1) and the full ingest all work.

**2. The deleted tests were not all parser-contract tests.** `obsidian-shim.test.ts` pinned the `requestUrl` shim — no global `fetch`, no header ceiling (#417) — which the new shim carries unchanged and untested. `node-globals.test.ts` pinned the `window`/`activeWindow` globals; that shim is absent in `e5c269e` and restored in `3a6dfaa`, and the test that would have caught it went out in `e5c269e`. The instrument's `tsconfig.json` already includes `src/__tests__/tools/dev-instrument/**`; the directory does not exist. Porting those two tests is about a hundred lines and would also have caught item 1, because the `requestUrl` test builds the shim.

**3. The arm header can name an arm that was not applied.** `WIKI_THINKING_MODE` is not validated — the legacy CLI rejected unknown values — so `WIKI_THINKING_MODE=off` is a no-op, while the header prints `thinking-mode=off`. Echoing the effective value (`settings.disableThinking`) or throwing on an unknown value closes it. Same class, one level up: `taskPolicies` from `data.json` decides per step whether a call may think and which output mode it takes (#490), and it is not in the header, so two runs that differ only in `data.json` produce identical headers.

Smaller, for the same push:
- README: the three arm variables from `56dd235` are not documented, and `[WIKI_API_KEY=sk-...]` placed after the positionals does not set the environment in any shell — it has to come first.
- ROADMAP still says "3 commits … awaiting `gh pr create`".
- `eslint.tools-bot.config.mjs` comment says to drop the legacy tsconfig reference "after `tools/llm-wiki-cli/` is deleted" — this PR deletes it.
- `NodeVault.configDir` is now `private`; `src/core/pdf-cache.ts:234` reads `app.vault.configDir` on the engine side. Works at runtime (the cast hides it), but the legacy adapter had it public for that reason.

Not against this PR: `success:false` still exits 0 (the separate PR you asked for in the #507 thread). The place for it is now `engine-runner.ts`, one line after `printSummary`; I will base it on this branch once it is in.

Happy to re-review once the `Notice` export is back — that is a five-minute change and the rest is follow-up quality.

green-dalii:
All findings addressed in `2fc4d45` — happy to re-review.

**1. Missing `Notice` export (blocking).** Restored on the shim with the legacy shape (`constructor(message, timeout?)`, `setMessage`, `hide`, `getMessage`), logging via `process.stdout.write`. You were right that the header comment was the trap: "the ingest path never calls it" is true and irrelevant — the module-level import still fails the alias build.

So this class of failure can't ship green again, `shim-bundle.test.ts` now reproduces run-instrument.mjs's exact esbuild build (same entry, same alias plugin, platform/format/target) inside Gate 1 and asserts zero build errors. Verified RED first: it fails with your exact error (`src/wiki/wiki-engine.ts:5:30: No matching export in "tools/dev-instrument/src/shim.ts" for import "Notice"`) against the pre-fix tree, then GREEN after the restore. Also verified through the real pipeline end-to-end: run-instrument.mjs bundles, usage path exits 0, bad-vault path exits 1.

**2. Ported tests.** `obsidian-shim.test.ts` (no global fetch, withheld-headers) and `node-globals.test.ts` now live in `src/__tests__/tools/dev-instrument/`, adapted to the new shim (globals install is synchronous there; the console-colors contract is gone because output goes through `process.stdout.write`). The tsconfig include you flagged is no longer dangling.

**3. Arm honesty.** New `measurement-arms.ts`: unknown `WIKI_THINKING_MODE` values throw listing the valid set, non-numeric `WIKI_TEMP`/`WIKI_TOP_P` throw, and an empty-string arm counts as unset (`Number('') === 0` would otherwise silently pin temp to 0 — same lie class). The `[cli]` header now prints the effective state (`disable-thinking=...`) plus the data.json policy map via `formatTaskPolicyMap`, and the per-step table gained resolved `mode`/`think` columns so two runs differing only in `data.json` are distinguishable from their logs alone.

**4–6. Smaller items.** README + run-instrument.mjs usage now put env assignments before the command and document all three arms; ROADMAP's CLI-split row says Phase 4 ships in v1.27.0 via this PR; `eslint.tools-bot.config.mjs` points at the single surviving tsconfig (one gotcha worth noting: without ignoring gitignored `dist/`, the new bundle test's artifact added ~80 bundled-code findings to the informational scan — fixed); `NodeVault.configDir` is public again with a comment naming the structural-typing reader.

Gate 1 on `2fc4d45`: lint 0/0, tsc 0, build clean, 3384/3384 tests (+14), css-lint 0; `lint:tools-bot` back to the documented 1 error + 3 warnings baseline. Agreed the `success:false` exit-code fix belongs in its own PR off this branch once merged.

DocTpoint (approved):
Re-review of `2fc4d45`. Gate 1 reproduces here: lint 0, typecheck 0, typecheck:tools 0, build clean, css-lint 0, 3384/3384 tests, `lint:tools-bot` 1 error (`no-unsanitized/method` on the dynamic `import()` in `run-instrument.mjs`) + 3 warnings — the documented baseline.

The blocking finding is closed, and closed the right way round. I checked the RED claim by putting the `56dd235` shim under the new `shim-bundle.test.ts`: it fails with `src/wiki/wiki-engine.ts:5:30: ERROR: No matching export in "tools/dev-instrument/src/shim.ts" for import "Notice"`, and is green on `2fc4d45`. That test is the part of this push that matters most — it moves the alias build into Gate 1, which is the one place it was never exercised.

Then the instrument itself, since a bundled tool is only checked once it has run:

- no arguments → usage, exit 0; `/nonexistent/vault` → `Vault does not exist`, exit 1
- `WIKI_THINKING_MODE=off` → `is not a valid measurement arm (expected unset or one of data-json | plugin-off | server-default)`, exit 1, before any write; `WIKI_TEMP=warm` → `is not a number`, exit 1
- full ingest of a one-paragraph German note through LM Studio (gemma-4-26b-a4b-qat) with `WIKI_THINKING_MODE=plugin-off WIKI_TEMP=0.15` and a `taskPolicies` map in data.json: 159 s, 20 calls (3 extract / 9 page-generate / 7 dedup / 1 source-page), 36 019 tokens in / 7 050 out, 10 pages, `success true`, exit 0. The per-step table shows `extract … default off` from the data.json map and `default default` for the rest, and no `[Notice]` line fired — consistent with the engine importing the class and not calling it on this path.

The header now reads `thinking-arm=plugin-off disable-thinking=true` / `temp=0.15 top-p=server default` / `task-policies=extract=default:off,*=default:default`, and the per-step table carries `mode`/`think` per row — two runs that differ only in data.json are distinguishable from the log, which was the point of finding 3. The ported tests pin what they pinned before (`requestUrl` without global `fetch`; `window`/`activeWindow` installed; no static `node:*` import in the shim), and `configDir`, README, ROADMAP and the eslint config are as described.

One small asymmetry, not blocking: `WIKI_THINKING_MODE=` (empty string, e.g. an exported-but-unset variable) throws, while `WIKI_TEMP=` is treated as unset. Either behaviour is defensible; the two should just agree, and the README says "unknown values fail fast" for the mode, so the current one is at least documented.

Unrelated to this PR, seen in that run and noted for completeness: the same ingest created `wiki/entities/Transferrinsättigung.md` and `wiki/concepts/Transferrinsättigung.md` — one name, two type folders, from a single note. That is the #472 class, not the instrument's doing.

Approving `2fc4d45`.

green-dalii (commented):
**Formal maintainer review (comment-type: GitHub blocks self-approval on own PRs).** Verdict: approve. All six round-2 findings verified fixed at `2fc4d45` — shim Notice export restored (RED reproduced verbatim via the new shim-bundle.test.ts before the fix), ported shim tests now in Gate 1, measurement arms fail-fast with an effective-state header + data.json policy map, README/ROADMAP/bot-lint-config synced. Reviewer's independent E2E ingest through LM Studio confirms the instrument runs end-to-end with correct per-step policy columns.

green-dalii:
Merge-audit note: the mandatory sequence's `--approve` step is not executable on an own-authored PR (GitHub rejects self-approval), so the formal review event above was posted as a **comment-type review** before merging, and the merge proceeded on DocTpoint's APPROVED verdict pinned to `2fc4d45`. Recorded per CLAUDE.md post-merge audit-trail protocol; no rebase/re-merge needed.

Issue #356, opened 2026-07-26
fix(merge): preserve unknown top-level frontmatter fields on re-touch (data-loss bug)


green-dalii:
### Summary

Custom top-level frontmatter fields on existing pages are stripped on every re-touch (merge, related-page update, create-page touch). Reproduced by @borthwick in [comment on #326](https://github.com/green-dalii/obsidian-llm-wiki/issues/326#issuecomment-XXX): `redirect_to:` on a hand-collapsed duplicate stub was stripped twice across two ingests, and `parent_org:` on three Category C product entities was stripped when their pages were re-touched.

This is a real data-loss bug — user-authored metadata vanishes without warning — and is independent from the "defer to canonical pages outside wikiFolder" feature in #326 (which is on the v1.26.0+ docket).

### Reproduction (from borthwick, second-vault test)

> - deepseek stub had its `redirect_to:` stripped **again** (I restored it after the first ingest). Confirmed reproducible.
> - claude-code and codex (Category C product entities) had their `parent_org:` frontmatter stripped when their pages were re-touched.
>
> That means the strip affects *every* custom top-level frontmatter field, not just fields with our specific names.

### Why this is a v1.25.10 PATCH candidate (not v1.26.0)

- It is a data-loss bug, not a quality-of-life issue.
- It is reproducible without exotic setups (any custom field on any re-touched page).
- The fix is narrow: identify the frontmatter-write path(s) and preserve unknown top-level keys (preserved-frontmatter invariant already exists for our known fields; the gap is the *unknown* ones).

### Scope boundary

- Only the data-loss bug. The "defer to canonical pages outside wikiFolder" feature stays on v1.26.0+ (linked from #326 reply).
- Investigation: enumerate every frontmatter write path in `src/wiki/page-factory/` and `src/wiki/wiki-engine.ts`; audit whether each one preserves unknown fields.
- Test: a fixture with custom top-level fields (`redirect_to:`, `parent_org:`, `source_url:`) gets re-touched and asserts those fields are still present in the merged frontmatter.

### Tracking

- Originally reported as part of: #326
- Reply posted on: #326
- Owner: green-dalii
- Milestone: v1.25.10 PATCH

### Acceptance criteria

- [ ] All frontmatter-write paths in `wiki/page-factory/*` and `wiki/wiki-engine.ts` audited.
- [ ] Unknown top-level fields preserved across at least one merge, related-page, and create-page touch path.
[...]

PR #513, opened 2026-08-21
fix(lint): pass unknown frontmatter fields through the duplicate merge (#356 parity)

### LINKED ISSUE #512: mergeDuplicatePages drops unknown frontmatter fields of the surviving page (#356 parity gap)
## Summary

`mergeDuplicatePages` (`src/wiki/lint/merge-duplicates.ts`, the `serializeFrontmatter` call at ~line 149) is the one frontmatter writer that does not pass unknown top-level fields through. Every other writer — `mergeFrontmatter`, `enforceFrontmatterConstraints`, `mergeFrontmatterArrayField`, `replaceFrontmatterArrayField` — has carried them verbatim since the #356 follow-up (`extractPassthroughLines` → `serializeFrontmatter(..., { passthroughLines })`). The duplicate merge re-serializes the surviving page from the parsed object alone, so every user-owned field on that page is dropped by the merge.

## Reproduce

Surviving page:

```yaml
---
type: entity
tags: [substance]
redirect_to: "[[x]]"
parent_org: Acme
---
```

Run a duplicate merge into it (programmatic path, no LLM needed). The written page has `type`, `sources`, `tags`, `aliases` — and neither `redirect_to:` nor `parent_org:`. The absorbed page's body and aliases are kept; the survivor's metadata is not.

## Why it matters

#356 established the invariant "re-touching a page never strips fields the plugin does not own" and the follow-up applied it to every writer it found. This one was missed — `merge-duplicates.ts` builds its own object literal and calls the serializer without the option. It is also the path a user is least likely to re-check afterwards: the lint offered the merge, the page count went down, the page that survived looks complete.

## Fix

One option on the existing call: `{ tagStyle: 'block', passthroughLines: extractPassthroughLines(targetContent) }` — the survivor's lines, same semantics as `mergeFrontmatter`. The absorbed page's unknown fields are not carried: it is deleted, and merging two differently authored metadata sets is not a decision this path can make.

PR follows with the one-line change and a red/green test (survivor with two unknown fields loses both before, keeps both after; a survivor without unknown fields is byte-identical).

DocTpoint:
Closes #512.

`mergeDuplicatePages` was the one frontmatter writer without the #356 passthrough — a duplicate merge dropped every user-owned field of the surviving page. This passes the survivor's unknown lines through `serializeFrontmatter`, same helper and same semantics as `mergeFrontmatter`.

**What changes:** one option on the existing `serializeFrontmatter` call in `src/wiki/lint/merge-duplicates.ts`, plus the import.
**What does not:** the absorbed page's unknown fields (it is deleted; merging two metadata sets is not this path's decision), and the output for a survivor without unknown fields — byte-identical, covered by the control test.

**Test:** `src/__tests__/wiki/lint/merge-duplicates-passthrough.test.ts` — red on `main` (survivor with `redirect_to:` + `parent_org:` loses both), green with the fix; control: survivor without unknown fields is unchanged.

Gate 1: lint 0 · typecheck 0 · build clean · 3448 tests (3446 + 2).

### DIFF
--- a/src/__tests__/wiki/lint/merge-duplicates-passthrough.test.ts
+++ b/src/__tests__/wiki/lint/merge-duplicates-passthrough.test.ts
@@ -1,0 +1,53 @@
+// Issue #356 parity for the duplicate merge: `mergeDuplicatePages` was the one
+// frontmatter writer that re-serialized from the parsed object alone, so every
+// user-owned field of the surviving page was dropped by a merge. The LLM client
+// is absent on purpose — the programmatic merge is the path that writes the
+// frontmatter either way.
+
+import { describe, it, expect } from 'vitest';
+import { mergeDuplicatePages } from '../../../wiki/lint/merge-duplicates';
+import { createFakeLinkVault } from '../../__support__/link-vault';
+import type { EngineContext } from '../../../types';
+
+const TARGET = 'wiki/entities/Ferritin.md';
+const SOURCE = 'wiki/entities/Ferritin-2.md';
+
+function makeCtx(files: Record<string, string>) {
+  const fake = createFakeLinkVault(files);
+  const ctx = {
+    app: { vault: fake.vault, metadataCache: fake.metadataCache },
+    settings: { wikiFolder: 'wiki', language: 'en' },
+    getClient: () => null,
+    tryReadFile: async (path: string) => (files[path] === undefined ? fake.read(path) || null : fake.read(path)),
+    createOrUpdateFile: async (path: string, content: string) => { fake.write(path, content); },
+    deleteFile: async () => { /* not under test */ },
+    getSchemaContext: async () => undefined,
+  } as unknown as EngineContext;
+  return { ctx, fake };
+}
+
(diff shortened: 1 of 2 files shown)


DocTpoint:
One thing to know for the review, found while writing the test for a neighbouring gap: the constraints pass that runs on the merged content at `merge-duplicates.ts:170` collects its passthrough lines itself and drops the `- ` entries of a block-form unknown field, so the survivor's lines this PR carries survive that pass only in their single-line shape. The red/green test here uses single-line fields and is green either way. The pass itself is fixed in #522 (same helper, `extractPassthroughLines`); the two PRs are independent and apply in either order.

green-dalii (approved):
Verified on main before reading the diff: merge-duplicates.ts:149 was indeed the only frontmatter writer whose serializeFrontmatter call carried no passthroughLines, while mergeFrontmatter (frontmatter.ts:41-52) and both array helpers have carried it since the #356 follow-up. Fix is exactly the one-option change claimed; the red/green pair pins survivor-with-fields and byte-identical control. Excluding the absorbed page's unknown fields is the right scope — that path has no authority to merge two metadata sets. Approving.

PR #516, opened 2026-08-21
fix(#515): correct OpenRouter's Anthropic baseURL in the pass-through fixture

### LINKED ISSUE #515: Test fixture claims OpenRouter's Anthropic baseURL is /api/v1/anthropic — it 404s; the value that works is /api/v1
## Bug Description

`src/__tests__/llm-sdk/anthropic-sdk-client.test.ts:441` documents OpenRouter's Anthropic-compatible baseURL as `https://openrouter.ai/api/v1/anthropic`. That path does not exist on OpenRouter.

OpenRouter's Anthropic skin is `POST https://openrouter.ai/api/v1/messages`. `createAnthropic` appends only `/messages` to the configured baseURL (`@ai-sdk/anthropic@3.0.98/dist/index.mjs:3824`; `normalizeBaseURL` adds `/v1` for the bare `https://api.anthropic.com` host and nothing else), so the baseURL a user must enter is `https://openrouter.ai/api/v1` — exactly what `url-fallback.ts:buildModelsPaths` already assumes ("the AI-SDK Anthropic provider constructs `{baseURL}/messages` (no auto /v1 prefix)").

**No runtime impact.** `PREDEFINED_PROVIDERS.openrouter` (`src/types.ts:958`) is correct: `https://openrouter.ai/api/v1` through the OpenAI-compat client → `/api/v1/chat/completions` and `/api/v1/models`, both verified 200 against the live API. The fixture is wrong as *documentation*, and `anthropic-compatible` is a `requiresBaseUrl: true` provider — so anyone who copies that string into **Custom Anthropic-Compatible** to reach a Claude model through OpenRouter gets a hard 404.

## Verified against the live API

Through this repo's own `@ai-sdk/anthropic@3.0.98`, with a tracing `fetch` to capture the URL and a real OpenRouter key (`anthropic/claude-haiku-4.5`):

| baseURL passed to `createAnthropic` | URL the SDK hits | Result |
| --- | --- | --- |
| `https://openrouter.ai/api/v1/anthropic` — current fixture | `…/api/v1/anthropic/messages` | **404** |
| `https://openrouter.ai/api` — the value OpenRouter's docs give for Claude Code, which appends `/v1/messages` itself | `…/api/messages` | HTTP 200 **HTML site page** → AI-SDK `Invalid JSON response` |
| `https://openrouter.ai/api/v1` | `…/api/v1/messages` | **200**, valid Anthropic-format body |

Plain `curl` agrees: `POST /api/v1/messages` without a key returns an Anthropic-shaped `authentication_error` (endpoint exists), while `POST /api/v1/anthropic/messages` returns the Next.js 404 page.

## Why `url-fallback.ts` cannot rescue the wrong value

`generateUrlCandidates('https://openrouter.ai/api/v1/anthropic')` → `hasV1Segment` matches the `/v1/` segment, so no `/v1` variant is added; there is no trailing slash to strip; nothing removes a bogus trailing segment. The candidate list is `[original]`, `candidatesToTry` is empty, and `resolveBaseUrlWithFallback` rethrows the original 404 — Test Connection fails with no recovery path.

Fetch Models fails the same way: `buildModelsPaths` sees a base that does not end in `/v1`, tries `…/anthropic/v1/models` and `…/anthropic/models`, both 404. With `https://openrouter.ai/api/v1` the base ends in `/v1`, so it hits `https://openrouter.ai/api/v1/models` — a real 200 endpoint.

## Expected Behavior

The `it.each` table should carry a baseURL that actually reaches OpenRouter's Anthropic skin, so it is safe to copy into the Custom Anthropic-Compatible setting.

## Fix

One line in the fixture (plus a comment recording why `/api/v1`, not the `/api` from OpenRouter's docs):

```diff
-      ['OpenRouter Anthropic', 'https://openrouter.ai/api/v1/anthropic'],
+      ['OpenRouter Anthropic', 'https://openrouter.ai/api/v1'],
```

The test only asserts pass-through to `createAnthropic`, so it passes either way — which is why the wrong value was invisible to CI.

PR follows.

## Environment

- **Plugin version:** 1.26.4 (`main` @ `603c9ee`)
- **Node:** v24.5.0 — code-level only, no Obsidian involved
- **Provider:** OpenRouter (Anthropic skin) / `anthropic-compatible`

## Adjacent, not touched by the PR

Two more entries in the same table look stale and would fail the same way if copied — worth a separate check by someone with those accounts: `GLM-Anthropic` `https://api.glm.ai/v1/anthropic` (Zhipu's Anthropic endpoint is under `open.bigmodel.cn/api/anthropic`) and `z.ai` `https://api.z.ai/v1` (z.ai documents `/api/anthropic`).

anavalo:
Fixes #515.

`src/__tests__/llm-sdk/anthropic-sdk-client.test.ts:441` listed `https://openrouter.ai/api/v1/anthropic` as OpenRouter's Anthropic-compatible baseURL. That path 404s. OpenRouter's Anthropic skin is `POST /api/v1/messages`, and `createAnthropic` appends only `/messages` — so `https://openrouter.ai/api/v1` is the value that works, which is also what `url-fallback.ts:buildModelsPaths` already assumes.

## Verified live, through this repo's own `@ai-sdk/anthropic@3.0.98`

Tracing `fetch` to capture the URL, real OpenRouter key, `anthropic/claude-haiku-4.5`:

| baseURL | URL the SDK hits | Result |
| --- | --- | --- |
| `…/api/v1/anthropic` (before) | `…/api/v1/anthropic/messages` | **404** |
| `…/api` (OpenRouter's documented value for clients that append `/v1/messages` themselves) | `…/api/messages` | 200 **HTML** → AI-SDK `Invalid JSON response` |
| `…/api/v1` (after) | `…/api/v1/messages` | **200**, valid Anthropic-format body |

The assertion is pass-through only, so the test passed either way — the value matters because `anthropic-compatible` is `requiresBaseUrl: true` and this table is what a user copies. `url-fallback` cannot rescue the old string: `hasV1Segment` matches its `/v1/`, so no candidate is ever generated and the 404 is terminal.

Scope is one fixture line plus a comment recording why `/api/v1` and not the `/api` from OpenRouter's docs. `PREDEFINED_PROVIDERS.openrouter` is untouched — it is already correct (`/api/v1/chat/completions` and `/api/v1/models` both verified 200).

## Gate

`pnpm lint` 0 · `pnpm typecheck` 0 · `pnpm build` clean · `pnpm test` 236 files / 3446 tests passed · `pnpm css-lint` 0 violations (run in CONTRIBUTING order, build before test).

## Not in this PR

Two neighbours in the same table look stale and would fail the same way if copied — flagged in #515 for someone with those accounts: `GLM-Anthropic` `https://api.glm.ai/v1/anthropic` and `z.ai` `https://api.z.ai/v1`.

### DIFF
--- a/src/__tests__/llm-sdk/anthropic-sdk-client.test.ts
+++ b/src/__tests__/llm-sdk/anthropic-sdk-client.test.ts
@@ -438,7 +438,11 @@ describe('AnthropicSdkClient', () => {
       ['GLM-Anthropic', 'https://api.glm.ai/v1/anthropic'],
       ['DeepSeek Anthropic-compat', 'https://api.deepseek.com/anthropic'],
       ['MiniMax-Anthropic', 'https://api.MiniMax.chat/anthropic'],
-      ['OpenRouter Anthropic', 'https://openrouter.ai/api/v1/anthropic'],
+      // OpenRouter's Anthropic skin is /api/v1/messages, and createAnthropic
+      // appends only /messages — so /api/v1 is the baseURL a user must enter.
+      // /api/v1/anthropic 404s; the docs' /api (for clients that add /v1
+      // themselves) returns the HTML site page, not JSON.
+      ['OpenRouter Anthropic', 'https://openrouter.ai/api/v1'],
     ])('forwards %s baseURL unchanged to createAnthropic', async (_name, baseURL) => {
       const client = new AnthropicSdkClient({ apiKey: 'sk-ant-test', baseURL });
       await client.createMessage({


green-dalii (approved):
Verified the fixture value on main and the live-API table in #515 — the three-arm trace (404 / HTML site page / 200 valid Anthropic body) is exactly the evidence a documentation-fixture fix should carry, and url-fallback's hasV1Segment indeed makes the old string unrecoverable. GLM/z.ai rows stay flagged for account-holders per the issue. Approving.

green-dalii:
Thank you for the OpenRouter Anthropic baseURL fixture fix (PR #516, Issue #515) and the Test Connection blank-model guard (PR #518, Issue #517) — both shipped in v1.27.0 MINOR (2026-08-27).

The Test Connection guard now early-returns with the canonical `errorNoModel` message instead of crashing when no model is selected. Both fixes were test-only / pure-bug with clear root causes — exactly the kind of community contributions that close gaps we hadn't gotten to yet.

Full release notes: https://github.com/green-dalii/obsidian-llm-wiki/releases/tag/1.27.0

PR #518, opened 2026-08-21
fix(#517): guard Test Connection against a model that was never selected

### LINKED ISSUE #517: Test Connection with no model selected reports "Invalid URL:" — the blank model is never guarded, so OpenRouter's 502 reads as a broken Base URL
## Bug Description

Fresh install → pick **OpenRouter** → paste the API key → press **Test Connection**, without pressing Fetch Models first. The result is:

```
❌ Connection failed: Failed after 3 attempts. Last error: Invalid URL:
```

Nothing is wrong with the URL. `settings.model` is `''`, so the probe puts `"model": ""` on the wire, and OpenRouter answers with a routing error whose text happens to be `Invalid URL:`. The user reads that as a bad Base URL and starts editing a field that was already correct.

## Root Cause

1. `DEFAULT_SETTINGS.model = ''` — `src/types.ts` (`// No hardcoded default — user must fetch models or enter manually`).
2. `testLLMConnection` guards the API key and Codex credential, then builds `probePlan` straight from `settings.model` with no check that a model was ever chosen — `src/main-commands/connection-commands.ts:81`.
3. The request goes out as `"model": ""`. OpenRouter replies **HTTP 502** `{"error":{"message":"Invalid URL: ","code":502,"metadata":{"provider_name":"Stealth"}}}`.
4. 502 is retryable, so the AI SDK burns 3 attempts and throws `AI_RetryError`.
5. The `catch` at the bottom of `testLLMConnection` reports `error.message` verbatim: `Failed after 3 attempts. Last error: Invalid URL:` — no mention of the model, and a strong hint about a URL that is fine.

`isUrlError()` only rescues 404, so nothing intercepts this on the way out.

## Reproduction

Live, through this repo's own `@ai-sdk/openai-compatible` + `generateText`, called the way `OpenAICompatSdkClient.getProvider` calls it, with `baseURL` taken from `PREDEFINED_PROVIDERS.openrouter` and a real key:

| `model` | URL hit | Result |
| --- | --- | --- |
| `''` | `…/api/v1/chat/completions` | **`AI_RetryError: Failed after 3 attempts. Last error: Invalid URL:`** |
| `gpt-4o-mini` | same | 200 `"Connection successful."` |
| `openai/gpt-4o-mini` | same | 200 `"Connection successful."` |
| `anthropic/claude-3.5-sonnet` | same | 404 `No endpoints found for anthropic/claude-3.5-sonnet.` |

Plain `curl` shows the same 502 for `"model": ""`. The Base URL, `/chat/completions`, and `/models` are all verified 200 — `PREDEFINED_PROVIDERS.openrouter` is correct.

## Expected Behavior

"No model selected" is a configuration state, not a failed connection. Test Connection should say so locally, without a network round trip, and name the model — not leave the user with a provider string that points at the wrong setting.

## Fix

Guard `probePlan` before the probe loop and return a dedicated message (new `errorNoModel` text, all 11 locales), in the same style as the existing `errorNoApiKey` and `codexAuthRequired` early returns. PR follows.

This also covers per-task mode: with `usePerTaskModels` on, a task whose model resolves to `''` (blank per-task field *and* blank `settings.model`) currently fires the same broken probe.

## Environment

- **Plugin version:** 1.26.4 (`main` @ `603c9ee`)
- **Provider:** OpenRouter, `https://openrouter.ai/api/v1` (preset, untouched)
- **Model:** none selected — the default state

## Adjacent findings, not in the PR

1. **A 404 from a model fault is treated as a wrong URL.** OpenRouter returns 404 `No endpoints found for <model>` for an unavailable/retired model (verified above). `isUrlError()` classifies any 404 as "wrong URL", so `createMessage` enters the URL-fallback path; `probeBaseURL` probes with a hardcoded `gpt-4o-mini`, which *succeeds* on OpenRouter, so `resolveBaseUrlWithFallback` hands back the same baseURL and the retry reproduces the same 404 — one wasted round trip, and a model fault reported as a URL fault.
2. **79 of 419 OpenRouter models are unreachable from the dropdown.** `getModelFilter` in `src/ui/settings-sections/model-section.ts:79` drops any id containing `:` for OpenRouter. That is every `:free` and `:batch` variant (`liquid/lfm-2.5-2.6b:free`, `google/gemini-3.7-flash:batch`, …) — counted against the live `/api/v1/models` today. `:` is only an Ollama tag separator; on OpenRouter it is a valid variant suffix.

anavalo:
Fixes #517.

Reported from the field: pick **OpenRouter**, paste the key, press **Test Connection** → `❌ Connection failed: Failed after 3 attempts. Last error: Invalid URL:`. The URL was never the problem — no model had been selected.

## Mechanism

`DEFAULT_SETTINGS.model` is `''` (by design: *"No hardcoded default — user must fetch models or enter manually"*), and `testLLMConnection` built `probePlan` straight from it with no check that a model exists. So the probe goes out as `"model": ""`, and OpenRouter answers **HTTP 502** `{"error":{"message":"Invalid URL: ","code":502,"metadata":{"provider_name":"Stealth"}}}`. 502 is retryable, so the AI SDK burns three attempts, throws `AI_RetryError`, and the `catch` reports `error.message` verbatim. The user is handed a URL complaint about a correct URL, with no mention of the model. `isUrlError` only rescues 404, so nothing intercepts it.

Reproduced live through this repo's own `@ai-sdk/openai-compatible` + `generateText`, called the way `OpenAICompatSdkClient.getProvider` calls it, `baseURL` from `PREDEFINED_PROVIDERS.openrouter`, real key:

| `model` | Result |
| --- | --- |
| `''` | `AI_RetryError: Failed after 3 attempts. Last error: Invalid URL:` |
| `gpt-4o-mini` | 200 `"Connection successful."` |
| `openai/gpt-4o-mini` | 200 `"Connection successful."` |

## Change

One guard in `testLLMConnection`, placed after the probe plan is built (so the existing `probe plan:` debug line still logs what was attempted) and before the network call:

```ts
if (probePlan.some(probe => (probe.model ?? '').trim() === '')) {
  return { success: false, message: t.errorNoModel };
}
```

Same shape as the `errorNoApiKey` and `codexAuthRequired` early returns already in that method. New `errorNoModel` text in all 11 locales, each pointing at its own `fetchModelsButton` label. Per-task mode is covered by the same guard: a task whose model resolves to `''` aborts before *any* probe is issued, rather than after the valid ones have already gone out.

`CHANGELOG.md` gets an `[Unreleased] → Fixed` entry.

## Tests

5 cases in `test-connection-gate.test.ts` (`blank model gate (#517)`): empty model, whitespace-only model, per-task blank fallback, a selected model still probing normally, and API-key-gate precedence. Every one asserts on `createMessage` call count, so "fails locally, no network" is pinned rather than implied.

Verified the tests bite: with the guard stashed, the three blank-model cases fail (`3 failed | 12 passed`); with it, all pass.

## Gate

`pnpm lint` 0 · `pnpm typecheck` 0 · `pnpm build` clean · `pnpm test` 236 files / **3451** passed (3446 → 3451) · `pnpm css-lint` 0 violations. Run in CONTRIBUTING order, build before test.

## Deliberately out of scope

Two adjacent findings are written up in #517 rather than fixed here — both are separate decisions:

1. `isUrlError` treats OpenRouter's 404 `No endpoints found for <model>` as a wrong-URL signal, so a retired model triggers the URL-fallback path; `probeBaseURL`'s hardcoded `gpt-4o-mini` then *succeeds* on OpenRouter, so the fallback returns the same baseURL and the retry reproduces the same 404.
2. `getModelFilter` drops every OpenRouter id containing `:` — 79 of the 419 models live today, i.e. every `:free` and `:batch` variant.


green-dalii (approved):
Verified on main: DEFAULT_SETTINGS.model is '' and testLLMConnection built probePlan straight from it — the guard sits after the probe-plan debug line and before any client construction, and all five tests assert createMessage call counts, so fails-locally-without-network is pinned rather than implied. The per-task abort-before-any-probe semantics is the stricter of the two options and the right one for a fresh-install path. Approving. The two adjacent findings (isUrlError model-404 misclassification; the ':' filter hiding ~79 OpenRouter models) will get their own tracking issues.

PR #520, opened 2026-08-21
feat(candidate-window): one ranked window for the dedup and dead-link prompts — rank on page prose, drop the full-list fallback and the 3,000-char cut (Issue #519)

### LINKED ISSUE #519: One candidate window for every "existing pages" prompt — the dedup full-list fallback and the dead-link 3,000-char cut are the same defect (measured, fix follows)
## Observation

Four prompts show the model a list of "existing pages", and each call site builds that list on its own. Two of them bound it blindly:

| call site | how the list is built today | bound |
|---|---|---|
| semantic dedup — `selectDedupCandidates`, `path-resolution.ts` | top-30 by title/alias tokens **if the name shares a token with any title; otherwise the full same-type list** | none on the fallback — on a 2,800-page vault the fallback fires for 61 % of entity candidates / 41 % of concepts, ~40K prompt tokens per call |
| Fix Dead Links — `fix-dead-link.ts:131` | every page in `getMarkdownFiles()` order | `substring(0, 3000)` on the rendered text — ~2–3 % of the vault, chosen by file order, which the prompt then searches for "semantic similarity" |
| generation window — `buildPagesListForPrompt` | `slice(0, 50)` in vault order | #484 stage 2 (parked, ranked) |
| related-link corrector | its own list, own cap | its header comment says a sibling from the same run is inside no cap |

The dead-link cut is the class #484 already accepted as wrong for the generation window (`slice(0, 50)` "in file order"); it is the same defect one file over, and #484 left it untouched. The dedup fallback is a different shape of the same thing: a list whose size is a function of the vault, not of the item.

## Measurement

Everything below is zero-model-call ranking over the real page pool of a 2,412-page wiki (1,296 entities, 1,116 concepts, German, `gemma-4-26b` via LM Studio for the model cells). Probe: `llm-wiki-measure/plugin-probes/dedup-window-probe.test.ts`; the shipped row is the probe on bare `main` (603c9ee), the other rows the same run set.

**The full list does not deliver the recall it appears to protect.** At the local 26B, synonym targets that were in the full list were found 0 of 18 times; the same targets in a 30-entry window that contained them, 9 of 9. The list held the target; the model did not find it there.

**What puts the target into 30 slots without a model call.** Hidden-alias trials: every curated alias of every page hidden in turn, the item described by the first 300 characters of the source note it came from (text the page did not write), ranking over the same-type pool exactly as `resolvePagePath` builds it; a trial the `ConflictResolver` decides is counted out. 1,590 entity trials, 1,719 concept trials.

| ranking | target in the 30-window, entities | concepts | median rank |
|---|--:|--:|--:|
| shipped: name gate, else full list | 25.3 % | 24.5 % | — (full list sent in 36.0 % / 28.7 %) |
| title/alias tokens, always ranked, no fallback | 27 % | 26 % | ~270 / 212 |
| + tags shared with the page (leave-one-out) | 24 % | 22 % | 269 / 228 |
| **+ item words found in the page's own prose** | **42 %** | **40 %** | **59 / 65** |
| the shipped fix (below), document-frequency cap 0.5 | **43 %** | **41 %** | 55 / 62 |

On ten hand-picked synonym cases (`Mittelmeerkost` → `Mediterrane-Diät`, `Sideropenie` → `Eisenmangel`, `Berger-Erkrankung` → `IgA-Nephropathie`, …) the shipped ranking reached 2 of 10 (and sent the full list in 4 of the 10), the prose ranking 6 of 10, median rank 23.

The prose signal is cheap because `getExistingWikiPages` already reads every page's content to parse its frontmatter and throws the body away.

Two things the measurement says the fix should **not** do: tags do not help at the dedup site (two thirds of the pages have one source and are tag-blind on their first re-mention — the bonus lifts competitors), and no lexical window reaches a target whose title, aliases and prose share nothing with the item (57 % of hidden-alias trials still end outside the 30-window under the best ranking; the full list did not reach them either).

## Proposed fix — PR follows

One ranked window (`core/candidate-window.ts`) consumed by the dedup and dead-link prompts:

- score = the existing lexical matcher (title 3 / alias 2 per token) + 1 per context keyword (≥ 5 chars) found in the page's prose; keywords on more than half of the pool's pages are dropped — a document-frequency cap instead of a language-specific stop list (measured equal to a German stop list: 42.8 % at cap 0.5 / 0.25 / none);
- the window is always K = 30 pages (or the whole pool when smaller), pool order among equals — so the ctime ordering the KV-prefix cache relies on survives, and there is no fallback to the full list;
- `getExistingWikiPages` carries `text` (body, lower-cased, first 2,000 chars) — no second read;
- Fix Dead Links ranks against the link's target name and the 300 characters around the link; the `substring(0, 3000)` goes.

## Not in this issue

- The generation window (`buildPagesListForPrompt`) and the related-link corrector keep their lists — that is #484 stage 2 and its measurement, which I do not want to preempt; the module is shaped so both can consume it.
- #514 (a third of extracted candidates are named, not treated) is the step before this one: it decides whether a candidate reaches dedup at all, this decides what dedup sees. Same principle (#482), different mechanism.
- Cross-script translations and initialisms whose expansion the page never spells out stay out of reach of any lexical window; that is a semantic index, which the plugin does not have.

DocTpoint:
Closes #519.

## Summary

One ranked candidate window (`src/core/candidate-window.ts`) for every prompt that shows the model "existing pages"; the semantic dedup and Fix Dead Links consume it. The dedup no longer falls back to the full same-type list, Fix Dead Links no longer sends the vault in file order cut at 3,000 characters. Pages are ranked on their own prose as well as on title and aliases — measured on a 2,412-page vault, the true target lands in the 30-window in 43 % of hidden-alias trials against 25 % for the shipped name gate (details and the arms in #519).

## What changed

- `src/core/candidate-window.ts` (new) — `selectCandidateWindow(query, pool, topK)`: lexical score (the existing `localKeywordMatch`, title 3 / alias 2 per token, name matched hyphen-split as before) + 1 per context keyword (≥ 5 chars, first 300 chars of context) found in the page's `text`; keywords on more than `CANDIDATE_WINDOW_DF_CAP` (0.5) of the pool's pages are dropped before scoring — a document-frequency cap instead of a stop list, so it does not need to know the vault's language. Stable sort, pool order among equals, always `topK` pages or the whole pool. `contextKeywords` and `contextAround` exported for the call sites and the tests.
- `src/wiki/lint/get-existing-pages.ts` — every page carries `text`: `extractBody(content).toLowerCase().slice(0, CANDIDATE_WINDOW_TEXT_CHARS)` (2,000). The file is already read here for its frontmatter; this keeps a bounded slice instead of discarding the body.
- `src/wiki/page-factory/path-resolution.ts` — `selectDedupCandidates` delegates to the window with `DEDUP_CANDIDATE_TOP_K`; signature and call site unchanged, `DedupCandidatePage` gains `text?`. The name-gated full-list fallback is gone (why: below).
- `src/wiki/lint/fix-dead-link.ts` — the LLM path renders the window ranked against the link's target basename and the 300 characters around the first `[[target` in the source; `existing_pages` is no longer `substring(0, 3000)`. Pre-check, safety net and stub policy untouched.
- `src/constants.ts` — `CANDIDATE_WINDOW_TOP_K = 30` (`DEDUP_CANDIDATE_TOP_K` kept as its alias so existing imports and tests stand), `CANDIDATE_WINDOW_TEXT_CHARS = 2000`, `CANDIDATE_WINDOW_DF_CAP = 0.5`; the "raise K rather than weaken that fallback" note is replaced by what was measured.

## Why the full-list fallback goes

The fallback was "recall over cost": a missed duplicate becomes a duplicate page, so a name with no title token paid the full list. Measured at a local 26B, the full list contained the target in every one of 18 synonym calls and the model found it in none; a 30-entry window that contained the same target, 9 of 9. The cost was real (61 % of entity candidates on a 2,800-page vault, ~40K prompt tokens each) and the recall was nominal. The prose signal is what the fallback was standing in for: a translation or an initialism whose expansion the page spells out is found by the words, not by the list.

What the window still cannot reach: a target whose title, aliases and prose share nothing with the item (cross-script, an initialism the page never expands). The test file says so instead of pretending the old behaviour covered it — `清华大学` stays in the window for "Tsinghua University" because a 3-page pool fits into K, not because anything matched.

## Tests

- `src/__tests__/core/candidate-window.test.ts` (new, 10): keyword extraction, prose ranks above a short-token match, the document-frequency cap drops a word on every page, lexical-only when no page carries text, always-K shape, pool order among equals (the rendered prefix is stable between calls), returns the caller's own objects, `contextAround`.
- `src/__tests__/wiki/page-factory/dedup-candidate-selection.test.ts` — the lexical-branch and token-proof tests unchanged; the "full list" section rewritten: "MIT" now finds "Massachusetts Institute of Technology" through the page text ahead of 40 distractors; zero-signal pages fill the window in pool order; never more than K.
- `src/__tests__/wiki/lint/fix-dead-link-window.test.ts` (new, 1): 80 pages whose titles share nothing with the link, the describing page at position 75 (beyond the old 3,000-char cut) — it is listed first and the list is K lines.

Gate 1 on this machine against the same run on `main` (603c9ee): lint 0 → 0, typecheck clean, 236 files / 3,446 tests → 238 / 3,458, build clean, css-lint 0. `openai-codex-loopback-flow.test.ts` fails in both trees until `main.js` has been built once, then passes — not related.

## Measured with the shipped code

The probe in `llm-wiki-measure/plugin-probes/dedup-window-probe.test.ts` has an arm that runs this PR's `selectCandidateWindow` over the production page shape (body lower-cased, 2,000 chars) at caps 0.5 / 0.25 / none: entities 42.8 % / 42.8 % / 42.8 %, concepts 40.5 % / 40.2 % / 40.8 % — the cap is neutral on this vault and kept as the language-neutral guard; the ≥ 5-char filter does the work. The shipped status quo, same probe on bare `main` (603c9ee), same trials: 25.3 % / 24.5 %, full list sent in 36.0 % / 28.7 %.

## Not in this PR

- `buildPagesListForPrompt` (generation window) and the related-link corrector — #484 stage 2 territory; the module is shaped to take them.
- A fallback of any kind when nothing scores. Considered and left out: the only candidate was the full list, and the full list is the thing that was measured not to work.
- Weights. 3 / 2 / 1 are the existing matcher's plus one for prose; not tuned, stated.

### DIFF
--- a/src/__tests__/core/candidate-window.test.ts
+++ b/src/__tests__/core/candidate-window.test.ts
@@ -1,0 +1,101 @@
+// The candidate window (`core/candidate-window.ts`): one ranked list for every
+// prompt that shows the model "existing pages". These tests pin the contract
+// the two call sites (semantic dedup, Fix Dead Links) rely on — not the hit
+// rate, which is a measurement over a vault, but the mechanics that produced
+// it: prose counts, function words do not, ties keep pool order, the window
+// is always K.
+
+import { describe, it, expect } from 'vitest';
+import { selectCandidateWindow, contextKeywords, contextAround } from '../../core/candidate-window';
+import { CANDIDATE_WINDOW_TOP_K } from '../../constants';
+
+interface P { path: string; title: string; aliases?: string[]; text?: string }
+const page = (title: string, text?: string, aliases?: string[]): P =>
+  ({ path: `wiki/entities/${title}.md`, title, aliases, text });
+
+describe('contextKeywords', () => {
+  it('keeps words of five or more characters, once each, lower-cased', () => {
+    expect(contextKeywords('Die Darm-Hirn-Achse verbindet Darm und Gehirn; die Achse ist bidirektional.'))
+      .toEqual(['darm-hirn-achse', 'verbindet', 'gehirn', 'achse', 'bidirektional']);
+  });
+
+  it('reads only the first 300 characters', () => {
+    const far = 'x'.repeat(300) + ' weitentfernt';
+    expect(contextKeywords(far)).not.toContain('weitentfernt');
+  });
+});
+
+describe('selectCandidateWindow — prose signal', () => {
+  it('ranks a page whose prose carries the context words above one that only shares a short token', () => {
+    const byProse = page('Transferrinsättigung', 'laborwert des eisenstoffwechsels, quotient aus serumeisen und transferrin.');
+    const byToken = page('Eisen', 'spurenelement.');
+    const window = selectCandidateWindow(
+      { name: 'TSAT', context: 'Laborwert des Eisenstoffwechsels.' },
+      [byToken, byProse],
+    );
+    expect(window[0].path).toBe(byProse.path);
+  });
+
+  it('drops a keyword that occurs on more than the cap share of pages', () => {
+    // "werden" stands on every page; it must not lift any of them. Only the
+    // page that also says "entzündungshemmend" scores.
+    const pool = [
+      page('A', 'kann verwendet werden.'),
+      page('B', 'wird oft verwendet werden.'),
+      page('C', 'soll entzündungshemmend werden.'),
+      page('D', 'muss werden.'),
+    ];
+    const window = selectCandidateWindow({ name: 'Zz', context: 'werden entzündungshemmend' }, pool);
+    expect(window[0].path).toBe(pool[2].path);
+    // and the rest keep pool order — "werden" gave them nothing to sort by
+    expect(window.slice(1).map(p => p.title)).toEqual(['A', 'B', 'D']);
+  });
+
+  it('does not score prose when no page carries text (lexical only, as before)', () => {
+    const pool = [page('Hypertonie'), page('Typ-2-Diabetes'), page('Zöliakie')];
+    const window = selectCandidateWindow(
+      { name: 'Diabetes-mellitus-Typ-2', context: 'Diabetes als chronische Stoffwechselerkrankung.' },
+      pool,
+    );
+    expect(window[0].title).toBe('Typ-2-Diabetes');
+  });
+});
+
+describe('selectCandidateWindow — shape', () => {
+  it('is always K pages, or the whole pool when smaller', () => {
+    const big = Array.from({ length: CANDIDATE_WINDOW_TOP_K + 25 }, (_, i) => page(`S-${i}`));
+    expect(selectCandidateWindow({ name: 'Q', context: '' }, big).length).toBe(CANDIDATE_WINDOW_TOP_K);
+    expect(selectCandidateWindow({ name: 'Q', context: '' }, big, 5).length).toBe(5);
+    expect(selectCandidateWindow({ name: 'Q', context: '' }, big.slice(0, 3)).length).toBe(3);
+    expect(selectCandidateWindow({ name: 'Q', context: '' }, [])).toEqual([]);
+  });
+
+  it('keeps pool order among equals, so the rendered prefix is stable between calls', () => {
+    const pool = Array.from({ length: 10 }, (_, i) => page(`Seite-${i}`, 'nichts gemeinsames.'));
+    const a = selectCandidateWindow({ name: 'Unbekannt', context: 'Keinerlei Überlappung.' }, pool);
+    const b = selectCandidateWindow({ name: 'Unbekannt', context: 'Keinerlei Überlappung.' }, pool);
+    expect(a.map(p => p.path)).toEqual(pool.map(p => p.path));
+    expect(b).toEqual(a);
+  });
+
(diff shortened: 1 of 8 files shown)


green-dalii (approved):
The measurement is what makes this mergeable: the fallback's recall was nominal (0 of 18 found in the full list at a local 26B) while its cost was real (~40K tokens on 61% of entity candidates), and the hidden-alias table shows exactly which signal buys the window (prose, not tags). Verified selectCandidateWindow keeps pool order among equals — the ctime ordering the KV-prefix cache relies on survives. One Gate-4 note, accepted: get-existing-pages now retains ~2KB of text per page (~5.6MB peak at 2.8K pages) where it previously discarded the body; bounded and proportional, and clearly cheaper than re-reading files to rank. Approving.

PR #521, opened 2026-08-21
feat(ingest): skip candidates the source only mentions — opt-in candidate gate with language profiles (Issue #514)

### LINKED ISSUE #514: Pages from mentions: a third of extracted candidates are named, not treated — measurement and a deterministic gate (offer, no PR yet)
## Observation

On a German vault (10 curated notes, 115 extracted entity/concept candidates, `gemma-4-26b` via LM Studio), roughly a third of what the extraction names is not something the source *treats*:

| where the candidate's name stands in the source body | share |
|---|--:|
| nowhere — the name does not occur in the text at all | 9.6 % |
| only inside parentheses, an enumeration, or a short list item | 19.1 % |
| in running prose at least once | 71.3 % |

The first row is a page named after something the source never says (`Metabolismus` for a text that says *Stoffwechsel*, `ATP-Produktion`, `Pectin`); the second is a page made from a mention (`Kohlenhydrate` from "Makronährstoffe (Kohlenhydrate, Proteine, Lipide)"). Both classes get a page, a dedup call and a generation call today, and their pages carry exactly one source with exactly one mention — on this vault 58 % of all entity pages look like that.

This is a deterministic property of the text, not a model judgment. Two regexes find it before any further call is spent. Measured against three prompt-side formulations of the same question ("defined / discussed / named"), the prompt answers were stable *within* a formulation (91–100 % self-agreement over three draws) and inconsistent *between* formulations (0.9 % to 13.9 % filtered) — the model is consistent with itself and only loosely coupled to the text; the deterministic check is the part that does not drift.

## Why this matters — the intent

The ingest should not produce a dictionary of the notes. A wiki page is a claim: *this source treats this thing*. A page made from a passing mention makes no such claim — it records that two words stood near each other, which is what an embedding already measures, cheaper. A graph whose edges come from mentions cannot beat a nearest-neighbour search because it asserts nothing the search does not; a page that exists only because the source actually treats the thing carries a judgment, and judgments are what a knowledge graph has that a co-occurrence statistic has not. The threshold is a product decision, and it has to live somewhere; I would rather it lived in code, where it is visible and can be moved, than in a prompt, where it drifts between formulations as the measurement above shows.

## What I built, and why I am not opening a PR yet

A gate between analysis and page planning: a candidate whose name is absent from the source body, or present only in parentheses / enumerations / short list items, gets no page and no further call; its name is pruned from the surviving candidates' `related_*` lists so the gate never produces a dead link. Dropped candidates are reported once per ingest.

Keyed on `settings.wikiLanguage`, because what a word boundary, an inflection or a connective is depends on the language the page names are written in. The module carries one profile per language — `de` (measured), `en` — and is a **no-op for every language without a profile**: CJK has no word boundaries to match on, and an unmeasured fallback would be a second claim, not a default. It is also a no-op when the note declares a `language:` that differs from the wiki language (your `isCrossLanguage` path): the names are translations then, and the source cannot contain them. An unprofiled or cross-language vault behaves exactly as today.

Size: one module (~150 lines, all thresholds as named constants), one block in `ingestSource`, 19 tests; the full suite passes unchanged. No setting, no prompt, no schema change.

What I would want before a PR:
- your view on whether "fewer pages, by a rule in code" fits the direction — this touches the same question as #485 (stubs from dead links) from the other side;
- whether this should sit behind a setting (`off | aside | strict`) or ride on `wikiLanguage` alone;
- a measurement on a non-German vault — I only have German ones. The `en` profile is cheap but unmeasured.

If useful I can push the branch for reading, or open the PR against `main` once the shape is agreed.

DocTpoint:
Closes #514.

## Summary

The gate from #514, as a PR: an extracted entity or concept whose name the note does not carry in running text — absent, or present only inside parentheses, enumerations or short list items — gets no page and no further model call, and its name is pruned from the other candidates' `related_*` lists so the gate never manufactures a dead link. **Opt-in** (`skipMentionOnlyCandidates`, off by default, bottom Advanced settings panel): fewer pages is a behaviour change, so it is the user's choice, not an upgrade's. This answers the second open question in #514 (setting vs. `wikiLanguage` alone) with "both": the language profile decides whether the gate *can* read the text, the setting whether the user *wants* it to.

Measured on one German vault (10 notes, 115 candidates): 9.6 % of candidates absent, 19.1 % aside, 71.3 % prose — the numbers and the prompt-side comparison are in #514.

## What changed

- `src/core/candidate-gate.ts` (new, 350 lines) — `classifyCandidate(text, name, profile)` → `prose | aside | absent`; `gateCandidates(analysis, sourceText, language)` keeps the prose candidates, prunes dropped names from the survivors' `related_*` lists, reports what was dropped and why, and returns `applied: false` when the language has no profile. All thresholds are named constants at the top of the file.
- **Language profiles.** Two shapes:
  - *word-script* — the name's words are matched at word boundaries, each with an optional additive inflection suffix; chunk length for the enumeration test is counted in words minus connectives. `de` is the **measured** profile. `en`, `fr`, `es`, `pt`, `nl`, `ko` are **estimated** from the language's inflection and have not been measured on a vault (Korean particles attach to the noun additively, so it fits this shape).
  - *character-script* — no word boundaries, no inflection: the whole name is matched as a substring, chunk length is counted in characters, and the language's dedicated list separator marks an enumeration outright — `、` in Chinese (`zh`, `zh-Hant`), `・` in Japanese (`ja`, where `、` is the general comma). **Estimated**; the character thresholds (12 for an enumeration chunk, 16 for a list item) are a guess from word-to-character ratios and are the first thing to measure on a Chinese vault.
  - No profile, on purpose: languages whose plural or case forms change the stem (Italian vowel alternation, Swedish/Danish/Norwegian `-a → -or/-er`, Slavic declension). A suffix profile under-matches there and drops candidates the source does treat — the wrong direction for a guess. Each estimated profile's known edge is pinned in the tests (Portuguese `-l → -is`, for one) so the limit is visible, not discovered.
- `src/wiki/wiki-engine.ts` — the gate runs between analysis and page planning when the setting is on. A note that declares a `language:` other than the wiki language is not gated (its names are translations, the source cannot contain them). A wiki language without a profile is **reported** on the progress channel once per ingest (`Candidate gate: no language profile for "it" — not applied`), not silently skipped — the user turned this on.
- `src/types.ts`, `src/ui/settings-sections/advanced-settings-section.ts`, `src/texts/*.ts` — the setting (default `false`), its toggle below the welcome-note toggle and above the dedup sub-group (an ingest policy, not an LLM sampling parameter), name and description in all eleven locales; closing the Advanced panel resets it to off like the other gated fields.

## What it does not do

- No prompt or schema change, no CLI change.
- It does not touch pages that already exist — a thing gated out of source A and found in prose in source B gets its page with B, and A is not among its sources. The page set stays order-independent; the evidence set does not. Stated in the module header.
- It is independent of #520 (candidate window): this decides whether a candidate reaches dedup, that decides what dedup sees. Mergeable in either order.

## Tests

- `src/__tests__/core/candidate-gate.test.ts` (22): classification on a German fixture (prose / parentheses / enumeration / nested clause that is not a list / short vs long list item / inflection but not compounds / case and NFC / headings), pruning of `related_*`, no-op and region-code resolution, and one block per estimated profile with its documented edge.
- `src/__tests__/wiki/wiki-engine-candidate-gate.test.ts` (5): pages written only for prose candidates when on; cross-language note not gated; no-profile language reported and not gated; **off by default — an upgrade writes the same pages as before**; silent when everything is prose.

Gate 1 on this machine against the same run on `main` (603c9ee): lint 0 → 0, typecheck clean, 236 files / 3,446 tests → 238 / 3,473, build clean, css-lint 0. `openai-codex-loopback-flow.test.ts` fails in both trees until `main.js` has been built once, then passes — not related.

## What I would ask of the review

- The character thresholds for `zh`/`ja` and the estimated profiles in general: you have the vaults to measure them on; I have German ones. If a profile is wrong for a language you know, dropping it is a one-line change and leaves that language as it is today.
- Whether the toggle's place (bottom Advanced panel) is where you want ingest policies to live; it is one `Setting` block to move.


green-dalii (approved):
Right shape for a behavior change: opt-in via the bottom Advanced panel (ingest policy, not LLM sampling), no-op without a language profile and reported rather than silently skipped, cross-language notes excluded for the correct reason, and stem-changing languages refused a profile rather than guessed at — the tests pin each estimated profile's known edge so the limit is discoverable. Approving. Follow-up I owe you: a zh/ja measurement pass on a Chinese vault against ENUM_MAX_CHARS=12 / LIST_ITEM_MAX_CHARS=16 before calling those two profiles measured; if either threshold misfires the fix is the one-line drop you built for.

PR #523, opened 2026-08-21
fix(frontmatter): carry block-form unknown fields through the constraints pass (#356 parity)

### LINKED ISSUE #522: enforceFrontmatterConstraints drops the entries of a block-form unknown field (#356 parity gap)
## Summary

`enforceFrontmatterConstraints` (`src/core/frontmatter.ts`, the passthrough collection at ~`:640–656`) is the one writer that does not take its passthrough lines from `extractPassthroughLines`. It collects them inside its own line walk — every line trimmed, every `- ` item skipped — so an unknown field survives only when it fits on one line. A block-form list under a user-owned key keeps its header and loses every entry: `related_links:` comes back as a bare key, i.e. a YAML null.

## Reproduce

```yaml
---
type: concept
tags: [a]
related_links:
  - "[[Foo]]"
  - "[[Bar]]"
---
```

`enforceFrontmatterConstraints(content, 'concept')` returns

```yaml
---
type: concept
created: 2026-08-21
updated: 2026-08-21
related_links:
tags: [a]
---
```

The single-line shape (`redirect_to: "[[x]]"`, `parent_org: Acme`) is carried correctly — that is the shape the #356 follow-up tests exercise, so the gap has no test.

## Where it bites

The pass is the last writer on three paths: page generation (`create-page.ts:224`), fill-empty (`fill-empty-page.ts:99`) and the duplicate merge (`merge-duplicates.ts:170`). The first two hand it the model's reply; the third hands it the survivor's serialized frontmatter — and with #513 that now includes the survivor's unknown fields, which this pass strips again one line later when they are block-form. So #513 closes the parity gap for single-line fields and leaves the list case open at this seam; its red/green test uses single-line fields and is green either way. `sources:` met the same mechanism once (#438 B) and was patched around it by re-emitting the parsed array; the mechanism itself stayed.

## Fix

Take the lines from the shared helper — `const passthroughLines = extractPassthroughLines(content);` — and drop the loop's own collection (the loop keeps collecting type/tags/aliases as before). `sources:` keeps its re-emit from the parsed array. One observable side difference: a literal `reviewed: false` line used to ride through the old loop verbatim and is now omitted, as `mergeFrontmatter` has always done (the serializer emits `reviewed` only when true).

PR follows with the change and a red/green test (block-form list loses both entries before and keeps both after; a single-line unknown field as the control, unchanged). Lint, typecheck, build and 3448 tests green.

DocTpoint:
Closes #522.

## What

`enforceFrontmatterConstraints` collected its passthrough lines inside its own line walk — every line trimmed, every `- ` item skipped — so a user-owned block list under an unknown key came back as its header alone (`related_links:` with no entries, a YAML null). Every other writer — `mergeFrontmatter`, both array helpers, and the duplicate merge with #513 — takes the lines from `extractPassthroughLines`, which keeps the indented continuation lines of an unknown key.

Same helper here, same semantics: `const passthroughLines = extractPassthroughLines(content);`, and the loop no longer builds its own list (it still collects type/tags/aliases). `sources:` keeps its re-emit from the parsed array (#438 B); the helper already treats it as known.

## Side difference, stated

A literal `reviewed: false` line used to be carried verbatim by the old loop and is now omitted — the same as `mergeFrontmatter` has always done, because the serializer emits `reviewed` only when true. No other canonical line changes shape; the two `- ` item skips that fed the bug were only ever reached for unknown keys.

## Test

`src/__tests__/core/frontmatter.test.ts`, in the #356 passthrough block:

- red → green: a block-form `related_links:` list loses both entries before, keeps both after (asserted on the lines and on `parseFrontmatter` of the result);
- control: a single-line `redirect_to:` + `parent_org:` page through the same pass, unchanged before and after.

Gate: lint, typecheck, build, 3448 tests.

## Relation to #513

Independent change in a different function; applies on `main` with or without #513. With #513 merged, the duplicate merge hands this pass the survivor's unknown fields and this is what keeps the block-form ones alive at `merge-duplicates.ts:170`.

### DIFF
--- a/src/__tests__/core/frontmatter.test.ts
+++ b/src/__tests__/core/frontmatter.test.ts
@@ -942,4 +942,40 @@ related_links:
     expect(result).toContain('  - "[[Foo]]"');
     expect(result).toContain('  - "[[Bar]]"');
   });
+  it('enforceFrontmatterConstraints preserves an unknown scalar field', () => {
+    // Control for the block-list case below: the single-line shape has been
+    // carried since the #356 follow-up and must stay byte-identical.
+    const content = `---
+type: concept
+tags: [a]
+redirect_to: "[[canonical]]"
+parent_org: Acme
+---
+
+# Body`;
+    const result = enforceFrontmatterConstraints(content, 'concept');
+    expect(result).toContain('redirect_to: "[[canonical]]"');
+    expect(result).toContain('parent_org: Acme');
+  });
+
+  it('enforceFrontmatterConstraints preserves an unknown list field (YAML block form)', () => {
+    // Same page shape as the array-helper case above, through the constraints
+    // pass. Before the fix the pass kept the header line and dropped every
+    // `- ` entry beneath it, so a user-owned block list came back as a bare
+    // `related_links:` — the key survived, its value did not.
+    const content = `---
+type: concept
+tags: [a]
+related_links:
+  - "[[Foo]]"
+  - "[[Bar]]"
+---
+
+# Body`;
+    const result = enforceFrontmatterConstraints(content, 'concept');
+    expect(result).toContain('related_links:');
+    expect(result).toContain('  - "[[Foo]]"');
+    expect(result).toContain('  - "[[Bar]]"');
+    expect(parseFrontmatter(result)?.related_links).toEqual(['[[Foo]]', '[[Bar]]']);
+  });
 });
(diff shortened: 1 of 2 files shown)


green-dalii (approved):
Verified on main: the constraints pass's own line walk did skip every '- ' item (frontmatter.ts:641), so block-form unknown fields came back as bare keys — and the #356 follow-up's single-line tests were exactly why this had no test. Swapping to extractPassthroughLines(content) matches every sibling writer; the disclosed reviewed:false side difference matches mergeFrontmatter semantics. With #513 this closes the parity chain end to end — approving both as one batch.

PR #404, opened 2026-08-03
feat: add MinerU online PDF conversion backend

XEurekaX:
## Summary

This PR implements the thin online-service integration described as **Option A** in #376.

- Adds an explicit PDF conversion backend selector: existing provider PDF support or MinerU online API.
- Keeps the existing provider path as the default, so current behavior is unchanged unless the user opts in.
- Converts PDFs through the official MinerU API and feeds the resulting Markdown into the existing wiki ingest pipeline through `contentOverride`.
- Stores the MinerU token in Obsidian SecretStorage and migrates the unpublished legacy plaintext fields if present.
- Adds localized settings and progress messages across all currently supported locales.

## Implementation

The MinerU path is intentionally narrow:

1. `convertPdfToMarkdown` dispatches to MinerU only when `pdfConversionBackend === "mineru"`.
2. The plugin checks the PDF size and cache before making any network request. The cache key includes the PDF content hash and `mineru:vlm:v1`, so converter changes can invalidate stale results.
3. It requests an upload lease from `POST /api/v4/file-urls/batch` with `model_version: "vlm"`.
4. It validates the returned upload URL, uploads the PDF with `PUT`, and polls the batch result endpoint until completion.
5. It validates and downloads the result archive, then extracts exactly one `full.md` using `fflate` with strict UTF-8 decoding.
6. The Markdown is returned as the normal `ConversionResult`; `WikiEngine` re-enters the existing `ingestSource(..., { contentOverride })` flow without introducing a parallel analysis pipeline.

The UI reports upload, conversion wait, and download phases. Cancellation and the shared 30-minute deadline return control to the existing ingestion lifecycle, including status cleanup.

## Safety and boundaries

- MinerU token is read from a fixed SecretStorage ID and is not persisted in plugin settings.
- Upload and download URLs must use HTTPS, contain no credentials, and cannot target obvious local/private hosts.
- PDF input is capped at 200 MB.
- Downloaded archives are capped at 256 MB.
- Archive scanning is capped at 10,000 entries.
- Only one `full.md` is accepted, capped at 10 MB; images and unrelated archive entries are ignored.
- Native PDF conversion, sidecar behavior, provider selection, and the downstream wiki generation pipeline are unchanged.
- This PR does not bundle the MinerU CLI, add a self-hosted endpoint, or persist MinerU images/artifacts in the vault.

## Testing

- `eslint src/ --max-warnings=0`
- `tsc --noEmit`
- Production build
- Vitest: **217 files / 2942 tests passed**
[...]

PR #525, opened 2026-08-22
fix(extract): text mode by default, repetition-loop guard, task-policy control (Issue #524)

### LINKED ISSUE #524: Extraction under `json_schema` degrades on LM Studio — often silently, as valid but near-empty JSON. Proposal: `text_prompt` default for `extract` plus batch-integrity guards (PATCH, not 1.27.0)
## Summary

Since `266d641` (#443, shipped in 1.26.3) the extraction call carries `response_format: json_schema`. On LM Studio with `gemma-4-26b-a4b-qat` (MLX) the extraction under that schema degrades in 3 of 3 replays of a real pipeline request; in 2 of the 3 the degradation is **silent** — `finish_reason: stop`, schema-valid JSON, 4–5 items where the text-mode arms return 17–30. The same request without `response_format` ran without a loop in 6 of 6 replays. LM Studio's own server logs show the same split across the whole history of this vault: 0 repetition loops in 2,355 extraction calls before the schema (May–13 Aug), 19 % `length` and a falling item count after it. The effect is not new content being destroyed but new content not arriving, and the plugin cannot see it: every guard on the extraction path triggers on *empty*, *truncated* or *unparsable*, never on *short and valid*.

We proposed the schema in #443 and approved the PR that migrated the extraction call; every gate along the way measured form (HTTP status, wire body, schema shape), none measured content. This is the first content measurement of the real request under the real schema.

## What we measured

**Replay of a captured pipeline request.** The exact `extract` request of the shipped plugin (1.26.4, captured with a local proxy: system 2,873 chars, user 39,007 chars, `max_tokens 16000`, no `temperature` set, note of 30 KB), three draws per arm, same machine, same hour:

| Arm | `response_format` | draw 1 | draw 2 | draw 3 |
|---|---|---|---|---|
| **A** — as shipped | `json_schema` (SourceAnalysis), reasoning 0 | `stop`, 2,047 tokens, **4 items**, 624-char repetition loop inside a quote, `coverage`/`domains` missing | `stop`, 2,184 tokens, **5 items**, key mangled to `"source_ slug"`, `key_points` empty | `length` at 15,999, unparsable, 33,669-char loop, 316 s |
| **B** — text, one JSON object per line | none | 18/18 lines valid, 17 items | 28/30 lines valid, 27 items | 20/20, 19 items |
| **C** — text, today's single-object format | none | 30 items | **0 items** — one invalid escape (`\ge`, the model turned “≥” into LaTeX) voids the whole object; 21 items would survive an escape repair | 18 items |

A is damaged 3/3, twice as *valid* output; B and C show 0 loops in 6. B and C were run with reasoning on (3.4–8.2 k reasoning tokens), which is why the C1 draw sits at 14,824/16,000 — see the log table below for why reasoning is not required.

**LM Studio server logs of this vault, May–August** (55,932 responses; extraction calls identified as prompt ≥ 5,000 tokens with `max_tokens 16000`; response bodies complete in the logs; loop = a repeated run ≥ 200 chars):

| Regime | calls | `length` | loops ≥ 200 chars | parsable | items (median) |
|---|---|---|---|---|---|
| before 14 Aug — no `response_format`, `temperature 0.15`, reasoning off (`reasoning_tokens 0` in 2,160 of 2,355) | 2,355 | 2 % | **0** | 93 % | 6 |
| since 14 Aug (1.26.3 deployed) — `json_schema` | 125 | 19 % | 3 | 80 % | 3 |

The pre-schema regime is exactly the cell “text mode, reasoning off”. So the factor is the constrained decoding, not the absent reasoning — turning the schema off is sufficient, reasoning need not be turned on (and when it is, it eats about half of the 16,000 budget, the #403 class).

**The pipeline itself, not a probe:** on 19 Aug four full ingests of the unmodified 1.26.4 ended 4/4 with `length` at 15,999 (loop in the output, batch discarded, extraction empty). On 17 Aug the same configuration had 38 `stop` / 2 `length`. The rate moves from day to day; the arm does not.

An earlier, weaker data point points the same way: on 3 Aug (#402, comment 2) two short-prompt probes on `gemma-4-12b` under `json_schema` returned empty or placeholder `entities`/`concepts` — a signal at the time, not a proof, and not followed up.

## Why we do not think this is a measurement artifact

Three independent sources agree and only one of them is our instrument: (1) git — the extraction call carries a schema since `266d641` and never before; (2) the server logs, written by LM Studio long before anyone asked the question; (3) the shipped pipeline on 19 Aug. The replay above is the fourth.

## Why it is silent

- `checkEmptyBatch` stops on **zero** items; a 4-item batch passes.
- `canHalveBatch` / the truncation retry fire on `length`; a `stop` passes.
- `parseJsonResponse` + repair fire on **unparsable** JSON; schema-valid JSON passes.
- The Zod schemas are `.passthrough()` with optional fields below the two top-level arrays — the question from our approval on #447 — so an item that omits fields validates.
- A source page is written, the source counts as ingested, and a re-ingest happens only on an explicit trigger. Nothing tells the user that most of the note never arrived.

## Proposal — four layers, the first two small enough for a PATCH

1. **Default `extract` / `extract-retry` to `text_prompt`.** The mechanism exists since #490 (`taskPolicies`); today `default` lets the `OutputModeProber` start at `json_schema`. Flip the built-in default for the long-output generative step; keep the schema for the short judgement calls (`lemma-classify`, `merge-triage`, the dedup pair), where nothing is measured yet and the outputs are short. Before 14 Aug every user ran extraction in text mode; the schema on extraction has had eight days and no content measurement in its favour. The body of #490 (ours) argued against a provider opt-out because “the schema is what made the typed output reliable on LM Studio” — we withdraw that sentence; it was a premise, and the measurement above contradicts it for the extraction step. **Also expose `taskPolicies` in the settings tab** — the field exists in `types.ts:343` but no UI reaches it, so today no user can opt out without editing `data.json`.
2. **Batch-integrity guards, backend-independent:** a repetition detector on the raw response (a repeated run ≥ 200 chars marks the batch damaged → treat like truncation: halve and retry) · an item-count plausibility check against the requested batch size on `stop` · required fields at item level instead of optional-under-`.passthrough()` · `finish_reason`, items per batch and `reasoning_tokens` in the ingest log. This is what protects users on backends nobody has measured.
3. **One JSON object per line for the analysis output** (arm B above). Only meaningful in text mode; a one-character defect then costs one line instead of the batch, and broken lines become countable. Larger change, separate issue if you want it.
4. **Prober demotion on content, not only on HTTP 400.** `length` or a detected loop under `json_schema` on a step that terminates in text → demote `(baseURL, model)` for that task, the way the 400-strip does today. Most speculative; raising it as a question.

We can open a PR for layers 1 + 2 against `main` at once. We would ask for a PATCH milestone rather than `v1.27.0 MINOR`: every user on 1.26.3+ with a grammar-constrained local backend is on arm A today (we measured LM Studio; Ollama and `custom` take the same Tier 0 path and are unmeasured), and the loss does not announce itself.

**Workaround until then** (edit `data.json`, no UI; wired through `create-plugin-llm-client.ts:53`, which we verified in code but have not yet exercised through the Obsidian UI path): `"taskPolicies": { "extract": { "outputMode": "text_prompt", "thinking": "off" }, "extract-retry": { "outputMode": "text_prompt", "thinking": "off" } }`.

## Caveats

One backend (LM Studio 0.4.21, MLX engine), one model (`gemma-4-26b-a4b-qat`, 4-bit), one note, n = 3 per arm on the replay; the rate drifts by machine day (17 Aug 38/40 clean, 19 Aug 0/5, 22 Aug 3/3 damaged) but always on the schema arm. The logs see loops and `length`, not mangled keys or missing fields — the quieter form in the pre-schema regime is unmeasured. We have not measured cloud providers, and the schema may well be the right default there. The replay script, the captured request and the log tabulation are available if useful.

Environment: macOS, Mac mini M4 Pro / 48 GB, LM Studio 0.4.21+2, `google/gemma-4-26b-a4b-qat`, plugin 1.26.4; code references are against `main` at `769e7bb`.

DocTpoint:
Closes #524. One commit, `57b9f9c`, against `main` at `769e7bb`.

## What this changes

**1. `extract` / `extract-retry` run in text mode by default** — `BUILTIN_TASK_POLICIES` in `src/core/task-policy.ts`, resolved below any user policy (a user entry for the task or a `*` wildcard still wins). No `response_format` on the wire for these two calls, the JSON-enforcement prefix in the system prompt instead — the wire shape every user had before 1.26.3 for the one long-output generative step. The short judgement calls (`lemma-classify`, `merge-triage`, the dedup pair) keep the prober's default; nothing is measured there yet and their outputs are short. The thinking axis is not touched: `disableThinking` passes through exactly as before.

**2. A repetition loop is no longer accepted as a short batch** — `src/core/repetition-loop.ts` (new) flags a unit of 3–120 characters repeated at least four times in a row over ≥ 200 characters, provided it contains a letter or digit. `source-analyzer.ts` treats a flagged batch like truncation: halve and retry while the budget allows, otherwise parse what arrived and warn. The debug log now carries `finish`, `reasoning_tokens` and the first batch's item counts, so a user can read the regime off the log.

**3. `taskPolicies` gets a control** in the LLM Advanced section (custom mode), next to *Disable thinking*: spec `step=mode:thinking`, parsed on change with `parseTaskPolicySpec`; an unreadable spec is reported on the input (`setCustomValidity`) and not saved, so the last good value stands. Blank = built-in baseline. Reset together with the other custom-mode fields. Two i18n keys in all 11 locales — the non-English strings are mine; please correct any that read wrong.

## Measurement behind the default

| Arm | `response_format` | draw 1 | draw 2 | draw 3 |
|---|---|---|---|---|
| as shipped | `json_schema`, reasoning 0 | `stop`, **4 items**, 624-char loop inside a quote | `stop`, **5 items**, key mangled to `"source_ slug"` | `length` at 15,999, 33,669-char loop |
| text, single object | none | 30 items | 0 items (one `\ge` escape voids the object) | 18 items |
| text, one object per line | none | 17 items | 27 items | 19 items |

Real captured request, LM Studio 0.4.21 / `gemma-4-26b-a4b-qat` (MLX), Mac mini M4 Pro. Server logs of one vault, May–Aug, extraction calls (prompt ≥ 5,000 tokens, `max_tokens 16000`): with the detector from this PR, **14 loops in 125 schema-mode calls (11 %) against 9 in 2,368 text-mode calls (0.4 %)**; `length` 19 % against 2 %. One correction to the issue text: its “0 loops” used a word-level criterion; the detector here also sees phrase loops, hence 9, all of them runs of ≥ 2,280 characters that ended in `stop`. The calibration of the 200-character threshold comes from the same archive: 0 of 2,319 clean text-mode `stop` responses flagged.

## Tests and gate

+18 tests: `task-policy` 5 (baseline, user entry wins, wildcard wins, thinking untouched), `llm-client-wrapper` 2 (`outputModeOverride: text_prompt` reaches the wire for `extract`/`extract-retry` with no policy configured; a configured policy moves it back), `repetition-loop` 9 (word loop, phrase loop inside valid JSON, multi-line unit, punctuation/fences/table rules ignored, minimum length, caller minimum, longest run, unit trim), `source-analyzer` 2 (looped first batch → halve → clean retry kept, damaged items absent; clean batch untouched). The existing “unset policy changes nothing” tests now exclude the two baseline steps and say why.

Gate 1 on `57b9f9c`: `lint` 0 · `typecheck` 0 · **3500/3500** · `build` · `css-lint` 0. Note for anyone reproducing on a fresh checkout: `openai-codex-loopback-flow.test.ts` reads `main.js`, so `pnpm build` has to run once before `pnpm gate:1`; and `#404` added `fflate`, so a worktree needs `pnpm install` — both independent of this PR.

## Left out on purpose

- **Item-level `required` fields in the Zod schemas.** Your #463 stance (“no scope creep”, redundancy room for variable fields) stands; if you want `summary`/`type` required at item level for the schema-pinned path, that is a one-line follow-up.
- **An item-count plausibility check.** A short note produces the same signal as a degraded batch; without a reference it is noise. The loop detector is the signal that separates the two.
- **Prober demotion on content** (issue layer 4) — raised as a question there, not built.
- **One object per line for the analysis output** (issue layer 3) — separate issue if you want it.

## Compatibility

No `data.json` migration. A user who already set `taskPolicies` in `data.json` keeps exactly that; a user who never did moves from `json_schema` to `text_prompt` on `extract`/`extract-retry` only. Anthropic / OpenAI / Codex clients ignore `outputModeOverride` and are unchanged. The 1.26.3 migration comment in `source-analyzer.ts` is left in place — it describes the typed path, which is still the path; only the mode it resolves to changed.


DocTpoint:
Rebased onto `8376ba1` after today's merges — `57b9f9c` → `173c47c`. The only conflict was the CHANGELOG bullet next to #517's; no code changes. Gate 1 against the new base: lint 0, typecheck 0, build clean, tests 3493 → 3511.

DocTpoint:
Rebased onto `548ba99` after today's merge wave — `173c47c` → `63ec125`. Only conflict was the CHANGELOG bullet next to #528's; no code changes. Gate 1 against the new base: lint 0, typecheck 0, build clean, tests 3534 → 3552.

green-dalii (approved):
Approved after a max-effort adversarial review: all four touched test suites re-run green on the PR tree (63/63), tsc clean, i18n parity 110/110; the detector regex probed for catastrophic backtracking (polynomial, ~linear on adversarial inputs); lookup order, wire shape (no response_format leaks on the compat path), halving termination, and interplay with the just-merged type-repair intake (#528) all verified clean.

Non-blocking follow-ups for the tracker (none gate this merge):

1. `src/llm-sdk/openai-codex-sdk-client.ts:124` — the codex client never reads `outputModeOverride`, so extract stays JSON-output mode on that provider despite the built-in pin; honor `'text_prompt'` there, or scope the release-note wire-shape claim.
2. `src/wiki/source-analyzer.ts:437` — source-borne loops (legitimately echoed refrains quoted 4×) deterministically burn the single halve-retry and merge the same batch anyway; cheap mitigation: check whether the loop unit occurs verbatim in the source note before spending the retry.
3. `src/__tests__/wiki/source-analyzer.test.ts` — the exhaustion arm (retry budget spent → damaged batch still parsed and merged) has zero coverage; one test would pin the exact safety property users rely on.
4. `src/ui/settings-sections/advanced-section.ts:116` — hardcoded English placeholder in an otherwise fully localized control.
5. `src/ui/settings-sections/advanced-section.ts:66` — switching LLM Advanced mode to default now also erases hand-edited `taskPolicies`; worth one upgrade-note sentence.
6. Spec parser accepts unknown step names and silently last-wins duplicate entries — validate against the label list or document.

PR #526, opened 2026-08-22
fix(tools): dev-instrument exit code follows the ingest report (Issue #417 secondary)

DocTpoint:
## What

The dev instrument's process exit code now comes from the ingest report, not from whether something threw.

| code | meaning |
|---|---|
| `0` | the engine reported `success: true`. A requirements-gate skip (empty, unsupported type, duplicate) is reported by the engine as success + `skipped` and stays `0`; the summary names the skip |
| `1` | the report says `success: false`, the engine returned without a report, or the run threw (before or inside the engine) |
| `2` | `<vault> <source>` missing — usage goes to stderr, nothing to stdout |

## Why

PR #418's merge note set this aside: "the exit-code contract change deserves its own PR." Measured on `main` (`002da74`) before this change: no arguments → `0` with usage on stdout; a vault whose local provider points at a closed port → `1`, but only because `WikiEngine.ingestSource` rethrows after it has already handed `onDone` a `success: false` report; missing vault → `1`. The `success` field the summary prints was never read — so the failure case is right by accident today and wrong the moment the engine reports a failure without throwing (the cancel path already does; headless it is unreachable, which is why nobody has seen it).

## How

- `tools/dev-instrument/src/exit-code.ts` — `exitCodeForReport(report)` plus the three constants; the single seam.
- `engine-runner.ts`: `runIngest` returns the report, `main` returns the code. `run-instrument.mjs` assigns `process.exitCode`; its catch stays at `1`.
- README gains an "Exit codes" table; CHANGELOG `[Unreleased]` gets a **Changed** entry (it is a contract change, not a bug fix in the engine).

## Decisions taken here — say so if you want any of them different

- **Skip stays `0`.** The engine calls it success; the instrument reports what the engine reports. If a measurement script needs "nothing was measured" distinguishable from "measured fine", that is a fourth code and a separate decision — not taken without a need.
- **Setup errors stay `1`, not `2`** (missing vault, unreadable `data.json`, unknown measurement arm, missing API key, source not in the index): they are failures of the run, not of the invocation.
- **Usage moves to stderr.** Exit `2` declares it an error path; stdout stays reserved for the `[cli]` header and the summary.

## Tests

`src/__tests__/tools/dev-instrument/exit-contract.test.ts` — unit half pins `exitCodeForReport` (success / skip / failure / no report); process half spawns the real `run-instrument.mjs` (own esbuild step, ~0.15 s per start): no positionals → `2`, usage on stderr, empty stdout; lmstudio `baseUrl` on a closed port → summary prints `success false` and the process exits `1`.

Gate 1: lint 0 / tsc 0 / tsc tools 0 / build clean / 3433 → 3439 tests / css-lint 0; `lint:tools-bot` baseline unchanged (1 error + 3 warnings).

## Not in this PR

- The README cites `run-instrument.mjs:78` for the `no-unsanitized/method` finding; the line is 75 now (it was already off before this change). One-liner if you want it folded in here.

Refs #417, #418


green-dalii (approved):
Approved. Deriving the exit code from the report instead of the accident of a rethrow is the contract scripts actually branch on, and keeping requirements-gate skips at 0 matches engine semantics. Usage to stderr plus a process-level test through the real launcher closes both halves. Good follow-through on what #418 set aside.

PR #528, opened 2026-08-22
fix(ingest): resolve extracted types against the active vocabulary at intake (Issue #527)

### LINKED ISSUE #527: Custom tag vocabulary: one extracted item in ten carries a type outside the vocabulary — the page is born as a lint violation, and the source never decides its tag
Follow-up to #510 (your approval note: "file it when ready"). The order-dependence fixed there was, under a custom vocabulary, the symptom of this.

## The chain

Under `tagVocabularyMode: 'custom'` the extraction system prompt lists the user's terms and says "you MUST use one of the following … Do NOT emit a free-form type string — the frontmatter validator will reject it" (`system-prompts.ts:250-272`). Nothing on the wire enforces it: `type` is `z.string().optional()` in `output-schemas.ts:161/172`. From there:

1. `create-page.ts:190` renders `entity_type: info.type`; the generation template writes it as `tags: [{{entity_type}}]` (`generation.ts:46/104`).
2. `enforceFrontmatterConstraints` sees the tag is not in the active vocabulary and **keeps it**, with a `console.debug` (`frontmatter.ts:685-697`). The validator the prompt threatens the model with does not reject anything.
3. `scanTagViolations` reports the page (`scanners.ts:380`).
4. `runRetagViolations` — hand-triggered — asks the model again, with the page name, the invalid tag and **the first 400 characters of the page's own body** (`fix-runners.ts:525-552`). The source's summary, which is what the type was extracted from, is not in that prompt.

So for every item whose type falls outside the vocabulary, the page starts as a lint violation, and the tag it ends up with is decided by the page's own generated prose, not by the source. Under the default vocabulary the same chain is invisible: the model's fallback *is* the default taxonomy, so the values coincide.

## Measurement — how often the model ignores the vocabulary

LM Studio server logs of my own build, 3–19 July 2026 (gemma-4-26b-a4b-qat, text mode, reasoning off, temperature 0.15, 34-term German vocabulary). Extraction responses identified by their `entities`/`concepts` keys; `type` values read from the items inside those two arrays only.

| | |
|---|---|
| extraction responses | 1 555 |
| items (entities + concepts) | 10 669 |
| type in the active vocabulary | 9 430 (**88.4 %**) |
| type from the built-in default taxonomy | 1 069 (**10.0 %**) — `person` 231 · `organization` 222 · `theory` 201 · `method` 191 · `term` 124 · `other` 37 |
| free-form | 170 (1.6 %) — mostly near-misses: `Methoden`, `Aminosäure`, `Pathophysiologie`, `psychologie` |
| responses with ≥ 1 item outside the vocabulary | 493 (**32 %**) |

Two things stand out. The fallback is not `other` but `person` / `organization` / `theory` / `method` — the taxonomy the model knows from training wins over the setting about one time in ten. And the free-form tail is small and mostly vocabulary neighbours (plural, case), which a deterministic fold would catch; the default-taxonomy share would not be caught by anything deterministic.

Before the vocabulary was switched on (1–10 June) the same vault shows 0 % vocabulary terms; from 11 June the picture above. The request bodies in the log are truncated, so the per-request prompt cannot be checked — the date boundary is the evidence that the setting was active.

## What #510 changes, and does not

#510 makes merges carry the source's type into `tags:` — under a custom vocabulary only when the value is admitted (`incomingTypeTag`). So #510 is right to gate, and for the 10 % it is a no-op: the merge cannot pass on what the extraction did not produce. This issue is the upstream half of that gate.

## Proposal

The repair already exists — `runRetagViolations` is the mechanism — it just runs at the wrong time with the wrong input. One seam, in the analyzer where `firstActiveTag` (`source-analyzer.ts:818`) already handles the same problem for lemma pages:

- **(a) resolve at intake:** when `info.type` is outside the active vocabulary, one short call with the item's own `summary` and the vocabulary list, same prompt shape as retag. Cost: one small call per out-of-vocabulary item (~10 % of items on my numbers); the source decides the tag; the page is never born as a violation.
- **(b) deterministic fold first** (case, diacritics) against the vocabulary — catches part of the 1.6 % free-form tail, not the 10 %.
- **(c) leave the chain as is** and let lint own it — then the README should say that under a custom vocabulary a share of new pages will show up as tag violations until retag runs, and that the retag decides from the page, not the source.

A PR with (b) + (a) follows in this session, fail-safe on every path (parse failure, answer outside the vocabulary, call error → the item keeps its extracted type, i.e. today's behaviour) and with the `console.debug` kept so the intake repair is visible in the log. If you would rather have (c), say so and I will close the PR.

## Not claimed

One model, one backend, one vocabulary; the ratio will differ elsewhere. The log's request side is truncated, so the classification runs on responses. The `enforceFrontmatterConstraints` retain-and-debug behaviour itself is not in question here — this is about what reaches it.

DocTpoint:
Closes #527 — the (b) + (a) proposal from the issue, as one seam in the analyzer.

## What it does

`SourceAnalyzer.repairTypesAgainstVocabulary` runs after accumulation, next to the lemma guarantee, over every extracted entity and concept:

1. **Fold** — `foldToVocabulary` (`core/tag-vocab.ts`): exact, then case-insensitive, then diacritic-insensitive match onto the active vocabulary; returns the vocabulary's own spelling. Plural and inflection are deliberately not folded — `Aminosäuren` is not `Aminosäure` by any rule that is safe across the languages a vocabulary may be written in.
2. **Ask once** — for what the fold cannot place, one short `type-repair` call with the item's own `summary` and the allowed terms (`TypeRepairLLMSchema`, `TOKENS_TYPE_REPAIR = TOKENS_LEMMA_CLASSIFY`; same shape as `classifyLemmaType`, typed-output path, `enableThinking` follows `disableThinking` like its siblings). The answer is folded too.

Every doubt — parse failure, answer outside the vocabulary, call error — keeps the extracted value, which is exactly today's behaviour. Repairs and kept values are logged under `[Type repair]`. Under the default vocabulary the step is a no-op for in-vocabulary items and only touches free-form types.

## Why here

The page template writes `info.type` into `tags:` and `enforceFrontmatterConstraints` retains whatever arrives; the first place that holds the source's words *and* the vocabulary is the analyzer, and `firstActiveTag` already solves the same problem there for lemma pages. Retag stays the normalizer of last resort, as you put it in #510 — it just stops being the first and only decider for one item in ten.

## Cost

One short call per item the fold cannot place: ~10 % of items on my numbers (1 069 of 10 669; the fold would have caught part of the 1.6 % free-form tail — `psychologie`, `Pravention` — none of the 10 % built-in-taxonomy share). Calls for one source run concurrently (`Promise.all`), same as retag's batch.

## Tests

- `core/tag-vocab.test.ts` — fold: exact, case, diacritics, no plural/inflection, built-in taxonomy → null, blank → null (5)
- `llm-sdk/output-schemas.test.ts` — `TypeRepairLLMSchema` widening + required (2)
- `core/task-policy.test.ts` — `type-repair` added to the list of labels production passes
- `wiki/source-analyzer-type-repair.test.ts` — wiring through `analyzeSource`, with a call counter on the mock client: fold + model answers reach the analysis and cost exactly two calls beyond extraction; answer outside the vocabulary / non-JSON keeps the extracted value; the model's own near-miss (`laborwerte`) lands on the vocabulary spelling; default vocabulary with in-vocabulary types makes no repair call (4). **RED on `main` for the three repair cases, GREEN after.**

Gate 1: lint 0 / tsc 0 / tsc tools 0 / build clean / 3433 → 3444 tests / css-lint 0.

## Also touched

- README "Custom tag vocabulary" bullet: one sentence on what now happens to a drifted type before the page is written; the "schema injection hint, not a wire-level gate" framing stays.
- CHANGELOG `[Unreleased]` Fixed entry.

## Not in this PR

- The settings hint (`tagVocabularyNotEnforcedHint`, 11 locales: "the LLM may still emit out-of-vocabulary types. Run Lint to find and fix those pages.") is now half true — lint still catches what slips through, but most of it no longer slips. I left the locale strings alone; your call whether to reword them here or at release.
- `enforceFrontmatterConstraints`' retain-and-debug behaviour is unchanged; this PR is about what reaches it.
- The wire schema keeps `type: z.string()` — an enum of the active vocabulary would enforce at the wire, but it would also force the model to pick under `json_schema` decoding, which is the regime #524 is moving the extraction away from.


DocTpoint:
Rebased onto `8376ba1` — `5e5535a` → `c9d9deb`. Three textual conflicts, all additive: the CHANGELOG bullet, `TOKENS_TYPE_REPAIR` now sits above #520's new `CANDIDATE_WINDOW_TOP_K` comment in `constants.ts`, and `tag-vocab.test.ts` carries both #510's `incomingTypeTag` suite and the `foldToVocabulary` one. The two fixes are complementary — #510 withholds an out-of-vocabulary type at merge time, this PR repairs it at intake — and the #510 tests stay green next to ours. Gate 1 against the new base: lint 0, typecheck 0, build clean, tests 3493 → 3504.

green-dalii (approved):
Approved. The fold-then-one-short-call shape is the right intake altitude — the retag question finally gets asked while the source's words are still at hand, and every doubt path keeps the extracted value exactly as the tests pin. Verified the typed-output call mirrors classifyLemmaType byte-for-byte in pattern and that the default-vocabulary no-op holds for all default tags. A max-effort correctness scan returned CLEAN with two non-blocking notes worth a small follow-up: (1) `repairTypesAgainstVocabulary` fires `Promise.all` over all pending items unbounded — every other concurrent LLM path here chunks to a 2–4 window (dedup-phase pattern), and a wholesale vocabulary miss on a long source could fan out to hundreds of simultaneous calls that local single-slot backends queue painfully; (2) the method doc promises more fail-safety than the `buildSystemPrompt`-outside-the-`try` placement delivers (same precedent as classifyLemmaType, so a comment suffices).

Issue #258, opened 2026-07-09
createNewPage prompt drift: LLM emits non-schema sections (`# Title`, `## 基本信息`)


green-dalii:
**Problem**

When LLM generates a new entity/concept page via `createNewPage()`, it sometimes emits sections NOT in the schema (e.g. `# Title` H1 heading, `## 基本信息` with hard-coded `Source: [[...]]` line). This is the existing-page path's complement to #216's `mergePage` triage: the drift originates on creation, then persists across merges because the merge prompt assumes schema-standard sections.

**Reproduction** (v1.24.0, DeepSeek, Chinese-language wiki)

1. Ingest a source mentioning a new entity "SiliconFlow".
2. Open `wiki/entities/siliconflow.md`:
   ```
   # SiliconFlow                    ← non-schema H1 (others lack this)
   ## 基本信息                       ← non-schema section (others lack this)
   - Type: organization
   - Source: [[sources/...]]         ← hard-coded single source (others don't)
   ## 描述
   ...
   ```
3. Compare to `wiki/entities/openrouter.md` (created in same ingest):
   ```
   ## 描述
   OpenRouter 是一个 AI 模型路由平台...
   ```
   OpenRouter has no `# Title` and no `## 基本信息` section — schema-conformant.

**Why it matters**

- Wiki pages are not uniform → queries, summaries, and lint scanners don't know which sections exist.
- Tier-2 complementary append (#216) targets `target_section: "描述"` but on SiliconFlow-style pages the `## 描述` section is preceded by non-standard `# Title` + `## 基本信息` content. The 4-layer fallback resolves the section, but the front-loaded non-schema sections stay forever.
- Linter (`core/section-header-canonicalizer`) only repairs near-misses of canonical labels; it doesn't drop wholly non-schema sections.

**Hypothesis on root cause**

`createNewPage` uses a generation prompt that lists schema sections but doesn't explicitly forbid the model from adding its own scaffolding (`# Title` H1, `## 基本信息` summary table, etc.). DeepSeek appears to add these as a "rounded" output even when not asked.

**Possible fixes** (evaluate, don't apply yet — scope unclear)

1. Prompt engineering: add "Output ONLY the body content starting with the first schema section (`## 描述`). Do NOT include a top-level `# Title` heading or any non-schema sections like `## 基本信息`."
[...]

PR #529, opened 2026-08-22
fix(ingest): strip non-schema sections from generated pages (#258)

DocTpoint:
## What

The schema decides which `## …` sections a page has. `canonicalizeSectionHeaders` already re-asserts that for garbled canonical labels; it leaves invented sections untouched. Local models invent them regularly: the tag-vocabulary block from the system prompt copied verbatim into the page body, a `## 基本信息` or `# Title` heading (#258), a `## Notes` afterthought. #283 removed one prompt-side cause; this PR is the deterministic guard for the class.

## How

- New pure `stripUnknownSections` in `core/section-header-canonicalizer.ts`: drops any `## …` whose header does not snap to a canonical label (exact or near-miss, sharing `snapHeaderToCanonical` with the canonicalizer).
- Runs right after the canonicalizer on the three generation paths (`createNewPage` / `mergePage` / `updateRelatedPage`).
- Content before the first `##` is untouched. Reviewed pages route through `appendToReviewedPage` and never reach this step, so no curated section is at risk.
- Conservative on purpose: the whitelist is every known section label, so only a truly foreign section is dropped and a real one never is.

## Measurement

German vault, local 26B model, v1.2x (the vocabulary block sat in the user prompt then): ~36 % of generated pages carried the copied block as a body section. The block has since moved into the system prompt (#328 follow-up); how often it is still echoed was not re-measured — the guard does not depend on where the block sits.

## Not in this PR

- The misplaced-`## Grundlegende Informationen` variant of #258 (a valid label in the wrong place) — different mechanism.
- Any prompt wording change.

## Tests

6 new (the 3 loss cases verified red under mutation). Against `main` (002da74): tests 3433 → 3439, lint unchanged (35), `tsc` clean, build clean.

### DIFF
--- a/src/__tests__/core/section-header-canonicalizer.test.ts
+++ b/src/__tests__/core/section-header-canonicalizer.test.ts
@@ -4,6 +4,7 @@ import {
   classifyHeader,
   preserveExistingSections,
   reassertH1,
+  stripUnknownSections,
 } from '../../core/section-header-canonicalizer';
 
 describe('canonicalizeSectionHeaders (deterministic repair of LLM-garbled section headers)', () => {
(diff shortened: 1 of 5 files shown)


green-dalii (approved):
Approved. Running `stripUnknownSections` after `canonicalizeSectionHeaders` is the right composition — near-misses are repaired first so only true foreign sections die, and the reviewed-path bypass (`appendToReviewedPage`) keeps hand-curated content out of scope entirely. Passing `Object.values(getSectionLabels(settings))` means custom section labels stay canonical for the strip pass too. The gap-trimming and lead-paragraph tests pin the edges I would have worried about.

Issue #366, opened 2026-07-28
Slug derivation mismatch between link-writing and file-naming produces broken [[wikilinks]] at scale — degrades PPR retrieval


Guru35:
## Bug Description
The slug used when *writing* a [[wikilink]] and the slug used when *naming* the target file are derived inconsistently (observed around case and Turkish character folding: ı/i, İ/i, ç, ş, ğ, ö, ü). Result: links that point to files that exist but under a slightly different spelling — unresolved in Obsidian and dead edges in the plugin's own link graph.

## Environment
- **Plugin version:** 1.25.9
- **Obsidian version:** 1.12.x
- **OS:** macOS
- **Wiki size:** ~11,610 pages / ~198k wikilinks (Turkish-language vault)

## LLM Configuration
- **Provider:** DeepSeek
- **Model:** deepseek-v4-flash (ingest/query), deepseek-v4-pro (lint)
- **Local or cloud:** Cloud

## Steps to Reproduce
1. Ingest Turkish-language sources at scale (non-ASCII entity/concept names: ı, ç, ş, ğ, ö, ü)
2. Let the plugin both create pages and write [[wikilinks]] to them across many ingests
3. Audit link resolution against the actual file index
4. Observe links like `[[entities/dr-doga-demir-keles]]` pointing at a file actually named `dr-doğa-demir-keleş.md` (and the reverse direction too)

## Expected Behavior
One shared slugify function for both file naming and link writing: a link written by the plugin always resolves to the file the plugin itself created.

## Actual Behavior
Measured on our vault — two independent audits, different methods, reported separately (not merged):
- Fuzzy nearest-match audit (difflib): 25,770 broken links (12.9% of 200,030)
- Exact-match audit (lowercase + Turkish folding): 32,702 broken links (16.5% of 198,159)

Repair run using ONLY safe normalization (lowercase + Turkish char folding, then whitespace/underscore→dash), fixing a link only when it resolved to exactly one existing file:
- 4,408 links repaired across 1,856 files → broken rate 16.5% → 14.28%
- i.e. ~4.4k links pointed at files that EXISTED but under a differently-derived slug — pure derivation mismatch
- Remaining breakage is provably out of slug scope: ~20k targets are pages never created, ~7.8k are non-wiki paths

## Screenshots/Logs
Sample repairs (before → after, display text untouched):
[...]

PR #530, opened 2026-08-22
fix(alias): NFC-normalise and Turkish-fold the alias comparison keys (#366, phase 2)

DocTpoint:
## What

#366 shipped phase 1 in v1.25.10 (`slugKeys` with the opt-in Turkish fold). This is the comparison-key half announced as phase 2.

`filterRedundantAliases` compares with bare `toLowerCase()` on three keys (basename, cross-page aliases, candidate); `slugKeys` folds but never normalises. Two gaps, both observed on a German vault:

- no NFC/NFD normalisation anywhere in the comparison path — a basename read back from disk decomposed (macOS) and a composed alias from the model are two keys for one umlaut, so the redundancy/uniqueness gate lets the alias through as new;
- `İstanbul` / `istanbul` pass as two aliases, because `toLowerCase()` turns `İ` into `i` + COMBINING DOT ABOVE; `turkishCaseFold` exists for exactly this and was wired only into `slugKeys`.

## How

Both functions now key on `turkishCaseFold(s.trim().normalize('NFC'))`. The file-naming path (`computeSlug`) is untouched; filenames stay byte-stable.

## Not in this PR

- Any change to how files are named or to existing pages.

## Tests

Three new: NFD basename vs NFC alias → redundant; `İ`/`i` cross-page → one alias; `slugKeys` NFD/NFC equal with and without the fold. Red before (3 failures), green after. Against `main` (002da74): tests 3433 → 3436, lint unchanged (35), `tsc` clean, build clean.

### DIFF
--- a/src/__tests__/core/slug.test.ts
+++ b/src/__tests__/core/slug.test.ts
@@ -225,6 +225,22 @@ describe('filterRedundantAliases', () => {
     expect(result).toEqual([]);
   });
 
+  it('compares NFC-normalised: a decomposed basename from disk and a composed alias are the same name', () => {
+    // macOS hands out NFD filenames ("A" + COMBINING DIAERESIS); the model writes
+    // the composed "Ä". Bare toLowerCase() keeps them apart and the alias would
+    // be kept as if it were new.
+    const decomposedPath = 'wiki/entities/' + 'A\u0308rzte'.normalize('NFD') + '.md';
+    const result = filterRedundantAliases(decomposedPath, ['Ärzte', 'Mediziner']);
+    expect(result).toEqual(['Mediziner']);
+  });
+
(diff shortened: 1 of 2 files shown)


green-dalii (approved):
Approved. NFC-before-keying fixes the macOS-NFD class at the comparison layer where it belongs, and `aliasKey` sharing the Turkish fold keeps one definition of "same name" across both gates. File-naming path untouched and v1.25.9 compat pinned — clean phase 2.

green-dalii (approved):
Approved. NFC-before-keying fixes the macOS-NFD class at the comparison layer where it belongs, and `aliasKey` sharing the Turkish fold keeps one definition of "same name" across both gates. File-naming path untouched and v1.25.9 compat pinned — clean phase 2.

PR #531, opened 2026-08-22
fix(links): key the typed-list folder map on the comparison slug in correctRelatedLinkPrefixes

DocTpoint:
## What

`folderBySlug` — the map from the typed related lists to a folder — was keyed `slugify(name, preserveCase)` on both sides. Consistent with itself, so it never misfired; but under `slugCase: 'preserve'` a link whose spelling differs from the typed list ("Mediterrane Ernährung" in the list, `[[mediterrane ernährung]]` in the body) gets a different key, the lookup misses, and the section context wins over the known type.

## How

#484 already moved the vault index in the same function onto the comparison contract stated in `slug.ts` (comparison callers do not pass `preserveCase`). This applies it to the remaining path. The `preserveCase` parameter is thereby unused and removed from the signature and its three callers (`create-page`, `merge-page`, `related-page`).

## Not in this PR

- Any change to filename casing (`computeSlug` with `preserveCase` is untouched).

## Tests

One new: typed entity "Mediterrane Ernährung", lowercase link in the concepts section → re-typed to `entities/` (red before, green after). Against `main` (002da74): tests 3433 → 3434, lint unchanged (35), `tsc` clean, build clean.

### DIFF
--- a/src/__tests__/core/related-link-corrector.test.ts
+++ b/src/__tests__/core/related-link-corrector.test.ts
@@ -12,7 +12,7 @@ describe('correctRelatedLinkPrefixes (root-cause fix for sources/-prefixed relat
     // The related entity exists in the vault but outside the MAX_PAGES window, so the
     // model could not see its path and guessed the most salient prefix — sources/.
     const c = ['## Related Entities', '- [[sources/Hippocampus|Hippocampus]]'].join('\n');
-    const r = correctRelatedLinkPrefixes(c, ['Hippocampus'], [], ENT, CON, true);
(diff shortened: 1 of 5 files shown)


green-dalii (approved):
Approved. Folding `folderBySlug` onto the comparison slug closes the spelling-mismatch hole where the typed list and the body link disagreed; the ambiguity guard already owns the collision case, so nothing new leaks through. Removing the parameter beats threading it — one comparison contract, zero caller discretion. Display alias untouched per tests.

PR #532, opened 2026-08-22
feat(settings): expose the alias floor as `minAliasLength` (Settings → Advanced, default 2)

DocTpoint:
## What

`MIN_ALIAS_LENGTH` (v1.25.10) is documented as the one-line per-vault edit: "Raise to 3 only if a specific vault surfaces alias clutter." One vault did: at two characters the case-insensitive alias comparison destroys the only discriminator — `Cr` (chromium) and `CR` both fold to `cr`, so the model's `CR` lands as an alias on the Chrom page and resolves links there. A setting makes that edit possible without a code change. The default stays 2; behaviour without the setting is unchanged.

## How

- `types.ts`: `minAliasLength?: number` — undefined → `MIN_ALIAS_LENGTH` (2). Accepted range `MIN_ALIAS_LENGTH_MIN..MAX` = 2..6 (`constants.ts`): one-character aliases stay out, above six the floor starts to drop real names.
- `core/slug.ts`: `filterRedundantAliases` takes `minLength` as a fourth argument (default = constant) and applies it to the candidate and the cross-page keys alike. Existing callers are untouched.
- `page-factory/aliases.ts`: `AliasesContext.settings?` (only `minAliasLength` is read) and `resolveMinAliasLength` — the setting when it is an integer in range, the constant otherwise. `PathResolutionContext` already carries `settings`, so no call site changes.
- UI: text input next to "File Name Casing" in the Advanced panel, clamped to the range with the usual Notice. Texts in all eleven locales.

## Open for review

- Range 2..6 and the placement next to slug case are proposals. Happy to move it or widen the range.

## Not in this PR

- A different default. Two stays right for the technical vaults named in the constant's comment (ML, HD, AI, UI).
- Re-filtering aliases already on disk.

## Tests

Four new: floor honoured on candidate and cross-page keys (and the v1.25.10 default unchanged without the argument); setting applied through `appendAliases`; absent setting keeps 2; out-of-range values fall back. Against `main` (002da74): tests 3433 → 3437, lint unchanged (35), `tsc` clean, build clean.


green-dalii (dismissed):
Approved. Correct panel (Advanced settings beside slugCase — naming policy, not LLM sampling), optional field with guarded fallback, clamp Notice, all 11 locales. Mirror of the #537 note: once both land, routing the generation paths' floor through `resolveMinAliasLength` too would keep every alias writer on one contract.

PR #537, opened 2026-08-23
fix(#536): drop aliases that repeat the page filename on the create path

### LINKED ISSUE #536: Create path persists an alias that equals the page filename — the rule from appendAliases is not applied where the model writes aliases:
**Observed.** On a vault built with the ingest (2401 entity/concept pages created 2026-07), **488 pages (20 %)** carry an alias identical to their own basename — `CD44.md` → `aliases: ["CD44"]`, `Interleukin-10.md` → `["IL-10", "Interleukin-10"]`. The 11 pages created in 2026-08 under 1.26.x show the same: **8 of 11**. Across entities/concepts/sources that is 521 exact repeats plus 6 case-only ones (`D-Mannose` → `D-mannose`). Regime: initial build, LM Studio / gemma-4-26b, default prompts — the model lists the page's own name among the aliases as a matter of course.

**Mechanism (read against `8376ba1`).** The field has two writers and one gate.
- `appendAliases` (`page-factory/aliases.ts:50`) runs `filterRedundantAliases`, which drops an alias whose lower-cased value equals the basename — the comment says exactly why: "so we don't add *Vigilanz* to vigilanz.md".
- `createNewPage` (`create-page.ts:224`) hands the model's frontmatter to `enforceFrontmatterConstraints`, whose serializer (`frontmatter.ts:450-454`, "parity across all writers") only removes literal duplicates. The function does not know the page path, so it cannot apply the rule even if it wanted to. The two lint rewriters that call it (`fill-empty-page.ts:99`, `merge-duplicates.ts:175`) have the same gap.

**Not in scope.** Space/hyphen variants (`Angina pectoris` on `Angina-pectoris.md`, 245 in this vault) are deliberately kept by `filterRedundantAliases` (`slug.ts:98-106`) and stay untouched here.

**Proposed fix** (PR follows): give `enforceFrontmatterConstraints` an optional `pagePath` and apply `filterRedundantAliases` there — one rule, one function, both writers agree; callers that do not pass a path keep today's behaviour.

DocTpoint:
Closes #536.

`enforceFrontmatterConstraints` takes an optional `pagePath` (new `EnforceFrontmatterOptions`, extends the existing date options) and, when it has one, runs the collected aliases through `filterRedundantAliases` before serializing — the same gate `appendAliases` uses. `createNewPage` passes the resolved path; `fillEmptyPage` and `mergeDuplicatePages` pass theirs. Callers without a path are unchanged (pinned by a test).

**Consequence worth naming:** `filterRedundantAliases` also applies the `MIN_ALIAS_LENGTH` floor and a case-insensitive in-batch dedup, so on these three paths a one-character alias and a `Foo`/`foo` pair now fall as well — the rule `constants.ts` describes as what the plugin "will ever accept". If you would rather have only the basename check here, it is a two-line change.

Tests: 4 on `enforceFrontmatterConstraints` (basename, case-insensitive, space-variant and other aliases kept, legacy no-path unchanged) + 1 on `createNewPage` end to end. RED before the change (4 of 5; the legacy case is green by construction), GREEN after. Gate against `8376ba1`: lint 0 · tsc 0 · build · 3493 → 3498 · css-lint 0.

Touches `core/frontmatter.ts`, `page-factory/create-page.ts`, `lint/fill-empty-page.ts`, `lint/merge-duplicates.ts`, CHANGELOG. No overlap with the open #529–#532 beyond #532 adding a trailing optional parameter to `filterRedundantAliases`, which composes with this call.


green-dalii (approved):
Approved. Applying `filterRedundantAliases` at both writers of `aliases:` is the right altitude — the rule travels with the field rather than with whichever path noticed it first, and keeping the full filter (basename + floor + in-batch dedup) matches what constants.ts already promises. Legacy no-path callers pinned unchanged. One follow-up once #532 lands: `enforceFrontmatterConstraints` could take its floor from `resolveMinAliasLength(settings)` so appendAliases and the generation paths agree when a vault raises minAliasLength — fine as a small follow-up.

PR #538, opened 2026-08-23
fix(ui): keep OpenRouter model variants in the catalog

### LINKED ISSUE #534: fix(ui): getModelFilter hides ~79 OpenRouter models behind the Ollama ':' rule — every :free/:batch variant invisible
## Summary

`getModelFilter` (`src/ui/settings-sections/model-section.ts:79`) drops every model id containing `:` when the provider is OpenRouter. The rule exists for Ollama, where `:` separates tag from name — but on OpenRouter `:` is a valid **variant suffix**, so the filter hides a large share of the live catalog from the dropdown.

Measured by @anavalo in #517 (adjacent finding 2) against the live `/api/v1/models` on 2026-08-21: **79 of 419 models invisible** — including every `:free` and `:batch` variant (e.g. `liquid/lfm-2.5-2.6b:free`, `google/gemini-3.7-flash:batch`).

## Why it matters

`:free` variants are the only zero-cost entry point for users evaluating the plugin against OpenRouter's catalog. They are currently reachable only by typing an id manually, which the UI gives no hint is possible or necessary. The filter also silently diverges the dropdown from what the same user sees on OpenRouter's own site.

## Expected

The Ollama tag-separator rule should not apply to providers where `:` carries meaning. Options:

1. Scope the `:` filter to providers that actually use tag syntax (Ollama / LM Studio local catalogs), remove it for OpenRouter.
2. Or drop the character-based heuristic entirely and rely on the per-provider catalog as returned.
3. Or keep a filter but make it visible/c configurable instead of silent.

Option 1 is the smallest change with the clearest rationale; option 2 is simpler if nothing else depends on hiding suffixed ids anywhere else.

## Notes

- Free-tier rate limits and model quality on `:free` variants are OpenRouter-side concerns and out of scope here — the ask is only that the catalog not be filtered invisibly.
- Related but separate: #533 (model-fault 404 misrouted into URL fallback).

pttydou:
Fixes #534.

`getModelFilter` treated `:` as invalid for OpenRouter even though the provider uses suffixes such as `:free` and `:batch`. This keeps every valid string ID from the OpenRouter catalog while preserving the existing Ollama and LM Studio filters.

Tests (pnpm 10.14.0):
- `pnpm lint`
- `pnpm typecheck`
- `pnpm build`
- `pnpm exec vitest run src/__tests__/ui/settings-codex-sections.test.ts` — 8 passed
- `pnpm exec vitest run --exclude src/__tests__/wiki/query-engine/custom-instructions-injection.test.ts` — 243 files / 3488 tests passed
- `pnpm css-lint` — 0 violations

The full Windows `pnpm gate:1` run reaches one existing default-branch failure in `custom-instructions-injection.test.ts`: `import.meta.url.pathname` is resolved as `D:\D:\...`. The failing file and its target are unchanged by this patch.

## Gate 4: Performance

| Dim | Status | Notes |
|---|---|---|
| CPU | ✅ | The catalog still uses one linear filter and the existing sort. |
| Memory | ✅ | Only catalog entries that should be displayed are retained. |
| IO | N/A | No IO changes. |
| Network | ✅ | No request, retry, or fallback changes. |
| Token | N/A | No LLM calls are involved. |


green-dalii (dismissed):
Thanks @pttydou — approved. Keeping every valid string ID for OpenRouter restores the :free/:batch variants while the Ollama and LM Studio rules stay put; the three-provider matrix plus the malformed-entry drop test pin the boundary well. Verified on the maintainer side that nothing downstream splits selected model IDs on ':', so the wider catalog is inert outside the picker. The full Gate 4 table in the description is appreciated.

green-dalii:
Thank you for the OpenRouter `:` variants fix (PR #538, Issue #534) — shipped in v1.27.0 MINOR (2026-08-27).

All ~79 models (including every `:free` variant) now show in the model catalog. Your verification that no consumers split on `:` was the key piece of evidence that made the change safe to ship.

Full release notes: https://github.com/green-dalii/obsidian-llm-wiki/releases/tag/1.27.0

PR #539, opened 2026-08-24
fix(llm): honour a pinned text mode on Codex, spare the retry on source-borne loops (PR #525 review)

DocTpoint:
Follow-ups from your review of #525 — the six items you filed for the tracker. Grouped into one PR because they are all consequences of that PR; happy to split if you prefer them separately.

**1. Codex ignored the pin (`openai-codex-sdk-client.ts:124`).** The client never read `outputModeOverride`, so the built-in `extract` → `text_prompt` pin was honoured everywhere except the one provider routed through it. It kept `Output.json()` — which is also the SDK's structured-output path, so the reply bypassed the plugin's own JSON repair on top of it. Now: format dropped when a mode is pinned, JSON instruction carried in the system prompt instead. `forcedTextPromptSystem` moved from `openai-compat-sdk-client.ts` to `json-prompt-prefix.ts` so both clients apply one rule rather than one of them holding a private copy. Two wire tests (pinned / unpinned).

**2. Source-borne loops (`source-analyzer.ts:437`).** `isSourceBorneLoop` checks the note before the retry is spent. One deviation from your suggestion, deliberate: it requires the note to repeat the unit **four times**, not to contain it once. Skipping the retry on a genuinely damaged batch loses items silently; spending it on an echo costs one call — the errors are not symmetric, so the test has to be the conservative one. `Vitamin D, ` occurs in half a vault and a containment check would suppress the retry for every real loop built from it. This also needed `RepetitionLoop.rawUnit`: `unit` is trimmed to 40 characters for log lines, so a lookup through it would silently never match a longer refrain — worth flagging since that trap is invisible in review.

**3. Exhaustion arm.** Pinned: budget spent → damaged batch parsed and merged rather than dropped, and the log says so. A change to `return null` there would lose a whole source without a test going red.

**4. Placeholder.** Text key in all 11 locales, parity 110/110.

**5. Duplicate entries.** `parseTaskPolicySpec` now throws on a task named twice — two entries state two arms, and picking one quietly is the manifest lie the parser refuses everywhere else.

**6. Unknown step names — the "or document" branch.** Left unvalidated on purpose: every label is a string literal at its own call site (`task: 'merge-triage'`), so a list in the parser is a second copy that drifts, and a stale list rejecting a policy for a step that *does* exist is worse than the typo it catches. An unknown name is inert. Documented where the parser explains itself. If you want the registry instead, that is a separate change — give the labels one home and this becomes a two-line check.

Gate 1 against `65ebdbd`: lint 0, typecheck 0, build clean, i18n parity 110/110, tests 3552 → 3565.


green-dalii (dismissed):
Approving after the dual-pass independent review (simplify 4-angle + code-review 5-angle with mutation-tested assertions): **0 blocking across all six items.** Gate 1 green (3565 tests, i18n parity 110/110).

Verified clean: the Codex wire tests drive the real AI SDK `streamText` pipeline against an injected fetch and assert on the serialized body; unpinned calls are byte-identical on the wire; streaming is structurally unable to bypass the pin; the new throw cannot trip any existing `data.json` load path; `forcedTextPromptSystem` has exactly one home with both clients consuming it.

Non-blocking follow-ups filed:
- #542 — `isSourceBorneLoop` blind spot (repetition-loop.ts:97 / source-analyzer.ts:441): common-word degeneration misread as echo suppresses the retry; carries the exact-4/exact-3 boundary pin, >40-char analyzer fixture, and prose-branch coverage gaps as acceptance criteria.
- #543 — task-policy.ts:180 `__proto__` key bypasses the duplicate guard; one-line `Object.create(null)` fix.

None block: shipping the six fixes outweighs the heuristic's documented conservative tradeoff.

green-dalii (approved):
Approving after the dual-pass independent review (simplify 4-angle + code-review 5-angle with mutation-tested assertions): **0 blocking across all six items.** Gate 1 green (3565 tests, i18n parity 110/110).

Verified clean: the Codex wire tests drive the real AI SDK `streamText` pipeline against an injected fetch and assert on the serialized body; unpinned calls are byte-identical on the wire; streaming is structurally unable to bypass the pin; the new throw cannot trip any existing `data.json` load path; `forcedTextPromptSystem` has exactly one home with both clients consuming it.

Non-blocking follow-ups filed:
- #542 — `isSourceBorneLoop` blind spot (repetition-loop.ts:97 / source-analyzer.ts:441): common-word degeneration misread as echo suppresses the retry; carries the exact-4/exact-3 boundary pin, >40-char analyzer fixture, and prose-branch coverage gaps as acceptance criteria.
- #543 — task-policy.ts:180 `__proto__` key bypasses the duplicate guard; one-line `Object.create(null)` fix.

None block: shipping the six fixes outweighs the heuristic's documented conservative tradeoff.

PR #540, opened 2026-08-24
feat(bedrock): AWS SSO / IAM-key auth for Bedrock via hand-rolled OIDC + SigV4 (#425)

### LINKED ISSUE #425: Bedrock Stage 2: SSO/Profile auth via hand-rolled OIDC + SigV4 → bedrock-mantle (zero AWS SDK)
## Summary

Add **AWS SSO / Profile login** to the existing `bedrock-anthropic` / `bedrock-openai` providers (Stage 1 = bearer `bedrock-mantle`, already shipped +3 KB) via a **zero-AWS-SDK** path.

**Decision 2026-08-07:** cancels the prior "≥3 user requests" gate (was in the v1.26.0 design). Implementation window = **v1.26.x PATCH or v1.27.0**.

## Mechanism (4th path — Codex-style hand-rolled)

1. **Login layer:** reuse the Codex OAuth skeleton (`src/llm-sdk/openai-codex/device-flow.ts` / `auth-manager.ts`) → IAM Identity Center OIDC device-authorization-grant (`RegisterClient` → `StartDeviceAuthorization` → browser login → `CreateToken`) → SSO token (8h) in SecretStorage.
2. **Credential exchange:** hand-written SigV4 call to `GetRoleCredentials` (`sso.<region>.amazonaws.com`, Service = `sso`) → temp IAM creds (AKID/SK/SessionToken, ~1h).
3. **Signer:** new `sigv4-signer.ts` (~300 LOC, `crypto.subtle`, canonical request + HMAC-SHA256 key derivation + `x-amz-date` + `x-amz-security-token`), validated with AWS official test vectors.
4. **Call:** temp creds SigV4-sign the existing `bedrock-mantle` endpoint (`bedrockMantleMessagesUrl` / `bedrockMantleChatCompletionsUrl`). No change to the openai-compat / anthropic client structure.
5. **Refresh:** temp creds expire in ~1h → auto re-run `GetRoleCredentials` inside the SSO token's 8h window.

**Cost:** ~500-800 LOC, **+~10 KB bundle, zero new npm deps**.
**Vs PR #263** (closed, rejected): that approach shipped `@ai-sdk/amazon-bedrock` + `@aws-sdk/credential-providers` at **+1.2 MB** (bearer users pay it too — esbuild single-file CJS can't lazy-load). The `bedrock-mantle` endpoint accepts AWS credentials (SigV4) and speaks standard OpenAI/Anthropic protocols over plain SSE — so no native ConverseStream `STREAMING-AWS4-HMAC-SHA256-PAYLOAD` event-stream signing is needed.

## Key facts (verified 2026-08-07)

- AWS docs: `bedrock-mantle` authentication = "Amazon Bedrock API key … **AWS credentials (supported for HTTP requests)**".
- Codex provider proves zero-SDK OAuth is viable in this plugin (pure fetch + SecretStorage, mobile-capable device-code).
- Mobile: technically possible (pure HTTPS + `crypto.subtle`) but SSO login needs a browser → **desktop first**.

## Implementation checklist

- [ ] Confirm IAM Identity Center OIDC endpoints (`/client/register`, `/device_authorization`, `/token`) + scopes
- [ ] Confirm `GetRoleCredentials` SigV4 service = `sso`, region = identity-store region
- [ ] `sigv4-signer.ts` + AWS test-vector unit tests
- [ ] SecretStorage schema for SSO token / temp creds + 1h auto-refresh scheduler
- [ ] Settings UI: SSO login button + `sso_region` / `account_id` / `role_name` inputs (desktop-only)
- [ ] Reuse-vs-generalize evaluation of the Codex auth-manager skeleton

## Design record

`~/.claude/projects/-Users-greener-project-obsidian-llm-wiki/memory/project_bedrock_stage2_codex_style_sigv4.md` + [[project_v1_26_x_patch_scope]] item 15. PR #263 author (@dmsessions) notified of the new decision ([comment 5218259440](https://github.com/green-dalii/obsidian-llm-wiki/pull/263#issuecomment-5218259440)).

green-dalii:
## Summary

Implements Issue #425 (Bedrock Stage 2) — the approved "4th path": **zero AWS SDK, zero new npm dependencies**. Adds two auth modes to the existing `bedrock-anthropic` / `bedrock-openai` providers alongside the unchanged Stage-1 API-key mode:

- **AWS SSO** — IAM Identity Center OIDC device flow (`RegisterClient` → `StartDeviceAuthorization` → browser approve → `CreateToken` polling), then `GetRoleCredentials` against the portal API using the `x-amz-sso_bearer_token` header (botocore scheme, no SigV4). Temp IAM credentials (~1 h) are cached in memory under the SSO token's ~8 h window.
- **IAM keys** — static access key + secret (+ optional session token), buffered in-memory and flushed to SecretStorage once per edit session.

Both modes sign every bedrock-mantle request with a **hand-written SigV4 signer** (~200 LOC, pure Web Crypto — no Node builtins, desktop and mobile webview capable), conformance-pinned byte-exact against the official AWS SigV4 test-suite vectors vendored in botocore (`get-vanilla`, `post-vanilla`, `post-x-www-form-urlencoded` incl. real-body payload hash).

Also fixes a Stage-1 prerequisite bug: the plugin-level factory omitted `bedrockRegion` from the sync-factory literal, so production Bedrock calls silently ran on us-east-1 regardless of the Settings dropdown.

## Design anchors

- Default `bedrockAuthMethod: 'api-key'` = Stage-1 wire shape byte-for-byte (Gate 3).
- Secrets live ONLY in Obsidian SecretStorage (`karpathywiki-bedrock-sso` / `-iam`) — never data.json/logs/Notices. Sign-out clears stores; plugin unload drops in-memory temp creds only (`dispose()` ≠ `signOut()`).
- Signing wrapper is injected on BOTH fetch seams (non-streaming + streaming); foreign hosts pass through byte-identical; SDK bearer headers are stripped before signing; unsigned requests never hit the wire when credential resolution fails.
- Empirically-unconfirmed values isolated as named constants (`BEDROCK_MANTLE_SIGNING_SERVICE = 'bedrock'`, portal host, content-sha toggle) with `[SIGV4]` console.debug — an E2E mismatch is a one-line fix.
- Test Connection gate: sso/iam modes require credential presence, never surface a missing-API-key error; `isProviderConfigured` accepts AWS-credential presence so the client initializes.
- Post-login prefill: discovery runs only when BOTH account/role fields are empty AND exactly one account+role is visible; ambiguity keeps manual entry authoritative.
- 22 new i18n keys ×11 locales, parity test green.

Closes #425

## Gate 4: Performance

| Dim | Status | Notes |
|-----|--------|-------|
| CPU | ✅ | ~10 crypto ops per signed request; negligible next to TLS+LLM latency |
| Memory | ✅ | bounded: one temp-cred cache entry per account+role+region in use |
| IO | ✅ | SecretStorage reads/writes only on auth actions |
| Network | ✅ | single-flight dedupes concurrent exchanges; device polls honor server interval + slow_down |
| Token | N/A | no prompts touched |

Bundle delta vs main: +62 KB raw minified (3,947,565 → 4,011,459). Larger than the ≤15 KB estimate because 22 i18n keys × 11 locales contribute ~20 KB of user-facing strings plus UI plumbing; still ~20× lighter than the rejected +1.2 MB AWS-SDK path.

## Test plan

3613 tests green (was ~3434 at v1.26.4); new suites: sigv4 vectors, signing-fetch wire shapes, OIDC polling state machine, role-credentials, credential-manager (skew/single-flight/dispose/discovery), credential-store round-trips, factory wiring, connection gate, provider-auth gate, auth-controls truth tables. TDD RED→GREEN throughout except the factory-wiring block (characterization tests added after implementation — noted honestly).


green-dalii:
"@dmsessions since you built the earlier Bedrock path in #263 and have the real AWS account(s), I'm hoping you can test this. We went with the zero-AWS-SDK fourth path you proposed — pure Web Crypto SigV4 (~200 LOC) + hand-rolled IAM Identity Center OIDC device flow + portal bearer-token auth for GetRoleCredentials. Conformance is pinned byte-exact against the official AWS SigV4 test suite (get-vanilla, post-vanilla, post-x-www-form-urlencoded with real-body payload hash). No new npm deps, +62KB raw (mostly i18n strings ×11 locales; ~20× lighter than the +1.2MB @aws-sdk approach).\n\nThree isolated constants intentionally left for real-account verification (each is a one-line flip if your credentials disagree):\n\n- `BEDROCK_MANTLE_SIGNING_SERVICE = 'bedrock'` — sign-name for the `bedrock-mantle.{region}.api.aws` endpoint; if SigV4 returns SignatureDoesNotMatch this is the first knob.\n- `bedrockPortalBaseUrl(region)` → `https://portal.sso.{region}.amazonaws.com` — modern host; the legacy `sso.` host still resolves if the modern one fails.\n- `BEDROCK_INCLUDE_CONTENT_SHA_HEADER = false` — toggle flips both the header AND its SignedHeaders entry atomically; flip true if the endpoint demands it.\n\nTo exercise: Settings → Provider Bedrock-Anthropic → Authentication AWS SSO → enter your Start URL → Sign in. Device-code prompt opens browser pre-filled; account/role auto-prefill runs only when both fields are empty AND exactly one of each is visible. Test Connection then runs an ingest/query through the signed path. Any of the three constants mis-named surfaces as an immediate failure with a `[SIGV4]` debug line. Curious how this lands versus your original design."

green-dalii:
@dmsessions following up the 2026-08-07 invitation you didn't get to take — we implemented #425 / #540 (Bedrock Stage 2) on the zero-SDK path your #263 bundle-size math implicitly pointed toward: no @aws-sdk, pure Web Crypto SigV4 (~200 LOC, byte-exact against the AWS conformance test suite), and we reused the Codex OAuth skeleton for the OIDC device flow exactly as the 2026-08-07 design doc proposed.

Three empirical choices are left isolated as one-line flips in src/llm-sdk/bedrock-sso/constants.ts since they need a real AWS account to verify:
- `BEDROCK_MANTLE_SIGNING_SERVICE = 'bedrock'` (constants.ts:52) — service sign-name for `bedrock-mantle.{region}.api.aws`; if SigV4 returns SignatureDoesNotMatch this is the knob.
- `bedrockPortalBaseUrl(region)` (constants.ts:36) — modern `portal.sso.{region}.amazonaws.com`; legacy `sso.` host still resolves as fallback.
- `BEDROCK_INCLUDE_CONTENT_SHA_HEADER = false` (constants.ts:52) — flips the header AND its SignedHeaders entry atomically.

If you have AWS federation at hand, exercising the AWS SSO path under Settings → Provider Bedrock-Anthropic → Authentication AWS SSO is the single highest-value test; account/role auto-prefill fires only when both fields are empty AND exactly one of each is visible. Curiously waiting to hear how the three constants compare to your real setup.

DocTpoint (changes requested):
**Review of b084e92** — I read all eight `bedrock-sso/` modules plus the wiring/UI diffs, independently recomputed the three conformance vectors (plus the form-body payload hash) with a separate SigV4 implementation — all byte-identical — and ran the 12 new suites locally (95 tests green).

The security posture holds up well: fail-closed on credential resolution, bearer stripping before signing, byte-identical foreign-host passthrough, both fetch seams wrapped, secrets confined to SecretStorage with rebuild-on-save, and deriving the signing region from the host is exactly right.

**One change requested — the SSO region is coupled to the inference region.** The device flow runs against `oidc.{bedrockRegion}` and `getCredentials()` calls the portal with `config.region` (the Bedrock dropdown). IAM Identity Center is homed in exactly one region per org — that's why AWS config separates `sso_region` from `region`. Two consequences: an org with IdC in us-east-1 can't use SSO to run inference in eu-central-1 at all, and switching the region dropdown after login turns into a 401 → a misleading "SSO expired, sign in again". The code already knows the answer: `discoverAccountRole()` correctly uses `token.region`, and the token store persists `region` for precisely this purpose. Minimal fix: key the `getRoleCredentials` exchange on `token.region` instead of `config.region` (one line, makes the session survive region switches). Full fix, fine as a follow-up: a dedicated SSO-region setting for the login itself, mirroring `sso_region`.

**Non-blocking:**
- `signOut()` clears both stores, but the button lives in the SSO section and the dialog only asks "Sign out?" — a user who configured both modes silently loses their IAM keys. Either clear only the SSO store (IAM has its own clear button) or say so in the dialog.
- Gate-4 "SecretStorage reads only on auth actions" isn't quite true in iam mode: `getCredentials('iam')` reads the store on every signed request. Cheap, but the claim is off by a line — or cache the keys.
- `canonicalUri = url.pathname` skips segment double-encoding — correct for today's static paths, but worth one warning sentence in the sigv4.ts header for whenever a path ever carries a `:`-style segment (non-S3 canonicalization double-encodes).
- `listAccounts` maps missing fields to `''` instead of dropping the entry; ends harmlessly in the discovery catch, but dropping would be cleaner.

The `raceWithBounds` unhandled-rejection containment and the RFC 8628 slow_down/deadline handling are careful work, and pinning the vectors with a "never fabricate expected signatures" note is the right discipline.

DocTpoint (approved):
Re-reviewed the delta b084e92..c6f901a against my CHANGES_REQUESTED review — approving on c6f901a.

- **SSO region (the blocking item): resolved with the fuller fix.** Homing `getRoleCredentials` on `token.region` and keying the temp-credential cache on it are both right — portal-issued role credentials are not scoped to the inference region, so sharing the cache across a region switch is correct, not merely permissible. The new test seeding an eu-central-1 token against a us-west-2 inference region pins exactly the failure I described.
- **signOut/IAM:** verified against the UI call sites — provider-section.ts keeps the two-button model (SSO sign-out :252, IAM clear :287), so the narrowed wipe matches the surface.
- **IAM mirror:** the read-count instrumentation is the right assertion, and returns are defensive copies. One optional nano-nit: `saveIamKeys` keeps the caller's object as the mirror — `{ ...keys }` would make it immune to later caller-side mutation, matching the copies on the way out.
- **canonicalUri note and the listAccounts filter:** as discussed; the junk-entry tolerance via optional chaining is a bonus beyond the ask.

Ran the llm-sdk suites locally at c6f901a: 473/473 green (the codex-loopback bundle test wants a prior `pnpm build` — environment, not this PR).

green-dalii (commented):
Follow-up on @DocTpoint's approval (c6f901a round 2): the review findings are addressed — SSO region now decoupled via token.region, signOut is SSO-only, IAM keys memoized, listAccounts drops invalid entries. The three empirically-pending constants (mantle signing service name, portal host, content-sha header) stay isolated as named constants so any real-AWS correction is a one-line flip during @dmsessions' E2E. Merging ahead of that E2E per release-train timing; the constants carry the risk boundary.

PR #544, opened 2026-08-25
fix(llm-sdk): recover the in-scope composite on NoOutputGeneratedError; pin thinking off for welcome-translate

### LINKED ISSUE #506: fix(llm-sdk): recover reasoning_content on NoOutputGeneratedError + force-disable thinking for translation
## What happens

When a business path uses the typed-output SDK surface (`createMessageWithOutput` + `response_format.schema`) and the active provider model emits **visible `content` of length 0** — typically a reasoning-capable model (DeepSeek V4, Qwen 3.7/3.8 reasoning variants, GLM-5, Kimi K3, Gemini 3.6 series, Grok 4, Step 3, Hunyuan Hy3, MiMo V2.5, Gemma 4 — any provider whose SDK surface uses a separate `reasoning_content` channel), the AI SDK 6.0.230 stream flushes without a terminal chunk and throws `NoOutputGeneratedError`. Our compat client catches the sibling class `NoObjectGeneratedError` and **recovers the JSON payload from `err.response.body.choices[0].message.reasoning_content`** before returning to the caller — but the `NoOutputGeneratedError` branch returns `{ text: '' }` directly and the recovery logic is never attempted.

The caller (e.g. `localize-welcome-note`, but the bug is **generic across all 6 P0 typed-output business paths** migrated in v1.26.3 PATCH Issue #443) then runs `parseJsonResponse` on a 0-byte string, the JSON parse falls through, the schema lookup returns `undefined`, and the caller falls back to its "no useful content" branch — which for Welcome means writing the English template instead of the user's translated markdown, and surfacing the error so the user sees a "Run Configuration Test" Notice on first-run onboarding.

## Why this is a generic reasoning-model bug, not a per-model one

Reasoning models share one property across providers: they partition their generation budget between a `reasoning_content` channel and a `content` channel. The model has a single `max_tokens` budget; once the reasoning channel spends it, the content channel is empty. AI SDK 6.0.230 promotes the reasoning channel into `result.reasoning` and the content channel into `result.text` — but only when both reach the success path. When the content channel is empty **and** the Output-object parser (which validates against our Zod schema) cannot extract a valid object from `reasoning_content` (because the model emits reasoning prose plus an embedded JSON literal that doesn't strictly match the schema's required fields), the SDK rejects the result with `NoOutputGeneratedError`. The `reasoning_content` is still on the wire body — we just don't read it on this code path.

This will reproduce on every reasoning-model backend we don't have an explicit "disable thinking" wire parameter for. The current 12+ provider list per `README.md §🤖 Models` (cross-checked against models.dev 2026-07) is:

- DeepSeek V4 series
- Alibaba Qwen 3.7/3.8 series (reasoning variants)
- Google Gemini 3.6 series
- xAI Grok 4 series
- Moonshot Kimi K3 series (2.8T MoE frontier)
- Zhipu GLM-5 series
- Step (阶跃星辰) Step 3 series
- Tencent Hunyuan Hy3 series
- Xiaomi MiMo V2.5 series
- Google Gemma 4 series (when loaded with reasoning enabled)
- Plus any future reasoning-capable provider we add (cloud or local: Ollama, LM Studio, OpenRouter)

## Reproduction

1. Configure a reasoning-capable model on the openai-compat SDK (e.g. DeepSeek V4 via `api.deepseek.com/v1`, or any other entry from the list above)
2. Trigger a typed-output business path — first-run Welcome onboarding on a non-English `wikiLanguage` is the simplest reproducible case
3. Observe `[NO-OUTPUT-GENERATED] (typed) ... returning empty quiet path` followed by `parseJsonResponse ... response length: 0` and `LLM response did not contain a valid \`translated\` field`
4. Welcome is written in English instead of `wikiLanguage`

Same root cause, same wire shape, same log breadcrumbs will surface for `source-analyzer.ts` extract / extract-retry (entities / concepts arrays will be empty → batch aborts), `dedup-phase.ts` (duplicates always empty), and `merge-triage.ts` (strategy undefined → defaults to insert) when those paths run against a reasoning model.

## Why the catch block is incomplete

`src/llm-sdk/openai-compat-sdk-client.ts` lines 956-985 handle `NoObjectGeneratedError` by reading `err.response.body.choices[0].message.reasoning_content`, prepending it to `err.text`, then returning the composite text so `parseJsonResponse`'s thinking-block fallback (Layer 3) can recover the embedded JSON literal. Lines 1062-1075 handle `NoOutputGeneratedError` by returning `{ text: '' }` immediately — the reasoning-channel recovery is skipped because the two error classes look superficially similar (both `extends AISDKError`) but are thrown at different points in the SDK's stream lifecycle (`NoObjectGeneratedError` after parse fails; `NoOutputGeneratedError` at flush when no terminal chunk arrived). The wire body shape is identical for both, so the recovery logic should run in both branches.

## Suggested fix

Two-line defense-in-depth per the [[feedback_force_disable_thinking_openai_compat_noop]] chain — fix the root cause AND add the missing recovery path:

1. **Reasoning-channel recovery** (matches the existing `NoObjectGeneratedError` branch at lines 956-985): in the `NoOutputGeneratedError` catch block, attempt to read `err.response.body.choices[0].message.reasoning_content` and prepend it to `text` before returning. If the composite text contains valid JSON via `parseJsonResponse` Layer 3, the call succeeds; otherwise the caller sees `text=''` and falls back exactly as it does today. Zero regression risk on the happy path.

2. **Per-call `enableThinking: false`** in the `localize-welcome-note.ts` `translateArgs` object (the only current Welcome-translation call site, per [[feedback_force_disable_thinking_openai_compat_noop]] — the per-call policy is the project's ironclad invariant; translation is a text task, no reasoning needed). Reference: the v1.26.0 Batch 6 PR #411 four-layer fallback. Add a wire-shape test asserting `reasoning_effort: 'none'` is on the body for the typed-output path, mirroring `openai-compat-request-body.test.ts`.

A third defensive item — extracting the two reasoning-recovery catch blocks into a single `recoverReasoningFromError(err)` helper — is already in [[feedback_llm_retry_extraction]] as a v1.27.0+ refactor and does not need to ship in the same PR.

## Acceptance criteria

- Reproduction step above no longer surfaces `localizeResult.error` for reasoning models
- Wire-body test asserts `reasoning_effort: 'none'` on the Welcome translation request, regardless of model family
- Regression test exercises the `NoOutputGeneratedError` catch branch with a stub `err.response.body.choices[0].message.reasoning_content` containing valid JSON; asserts the recovered text reaches `parseJsonResponse`
- Other 5 typed-output business paths (source-analyzer extract, extract-retry, dedup-phase, merge-triage, fix-dead-link, link-orphan, conversation-ingest save-dedup, schema-manager suggest-update, path-resolution resolve-dedup, fix-runners alias-generate, fix-runners tag-fix, lemma-classify) verified to NOT regress on the same reasoning model setup

## Why this is in MINOR not PATCH

- The bug is currently masked by the English-fallback in `localizeWelcomeNote` and the abort+retry in `source-analyzer.ts` extract — both produce correct end-user outcomes (English Welcome; round-2 retry on extract). No data loss, no security implication, no performance regression.
- The wire-shape test (item 2 above) is a 6-line addition; the reasoning-channel recovery (item 1) is a 30-LOC change + 1 test. Total ~50 LOC — well within MINOR scope.
- Co-shipping with v1.27.0 P0 batch amortizes the release-cycle overhead and lets the test contract stay consistent with the P0 reasoning-disable work.

green-dalii:
Closes #506.

## What this changes

**1. Typed-path recovery, corrected mechanism.** The issue sketched reading `err.response.body.choices[0].message.reasoning_content` in the `NoOutputGeneratedError` catch — that cannot work: the class never carries `.response` (constructor is `{message, cause}` only; throw sites at ai@6.0.230 `dist/index.mjs` :5146/:7232/:7933 attach none), so that read would be dead code and its test could only pass by fabricating the field. The class does have two origins on our paths, and the fix follows the split:

- **Origin (b): generateText resolved, `.output` getter threw.** `createMessageWithOutput` has already composed reasoning-channel + visible text into a local before `readOutput()` touches the getter — the old catch discarded it and returned `{ text: '' }`. The composite is now hoisted and returned, so caller-side `parseJsonResponse` Layer 3 recovers the embedded JSON.
- **Origin (a): generateText rejected outright** (stream flush without a terminal chunk — the issue's repro). No result in scope, wire unreachable without a fetch-level tee (disproportionate here) → empty quiet path stays correct. Both quiet-path debug lines gained an actionable hint naming the per-step policy escape hatch.

**2. Root cause unchanged from the issue:** `welcome-translate` pins `enableThinking: false` (translation is a text task; reasoning budget spent there caused the first-run onboarding symptom).

## Tests (+5, TDD red→green)

- typed origin-(b): result with throwing `output` getter + JSON-bearing reasoning → composite reaches caller (`text` contains payload)
- origin-(a) pinned: generateText rejection still returns the empty quiet shape, exactly one call
- args contract: `createMessageWithOutput` receives `enableThinking: false`
- full-pipeline wire test: `localizeWelcomeNote` driven through a real `OpenAICompatSdkClient` against captured fetch asserts `reasoning_effort: "none"` on the body
- existing NoOutputGeneratedError suite unchanged green

A correction comment documenting the `.response` finding lands on #506 alongside this PR.

## Gate 1

lint 0 · tsc 0 · build clean · **3569/3569 tests** · css-lint 0

## Gate 4: Performance

| Dim | Status | Notes |
|-----|--------|-------|
| CPU | N/A | Error branch only |
| Memory | ✅ | One extra string reference per call scope |
| IO | N/A | No new reads/writes |
| Network | ✅ | welcome-translate stops spending budget on reasoning |
| Token | ✅ | Same call count; translation no longer generates reasoning tokens |


DocTpoint (changes requested):
Review of 5ea4f2e.

The corrected mechanism is right, and thank you for the `.response` autopsy on #506 — I verified the `{message, cause}` constructor and all three throw sites against ai@6.0.230 locally; the issue's sketched read would indeed have been dead code, and the in-scope hoist is the only real recovery. The welcome-translate pin plus the wire test match what the issue asked for, and the origin-(b) test's throwing getter is faithful to :5146.

One item I'd like addressed before merge, because it sits exactly on the path the recovery is meant to help:

**The recovery branch masks the true finish reason (typically `length`) as `stop`.** In origin (b) the result is in scope, and the early `reportFinish` right after `generateText` resolves (~:901) has already given `onFinish` the true `finishReason`/`usage`; the catch then reports a second time with `('stop', undefined)` — last write wins in consumers like `source-analyzer.ts:399` (`finish.reason = meta.finishReason`) — and the returned object hardcodes `finishReason: 'stop'` as well. This masking is pre-existing on the empty quiet shape, but there the empty text forced the caller onto its error branch anyway. With the recovery returning substantive text it becomes load-bearing: the issue's own scenario (reasoning budget exhausted → `finish_reason: length`, truncated JSON in the reasoning channel) now returns labeled as a clean `stop`, so length-keyed remedies never fire and a budget-truncated composite is accepted as complete. Suggested fix, same shape as your text/usage hoist: hoist `result.finishReason` alongside them, return it, and either drop the recovery branch's second `reportFinish` (the early one already told the truth) or re-report the hoisted values — `recoveredUsage` should ride along either way.

Verified locally at 5ea4f2e: llm-sdk suites + localize-welcome-note 446/446 green; the new tests are mock-faithful (real `NoOutputGeneratedError` from a throwing getter, no fabricated `.response`; the wire test drives a real client).

green-dalii:
Fixed in c2b7769 — the recovery branch now hoists `result.finishReason` alongside `recoveredText`/`recoveredUsage`, returns it instead of the hardcoded `'stop'`, and no longer re-reports finish (the early `reportFinish` after `generateText` already delivered the true reason/usage; last-write-wins consumers keep seeing it).

Regression test added: a real `NoOutputGeneratedError` thrown from an `output` getter on the resolved result, with `finishReason: 'length'` — asserts the returned `finishReason === 'length'`, usage rides along, and `onFinish` fired exactly once with `'length'`. Verified RED on unfixed code (returned masked `'stop'`), green after.

DocTpoint (approved):
Verified on c2b7769: the recovery branch returns the hoisted `result.finishReason`, and the catch no longer re-reports — the early `reportFinish` after `generateText` remains the single delivery of the true reason/usage, so last-write-wins consumers see `length` when it was `length`. Ran the test file locally: 67/67 green on c2b7769; as a cross-check I placed the new regression test over the 5ea4f2e source, where it fails with `expected 'stop' to be 'length'` — so it bites.

One adjacent, pre-existing case outside this PR's scope: the empty-composite quiet path still re-reports `('stop', undefined)` after the early true report. With no text to mistake for a complete result the harm is minimal — fine to leave, mentioning it only for the record.

Approving on c2b7769.

green-dalii (commented):
Self-authored fix for Issue #506 (reasoning-channel recovery on `NoOutputGeneratedError`). Reviewed against the SDK boundary:

1. **Origin-(b) recovery preserved**: When `generateText` resolves but nothing parses into `output` (throwing `NoOutputGeneratedError` on accessor), the composite text built from the reasoning channel is returned for caller-side repair (`parseJsonResponse` Layer 3) instead of dropping into the empty quiet path.
2. **True finishReason survives**: Hoisted `recoveredFinishReason` preserves `result.finishReason` (typically `length`) through recovery; single `reportFinish` early delivery avoids last-write-wins masking.
3. **Thinking pinned off**: Welcome-translate explicitly sets `enableThinking: false` to avoid reasoning budget consumption during startup.

Gate 1 green locally (3570/3570) and rebased cleanly on current main (`ccc1ae2`).

Issue #197, opened 2026-06-23
Fix Dead Links fabricates AI-expanded stub pages for unresolvable links — reintroduces the blank-source hallucination #174 prevents


DocTpoint:
## Summary

When `fixDeadLink` can't resolve a dead link to an existing page, it doesn't leave it or stop — it **creates a new stub page and immediately AI-expands it** (`fillEmptyPage`). For a dead link that's an honest forward-reference (the target has no source note yet), this manufactures an unsourced, hallucinated page. On my vault, "Fix Dead Links" dropped the dead-link count ~250 → 10 — but ~40 of those "fixes" were brand-new fabricated stubs, not retargets.

## Root cause

`src/wiki/lint/fix-dead-link.ts`, two stub-creating paths after retargeting fails:

- LLM `create_stub` action (L128–162): create stub → `fillEmptyPage()`.
- **Deterministic fallback (L189–205): "No match — create a basic stub and expand it."** Unconditionally creates a stub and calls `fillEmptyPage()`.

The stub is stamped `sources: ["[[<referring page>]]"]` (L196) — the page that merely *mentioned* the term becomes its "source", though it isn't a source note about it. `fillEmptyPage` then generates frontmatter with no real source content.

## Why this is wrong

1. **It fabricates.** A dead forward-ref is an honest signal ("no page yet"). Replacing it with an AI-filled page invents content — e.g. the auto-created `entities/VCAM-1` got alias `"ICAM-2"`. VCAM-1 is not ICAM-2: an unsourced false-identity claim.
2. **False provenance.** Stamping the referring page as the stub's `source:` asserts a sourcing relationship that doesn't exist.
3. **It bypasses #174.** The pre-ingest `source-requirements` gate (#164/#174) exists precisely to stop blank/thin sources from producing hallucinated pages. This path creates a page with no real source and AI-fills it — reintroducing that exact failure mode through the lint instead of ingest.

The dead-link metric improves only because honest dead links are swapped for hallucinated stubs.

## Repro

1. Ingest a cluster that references not-yet-ingested entities (forward-refs).
2. Run Fix Dead Links.
3. New pages appear in entities/concepts: empty description ("No source content available"), AI-generated aliases/related links, `sources:` pointing at the referring page.

## Suggested fix

The retarget path (alias/title match) is the good part — keep it. For the unresolvable case, don't fabricate:

- **leave the dead link** (honest forward-ref), or
[...]

PR #545, opened 2026-08-25
feat(lint): leave-it outcome for Fix Dead Links — stub creation is optional

### LINKED ISSUE #485: Fix Dead Links has no "leave it" outcome: an unresolvable link always becomes an empty page
# Fix Dead Links has no "leave it" outcome: an unresolvable link always becomes an empty page

## Summary

When `fixDeadLink` cannot match a dead link to an existing page, it creates a stub page — in both branches, and without a way to decline. #197 established that such a stub must never be LLM-expanded, and that gate holds: `shouldFabricateStubForUnresolvableLink` returns `false` on both arms, and neither call site hands the stub to `fillEmptyPage`. What #197 left in place is the page itself. This issue is about that remaining half: the empty page is created even when the honest answer is that the vault does not know this name yet.

## What happens

`src/wiki/lint/fix-dead-link.ts` reaches a stub in two places:

- `:223` — the model answers `action: "create_stub"`. Two alias safety nets run first (`:121` pre-check, `:229` re-check), so this branch is already conservative about *whether* the target really is unknown.
- `:304` — the deterministic fallback. When the LLM path yields nothing and no title/slug match is found, a stub is created **unconditionally**.

The result is `buildStubContent` (`:70`): frontmatter with `generation_complete: false`, `sources:` pointing back at the referring page, and a default tag — `other` for entities, `term` for concepts. The referring link is then rewritten to point at the new page.

Tracing the terminal outcomes of `fixDeadLink`, there are three: the link is retargeted to an existing page, a stub is created, or nothing happens because no client is configured. There is no outcome in which a dead link is recognised as a legitimate forward reference and left untouched for the user to resolve later.

## Why the empty page costs more than it gives

The stub's own comment states its purpose — it "will be filled by next ingest of an actual source that defines this entity". That is true when such a source is actually ingested later, and inert otherwise. Meanwhile the page has effects that do not wait:

1. **It removes the signal it was created from.** A dead link is a visible piece of unfinished work and appears in the lint report. After the fix, the same fact is a page that renders normally and links correctly, and nothing in the vault view distinguishes it from a page with content.
2. **The tag is a guess with no source behind it.** `other` / `term` are defaults chosen without any content to classify, and on vaults using a custom tag vocabulary they are outside the configured set (the neighbouring problem in #459).
3. **It changes what measurements mean.** Page counts, orphan and hub statistics, and any coverage ratio computed over the wiki folder count pages that carry no knowledge.
4. **Since #482 stage 2, a stub is reachable vault-wide.** In the PR for that issue (#484) every page's title and curated aliases enter one resolution index, replacing the candidate window that generation prompts used to carry. A stub's title therefore now captures related-link names from anywhere in the vault, where before it could only be linked if it happened to appear in the window. The empty page can absorb references that would otherwise have surfaced as dead links and been visible as such.

Point 4 is why we raise this now rather than later: the two changes point in opposite directions, and the resolver makes the stub's reach larger than it was when #197 was decided.

## Proposed direction

`shouldFabricateStubForUnresolvableLink` was deliberately written as, in its own words, "a single, greppable switch so any future PR that wants to re-introduce auto-fabrication has to change this gate explicitly". Creation belongs at that same switch rather than in two separate call sites, so we would suggest routing both `:250` and `:304` through it and giving it a real decision to make.

For behaviour, a setting seems more appropriate than a removal, since existing vaults may rely on the current outcome:

- **`createStubsForUnresolvableLinks`** (or a similar name), controlling whether a stub is written.
- When off, the link is left as it is and reported. The lint report already lists dead links, so the information is not lost; it simply stays in the report instead of being converted into a page.

We do not have a view on which default the project wants, and would rather ask than assume. Our own preference for a large vault is off, for the reasons above; a new vault may well be better served by the current behaviour.

We are happy to open the PR either way once the shape and the default are decided.

## Environment

Observed on 1.26.3, vault of ~2,840 wiki pages.

green-dalii:
Closes #485.

## What this changes

An unresolvable dead link always became an empty placeholder page — in both branches (the model's `create_stub` answer and the deterministic fallback), with no way to decline. #197 established the stub must never be LLM-expanded but left the page itself in place; since #484's vault-wide title-resolution index, such a page also absorbs same-name references that would otherwise have surfaced as visible dead links.

**`createStubsForUnresolvableLinks` (default ON — every existing vault keeps its current outcome)** routes both creation sites through a single greppable sibling of the #197 gate:

- `shouldFabricateStubForUnresolvableLink` → "may an LLM FILL the stub?" (never — unchanged)
- `shouldCreateStubForUnresolvableLink` → "is the stub PAGE written at all?" (yes, unless opted out)

When off: zero writes, the link stays as it is and keeps surfacing in every lint report; ingest creates pages through normal channels and never needed the pre-existing stub. **Correction outranks the gate either way** — the deterministic pre-check and the post-model alias safety net keep retargeting regardless of the setting.

Also in this change (approved UX follow-up): the per-step `taskPolicies` control's description now enumerates common step labels across all 11 locales — previously a user had to know internal pipeline label strings to configure it at all, and a misspelled label silently matched nothing.

## Tests (+4, TDD red→green)

- gate off + LLM create_stub branch → no page written, `"left as dead link"`
- gate off + deterministic fallback branch → same
- setting unset behaves as ON (default-behaviour pin; stub page + rewritten link)
- post-LLM alias safety net still retargets when the gate is off (correction is not stub creation)

## Gate 1

lint 0 · tsc 0 · build clean · **3569/3569 tests** · css-lint 0

## Gate 4: Performance

| Dim | Status | Notes |
|-----|--------|-------|
| CPU | N/A | One boolean check per unresolvable link |
| Memory | N/A | No new retained state |
| IO | ✅ | Setting off skips two file writes per left link |
| Network | N/A | |
| Token | N/A | No prompt changes |


DocTpoint (approved):
Review of 74f5f1c — approving.

The shape is exactly what #485 hoped for: one greppable sibling of the #197 gate with a real decision to make, both creation sites routed through it, and correction outranking the gate in both branches — the fourth test (an alias that deliberately dodges the pre-check so only the post-model safety net can catch it) pins that property well. Default ON with `!== false` semantics is consistent across DEFAULT_SETTINGS, the use site, and the toggle, and the default-pin test keeps the missing-key case honest. I also checked the eight step labels the taskPolicies description now enumerates against the `task: '…'` literals in the code — all eight exist verbatim, so the description won't recreate the trap it fixes.

One non-blocking observation for the tracker: the leave-it outcome is counted as a fix. `runDeadLinkFixes` excludes only results containing `lintFixNoAction` ("No action taken (no client)") from its `fixed` counter (fix-runners.ts:242), so with the gate off every unresolvable link increments `fixed` on every run — and since the link intentionally stays, it will be "fixed" again next run. The results list is honest (it shows the outcome string); only the counter inflates. Cheap fix if you want it: give leave-its their own count (or exclude them from `fixed`) while keeping the line in the results list.

Verified locally at 74f5f1c: lint suites 259/259, i18n parity 118/118 green.

green-dalii (commented):
Follow-up on @DocTpoint's approval (74f5f1c): verified the leave-it outcome end-to-end — `shouldCreateStubForUnresolvableLink` gates both the LLM create_stub branch and the deterministic fallback, default stays ON so existing vaults see zero behavior change, and the wire tests cover stub/left-as-dead on both paths including the title-collision case. A1 taskPolicies description strings rode along ×11 locales with parity green. Merging.

Issue #496, opened 2026-08-19
Source pages never receive verbatim quotes — Mentions Format and Source Page Template are excluded from TASK_SECTIONS, and generateSummaryPage sees only the first 500 characters


rexplx:
Thanks for the plugin — it's doing real work on a large vault here (2,726 generated pages).

Reporting one defect with two independent causes. All findings are from reading the shipped main.js in v1.26.3 and measuring generated output, not inference.

## What I observed

`## Mentions in Source` — the section carrying verbatim quotes back to the source — appears on:

| Page type      | Pages | Carrying the section |
|----------------|-------|----------------------|
| wiki/concepts/ |   732 | 704 (96.2%)          |
| wiki/entities/ |   942 | 416 (44.2%)          |
| wiki/sources/  | 1,045 | 0 (0.0%)             |

The source page — the page naming the underlying document — is the only type that never carries a verbatim quote. It's also the page a reader most likely treats as representing the document.

## Cause 1 — TASK_SECTIONS filters the verbatim mandate out of every ingest call

wiki/schema/config.md is injected into the system prompt by SchemaManager.getSchemaContext(). Mine states: "Quotes are VERBATIM — never paraphrased, summarized, or translated away."

That instruction never reaches a model on the ingest path. TASK_SECTIONS admits only:

```
entity / concept  -> Entity/Concept Page Template, Naming Conventions, Classification Rules
summary (sources) -> Wiki Structure, Classification Rules
analyze           -> Wiki Structure, Classification Rules, Naming Conventions
```

Never admitted to analyze/entity/concept/summary/related/conversation/merge: Mentions Format, Source Page Template, Extraction Rules, Linking Rules, Aliases, Frontmatter Rules, Merge & Accumulation Rules. (`task === "full"` does return the whole file, but that path is only used by lint's empty-page repair — two call sites total.)

That list matters more than it first looks. Extraction Rules carries the entire source-grounding layer — "Every factual claim must be a direct paraphrase or verbatim quote of a cited source sentence", "No added knowledge", "No umbrella language" — and its own opening line says "These govern every page type and are not repeated in the templates below." The file explicitly depends on a delivery that never happens on the ingest path. Aliases carries "Every page MUST carry at least one alias", which the Purpose section calls load-bearing for retrieval.

[...]

Issue #244, opened 2026-07-05
Mention citations point to the raw source-note path instead of the canonical [[sources/<slug>]] page


DocTpoint:
## Summary

Every `## Mentions in Source` citation is stamped with the **raw source-note path** (e.g. `[[MyNote|MyNote]]`) instead of the canonical `[[sources/<slug>]]` page. The same file already documents the intended behavior — the #155 comment at `page-factory.ts:368`:

> `// Issue #155: entity/concept pages cite the canonical source PAGE`
> `// ([[sources/<slug>]]), not the raw note path`

— and `{{source_file}}` (`:371`) and the frontmatter stamp (`:423`) both use it. The mention citation is the one path that still passes the raw note path.

It's a single root cause with two effects — one deterministic, one that the weak-local-model path then amplifies. (Line numbers against `main` @ v1.23.2.)

## Root cause (one line, three call sites)

`page-factory.ts:361` (and the merge/update twins `:433`, `:495`) pass `sourceFile.path` to `truncateMentions`:

```ts
.replace('{{mentions}}', truncateMentions(info.mentions_in_source, 500, sourceFile.path))
```

`sourceSlug` is in scope and is used for exactly this purpose a few lines down (`:371`, `:423`). It just isn't threaded into the citation.

## Effect 1 — deterministic (the citation itself)

`truncateMentions` (`core/report.ts:43`) formats each quote (`:51-54`) as:

```ts
const leftPath = sourcePath.replace(/\.md$/, '');
const displayName = leftPath.split('/').pop() || leftPath;
const line = `- ${m} — [[${leftPath}|${displayName}]]`;
```

With the raw note path in, every citation becomes `[[<note-path>|<name>]]` rather than `[[sources/<slug>|<name>]]`. No LLM involved — pure assembly.

## Effect 2 — LLM-amplified (related links)

The leaked mentions block is injected into the page-generation prompt as `{{mentions}}` (`:361`). The corrector already documents this behavior — `related-link-corrector.ts:5-7`:

> `the model ... defaults to the most salient prefix in the prompt — sources/ (which appears in ... every mention citation)`

Because the mention citations now carry the note-folder prefix, that becomes the "most salient prefix," and the model copies it into the Related sections. `correctRelatedLinkPrefixes` is supposed to re-assert the type, but its match is scoped to `/\[\[(entities|concepts|sources)\/.../` (`:55`) — a note-folder prefix falls straight through, so those related links survive uncorrected.

[...]

PR #546, opened 2026-08-25
fix(schema,wiki): deliver orphaned schema sections, route captured quotes into source pages, strip append-time thinking (#491, #496)

### LINKED ISSUE #491: Five default-schema sections reach only the two lint-repair prompts, not ingest, generation, or merge
> **Corrected after filing.** The first version of this issue claimed no call site passes
> `'full'`. That was wrong — two do, and my check could not have found them: the grep pattern
> matched only single-line calls, and both are spread over four lines. The mechanism below is
> rewritten around what is actually there. See the correction comment for what changed.

## Mechanism

`SchemaManager.getSchemaContext(task)` does not deliver the schema body. It delivers the
sections that `TASK_SECTIONS[task]` names (`src/schema/schema-manager.ts:105-116`), via
`selectSections` (`:295-301`). The one value that returns the body unfiltered is `'full'`
(`:296`, the early return).

Exactly two call sites pass `'full'`:

- `src/wiki/contradictions.ts:186-190` — contradiction resolution
- `src/wiki/lint/fill-empty-page.ts:63-67` — filling an empty page

Both are lint-side repair paths. Every other caller passes a named task, and the union of the
headings those name is:

`Wiki Structure`, `Entity Page Template`, `Concept Page Template`, `Naming Conventions`,
`Classification Rules`, `Maintenance Policies`.

`buildDefaultSchemaBody()` writes eleven. The other five — `Source Page Template`,
`Date Fields`, `Mentions Format`, `Content Rules`, `Multi-Source Merge Rules` — are delivered
on those two repair paths and nowhere else. They never reach ingestion, page generation, or
merge.

## Why that is worth deciding

Two of them name the path they do not reach:

- **`Multi-Source Merge Rules`** is not in `TASK_SECTIONS.merge`. The merge task instead
  carries the two page templates and the naming conventions. A user who writes their merge
  policy into the section named for it is writing to the one task that will not see it.
- **`Source Page Template`** never reaches a call that writes a source page. Source pages are
  produced on the `analyze` path (`source-analyzer.ts:447,535,919,937`) and through
  `create-page.ts:236`, which passes `pageType`. Neither selects it.

`ensureSchemaExists()` writes all five into every user's `schema/config.md`. They read as
editable policy, they sit beside sections that are delivered on every path, and nothing marks
the difference. Editing them changes ingest behaviour not at all, and lint-repair behaviour
only.

That may be part of what #317 reports ("Schema.md changes are ignored"). The answer there is
about the type-system wall — closed enums, no `risks/` folder, no per-type handler — and it
holds for what that user asked for. But its opening, "Schema content *is* injected into the LLM
context … but only as advisory text", is true per task rather than globally: for these five
sections, ingestion never sees them at all. A user testing their edits by running an ingest
would observe exactly the reported symptom, for a second and much simpler reason.

## Two readings, and the code cannot tell them apart

1. **Oversight** — the sections were meant to be delivered and the task lists were never
   extended. The fix is to name them where they belong: `Multi-Source Merge Rules` on `merge`,
   `Source Page Template` on the source-writing tasks, and so on.
2. **Deliberate** — they are documentation for the human reader rather than instruction for the
   model, in the spirit of the user-knowledge / engine-internals split in #328. Then the file
   should say so, because today nothing marks them.

What should not stay is the present state, where the file promises and the filter silently
declines. Happy to open a PR for whichever reading is intended — (1) is a change to
`TASK_SECTIONS`, (2) is a change to `buildDefaultSchemaBody()`.

I have not measured whether delivering them improves output. That is a separate question, and
it can only be asked once they arrive at all.

green-dalii:
Closes #491. Refs #496.

Four commits on top of main.

## 1. Deliver the five orphaned schema sections to the tasks that own them (#491)

`schema/config.md` writes eleven sections and promises "edit it freely", but the per-task delivery whitelist named only six — Source Page Template, Date Fields, Mentions Format, Content Rules, Multi-Source Merge Rules reached just the two `full` callers and never ingest, generation, or merge. A user writing merge policy into the section named Multi-Source Merge Rules was editing the one task that would not see it. Ruled an oversight after a full call-site audit; the map now delivers Content Rules to analyze/entity/concept, Mentions Format to entity/concept/summary, Source Page Template to summary (audited as the task whose call writes `sources/<slug>.md`), and Multi-Source Merge Rules + Date Fields to merge. Measured prompt growth +73..336 tokens/call (<10%). Unchanged tasks pinned by a snapshot matrix.

## 2. Source pages now carry verbatim quotes (#496)

A measured vault showed 0% of 1,045 `sources/` pages with the Mentions section that ~96% of concept pages carry. Two deeper causes found beyond the issue's window:

- `buildSourceAnalysis` **never populated** the top-level `mentions_in_source` field — quotes existed only per extracted item. It now aggregates provenance-or-legacy quotes into a deduplicated pool.
- `createSummaryPage` injects that pool via the #244 programmatic route; the summary model still sees only 500 chars of source, so it cannot fabricate beyond them; with nothing captured, any Mentions section the model wrote is stripped. `[MENTIONS-CAPTURE]` debug line per ingest makes field-skip visible without making it required (weak-model hard-fail risk).

## 3. Verification round (all #496 claims checked against code)

- **Quote-grounding scanner finding — TRUE and load-bearing**: the lint phase grounded quotes only against generated `sources/` pages, so raw-note citations were checked against nothing → flagged ungrounded; after commit 2's raw-note citations, every fresh capture would have been flagged on re-lint. The phase now reads each cited primary note once and grounds against it; also passes the localized Mentions label (non-English wikis were silently no-oping the whole check).
- **Formatter 500-char cap — TRUE**: would ellipsize the very quotes this feature preserves; source-page budget raised to 2000 via a `maxChars` pass-through (mutation-verified), entity paths unchanged.
- **Granularity description nit — TRUE**: "caps items per source, not detail per item" now stated in all 11 locales.
- **"Extraction Rules" section names** the reporter quoted never existed in this repo's history — from an older/customized vault schema; no migration needed.

## 4. E2E bug: complementary appends leaked visible reasoning into wiki bodies

Found during this branch's E2E: the per-section append was the ONE raw-prose LLM consumer without thinking-block stripping, so a reasoning model's full `<think>` planning prose was spliced into the target section on page updates — and a think-wrapped `NO_NEW_CONTENT` missed its short-circuit. `stripThinkingBlocks` is now the single home in core/markdown (shared with `cleanMarkdownResponse`), applied at the fragment, and the append call pins `enableThinking: false` unconditionally.

## Tests & Gate

+19 tests across the four commits (delivery matrix 7, quote aggregation 4, summary-page injection 5, grounding-vs-primary-notes 2, think-strip 3, mutation-verified). Gate 1: lint 0 · tsc 0 · build clean · **3590/3590 tests** · css-lint 0. Full consumer audit documented in commit 4 (all other writers covered by JSON parsers or cleanMarkdownResponse).

## Gate 4: Performance

| Dim | Status | Notes |
|-----|--------|-------|
| CPU | ✅ | O(n) aggregation; one regex pass for grounding targets |
| Memory | ✅ | Cited-note contents transient per lint run |
| IO | ⚠️ | One read per uniquely-cited primary note per lint run — bounded by citations, not vault size |
| Network | ✅ | Append calls stop generating reasoning tokens |
| Token | ⚠️ documented | Prompt growth +73..336/call for delivery correctness; quotes programmatic (flat) |


DocTpoint (changes requested):
Reviewed all four commits on 4cb71df; ran the suite locally (3590/3590, tsc clean) and re-verified the review-resistant parts by hand: the five delivered headings byte-match `buildDefaultSchemaBody()`, your summary-task audit is correct against `createSummaryPage` (my issue pointed at the analyze path — yours is the better finding), the item-level aggregation loses nothing (the LLM schema has no top-level mentions field), and collector/scanner agree on line shape and target normalization in all three citation forms. The think-strip ordering (strip before trim) makes the wrapped `NO_NEW_CONTENT` short-circuit — verified in the tests.

One blocker, one line: `programmatic.ts` reads cited notes via `ctx.app.vault.read({ path: target })`. At runtime `phaseCtx.app` is the real `App` (controller.ts), and the declared API is `read(file: TFile)` — the object literal works only if `Vault.read` happens to duck-type, and if any Obsidian version stops tolerating it, the per-note `catch` swallows the break: the feature silently degrades into the exact mis-flagged "ungrounded" noise it exists to remove, and no test can see it (the mocks read by `{path}`). You already hold the `TFile` from the `getAbstractFileByPath` call on the previous line — pass that instead (the `instanceof TFile` pattern your own fix-runners.ts:519 documents as the correct API).

Non-blocking, for the record: a truncation-unclosed `<think>` block (no closing tag) passes `stripThinkingBlocks` untouched — on backends that honor the `enableThinking` pin this can't occur, but on those that ignore it, a `length`-truncated reasoning stream still splices; a trailing-unclosed strip after the paired one would close that. The "shared" mention-line regex exists twice (module constant + the scanner's inline literal) — byte-identical today, two copies to drift tomorrow. Translation-bearing bullets (`- "q" (t) — [[x]]`) match neither regex, so they escape both collection and grounding — consistent on both sides and pre-existing, just naming it. The injector's docblock still describes the old "empty header-only section" behavior its implementation no longer has — stale doc, also pre-existing.

DocTpoint:
Looked at 492f359 (it crossed my review by minutes): the `.md` normalization fix is real — my review wrongly cleared the old `hasMd ? '' : '.md'` as consistent with the scanner's candidates; it dropped the extension exactly in the with-`.md` citation form. The shared-regex and leave-it-counter fixes both check out, and the 3→5 alignment matches the enforced `maxTotalItems: 5` in batch-limits. The one open item from my review is unchanged: `vault.read({ path: target })` in programmatic.ts — the `TFile` from the previous line is the object to pass. Suite green on 492f359 locally (3590/3590).

DocTpoint (dismissed):
The blocker is fixed on fe7beb4: the abstract file from `getAbstractFileByPath` now flows into `vault.read` itself, honoring the declared TFile contract instead of relying on duck-typing — and the per-note catch now covers only the honest failures (a folder at a `.md` path reads like an unreadable note: flagged, not silently cleared). Verified the force-push scope: 4cb71df remains the ancestor and the amendment's only delta against 492f359 is this fix. Suite 3590/3590 and tsc clean locally on fe7beb4. Approving on fe7beb4.

green-dalii (commented):
Rebased onto current main (`bdb54de`) to clear the CHANGELOG conflict; force-pushed as `ec16844`. Code content identical to what was reviewed and approved on `fe7beb4` — the rebase resolution touched CHANGELOG.md only (union of both sides' Unreleased entries: #496 + think-leak + #491 kept alongside main's #506/#469/#485/#514), and the already-merged #485 commit (`74f5f1c`, landed via #545) was correctly skipped by git, which is why `advanced-settings-section.ts` no longer appears in this PR's delta.

Follow-up commit `fe7beb4` (carried through the rebase) resolves the review blocker — `runProgrammaticPhase` now holds the `TFile` from `getAbstractFileByPath` and passes it to `vault.read` (no duck-typed `{ path }` object, non-TFile returns skipped before IO) — plus the six internal-review findings: minimal-granularity prompt/copy aligned to 5 across all 11 locales, `stripThinkingBlocks` delegates to `extractThinkingBlocks().visibleContent` (covers the bare Thinking/Response form), `.md` citation normalization always appends the extension, `scanQuoteGrounding` consumes the module-level `MENTION_LINE_REGEX`, cited-note pre-read skips targets already in `pageMap`, and the leave-it outcome no longer counts toward the fixed tally.

Gate 1 green locally on `ec16844`: lint 0/0, tsc 0/0, build clean, 3673/3673 tests, css-lint 0. CI Gate 1 green on the pushed head.

PR #547, opened 2026-08-25
feat(llm): carry a task label through the stream path — Query Wiki accounts under its own step

### LINKED ISSUE #469: Extend LLMClient.createMessageStream interface to carry task label

green-dalii:
Closes #469.

## What this changes

`createMessageStream` predated per-task LLM accounting, so every streamed Query answer recorded under the single merged `untagged` row — a known step indistinguishable from genuinely unlabelled calls, and any future stream call site was type-forbidden from passing a label.

- `createMessageStream` interface now carries `task?: string` (same contract as `createMessage`)
- the wrapper's stream block drops its hardcoded `undefined` in favour of the caller's label
- the sole stream caller passes `task: 'query-wiki'`, distinct from its separate `query-view-evaluate` save-judgement call
- SDK clients forward the field for interface conformance (accounting stays in the wrapper, not the client — unlabelled stream calls still record as `untagged`, no fabricated suffix)

## A deliberate deviation from the issue's step 5

The issue suggested "clean up chatTemperature / enableThinking forwarding in QueryView post-extraction". Audited: those forwards are **load-bearing**, not redundant — QueryView passes `chatTemperature` itself as `temperature` (the wrapper only injects `extractionTemperature`), and `enableThinking` is a per-call concern the wrapper does not synthesize. Removing them would re-open the #451 silent-drop class. Kept, with the comment already noting why.

## Tests (+3, TDD red→green)

- labelled stream call records under `query-wiki`, not `untagged`
- unlabelled stream call records under `untagged` (no fabricated suffix)
- caller-provided temperature survives the stream wrapper (no silent drop)

## Gate 1

lint 0 · tsc 0 · build clean · **3568/3568 tests** · css-lint 0

## Gate 4: Performance

| Dim | Status | Notes |
|-----|--------|-------|
| CPU | N/A | Label pass-through |
| Memory | N/A | |
| IO | N/A | |
| Network | N/A | No wire change |
| Token | N/A | No prompt/response change |

Preview: also open in the PR body — the only behavior change is the accounting label; the wire is byte-identical.


green-dalii (commented):
Self-authored fix for Issue #469 (stream-path task accounting). Reviewed against the wrapper seam: `createMessageStream` now flows through the same `withTaskAccounting` + advanced-settings injection as `createMessage`, so QueryView's chat stream lands under its own `query-wiki` label instead of inflating 'untagged'. Caller-wins contract preserved on the stream path (QueryView's chatTemperature forwarding is load-bearing, regression-tested). Interface parity added on all three SDK clients (openai-compat / anthropic / openai). Gate 1 green 3568/3568 locally; CI re-run green after clearing an unrelated historical timing flake (fixed separately in #549).

green-dalii (commented):
Self-authored fix for Issue #469 (stream-path task accounting). Reviewed against the wrapper seam: `createMessageStream` now flows through the same `withTaskAccounting` + advanced-settings injection as `createMessage`, so QueryView's chat stream lands under its own `query-wiki` label instead of inflating 'untagged'. Caller-wins contract preserved on the stream path (QueryView's chatTemperature forwarding is load-bearing, regression-tested). Interface parity added on all three SDK clients (openai-compat / anthropic / openai). Gate 1 green 3568/3568 locally; CI re-run green after clearing an unrelated historical timing flake (fixed separately in #549).

PR #549, opened 2026-08-26
test(ci): tolerate libuv timer granularity in url-fallback delay assertion

green-dalii:
Single test-only change: the wall-clock assertion in `url-fallback.test.ts` required `elapsed >= RETRY_DELAY_MS * 2` exactly (600ms), and one CI run measured **599ms** — libuv caches loop time (~1ms granularity), so on a loaded shared runner `setTimeout(300)` can resolve marginally early against a freshly captured `Date.now()`.

The 30ms tolerance keeps the assertion's intent intact: if either delay were removed entirely, elapsed drops to single-digit milliseconds and still fails by two orders of magnitude (mutation-verified during development: production delay removed → test fails measuring 3ms).

First observed on PR #547's Gate 1 run; the same test code had passed ~40 prior CI runs unchanged — intermittent by nature, not tied to any branch's changes.

### DIFF
--- a/src/__tests__/core/url-fallback.test.ts
+++ b/src/__tests__/core/url-fallback.test.ts
@@ -230,8 +230,13 @@ describe('resolveBaseUrlWithFallback', () => {
       testFn,
     });
     const elapsed = Date.now() - start;
-    // 2 candidates tried after original fails: at least 2 × RETRY_DELAY_MS
-    expect(elapsed).toBeGreaterThanOrEqual(RETRY_DELAY_MS * 2);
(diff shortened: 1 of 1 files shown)


green-dalii (commented):
Self-authored test-only fix, reviewed against the CI log: the assertion now tolerates libuv's ~1ms timer granularity (30ms budget) while still failing by two orders of magnitude if either delay is removed — mutation-verified at 3ms during development. Gate 1 green 3643/3643 locally and on CI.

PR #552, opened 2026-08-26
fix(llm-sdk): key the token-key and reasoning-strip probe caches per (baseURL, model)

### LINKED ISSUE #551: TokenKeyProber and ReasoningStripProber cache per baseURL, but the model arrives per call — one model's probe verdict binds every sibling on the gateway
One long-lived client serves every task, and the model is chosen per call (`resolveModelForTask`), so several models share one baseURL and one set of probe caches. Of the three caches, `OutputModeProber` keys on `(baseURL, model)` — its own header explains why per-model granularity is required (a Qwen demotion must not demote gemma on the same LM Studio). `TokenKeyProber` and `ReasoningStripProber` key on baseURL alone, each with the stated rationale "same gateway typically uses the same wire format". This issue is about the word *typically*: the wire format is a property of the backend behind the model, not of the gateway URL — any proxy that routes per model (LiteLLM, one-api) and any multi-model LM Studio can host both answers at once.

Two mechanisms make the shared key bite, both verified in the shipped code:

1. **The verdict is inherited unprobed.** `transformRequestBody` reads `getCachedKey(this.baseURL)` per request — after model A's rejection populates the cache, every sibling request is rewritten to A's key without ever being probed.
2. **The repair path is disabled for the whole gateway.** All three 400-retry branches are gated on `!this.tokenKeyProber.getCachedKey(this.baseURL)`. Once the cache holds A's verdict, a token-key 400 from model B falls through to `throw` — the retry that exists precisely for this error can never run again on that baseURL.

The `ReasoningStripProber` analog is quieter: `markStrip(baseURL)` after one model's reasoning-field 400 silently drops `reasoning_effort`/`enableThinking` pass-through for siblings that support them — a capability loss with no error to see.

Today this hits nobody running one model per baseURL, which is why it has sat unnoticed; it becomes a precondition the moment two models share a gateway — the exact setup the `OutputModeProber` header already treats as the norm. Happy to open a PR adopting the same composite key for both probers (mirroring the `OutputModeProber` design, `invalidate()` clearing the baseURL's entries), test included: ~12 call sites in two files plus threading `model` into `buildProviderOptions`.

DocTpoint:
Closes #551.

One commit. One long-lived client serves every task and the model arrives per call, so several models share one baseURL and one set of probe caches. `OutputModeProber` already keys per `(baseURL, model)` — its header documents why. `TokenKeyProber` and `ReasoningStripProber` keyed on baseURL alone, with the three consequences from the issue: the first probed model's token key rewrote every sibling's requests unprobed; the 400-retry branches are gated on cache emptiness, so a sibling's token-key 400 was thrown with the repair path permanently disabled for the whole gateway; and `markStrip(baseURL)` silently dropped the `reasoning_effort`/`enableThinking` pass-through for siblings whose backend supports them.

## What changed

- Both probers adopt the `OutputModeProber` composite key (`baseURL::model`). Single-model setups produce identical behavior before and after — the key differs only when a second model actually shares the gateway.
- `TokenKeyProber.invalidate(baseUrl)` clears every model probed behind the baseURL; the `::` join keeps `https://a` from matching `https://a.example.com`.
- `ReasoningStripProber` keeps its no-invalidate design (removed in the PR #411 simplify round — not reintroduced).
- `buildProviderOptions` in both clients takes the model to key the strip check; all 20 call sites thread it. `transformRequestBody` keys on `(baseURL, modelId)`.

## Tests

+4 unit (sibling isolation for both probers, per-baseURL invalidate, prefix separation) and +1 client-level regression: model B still gets its own probe after model A populated the cache — the exact throw path from the issue. All verified RED against the per-baseURL sources (7 failures), green after.

## Gate 1

lint 0 · tsc 0 · build clean · 3656/3656 tests · css-lint 0

## Gate 2: Side effects

- Call-site audit: every `shouldStrip`/`markStrip`/`getCachedKey`/`setCachedKey` caller threads a model already in scope (tsc-enforced — the new required parameter enumerated them). No consumer outside the two clients.
- Cache capacity grows from O(gateways) to O(gateways × models actually used) — a handful of in-memory strings per client instance.

## Gate 3: Breaking changes

None detected. The caches are in-memory per client instance; no settings, schema, or ID changes.

## Gate 4: Performance

| Dim | Status | Notes |
|-----|--------|-------|
| CPU | N/A | Same lookups, one string concat per read |
| Memory | ✅ | One entry per (gateway, model) actually probed |
| IO | N/A | |
| Network | ✅ | Siblings stop inheriting wrong verdicts — fewer doomed requests |
| Token | N/A | |

### DIFF
--- a/src/__tests__/llm-sdk/reasoning-strip-probe.test.ts
+++ b/src/__tests__/llm-sdk/reasoning-strip-probe.test.ts
@@ -13,18 +13,29 @@ import { OutputModeProber } from '../../llm-sdk/output-mode-prober';
 // Batch 6: if the cache or the classifier ever drift, the tests fail.
 
 describe('ReasoningStripProber', () => {
-  it('starts empty for any baseURL', () => {
+  const MODEL = 'qwen3-30b';
+  const SIBLING = 'gemma-4-26b';
+
+  it('starts empty for any (baseURL, model)', () => {
     const prober = new ReasoningStripProber();
-    expect(prober.shouldStrip('https://api.deepseek.com/v1')).toBe(false);
-    expect(prober.shouldStrip('https://api.example.com/v1')).toBe(false);
+    expect(prober.shouldStrip('https://api.deepseek.com/v1', MODEL)).toBe(false);
+    expect(prober.shouldStrip('https://api.example.com/v1', MODEL)).toBe(false);
   });
 
-  it('markStrip + shouldStrip round-trip per baseURL', () => {
+  it('markStrip + shouldStrip round-trip per (baseURL, model)', () => {
     const prober = new ReasoningStripProber();
-    prober.markStrip('https://api.deepseek.com/v1');
-    expect(prober.shouldStrip('https://api.deepseek.com/v1')).toBe(true);
+    prober.markStrip('https://api.deepseek.com/v1', MODEL);
+    expect(prober.shouldStrip('https://api.deepseek.com/v1', MODEL)).toBe(true);
     // Different baseURL unaffected
-    expect(prober.shouldStrip('https://api.openai.com/v1')).toBe(false);
+    expect(prober.shouldStrip('https://api.openai.com/v1', MODEL)).toBe(false);
+  });
+
+  it('does not strip for a sibling model on the same gateway (#551)', () => {
+    // One backend behind the gateway rejects reasoning_effort, the
+    // sibling supports it — the sibling must keep the pass-through.
+    const prober = new ReasoningStripProber();
+    prober.markStrip('https://api.deepseek.com/v1', MODEL);
+    expect(prober.shouldStrip('https://api.deepseek.com/v1', SIBLING)).toBe(false);
   });
 
   // v1.26.0 Batch 6 review (PR #411 simplify 2026-08-05): the previous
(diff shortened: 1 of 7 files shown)


green-dalii (approved):
Approving — landing in the v1.27.0 MINOR line.

Dual review pass (4 simplify angles + 2 module-slice correctness reviews) verified the mechanics independently: all 20 `buildProviderOptions` call sites thread the request's own `model`; all four reasoning-strip retry blocks use the same binding across gate → retry → markStrip; no consumers outside the two clients (anthropic/codex/bedrock have zero references); composite-key collision is practically unconstructible and OutputModeProber has shipped this exact scheme since v1.26.3; single-model wire output stays byte-identical. The client-level regression test is genuinely discriminative — under per-baseURL caching model-b's retry guard is already false and the awaited call rejects.

Three non-blocking findings for the record (follow-up being filed):

1. `src/llm-sdk/openai-compat-sdk-client.ts:258` — `probeBaseURL` probes via `getProvider('gpt-4o-mini', …)`, so its `(baseURL, 'gpt-4o-mini')` cache slot is never populated and URL-fallback probes no longer inherit a cached `max_completion_tokens` swap (they did under per-baseURL keying). Narrow triple-condition window (token-key verdict cached → later 404-class URL fallback → gateway rejects `max_tokens`), and the failure mode is visible (probe rethrows, fallback aborts), not silent.
2. `src/llm-sdk/reasoning-strip-probe.ts:47-49` — the untouched header block still argues "Why per-baseURL not per-model", contradicting this PR's own rewritten class JSDoc; same drift class at `reasoning-strip-probe.ts:12`, `token-key-probe.ts:13-14`, the `[REASONING-STRIP-DEBUG]` log wording (`openai-compat-sdk-client.ts:601`), and ~8 adjacent mechanism comments still phrased per-baseURL.
3. `src/llm-sdk/token-key-probe.ts` — `invalidate()` has zero production callers (test-only); the prefix scan grows a dead API where the sibling precedent (PR #411 simplify round) deleted it.

PR #553, opened 2026-08-26
docs(llm-sdk): clear stale per-baseURL wording after Issue #551 composite-key fix

### LINKED ISSUE #551: TokenKeyProber and ReasoningStripProber cache per baseURL, but the model arrives per call — one model's probe verdict binds every sibling on the gateway

green-dalii:
Follow-up to #552 (closes #551). Review-only polish: no behavioural wire change on single-model users.

The #552 composite-key fix moved the two probe caches from per-baseURL to per-(baseURL, model) and rewrote the class-JSDocs, but ~10 adjacent prose blocks still said per-baseURL — a header that now argues the exact opposite of the merged change. This sweep clears them:

- `reasoning-strip-probe.ts` — the untouched "Why per-baseURL not per-model" header block now argues per-(baseURL, model) with Issue #551 rationale.
- `token-key-probe.ts` — step-3 header phrasing + `invalidate()` JSDoc ("Called when user changes baseURL/API key" → test helper; production re-probes by client reconstruction).
- `openai-compat-sdk-client.ts` — transformRequestBody comment, Guard phrases, [REASONING-STRIP-DEBUG] log wording, stream-ordering notes; note inside `getProvider(transformRequestBody)` for probeBaseURL's synthetic `gpt-4o-mini` (that slot is never populated — safe, it's a connectivity check).
- `openai-sdk-client.ts` — Layer-3 cache description + gate guards.

The narrow triple-condition regression from #551 review (cached swap + URL fallback + gateway rejects max_tokens) is documented as a visible abort; threading the request model into `probeBaseURL` is not justified this cycle.

## Gate 1
lint 0 · tsc 0 · build clean · 3677/3677 tests · css-lint 0

## Gate 2: Side effects
Comment/log-only — no code path, no call-site, no error-propagation change. No behavioural change for single-model or multi-model users (wire bytes identical).

## Gate 3: Breaking changes
None.

## Gate 4: Performance

| Dim | Status | Notes |
|-----|--------|-------|
| CPU | N/A | Comment-only |
| Memory | N/A |  |
| IO | N/A |  |
| Network | N/A |  |
| Token | N/A |  |

### DIFF
--- a/src/llm-sdk/openai-compat-sdk-client.ts
+++ b/src/llm-sdk/openai-compat-sdk-client.ts
@@ -103,7 +103,7 @@ export class OpenAICompatSdkClient implements LLMClient {
    * backends (Gemini-via-shim per #137) reject `reasoning_effort: 'none'`
    * with HTTP 400. On 400 with a reasoning-related field name in the
    * error message, we strip the field and retry exactly once; the
-   * strip decision is cached per baseURL so subsequent calls skip
+   * strip decision is cached per (baseURL, model) so subsequent calls skip
    * the probe. Mirrors the [[TokenKeyProber]] design.
    */
   private readonly reasoningStripProber = new ReasoningStripProber();
@@ -226,13 +226,17 @@ export class OpenAICompatSdkClient implements LLMClient {
       // cached on the instance in the constructor.
       supportsStructuredOutputs: this.supportsStructuredOutputs,
       // v1.23.0 P1.5 follow-up: token-key probe hook. Read the
-      // current cached key for this baseURL at request time — this
+      // current cached key for this (baseURL, model) at request time — this
       // closure captures `this` so each request consults the latest
       // probe result. If we have no cached entry, the transform is
       // a no-op and the request goes out with the AI-SDK default
       // (max_tokens). On rejection we probe + cache + next request
       // uses the swapped key.
       transformRequestBody: (args: Record<string, unknown>) => {
+        // Issue #551 follow-up: keyed per (baseURL, modelId). The URL-fallback
+        // probe uses synthetic 'gpt-4o-mini' (probeBaseURL) — its slot is
+        // never populated; no swap applied there (safe — the probe is a
+        // separate connectivity check, not a request that needs the token-key fix).
         const cached = this.tokenKeyProber.getCachedKey(this.baseURL, modelId);
         if (!cached) return args;
         if (cached === 'max_tokens') return args;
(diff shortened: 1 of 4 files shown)


green-dalii (commented):
Approving — docs-only follow-up to #552 (closes #551).

DocTpoint-style review pass verifies: this PR touches only adjacent prose blocks around the composite-key change, leaving wire format, cache keys, control flow, tests, and settings untouched — no behavioural change for single-model users (wire bytes identical) and no wire change for multi-model either.

Verified site-by-site: `src/llm-sdk/reasoning-strip-probe.ts:47-53` "Why per-baseURL not per-model" header → per-(baseURL, model) Issue #551; step-3 header and invalidate doc clarified; `src/llm-sdk/token-key-probe.ts:13-14,62-67` baseURL → (baseURL, model) and invalidate JSDoc; `src/llm-sdk/openai-compat-sdk-client.ts:229,536,601,768,1610` transformRequestBody/Guard/debug/stream-ordering phrasing plus the synthetic `probeBaseURL('gpt-4o-mini')` isolation note; `src/llm-sdk/openai-sdk-client.ts:81,289` Layer-3 guards. The `probeBaseURL` note documents the accepted narrow abort (cached swap + URL fallback + gateway rejects max_tokens) rather than threading the request model into the probe — proportionate this cycle.

Gate 1 green (lint 0 · tsc 0 · build clean · 3677/3677 · css-lint 0). No side effects, no breaking surface, no performance delta.

PR #559, opened 2026-08-27
fix(alias): thread the minAliasLength floor to the create path; feed f9a680e's cross-page gate at path resolution

### LINKED ISSUE #558: `minAliasLength` never reaches the create path, and f9a680e's cross-page uniqueness gate has no caller
Two findings on the alias gate, found while preparing a vault rebuild; they share one file, so one issue.

**1. The floor stops at the append path.** `filterRedundantAliases` takes a `minLength` parameter. The append path (`page-factory/aliases.ts:65`) resolves it from the `minAliasLength` setting (#532). The create path — `enforceFrontmatterConstraints`, `core/frontmatter.ts:773` — calls the filter with two arguments and silently falls back to the `MIN_ALIAS_LENGTH` constant (2), although the settings object is already a parameter of that very function. A vault with `minAliasLength: 3` therefore still gets two-character aliases on every model-written alias at page birth; the setting *appears* to work because the append path honours it. Same failure shape as #536: the create path not knowing the append path's rule.

**2. The cross-page gate has never had a caller.** `f9a680e` (29 Jul, "alias hardening — 3-char floor + **cross-page uniqueness**") added `existingAliasesAcrossPages` to `filterRedundantAliases` for "alias completion, merge triage". A `git log -S` over the full history finds no call site that ever passes it — the capability was born and never wired. The harm it was designed to stop is real: when path resolution appends an incoming name as an alias to a matched page, nothing checks whether ANOTHER page already claims that name, and an alias carried by two pages silently routes every future occurrence of the name into whichever page matches first. Measured on our vault before curation: 47 such merge trials from 25 duplicated claims in one batch; in an earlier cleanup, 16 of 16 removed collision aliases were shadowed by a same-name page and had never carried a resolution.

PR follows: it threads the floor at the create path (the settings are already in scope there) and feeds the gate at both `appendAliases` call sites in path resolution, where `getExistingWikiPages`' snapshot is already in scope.

DocTpoint:
Closes #558.

**What changes.** Three moves, no behaviour change for callers that pass nothing:

- `resolveMinAliasLength` moves to `core/slug.ts`, next to the filter it parameterises (re-exported from `page-factory/aliases.ts`, existing importers unchanged). `enforceFrontmatterConstraints` now resolves the floor from the settings it already receives — the create path and the append path apply the same floor from the same place.
- `appendAliases` takes an optional `existingAliasesAcrossPages` and threads it into `filterRedundantAliases` — the parameter `f9a680e` introduced, fed for the first time.
- Both `appendAliases` call sites in `path-resolution.ts` (deterministic merge, LLM-decided match) pass `aliasClaimsFromPages(allPages, targetPath)`: the basenames and aliases of every OTHER page, built from the `getExistingWikiPages` snapshot already in scope at both sites — no extra vault read.

**One named semantic choice.** The claim set includes basenames alongside aliases, because Obsidian resolves `[[X]]` to basenames first: an alias equal to another page's basename is shadowed and never carries a resolution. Claims below the floor do not block — `f9a680e`'s own rule, unchanged.

**Not in this PR.** The cross-page gate on the create path: `enforceFrontmatterConstraints` has no pages snapshot in scope, and wiring one would mean a vault read per created page — that cost belongs to the run-wide snapshot question, not to this fix.

Tests: +7 (floor at the create path with and without the setting; gate drops a claimed alias and keeps the rest; case-insensitive claim matching; no-argument call unchanged; claims helper excludes the target page and blank entries). Five-gate green locally: lint, typecheck, build, test 3684/3684, css-lint.

### DIFF
--- a/src/__tests__/core/frontmatter.test.ts
+++ b/src/__tests__/core/frontmatter.test.ts
@@ -432,6 +432,23 @@ describe('enforceFrontmatterConstraints', () => {
   });
 });
 
+describe('enforceFrontmatterConstraints: minAliasLength reaches the create path', () => {
+  const input = '---\ntype: entity\naliases: ["ML", "Maschinelles Lernen"]\n---\n\nBody';
+  const opts = { pagePath: 'wiki/entities/machine-learning.md' };
+
+  it('applies the settings floor to model-written aliases at page birth', () => {
+    const settings = { minAliasLength: 3 } as unknown as LLMWikiSettings;
+    const result = enforceFrontmatterConstraints(input, 'entity', settings, opts);
+    expect(result).toContain('Maschinelles Lernen');
+    expect(result).not.toContain('"ML"');
+  });
+
+  it('keeps the constant floor of 2 when the setting is absent', () => {
+    const result = enforceFrontmatterConstraints(input, 'entity', undefined, opts);
+    expect(result).toContain('"ML"');
+  });
+});
+
 describe('serializeFrontmatter', () => {
   it('emits fields in canonical order: type, created, updated, passthrough, sources, tags, reviewed, aliases', () => {
     const block = serializeFrontmatter(
--- a/src/__tests__/wiki/page-factory/aliases.test.ts
+++ b/src/__tests__/wiki/page-factory/aliases.test.ts
@@ -6,7 +6,7 @@
 // how aliases get added, deduplicated, or rewritten into the frontmatter.
 
 import { describe, it, expect, vi } from 'vitest';
-import { appendAliases, resolveMinAliasLength, type AliasesContext } from '../../../wiki/page-factory/aliases';
+import { aliasClaimsFromPages, appendAliases, resolveMinAliasLength, type AliasesContext } from '../../../wiki/page-factory/aliases';
 
 function makeContext(initial: Record<string, string>): AliasesContext & {
   written: Map<string, string>;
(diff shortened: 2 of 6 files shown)


green-dalii (approved):
Verified against source — all three claims hold.

1. filterRedundantAliases at src/core/slug.ts:139-169 confirms the optional 3rd/4th-arg signature; aliases.ts:65 passes undefined for the cross-page gate; frontmatter.ts:730's 2-arg call silently falls back to MIN_ALIAS_LENGTH=2 (constants.ts:831). Both halves of the finding reproduce against the compiled module.

2. PR #559 is internally consistent: resolveMinAliasLength moves from page-factory/aliases.ts:36 to core/slug.ts, with the existing re-export preserving importers (aliases.test.ts:9 keeps compiling). Six files, +125/-25, and the two path-resolution call sites at path-resolution.ts:151 and :263 both feed aliasClaimsFromPages(allPages, targetPath) — verified in the diff.

3. Named semantic choice (no cross-page gate on the create path) is correct — enforceFrontmatterConstraints has no pages snapshot in scope, and a per-created-page vault read is the wrong cost.

Five-Gate green locally (lint/tsc/build/test 3684/3684/css-lint) — 7 tests added beyond v1.27.0's 3677, matches the PR body's count.

Approving. Awaiting your "merge it" / 合并 signal per CLAUDE.md Git Safety Protocol before `gh pr merge --admin --squash --delete-branch`.

Issue #563, opened 2026-08-27
#560 added a relative link to tools/dev-instrument/README.md in all 11 READMEs, which #375 forbids


DocTpoint:
#560 added `](tools/dev-instrument/README.md)` to `README.md` and the ten locale READMEs under `docs/`.

The #375 guard rejects relative cross-file links in READMEs, and `src/__tests__/root/readme-links.test.ts` has been failing 11× on `main` since `bd0da25`.

The rule is not cosmetic: as its own comment records, Obsidian's community plugin browser strips relative anchors during the marketplace render. For readers who reach the plugin that way — most of them — the reference to the dev instrument is simply gone.

Every other cross-file link in those files already uses the absolute `https://github.com/green-dalii/obsidian-llm-wiki/blob/main/<path>` form; this one was missed.

Fix in the linked PR: same URL in absolute form, 11 files, no prose changed.

green-dalii:
You're right that #375 forbids relative cross-file links, and your measurement is exact: the readme-links guard went red on 11 files at bd0da25. The equivalent fix already shipped as #566 (a4c90c46) at 13:02 UTC, so #565 is now duplicate-of-#566 — I'll close both. Adding to root MEMORY.md as a lesson: path-changing PRs (#511 type) should sync-audit the readme-links guard as a mandatory checklist item.

green-dalii:
Completed by #566 (a4c90c46, 13:02 UTC) — same 11-file single-link absolute-URL rewrite shipped before this issue's PR (#565) was reviewed. Closing as superseded. Lesson logged to root MEMORY.md: path-changing PRs (#511 family) must sync-audit the readme-links guard as a mandatory checklist item.

Issue #375, opened 2026-07-29
Bug: README internal links non-functional in Obsidian community plugin browser


Gojob2987:
## Bug Description

On the Obsidian community plugin details page (Settings → Community plugins → Browse → Karpathy LLM Wiki), internal links in the plugin's README are non-functional:

- **Heading anchor links** (quick-nav bar, table of contents, inline cross-references) — clicking does nothing
- **Relative file links** (language switchers to `docs/README_*.md`, doc references like `docs/PDF-OCR-GUIDE.md`) — clicking does nothing

External `https://` links work. `/docs/assets/*` images render correctly.

The same README works correctly on GitHub — all links function as expected.

---

## Environment

- **Plugin version:** 1.25.9 (also affects earlier versions — the README structure has been stable)
- **Obsidian version:** Any version with a community plugin browser (the issue is platform-level, not version-specific)
- **OS:** Cross-platform (tested on Windows; reproduces on macOS and Linux)
- **Wiki size:** N/A (this is a README rendering issue in the Obsidian plugin browser, not a plugin functionality issue)

---

## LLM Configuration

N/A — this issue is about the plugin's documentation display in the Obsidian community plugin browser, not about plugin functionality. No LLM configuration is involved.

---

## Steps to Reproduce

1. Open Obsidian Desktop → Settings → Community plugins → Browse
2. Search "Karpathy LLM Wiki" → click the plugin entry → read the Details page
3. In the quick-nav bar, click `🤔 Why this plugin?` or `🚀 Quick Start`
4. In the Table of Contents, click any item (e.g., `Features`, `FAQ`)
5. Click a language switcher link: `简体中文`, `日本語`, `Français`, etc.
6. In the Features section, click the cross-reference `[PDF OCR Paths](#-pdf-ocr-paths)`

**Verification on GitHub:**
7. Open `https://github.com/green-dalii/obsidian-llm-wiki` in a browser
8. Perform the same clicks — all links work correctly

---

## Expected Behavior

- Heading anchor links (`#-anchor`) scroll the page to the target heading
- Language switcher links navigate to the corresponding localized README
- Doc reference links navigate to the corresponding file
- All links work the same way as they do on GitHub

---

## Actual Behavior

- All heading anchor links, language switcher links, and doc reference links silently do nothing (they appear interactive but produce no action)
- Only absolute `https://...` links are functional
[...]

PR #564, opened 2026-08-27
fix(gate): link markup is not an aside — exclude its brackets from parenSpans

### LINKED ISSUE #562: Candidate gate reads [[wikilinks]] as parentheses, so linked names are dropped as asides
`PAREN_RE` in `src/core/candidate-gate.ts` treats `[...]` as a bracket pair that marks an aside. In a `[[wikilink]]` it matches the inner `[X]`, and in a markdown link it matches both `[text]` and `(url)`. `classifyCandidate` therefore finds the name inside a "parenthesis" and returns `aside`.

With `skipMentionOnlyCandidates` on, that candidate gets no page. `pruneDropped` then also removes the name from every surviving candidate's `related_*` lists, so the pages that *are* written lose their edges to it as well.

For an Obsidian vault this inverts the signal the gate exists to read. A wikilink is the most explicit statement a note makes that something matters; the gate reads it as a passing mention.

**Measured** on a German vault, 16 notes, 321 candidates, `wikiLanguage: de`:

| cause of the `aside` verdict | candidates |
|---|---|
| link markup only | 33 |
| real parentheses | 37 |
| mixed | 1 |
| list item / enumeration | 16 |

38 % of all `aside` verdicts came from link markup and nothing else. Examples from one note on advanced glycation end products, where the whole mechanism chain is wikilinked and every member was dropped: `[[NF-κB]]`, `[[Oxidativer Stress]]`, `[[Insulinresistenz]]`, `[[Inflammaging]]`, `[[Endotheliale Dysfunktion]]`, `[[Neurodegeneration]]`.

28 of the 31 distinct names had a note of their own, so most of them still reach the wiki through their own ingest — what is lost in those cases is the edge, not the page. Three had no note and lost the page outright.

Introduced with the gate in #521 (#514). Affects every user who turns the setting on.

Fix in the linked PR.

DocTpoint:
Closes #562

`parenSpans` now drops any bracket pair that lies inside a link span (`[[wikilink]]`, `![[embed]]`, `[text](url)`). The text itself is untouched.

**Why not rewrite the text to the display form.** That loses the other half of a piped link — `[[ACE-Hemmer|ACEi]]` names the thing twice and a candidate may carry either name. Measured on the same sample it turned 7 of 87 `aside` verdicts into `absent`: a different drop, not a fix.

**Why not rescue `aside` candidates by their `coverage` value.** It would make a model judgement load-bearing for a gate whose whole point is that its threshold sits in code. On inspection `discussed` marked plain enumeration entries (`(*Lactobacillus*, *Bifidobacterium*, …)`) as readily as it marked real ones.

**The rule cuts both ways.** A parenthesis that merely *contains* a link stays an aside, because it reads as one without the link's brackets too. A bare bracketed citation carries no `(url)` and stays an aside as well. Both are pinned by tests.

Re-measured on the vault above after the change: 32 verdicts move `aside` → `prose`, none move the other way.

Six tests added. `pnpm lint` and `npx tsc --noEmit` clean.

---

**On the first commit.** This branch carries `docs: make the dev-instrument README link absolute in all 11 READMEs` ahead of the fix. That is #565, and it is here only so CI on this branch can run green: `main` is red at `89118bc` because #560 introduced a relative link the #375 guard forbids (#563), and every branch cut from `main` inherits those 11 failures.

Merge #565 first and this commit drops out of the branch on rebase; merge this PR first and #565 becomes empty. Either order is fine — the two changes touch no common file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -219,7 +219,7 @@ The plugin composes with the rest of your Obsidian stack — each tool below plu
 
 **Most users should ignore this section.** The plugin's user-facing CLI lives in the sibling repo [green-dalii/obsidian-llm-wiki-cli](https://github.com/green-dalii/obsidian-llm-wiki-cli) — install with `npm i -g karpathywiki-cli` and run `karpathywiki-cli ingest --sources <path> --wiki <path> --provider <id> --key <key>`.
 
-What ships in this repo at `tools/dev-instrument/` is the **dev-only headless measurement instrument** for engine contributors — it runs the real `WikiEngine.ingestSource` against a vault on disk with no Obsidian runtime, prints per-task token + wall-clock accounting — same numbers that drive the perf evidence in CLAUDE.md and release notes. See [`tools/dev-instrument/README.md`](tools/dev-instrument/README.md) for the entry command, env vars, measurement modes, and exit-code spec.
+What ships in this repo at `tools/dev-instrument/` is the **dev-only headless measurement instrument** for engine contributors — it runs the real `WikiEngine.ingestSource` against a vault on disk with no Obsidian runtime, prints per-task token + wall-clock accounting — same numbers that drive the perf evidence in CLAUDE.md and release notes. See [`tools/dev-instrument/README.md`](https://github.com/green-dalii/obsidian-llm-wiki/blob/main/tools/dev-instrument/README.md) for the entry command, env vars, measurement modes, and exit-code spec.
 
 ---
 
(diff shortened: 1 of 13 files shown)


green-dalii (approved):
Verified against source — PAREN_RE at src/core/candidate-gate.ts:67 with the inner [X] of [[wikilink]] matching at positions 1–len-1 and [text](url] matching both halves is exactly the failure shape your 38% measurement describes. The 5 new test cases pin both boundary directions: a real parenthesis containing a link stays an aside, and a bare bracketed citation stays an aside — that stops the next person from overcorrecting.

The 11 README file changes in this PR are piggybacked from #565 so this branch's CI can run green while main is red at 89118bc. Will close #565 first (already superseded by #566 a4c90c46); then rebase this branch drops the README portion automatically. Merge order: #565 → rebase → this PR, or this PR → #565 becomes empty. Both work; prefer the first so this PR stays focused on the gate fix.

Approving. Awaiting your "merge it" / 合并 signal per CLAUDE.md Git Safety Protocol before `gh pr merge --admin --squash --delete-branch`.

Issue #568, opened 2026-08-27
Write-side domain axis: carry note domain tags onto generated pages, validated against the vault's tag vocabulary


DocTpoint:
Follow-up to #91, where you suggested the read-side framing deserves its own issue. Before a read side can exist, the pages need domain information that is actually reliable — this issue describes the write side we built and measured on our vault, and offers it as a PR.

**What it does.** Source pages receive the note's frontmatter tags 1:1 in a `domains:` field (the note is the truth; nothing to decide). Entity and concept pages receive a per-item subset, chosen by the extraction call — the one place where the source text is still in the window — and validated in code. `tags:` stays untouched: it is the plugin's type axis (#90, #114), and two vocabularies in one field drift apart.

**The first design failed, measurably.** We initially validated each item's choice against the *ingested note's own* tags. Two defects, both visible in one instrumented run over 20 notes (479 per-item answers): items whose kind lies athwart the note's axis — a gene in a disease note — can never receive their tag, which made 43 % of answers empty by construction; and the model drops the group prefix about every third answer (`Mikrobiom` for `Thema/Mikrobiom`), which the strict full-string comparison rejected even though the tag was correct (11 logged drops, all of them right answers).

**The correction.** The allowed list is the vault's tag vocabulary — every frontmatter tag any note carries, collected via the metadata cache — used both in the prompt (one static block, identical for every note, so the prompt-cache prefix survives) and in validation. The spelling fold accepts a bare value iff exactly one vocabulary entry carries it; anything else is dropped, not written. A value no note carries remains unwritable — a model-proposed tag is a candidate, not a fact.

**Measured after the fix** (same note re-ingested, same model and sampling): 42 items with 14 empty and 8 false rejections became 35 items with 7 empty and 0 rejections — and the remaining empties are honest (WHO, DSM-5, ICD-11: the vocabulary has no fitting value). A three-note run under default-like limits: 40 items, 7 empty, exactly 1 rejection — a value the model invented, correctly dropped while the item kept its valid second tag.

**What this is not.** No new setting, no nesting semantics (that stays #91's question), and no read side yet — this is the substrate your `path-resolution` framing needs. Evidence base: one German medical vault, ~420 notes, local model (Gemma-4-26B); the vocabulary quality is inherited from note hygiene, since the list is the union of note tags.

Implementation exists as three commits (~700 lines incl. tests, full suite green). Happy to open the PR.

green-dalii:
[...]

PR #569, opened 2026-08-27
feat(tags): the vault's tags are the vocabulary — folder+wiki harvest, one output field (#568)

DocTpoint:
Implements #568 — the write-side of the tag axis. The series keeps its corrections honest: later commits fix measured failures of the design the earlier ones introduce, and the final two re-shape the axis after live-vault measurement (numbers in the summary comment).

**Where it landed:** the vault's own tags are the vocabulary, and the axis has exactly one output field.

- **Harvest** from the user's declared source folders (`watchedFolders`, read regardless of the auto-watch toggle) plus the nested tags of existing wiki pages — the wiki is where a user manually corrects a tag, and those corrections feed back. `sources/` pages are excluded (auto-generated, written without the constraints pass — the summary model must not mint vocabulary). Flat extraction types are never harvested: they are the abstention fallback, not vocabulary.
- **Validation** (`selectDomains`): fold-keyed membership; a missing *or wrong* group prefix repairs deterministically iff the value part is unique in the vocabulary — 13 of 15 live rejections were wrong-group spellings of carried values.
- **Application:** the validated belonging values join the identity value in `tags:` of generated entity/concept pages, at every writer (create, merge, fill), all validating against the same vocabulary. The create path used to validate against the settings type lists instead — two vocabularies in one pipeline, measured at 19 rejected values surviving onto 58 pages in one run. `enforceFrontmatterConstraints` now strips vocabulary-foreign nested tags instead of retaining them.
- **`domains:` is no field at all.** It never shipped, so there is no legacy to carry: not canonical, never serialized, never merged — a user-authored `domains:` rides the #356 passthrough like any unknown key.
- **No new setting.** A vault without watched folders — or without tags in them — behaves exactly as today: an empty vocabulary renders as the prompt of a vault without the layer. `watchedFoldersDesc` (en/de) now names the dual role; the other locales still carry the shorter text.

3743 tests passing; exercised end to end against a live vault (three instrumented full-batch runs) plus a clean-room smoke vault.


green-dalii (changes requested):
Findings from simplify + code-review (max effort). Architecture is sound; intent reads cleanly; the 3-commit narrative + self-correction in commit 3 is the right way to ship a measured discovery. Pre-merge scope:

**Bugs (6)** — concentrated in one class: fold-aware dedup inconsistency between the validator (`selectDomains`, which uses `fold` = NFC + trim + toLowerCase) and three writers that do raw-string equality:

1. `src/core/frontmatter.ts:540-545` (`mergeFrontmatter`) — raw `includes(d)`.
2. `src/wiki/lint/merge-duplicates.ts:45-46` (`mergeDuplicatePages`) — raw `indexOf(d) === i`, inconsistent with its own `mergedSourcesSet` 6 lines above (which does `trim+lower`).
3. `src/core/batch-merger.ts:67-80` (`dedupStrings`) — raw Set, reused at `batch-merger.ts:747` for domains.

Fix: one fold-keyed helper, three call sites. The validator is already fold-aware, so this is a writer-side consistency pass.

4. `src/core/domain-axis.ts:901` — `collectDomainVocabulary` with empty `wikiFolder` (`prefix = '/'` excludes nothing, leaks wiki-internal tags). Short-circuit on empty, or skip the prefix check entirely.
5. `src/core/frontmatter.ts:539-545` — `mergeFrontmatter` existing-side trims but incoming-side doesn't (latent; mirrors today because `selectDomains` returns canonical strings).
6. `src/core/candidate-gate.ts:793` — `DropReason` union wider than `applyCoverageThreshold` ever emits; narrow to literal `'named'`.

**Tests (3 of 7)** — necessary to lock the bug fixes:
- mergeFrontmatter case-variant dedup (`Thema/Ernährung` + `thema/ernährung`).
- mergeDuplicatePages same.
- collectDomainVocabulary with `wikiFolder === ''`.

The remaining findings (per-ingest vocabulary caching on `EngineContext` + `metadataCache.on('changed')` invalidation, two-pass loop allocation, `EntityInfo`/`ConceptInfo` mixin, schema drift on `coverage` zod schema, prompt-naming `named`) are follow-up quality items — happy to file as separate issues and merge after B1-B6 + T1-T3 land. None are merge blockers; the PR is otherwise review-ready.

Approving once B1-B6 + T1-T3 land in this branch.

DocTpoint:
B1–B6 + T1–T3 landed in `be4da3c`.

The dedup class is now one fold-keyed helper (`unionDomains` in `domain-axis.ts` — NFC + trim + lowercase, matching `selectDomains`) at all three writers. It trims both sides, which closes B5 as well: the existing side comes off a parsed file where a stray space survives, and the incoming side is canonical only as long as every caller runs the validator first. `dedupStrings` stays raw where it belongs — two spellings of a quote are two quotes.

One deliberate deviation on B4. Rather than a short-circuit, the exclusion now goes through `isInFolderScope`, the boundary primitive the repo already has: it reads an empty folder as the vault root, so every file is a wiki page, no note is left to harvest, and the vocabulary is empty — which `buildDomainContext([])` already renders as the prompt of a vault without the layer. That avoids a second hand-rolled reading of "empty folder", and the sibling-prefix anchoring the primitive exists for comes along with it. Non-empty behaviour is byte-identical.

B6 is a generic parameter on `GateResult` / `DroppedCandidate`, so each gate's signature says what that gate can emit — the position gate never reports `named`, the coverage threshold reports nothing else — while the default parameter keeps the shared shape for the engine, which logs both through one line.

Tests: case-variant folding at `mergeFrontmatter` and `mergeDuplicatePages`, `collectDomainVocabulary` with an empty `wikiFolder`, plus the sibling folder that merely shares the name prefix. 3729 → 3733.

Happy to take the follow-up quality items (per-ingest vocabulary caching with `metadataCache` invalidation, the two-pass loop allocation, the `EntityInfo`/`ConceptInfo` mixin, the `coverage` schema drift, and the `named` prompt naming) as separate issues.

DocTpoint:
Please hold this one — I've moved it back to draft.

Running the branch against a real vault turned up behaviour the test suite cannot see. `selectDomains`' bare-value fallback is gated on `!raw.includes('/')`, so it catches a *missing* group prefix but never a *wrong* one. Over one full ingest run, 12 of 14 rejected values named a value part the vocabulary does carry — under a different group (`Sorte/Physiologie` for `Thema/Physiologie`, `Thema/Neurologie` for `Fach/Neurologie`, and so on).

I'm still measuring the axis end to end in a live vault and expect more of this kind, so I'd rather it not merge on my earlier "landed" comment — that one was true about B1–B6 + T1–T3 and silent about this. I'll report back with a fix here or a split, whichever fits better.

DocTpoint:
The live-vault measurement I put this PR on hold for is done — three instrumented runs later, the branch now carries the resulting design. What the vault said, in order:

1. **The two-field shape starved the field readers actually see:** full pages inherited the single-valued extraction `type` in `tags:`, and no step asked the belonging question — 52% of pages carried only the flat type. With the axis writing into `tags:` itself, 8% on the same batch.
2. **The wrong-group gap held:** rejected values overwhelmingly named a value part the vocabulary carries under a different group (13 of 15). `selectDomains` now repairs a wrong prefix the way it repairs a missing one — unique value part wins, two candidates stay a rejection.
3. **Two validation leaks surfaced under load:** the create path validated the model's template tags against the settings type lists instead of the harvest (19 rejected values survived onto 58 pages in one run), and the constraints pass retained foreign values with only a debug log. Both closed — one vocabulary at every writer, foreign nested values stripped.
4. **The vocabulary now comes from content, with no mode and no new setting:** the declared source folders (`watchedFolders`) plus the nested tags of existing wiki pages — the wiki is where a user manually corrects a tag, and corrections feed back (`sources/` excluded; flat fallback types never harvested).
5. **The correction loop, measured:** the first folder-scoped run was deliberately cold — 23 rejections, concentrated on exactly the values the curator had only ever kept in settings. One human pass over 19 flagged wiki pages (editing tags in place — the workflow this design assumes), then the next run: **rejections 23 → 2, vocabulary-foreign values on pages 19 → 0, zero empty tag lists, 64% of new pages multi-valued.**
6. **`domains:` never shipped, so it leaves without legacy** — the PR got smaller, not bigger.

Your CHANGES_REQUESTED was addressed in `be4da3c` and your approval was pledged to the `domains` design — so please treat this as a fresh review target rather than a delta. Taking it out of draft now; happy to split or reshape wherever you'd like it different.

DocTpoint:
Rebased onto `7c4d144` (main was 15 commits ahead, branch had gone CONFLICTING). One import-line conflict in `merge-page.ts` against #576/#579, resolved as the union; no code changes otherwise. Same ten commits as before, new head `d1e04a7`. Typecheck, lint and 3801 tests green locally, Gate 1 green.

DocTpoint:
Rebased onto `2bd4a6d` after this afternoon's merge wave. One import-line conflict in `merge-page.ts` against #602 (`appendContradictionNotes` no longer exists there), resolved as the union of the remaining imports; no code changes otherwise. Same ten commits, new head `acef6c3`. Typecheck, lint and 3852 tests green locally.

green-dalii (approved):
Approve. Full re-review after rebase onto wave-B main (8feb5fd + docs) + review fixes.

**B1-B6 from the 08-28 CHANGES_REQUESTED — all verified fixed:**
- B1/B2/B3: `unionDomains` fold-keyed helper now used at every writer (frontmatter.ts, merge-duplicates.ts, batch-merger.ts) — raw-equality dedup gone.
- B4: `collectDomainVocabulary` guards empty `wikiFolder`/`watchedFolders` (returns [] → no-layer prompt).
- B5: both sides trimmed in the fold.
- B6: `DropReason = Exclude<GateVerdict,'prose'> | 'named'` narrows the union; `GateResult<R>` parameterised so each gate's result says what it can emit.

**Rebase integration (new since 08-28):**
- Rebased onto current main incl. #612 (localDateStamp — F3: the PR's own date sites now local, no UTC regression), #621 (candidate-gate isKnownPage), #606/#610 (contradiction counting + two gates).
- F2 fix commit: `applyCoverageThreshold` now answers `isKnownPage` exactly like the deterministic gate (#620 parity) — a dropped `named` candidate the vault already has a page for keeps its edge and is reported in `linkedAnyway`; without the predicate behaviour is unchanged. Shared `pruneDroppedNames`, both gates carry `linkedAnyway`. 2 new tests lock the parity.
- `gateCandidates` reuses the already-read `rawSource` (no double vault read).

**Gate 1 verified:** 3892 tests (276 files), lint/tsc/build/css clean. Local + CI.

One design note for the release notes (not blocking): the coverage threshold (`named` → no page) runs regardless of the `skipMentionOnlyCandidates` opt-in — a new default filter for models that emit `coverage`. Worth an explicit note in the next PATCH release notes.

green-dalii (approved):
Approve. Full re-review after rebase onto wave-B main (8feb5fd + docs) + review fixes.

**B1-B6 from the 08-28 CHANGES_REQUESTED — all verified fixed:**
- B1/B2/B3: `unionDomains` fold-keyed helper now used at every writer (frontmatter.ts, merge-duplicates.ts, batch-merger.ts) — raw-equality dedup gone.
- B4: `collectDomainVocabulary` guards empty `wikiFolder`/`watchedFolders` (returns [] → no-layer prompt).
- B5: both sides trimmed in the fold.
- B6: `DropReason = Exclude<GateVerdict,'prose'> | 'named'` narrows the union; `GateResult<R>` parameterised so each gate's result says what it can emit.

**Rebase integration (new since 08-28):**
- Rebased onto current main incl. #612 (localDateStamp — F3: the PR's own date sites now local, no UTC regression), #621 (candidate-gate isKnownPage), #606/#610 (contradiction counting + two gates).
- F2 fix commit: `applyCoverageThreshold` now answers `isKnownPage` exactly like the deterministic gate (#620 parity) — a dropped `named` candidate the vault already has a page for keeps its edge and is reported in `linkedAnyway`; without the predicate behaviour is unchanged. Shared `pruneDroppedNames`, both gates carry `linkedAnyway`. 2 new tests lock the parity.
- `gateCandidates` reuses the already-read `rawSource` (no double vault read).

**Gate 1 verified:** 3892 tests (276 files), lint/tsc/build/css clean. Local + CI.

One design note for the release notes (not blocking): the coverage threshold (`named` → no page) runs regardless of the `skipMentionOnlyCandidates` opt-in — a new default filter for models that emit `coverage`. Worth an explicit note in the next PATCH release notes.

PR #570, opened 2026-08-28
feat(pdf): support vision-capable local providers (ollama, lmstudio) via per-page PNG rasterization

luxezreal:
# feat(pdf): support vision-capable local providers (ollama, lmstudio) via per-page PNG rasterization

> **PR title candidates** (copy one into the GitHub PR title field):
>
> 1. `feat(pdf): support vision-capable local providers (ollama, lmstudio) via per-page PNG rasterization`
> 2. `feat(pdf-vision): route ollama/lmstudio through per-page PNG instead of native PDF`

> **TL;DR**: today `UnsupportedProviderError` rejects `ollama` / `lmstudio` even with `forcePdfSupport: true`, because the AI SDK's openai-compat client doesn't emit `application/pdf` parts. This PR adds a parallel **vision path**: `pdfjs-dist` rasterizes each page to PNG and sends it as `image_url` content parts, which Ollama and LM Studio's `/v1/chat/completions` accept natively.

---

## Summary

Enables PDF ingestion on Ollama and LM Studio by rasterizing each page to PNG and sending it as an `image_url` content part. Native PDF (anthropic / openai / bedrock-anthropic / bedrock-openai) is unchanged.

## Motivation

- **Blocker**: every user on `provider: ollama` (or `lmstudio`) who tries to ingest a PDF sees `UnsupportedProviderError`. With `forcePdfSupport: true` the gate passes but the underlying AI SDK still doesn't produce a wire format those endpoints accept.
- **Provider support list**: `src/constants.ts:115-120` enumerates only `anthropic` / `openai` / `bedrock-anthropic` / `bedrock-openai`. Local vision users are stranded.
- **Capability gap, not a contract gap**: Ollama and LM Studio both accept multimodal `image_url` on `/v1/chat/completions`; they just don't accept the OpenAI-SDK-style `application/pdf` part. We control the wire format.
- **Local vision is mainstream in 2026**: llama3.2-vision, gemma-3, llava, qwen2-vl are first-class Ollama / LM Studio models. Drop a PDF, run ingest, get a wiki page — no cloud token needed.

### src/constants.ts

Insert a vision-provider list and version suffix right after `NATIVE_PDF_PROVIDER_IDS` (`src/constants.ts:115-120`):

```diff
 export const NATIVE_PDF_PROVIDER_IDS = [
   'anthropic',
   'openai',
   'bedrock-anthropic',
   'bedrock-openai',
 ] as const;
+
+/**
+ * v1.28.x PR — Provider IDs whose `/v1/chat/completions` accepts
+ * multimodal `image_url` parts. Used by `core/pdf-converter.ts` to
+ * route PDF ingestion through per-page PNG rasterization (via
+ * pdfjs-dist) instead of the `application/pdf` content type that
+ * openai-compat SDKs do not emit.
+ *
+ * Why a separate list (not an addition to NATIVE_PDF_PROVIDER_IDS):
+ * the conversion path is fundamentally different — the client must
+ * rasterize the PDF before sending, the message contains N image
+ * parts instead of one file part, and the cache key includes a
+ * DPI/version suffix to keep vision conversions isolated from
+ * native ones (a user switching providers cannot silently receive a
+ * stale conversion from a different backend).
+ */
+export const VISION_PDF_PROVIDER_IDS = [
+  'ollama',
+  'lmstudio',
+] as const;
+
+/** Version suffix for the vision-path cache key. Bump when the
+ *  rasterizer (DPI, image format, pdfjs version) changes. */
+export const VISION_PDF_VERSION = 1;
+export const VISION_PDF_DPI = 150;
```

### src/types.ts

Extend `MessageContentPart` (`src/types.ts:698-700`) to include an `image` variant:

```diff
 export type MessageContentPart =
   | { type: 'text'; text: string }
-  | { type: 'file'; data: string; mediaType: 'application/pdf'; filename?: string };
+  | { type: 'file'; data: string; mediaType: 'application/pdf'; filename?: string }
+  | { type: 'image'; image: string; mimeType: 'image/png' | 'image/jpeg'; filename?: string };
```

No change to `DEFAULT_SETTINGS` (`src/types.ts:1182`); `markdownConversionBackend: 'native'` stays as default and is used only by the MinerU route.

### src/core/pdf-converter.ts

**a. Update the imports at `src/core/pdf-converter.ts:42-45`:**

```diff
 import {
   TOKENS_PDF_CONVERSION,
   NATIVE_PDF_PROVIDER_IDS,
+  VISION_PDF_PROVIDER_IDS,
+  VISION_PDF_VERSION,
+  VISION_PDF_DPI,
 } from '../constants';
```

**b. Update the comment-header provider matrix at `src/core/pdf-converter.ts:21-24`:**

```diff
  * Provider support matrix:
  *   - anthropic / openai / bedrock-anthropic / bedrock-openai: native PDF support
+ *   - ollama / lmstudio: vision-path (PDF → per-page PNG via pdfjs-dist → image_url parts)
  *   - custom / anthropic-compatible: requires forcePdfSupport=true (else throw)
  *   - deepseek / glm / kimi / openrouter / gemini / future providers: never supported (throw)
```

**c. Extend the capability gate at `src/core/pdf-converter.ts:246-250`:**

```diff
 function providerSupportsPdf(settings: PdfConversionContext['settings']): boolean {
   if ((NATIVE_PDF_PROVIDER_IDS as readonly string[]).includes(settings.provider)) return true;
+  if ((VISION_PDF_PROVIDER_IDS as readonly string[]).includes(settings.provider)) return true;
   if (settings.forcePdfSupport === true) return true;
   return false;
 }
```

**d. Update the error message at `src/core/pdf-converter.ts:93-97`:**

```diff
       `PDF conversion is not supported by provider "${provider}". ` +
-        `Supported providers: anthropic, openai, bedrock-anthropic, bedrock-openai. ` +
+        `Supported providers: anthropic, openai, bedrock-anthropic, bedrock-openai, ` +
+        `ollama, lmstudio. ` +
         `For other OpenAI-compatible or Anthropic-compatible providers, enable ` +
         `"Force PDF Support" in Settings → LLM Configuration → Advanced (at your own risk).`
```

**e. Insert the dispatch in `convertPdfToMarkdown` between the metadata-extract block (`src/core/pdf-converter.ts:165`) and the existing LLM call (`src/core/pdf-converter.ts:175`):**

```diff
   const info = parsePdfInfoDictText(text);
+
+  // v1.28.x PR — dispatch by provider capability. Native providers
+  // get a single file content part; vision providers get per-page
+  // PNG rasterization + image content parts. The branch is on the
+  // same call site so metadata extraction and cache miss logic
+  // stay in one place.
+  const isVision = (VISION_PDF_PROVIDER_IDS as readonly string[]).includes(settings.provider);
+
+  // Cache key for the vision branch includes DPI + version so a
+  // user switching from anthropic to ollama cannot silently reuse
+  // an anthropic-rendered conversion.
+  let visionFileToken: string | null = null;
+  if (isVision) {
+    const visionLogicalKey = `${logicalKey}:vision:v${VISION_PDF_VERSION}@${VISION_PDF_DPI}dpi`;
+    visionFileToken = await hashCacheKey(visionLogicalKey, subtle);
+    const visionCached = await cache.get(visionFileToken);
+    if (visionCached) return visionCached;
+  }
+
+  // 7. Call LLM via the appropriate path.
+  const response = isVision
+    ? await convertPdfToMarkdownViaVision({
+        llmClient, model, system: PDF_PROMPTS.systemPrompt,
+        pdfBytes: bytes, info, pdfFileName: pdfFile.name,
+        abortSignal: ctx.abortSignal,
+      })
+    : (await llmClient.createMessage({
+        task: 'pdf-convert',
+        model,
+        max_tokens: TOKENS_PDF_CONVERSION,
+        system: PDF_PROMPTS.systemPrompt,
+        messages: [{
+          role: 'user',
+          content: [
+            { type: 'text', text: buildUserText(info) },
+            { type: 'file', data: bytesToBase64(bytes), mediaType: 'application/pdf', filename: pdfFile.name },
+          ],
+        }],
+        ...(ctx.abortSignal ? { abortSignal: ctx.abortSignal } : {}),
+      }));
```

Then change the cache-write line `src/core/pdf-converter.ts:219`:

```diff
-  await cache.set(fileToken, entry);
+  await cache.set(isVision ? visionFileToken! : fileToken, entry);
```

and the converter metadata at `src/core/pdf-converter.ts:216`:

```diff
       converter: isVision
-        ? `${settings.provider}/${model}`
+        ? `${settings.provider}/${model}:vision:v${VISION_PDF_VERSION}`
+        : `${settings.provider}/${model}`,
```

**f. Append the vision helper at the end of `src/core/pdf-converter.ts` (after `bytesToBase64` at `src/core/pdf-converter.ts:267`):**

```ts
// --- vision path ---

/**
 * v1.28.x PR — PDF → per-page PNG → image content parts.
 *
 * Rasterizes each PDF page with pdfjs-dist (legacy build, no worker)
 * using OffscreenCanvas. Works inside Obsidian's Electron renderer
 * without violating the obsidianmd/no-node-builtin ESLint rule.
 *
 * Message shape sent to llmClient.createMessage:
 *   user: [
 *     { type: 'text', text: <buildUserText(info)> + " (N page(s) attached)" },
 *     { type: 'image', image: <base64 PNG>, mimeType: 'image/png' }, // page 1
 *     { type: 'image', image: <base64 PNG>, mimeType: 'image/png' }, // page 2
 *     ...
 *   ]
 *
 * The AI SDK's openai-compat provider encodes image parts as OpenAI
 * Chat Completions `image_url`, which Ollama and LM Studio accept on
 * /v1/chat/completions natively and convert to their internal
 * images[] representation server-side.
 *
 * Throws: pdfjs-dist exceptions propagate verbatim (e.g. "Invalid PDF
 * structure" / "Password required") — caller can surface them.
 *
 * Abort: signal propagates to the LLM call but NOT to per-page
 * rasterization (cancellation mid-render would leak partial state).
 * For a worst-case 200-page cancel-at-page-50 the cost is 50
 * rasterized pages of CPU — acceptable, no external resources held.
 */
async function convertPdfToMarkdownViaVision(params: {
  llmClient: LLMClient;
  model: string;
  system: string;
  pdfBytes: Uint8Array;
  info: { title?: string; author?: string; pageCount?: number };
  pdfFileName: string;
  abortSignal?: AbortSignal;
}): Promise<string> {
  const { llmClient, model, system, pdfBytes, info, pdfFileName, abortSignal } = params;

  // pdfjs-dist v5 legacy build: synchronous rendering, no Worker. The
  // default worker import throws inside Obsidian's CSP, so we ship an
  // empty workerSrc to force the legacy fallback.
  // eslint-disable-next-line @typescript-eslint/no-require-imports
  const pdfjs: typeof import('pdfjs-dist') =
    await import('pdfjs-dist/legacy/build/pdf.mjs');
  // @ts-expect-error — empty workerSrc disables the worker
  pdfjs.GlobalWorkerOptions.workerSrc = '';

  // OffscreenCanvas factory — pdfjs calls create() per page, reset()
  // between renders. OffscreenCanvas is structurally compatible with
  // the HTMLCanvasElement surface pdfjs uses (width/height/getContext)
  // and stays inside the browser-safe API.
  const canvasFactory = {
    create(width: number, height: number) {
      const canvas = new OffscreenCanvas(width, height);
      return { canvas, context: canvas.getContext('2d')! };
    },
    reset(c: { canvas: OffscreenCanvas; context: OffscreenCanvasRenderingContext2D },
          width: number, height: number) {
      c.canvas.width = width;
      c.canvas.height = height;
    },
    destroy(c: { canvas: OffscreenCanvas; context: OffscreenCanvasRenderingContext2D }) {
      c.canvas.width = 0;
      c.canvas.height = 0;
    },
  };

  const scale = VISION_PDF_DPI / 72; // PDF spec: 72 DPI base
  const loadingTask = pdfjs.getDocument({
    data: pdfBytes,
    // @ts-expect-error — disableWorker is in v5 legacy options
    disableWorker: true,
    isEvalSupported: false, // Obsidian CSP forbids new Function(...)
    useSystemFonts: true,
  });
  const pdf = await loadingTask.promise;
  const pageImages: Array<{ image: string; mimeType: 'image/png' }> = [];

  try {
    for (let i = 1; i <= pdf.numPages; i++) {
      const page = await pdf.getPage(i);
      const viewport = page.getViewport({ scale });
      const ctxObj = canvasFactory.create(viewport.width, viewport.height);
      try {
        await page.render({
          canvasContext: ctxObj.context,
          viewport,
          canvasFactory,
        }).promise;
        const blob = await ctxObj.canvas.convertToBlob({ type: 'image/png' });
        const buffer = new Uint8Array(await blob.arrayBuffer());
        pageImages.push({ image: bytesToBase64(buffer), mimeType: 'image/png' });
      } finally {
        canvasFactory.destroy(ctxObj);
      }
    }
  } finally {
    await pdf.cleanup();
    await pdf.destroy();
  }

  const messages = [{
    role: 'user' as const,
    content: [
      { type: 'text' as const,
        text: buildUserText(info) + ` (${pageImages.length} page(s) attached as images)` },
      ...pageImages.map((p) => ({
        type: 'image' as const,
        image: p.image,
        mimeType: p.mimeType,
      })),
    ],
  }];

  return await llmClient.createMessage({
    task: 'pdf-convert',
    model,
    max_tokens: TOKENS_PDF_CONVERSION,
    system,
    messages,
    ...(abortSignal ? { abortSignal } : {}),
  });
}
```

### package.json

Add `pdfjs-dist` to `dependencies` (`package.json:57-64`):

```diff
   "dependencies": {
     "@ai-sdk/anthropic": "3.0.98",
     "@ai-sdk/openai": "3.0.86",
     "@ai-sdk/openai-compatible": "2.0.62",
     "ai": "6.0.230",
     "fflate": "0.8.3",
+    "pdfjs-dist": "5.3.93",
     "zod": "^3.25.76"
   }
```

> **bundling note for maintainers**: pdfjs-dist legacy build ≈ 1.3 MB minified. `esbuild.config.mjs` may need to mark `pdfjs-dist/legacy/build/pdf.mjs` as external, or we switch to dynamic `import()` (already used inside `convertPdfToMarkdownViaVision`). Bundle stays flat for non-vision users; vision users pay the +1 MB on first PDF call.

## Testing

**Unit / integration (vitest):**
- New file `src/core/pdf-converter.test.ts`:
  - `providerSupportsPdf({provider:'ollama'})` returns `true`
  - `providerSupportsPdf({provider:'lmstudio'})` returns `true`
  - Same PDF + same model → anthropic-path `fileToken` ≠ ollama-path `fileToken` (cache isolation)
  - Second call with same inputs returns the cached entry without invoking `llmClient.createMessage` (mock spy)
- Existing `src/core/pdf-converter.test.ts` native-path tests — must pass unchanged.

**Manual e2e (maintainer runbook):**
1. `pnpm install && pnpm gate:1` (lint + typecheck + build + test + css-lint).
2. With `provider: ollama`, model = a vision GGUF (`llama3.2-vision:11b` with mmproj). Drop a 1-page PDF in a watched folder → run ingest → assert wiki page created with non-empty markdown.
3. Re-ingest the same PDF → second call hits cache (no LLM call; verify via debug log).
4. Switch provider to `anthropic`, re-ingest same PDF → cache miss, native path runs.
5. Repeat steps 2–3 with `provider: lmstudio` and a vision model (e.g. `qwen2-vl-7b`).
6. With `provider: ollama` but a text-only GGUF → confirm a clear failure surfaces (empty / 400 from the server). `UnsupportedProviderError` MUST NOT be raised — the path is open; the user's model choice is their responsibility (release-note caveat).

**i18n check:**
- `src/texts/*.ts`: `UnsupportedProviderError` message now includes "ollama, lmstudio". Verify all 10 locales (en / zh / zh-Hant / ja / ko / de / fr / es / pt / it).

## Checklist

- [ ] Added CHANGELOG entry under `## [Unreleased]` with provider-list change
- [ ] Updated i18n strings for `UnsupportedProviderError` in 10 locales
- [ ] Added `pdfjs-dist` to `dependencies` and verified `pnpm install` resolves cleanly
- [ ] Added vitest cases in `src/core/pdf-converter.test.ts` (providerSupportsPdf + cache-key isolation)
- [ ] Ran `pnpm gate:1` locally (lint + typecheck + build + test + css-lint)
- [ ] Ran manual e2e on `ollama` (vision GGUF) and `lmstudio` per Testing §6
- [ ] Verified `package.json` version bump to `1.28.0` pre-release tag
- [ ] Confirmed no behavior change for users on `anthropic` / `openai` / `bedrock-*` (native path untouched)

## Risk & rollback

**Risk hotspots:**
- **Bundle size**: +1 MB if `pdfjs-dist` ships statically; +0 KB cold start if dynamic-`import()` is honored by the bundler. Maintainer can request the dynamic form; this PR ships the static form for simplicity (with the dynamic `import()` inside the helper as a safety net).
- **Concurrency on large PDFs**: N pages × ~0.6 MB base64. A 200-page paper transiently holds ~120 MB of strings in memory. No chunking — see Reviewer FAQ #3.
- **PDF rasterization fidelity**: 150 DPI text rendering through `llama3.2-vision` quant models varies; Chinese OCR is fragile on Q1_0 quant — release-note warns (release-note caveat 4).
- **Encrypted PDFs**: `EncryptedPdfError` (`src/core/pdf-converter.ts:102-111`) is unchanged — user must decrypt first (out of scope).

**Rollback steps:**
1. Revert the merge commit. Single PR, no settings migrations.
2. Users who already ingested PDFs through ollama/lmstudio keep their cached entries — those entries are inert post-revert (nothing reads them).
3. `DEFAULT_SETTINGS` (`src/types.ts:1182`) is unchanged; no required new fields.
4. Carry the release-note `KNOWN LIMITATIONS` block over to whatever supersedes this PR.

## Reviewer FAQ

| Concern | Reply |
|---|---|
| "pdfjs-dist is too large, plugin exceeds 5 MB" | `convertPdfToMarkdownViaVision` already uses dynamic `await import('pdfjs-dist/legacy/build/pdf.mjs')`. If the bundler tree-shakes the static import, the bundle stays flat for non-vision users. Vision users pay the +1 MB on first PDF call. Maintainer can request we drop the static fallback in `pdfjs-dist`'s package.json to force true lazy load. |
| "Why not just tell users to use MinerU?" | MinerU is a hosted API — requires token + internet. Local vision users explicitly want offline. Both backends now coexist (`markdownConversionBackend: 'native' \| 'mineru'`, default `native` in `src/types.ts:1182`). |
| "OffscreenCanvas isn't universally available" | Obsidian 1.4+ (2022) ships Electron 22+, which has OffscreenCanvas as a Web standard. Users reporting "OffscreenCanvas is not defined" are on a pre-2022 Obsidian; fallback path is `pdfjs-dist@4`'s `createImageBitmap` API. |
| "Why not bypass AI SDK and hit `/api/generate` directly for Ollama?" | Drops 8 fallback mechanisms (`enableThinking` probe, output-mode demotion, prompt-cache breakpoints, structured-output tier, `onFinish` finish-reason channel, …). The cost of maintaining a parallel SDK isn't worth it for one wire-format difference. |
| "Cache-key suffix is fragile — what if `VISION_PDF_VERSION` drifts out of sync?" | The suffix (`v${VISION_PDF_VERSION}@${VISION_PDF_DPI}dpi`) IS the cache-busting contract: bump it whenever DPI, format, or pdfjs version changes. The new vitest case asserts the key contains the suffix. Reviewers should grep for `VISION_PDF_VERSION` if they change DPI/format. |
| "Why per-page images instead of a tiled grid?" | Ollama and LM Studio both reject composite/tiled inputs on `/v1/chat/completions`; the AI SDK doesn't expose grid encoding. Per-page is the only contract the openai-compat provider produces that local servers accept. |

## Out of scope

- **Image-only sources (PNG / JPG) via the vision path** — not addressed here. The existing native path handles single-image files via `application/pdf`-style adapters. Future work.
- **PDF password removal** — `EncryptedPdfError` (`src/core/pdf-converter.ts:102-111`) is unchanged. Users decrypt with qpdf / Acrobat first.
- **Caching rasterized page images on disk** — currently we re-rasterize on every cache miss. Caching PNG intermediates would cut CPU on large PDFs but adds invalidation surface; deferred to v1.29+.
- **Per-page chunked sending** (split a 100-page PDF into 4 messages of 25) — noted in the helper docstring; deferred to a `PDF_VISION_MAX_PAGES_PER_CALL` setting in v1.29 if memory becomes a complaint.
- **Non-Latin-script OCR improvements** — relies on the local model's native OCR. Quality varies by quant level; release-note caveat 4.
- **Removing the `forcePdfSupport` escape hatch** — kept for custom / anthropic-compatible users who want to experiment with their own endpoints.

## Suggested commit message

> ```
> feat(pdf): support vision-capable local providers (ollama, lmstudio) via per-page PNG rasterization
>
> Adds pdfjs-dist-based PDF → image content part pipeline for providers
> whose /v1/chat/completions accepts image_url (Ollama, LM Studio). Native
> PDF path (anthropic/openai/bedrock-*) is unchanged. Closes the long-
> standing "provider cannot read PDF" error for local vision models.
> ```

### DIFF
--- a/esbuild.config.mjs
+++ b/esbuild.config.mjs
@@ -1,6 +1,18 @@
 import esbuild from "esbuild";
 import process from "process";
+import fs from "fs";
+import path from "path";
 import { builtinModules } from "module";
+import { createRequire } from "module";
+
+const require = createRequire(import.meta.url);
+// Read pdfjs worker source once at build time and inline via define.
+// We resolve through createRequire so the path is stable under pnpm's
+// .pnpm/<name>@<ver>/node_modules/<name>/ layout.
+const workerPath = require.resolve(
+  "pdfjs-dist/legacy/build/pdf.worker.mjs"
+);
+const workerSource = fs.readFileSync(workerPath, "utf8");
 
 const banner =
 `/*
@@ -19,6 +31,13 @@ const context = await esbuild.context({
     js: banner + prodBanner,
   },
   entryPoints: ['src/main.ts'],
+  // Inline pdf.worker.mjs source as a string constant so we can hand
+  // pdfjs a blob: URL without shipping a second .mjs file alongside
+  // main.js (relative URL resolution is unreliable in the Obsidian
+  // Electron renderer under CJS bundle — no import.meta.url).
+  define: {
+    __PDFJS_WORKER_SOURCE__: JSON.stringify(workerSource),
+  },
   bundle: true,
   external: [
     'obsidian',
--- a/package.json
+++ b/package.json
@@ -60,6 +60,7 @@
     "@ai-sdk/openai-compatible": "2.0.62",
     "ai": "6.0.230",
     "fflate": "0.8.3",
+    "pdfjs-dist": "5.3.93",
     "zod": "^3.25.76"
   }
 }
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -27,6 +27,9 @@ importers:
       fflate:
         specifier: 0.8.3
         version: 0.8.3
+      pdfjs-dist:
+        specifier: 5.3.93
+        version: 5.3.93
       zod:
         specifier: ^3.25.76
         version: 3.25.76
@@ -424,6 +427,81 @@ packages:
     peerDependencies:
       eslint: ^9
 
+  '@napi-rs/canvas-android-arm64@0.1.100':
+    resolution: {integrity: sha512-hjhCKhntPv9+t4ckHymdx0phYNcVW+GKQR6Lzw2zE+pOVjOplSmtx9nNNknTjbEDLcuLZqA1y8ufKg1XfgftzQ==}
+    engines: {node: '>= 10'}
+    cpu: [arm64]
+    os: [android]
+
+  '@napi-rs/canvas-darwin-arm64@0.1.100':
+    resolution: {integrity: sha512-2PcswRaC7Ly645DGt88///zuFDhJxJYdKAs1uU3mfk1atYkXufgcgLfBpk6Tm12nCQBaNt1wpybuPZ4qOhTo8A==}
+    engines: {node: '>= 10'}
+    cpu: [arm64]
+    os: [darwin]
+
+  '@napi-rs/canvas-darwin-x64@0.1.100':
+    resolution: {integrity: sha512-ePNZtj7pNIva/siZMg+HmbeozkIjqUIYdoymH8HaA3qK7LfzFN4WMBM8G6HQ9ZC+H3+Dnn5pqtiXpgLykaPOhw==}
+    engines: {node: '>= 10'}
+    cpu: [x64]
+    os: [darwin]
+
+  '@napi-rs/canvas-linux-arm-gnueabihf@0.1.100':
+    resolution: {integrity: sha512-d5cDB48oWFGU8/XPhUOFAlySgb/VAu7D+s8fi55K1Pcfg8aPplHWqMgibhVLU8ky7Pyg/fuiVLz4Nf3JrSTuUA==}
+    engines: {node: '>= 10'}
+    cpu: [arm]
+    os: [linux]
+
+  '@napi-rs/canvas-linux-arm64-gnu@0.1.100':
+    resolution: {integrity: sha512-rDxgxRu69RvDlX/bh9o22DxLsGr8EqsNgotL9+RwQE1S0b0cqeatqsw6aW45mukm0B42DIAaAacKaYQ8cqS1nw==}
+    engines: {node: '>= 10'}
+    cpu: [arm64]
+    os: [linux]
+    libc: [glibc]
+
+  '@napi-rs/canvas-linux-arm64-musl@0.1.100':
+    resolution: {integrity: sha512-K3mDW66N+xT2/V439u1alFANiBUjdEx2gLiNYnCmUsva5jZMxWTjafBYwTzYK+EMFMHrUoabuU+T1BIP5CgbYQ==}
+    engines: {node: '>= 10'}
+    cpu: [arm64]
+    os: [linux]
+    libc: [musl]
+
+  '@napi-rs/canvas-linux-riscv64-gnu@0.1.100':
+    resolution: {integrity: sha512-mooqUBTIsccZpnoQC4NgrC1v6C1vof39etLNMnBwCY+p0gajWJvAHLGQ6g/gGyS5YrpDW+GefSN4+Cvcr08UWw==}
+    engines: {node: '>= 10'}
+    cpu: [riscv64]
+    os: [linux]
+    libc: [glibc]
+
+  '@napi-rs/canvas-linux-x64-gnu@0.1.100':
+    resolution: {integrity: sha512-1eCvkDCazm7FFhsT7DfGOdSaHgZVK3bt/dSBl5EWHOWmnz+I7j8tPseJqqD81NF+MH21jKUK4wQSDjN0mdhnTg==}
+    engines: {node: '>= 10'}
+    cpu: [x64]
+    os: [linux]
+    libc: [glibc]
+
+  '@napi-rs/canvas-linux-x64-musl@0.1.100':
+    resolution: {integrity: sha512-20arT6lnI19S68qNlii73TSEDbECNgzMz2EpldC1V3mZFuRkeujXkcebRk0LRJe9SEUAooYiLokfMViY8IX7yA==}
+    engines: {node: '>= 10'}
+    cpu: [x64]
+    os: [linux]
+    libc: [musl]
+
+  '@napi-rs/canvas-win32-arm64-msvc@0.1.100':
+    resolution: {integrity: sha512-DZFFT1wIAg37LJw37yhMRFfjATd3vTQzjZ1Yki8u2vhO6Hi5VE6BVaGQ1aaDu7xb4iMErz+9EOwjpS7xcxFeBw==}
+    engines: {node: '>= 10'}
+    cpu: [arm64]
+    os: [win32]
+
+  '@napi-rs/canvas-win32-x64-msvc@0.1.100':
+    resolution: {integrity: sha512-MyT1j3mHC2+Lu4pBi9mKyMJhtP6U7k7EldY7sj/uS5gJA65gTXt8MefJQXLJo5d/vZbuWmfxzkEUNc/urV3pHA==}
+    engines: {node: '>= 10'}
+    cpu: [x64]
+    os: [win32]
+
+  '@napi-rs/canvas@0.1.100':
+    resolution: {integrity: sha512-xglYA6q3XO5P3BNJYxVZ1IV7DLVjp1Py6nwag88YntrS+3vKHyYcMqXVS4ZztJmwz2uGvz1FWhI/4LgbR5uQDA==}
+    engines: {node: '>= 10'}
+
   '@opentelemetry/api@1.9.1':
     resolution: {integrity: sha512-gLyJlPHPZYdAk1JENA9LeHejZe1Ti77/pTeFm/nMXmQH/HFZlcS/O2XJB+L8fkbrNSqhdtlvjBVjxwUYanNH5Q==}
     engines: {node: '>=8.0.0'}
@@ -470,36 +548,42 @@ packages:
     engines: {node: ^20.19.0 || >=22.12.0}
     cpu: [arm64]
     os: [linux]
+    libc: [glibc]
 
   '@rolldown/binding-linux-arm64-musl@1.2.3':
     resolution: {integrity: sha512-NHqjnxpsndf4MPymxteFAWHHfkTL8HjWh1KB7z23ofZ6QO2euONuxDXjat69dKZRALnGypg8k8SsK8vZJoXv1Q==}
     engines: {node: ^20.19.0 || >=22.12.0}
     cpu: [arm64]
     os: [linux]
+    libc: [musl]
 
   '@rolldown/binding-linux-ppc64-gnu@1.2.3':
     resolution: {integrity: sha512-6tbrbwfz5GB9DQ4Jwo6hy9v+vR31xZlvzZ6n5Xut6Hhx5PvrA9q/HsK8KMaYQp063iqZGXwNvZtYNLD7EM/x0w==}
     engines: {node: ^20.19.0 || >=22.12.0}
     cpu: [ppc64]
     os: [linux]
+    libc: [glibc]
 
   '@rolldown/binding-linux-s390x-gnu@1.2.3':
     resolution: {integrity: sha512-oyuXxXmoZHjXC917IAPFAAv4wWAa0cM9afk8nx1+9/jNNOX1uPf8yDA6p7G0RypOfw/X0PQt5IfoquY1um+zSg==}
     engines: {node: ^20.19.0 || >=22.12.0}
     cpu: [s390x]
     os: [linux]
+    libc: [glibc]
 
   '@rolldown/binding-linux-x64-gnu@1.2.3':
     resolution: {integrity: sha512-TytMwF2KVGqP2tgd0I1OY0PAv78dZRAYcF5ssDzjM34SUXCED3uXvSd5+lHoC0bTD6eEdFz7LdQNCO1y0oVk9w==}
     engines: {node: ^20.19.0 || >=22.12.0}
     cpu: [x64]
     os: [linux]
+    libc: [glibc]
 
   '@rolldown/binding-linux-x64-musl@1.2.3':
     resolution: {integrity: sha512-/E9m3qstrJFVPoULV25mVQblSNExY2+kBsYe4sy0Tn0yOOgJ8wZbZt3KnRbF/XeU2Gl1STKUQnDNTqhIE5MD4A==}
     engines: {node: ^20.19.0 || >=22.12.0}
     cpu: [x64]
     os: [linux]
+    libc: [musl]
 
   '@rolldown/binding-openharmony-arm64@1.2.3':
     resolution: {integrity: sha512-Kr0OcsoQI816i6HOl3vFHpd1K0eZyh76zgfj4c1nTyaTsd5r2Mj1lwM4R90y/qaCfmTn9eHy0SKwi98eitRxug==}
@@ -1357,24 +1441,28 @@ packages:
     engines: {node: '>= 12.0.0'}
     cpu: [arm64]
     os: [linux]
+    libc: [glibc]
 
   lightningcss-linux-arm64-musl@1.33.0:
     resolution: {integrity: sha512-yiO5ROMuYQgXbC60yjZU5CYSFZGKXL0HFATXt9mHJn1+zW55oCtMI9NfcVhYLMFDL7gV7oBPon/EmMMGg2OvtQ==}
     engines: {node: '>= 12.0.0'}
     cpu: [arm64]
     os: [linux]
+    libc: [musl]
 
   lightningcss-linux-x64-gnu@1.33.0:
     resolution: {integrity: sha512-ar+Ju7LmcN0Jo4FpL4hpFybwNG9/3A/Br5KW2n2jyODg3MEZXaDYADdemoNS+BDNfMgKvylJLj4S5tyRActuAg==}
     engines: {node: '>= 12.0.0'}
     cpu: [x64]
     os: [linux]
+    libc: [glibc]
 
   lightningcss-linux-x64-musl@1.33.0:
     resolution: {integrity: sha512-RYiYbkokw0trfKqqzfF55lginwEPrD3OJDfTuJzFs1MK6iFnDenaz1fqLLtX4ITG3OktJQXOeTaw1awrBAlZPw==}
     engines: {node: '>= 12.0.0'}
     cpu: [x64]
     os: [linux]
+    libc: [musl]
 
   lightningcss-win32-arm64-msvc@1.33.0:
     resolution: {integrity: sha512-1K+MPfLSFVpphzpdbfkhlWk6wBrTObBzS2T6db10PNOZgR9GoVsAWzwNyuhUYYbTp23j+4RrncfujZ4uAzXvwA==}
@@ -1526,6 +1614,10 @@ packages:
 
   pathe@2.0.3:
     resolution: {integrity: sha512-WUjGcAqP1gQacoQe+OBJsFA7Ld4DyXuUIjZ5cc75cLHvJ7dtNsTugphxIADwspS+AraAUePCKrSVtPLFj/F88w==}
+
+  pdfjs-dist@5.3.93:
+    resolution: {integrity: sha512-w3fQKVL1oGn8FRyx5JUG5tnbblggDqyx2XzA5brsJ5hSuS+I0NdnJANhmeWKLjotdbPQucLBug5t0MeWr0AAdg==}
+    engines: {node: '>=20.16.0 || >=22.3.0'}
 
   picocolors@1.1.1:
     resolution: {integrity: sha512-xceH2snhtb5M9liqDsmEw56le376mTZkEX/jEb/RxNFyegNul7eNslCXP9FDj/Lcu0X8KEyMceP2ntpaHrDEVA==}
@@ -2246,6 +2338,54 @@ snapshots:
       eslint-plugin-n: 17.10.3(eslint@10.2.1)
       eslint-plugin-react: 7.37.3(eslint@10.2.1)
       eslint-plugin-security: 1.4.0
+
+  '@napi-rs/canvas-android-arm64@0.1.100':
+    optional: true
+
+  '@napi-rs/canvas-darwin-arm64@0.1.100':
+    optional: true
+
+  '@napi-rs/canvas-darwin-x64@0.1.100':
+    optional: true
+
+  '@napi-rs/canvas-linux-arm-gnueabihf@0.1.100':
+    optional: true
+
+  '@napi-rs/canvas-linux-arm64-gnu@0.1.100':
+    optional: true
+
+  '@napi-rs/canvas-linux-arm64-musl@0.1.100':
+    optional: true
+
+  '@napi-rs/canvas-linux-riscv64-gnu@0.1.100':
+    optional: true
+
+  '@napi-rs/canvas-linux-x64-gnu@0.1.100':
+    optional: true
+
+  '@napi-rs/canvas-linux-x64-musl@0.1.100':
+    optional: true
+
+  '@napi-rs/canvas-win32-arm64-msvc@0.1.100':
+    optional: true
+
+  '@napi-rs/canvas-win32-x64-msvc@0.1.100':
+    optional: true
+
+  '@napi-rs/canvas@0.1.100':
+    optionalDependencies:
+      '@napi-rs/canvas-android-arm64': 0.1.100
+      '@napi-rs/canvas-darwin-arm64': 0.1.100
+      '@napi-rs/canvas-darwin-x64': 0.1.100
+      '@napi-rs/canvas-linux-arm-gnueabihf': 0.1.100
+      '@napi-rs/canvas-linux-arm64-gnu': 0.1.100
+      '@napi-rs/canvas-linux-arm64-musl': 0.1.100
+      '@napi-rs/canvas-linux-riscv64-gnu': 0.1.100
+      '@napi-rs/canvas-linux-x64-gnu': 0.1.100
+      '@napi-rs/canvas-linux-x64-musl': 0.1.100
+      '@napi-rs/canvas-win32-arm64-msvc': 0.1.100
+      '@napi-rs/canvas-win32-x64-msvc': 0.1.100
+    optional: true
 
   '@opentelemetry/api@1.9.1': {}
 
@@ -3609,6 +3749,10 @@ snapshots:
   path-parse@1.0.7: {}
 
   pathe@2.0.3: {}
+
+  pdfjs-dist@5.3.93:
+    optionalDependencies:
+      '@napi-rs/canvas': 0.1.100
 
   picocolors@1.1.1: {}
 
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -140,9 +140,29 @@ export const VISION_PDF_PROVIDER_IDS = [
 ] as const;
 
 /** Version suffix for the vision-path cache key. Bump when the
- *  rasterizer (DPI, image format, pdfjs version) changes. */
-export const VISION_PDF_VERSION = 1;
+ *  rasterizer (DPI, image format, pdfjs version, chunking strategy) changes. */
+export const VISION_PDF_VERSION = 2;
 export const VISION_PDF_DPI = 150;
+
+/**
+ * Max PDF pages sent to the LLM in one vision call.
+ *
+ * Empirically calibrated for a 32K-context local model (bonsai-27b Q1_0
+ * running through Ollama /v1/chat/completions). At 150 DPI each page
+ * rasterizes to ~2-3K image tokens; 6 pages ≈ 15K image tokens + ~700
+ * prompt overhead + 8K output reserve = ~24K, comfortably under 32K.
+ *
+ * Larger-context models (Claude 200K, GPT-4o 128K) trivially fit more
+ * pages per call but chunking never hurts — the bottleneck is the
+ * model's effective attention across many images, not raw context
+ * budget. Keeping a small constant here makes the worst-case LLM-call
+ * latency bounded regardless of which provider the user picks.
+ *
+ * Documents with more pages are split into chunks of this size;
+ * each chunk is sent as a separate LLM call and the markdown outputs
+ * are concatenated in page order.
+ */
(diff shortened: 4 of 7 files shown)


luxezreal:
## 主动说明 — 几处偏离 PR 描述的决定

下面这些是我**有意没改**的项,提前沟通避免来回:

1. **i18n `src/texts/*.ts` 没改**
   维护者的 `UnsupportedProviderError` 错误消息在 `src/core/pdf-converter.ts` 是单一英文硬编码字符串(已在 PR 中更新)。
   现存的 i18n key `rejectionReasonPdfUnsupported` / `sourceRejectedPdfUnsupported` 是 provider-agnostic 短标签,不列举 provider 名,在那加 "ollama, lmstudio" 反而 off-pattern。
   如果维护者坚持要在 i18n 模板里出现 provider 名,我可以补一个 PR,或在这次 PR 里加 — 等评审指示。

2. **CHANGELOG.md 没改**
   当前最新条目是 `## [1.27.0] - 2026-08-27` 完整 release block,未见 `[Unreleased]` 约定,惯例应是维护者 release 时写。
   PR_DESCRIPTION 里 checklist 的 `package.json` 版本号 bump 到 `1.28.0` 显式标注为 **maintainer step**,所以这里不动。

3. **本地没跑 `pnpm gate:1`**
   用户主动选择不在本地跑构建(校园网络拉 npm registry 不稳定),依赖 CI 验证。
   触发后如有 typecheck / lint 失败我会马上跟进。

## 测试建议
- 评审可以本地拉 `feat/pdf-vision-local-providers` 分支,跑 `pnpm install && pnpm gate:1`(按 PR_DESCRIPTION.md 的 Testing §1 描述)
- 手动 e2e: `provider: ollama` + vision GGUF(`llama3.2-vision:11b` + mmproj),丢 1 页 PDF,确认 wiki 页生成

## 配套工作
- 独立 utility `luxezreal/pdf-to-obsidian-ollama` 是这套能力的 Python CLI 桥接,纯本地离线,作为 ollama 用户当前的实际可用路径
- 在插件 PR 合并前,Ollama 用户可以用 utility 解 PDF → 喂 Obsidian vault

PR 已可合 (`mergeable: MERGEABLE`),但请先 review。如有改动需求我会在这条 review thread 直接 push 到 fork (`feat/pdf-vision-local-providers`)。

luxezreal:
## Update — fix runtime crash: `No 'GlobalWorkerOptions.workerSrc' specified`

Follow-up to comment #5453789051. The previous patch set `workerSrc = ''`, which pdfjs v5's getter treats as falsy and throws:

```js
static get workerSrc() {
  if (GlobalWorkerOptions.workerSrc) return GlobalWorkerOptions.workerSrc;
  throw new Error('No "GlobalWorkerOptions.workerSrc" specified.');
}
```

v5 has no public `disableWorker` option either — the auto-synchronous legacy path only triggers inside Node.js (`isNodeJS` check sets `_isWorkerDisabled = true`). Obsidian's renderer is Chromium, so the previous build crashed on `getDocument`.

### Fix (commit 8d4c482)

Read `pdfjs-dist/legacy/build/pdf.worker.mjs` at build time and inline it into main.js via esbuild's `define` map. At runtime, wrap the constant in a `Blob` and hand pdfjs the resulting `blob:` URL as if it were a remote worker module — pdfjs's standard `new Worker(src, { type: 'module' })` path then runs the worker in a real thread. `esbuild.config.mjs` resolves the worker path through `createRequire` so pnpm's `.pnpm/<pkg>@<ver>/node_modules/<pkg>/` symlinks still resolve.

### Files changed in this commit
- `esbuild.config.mjs` — read worker source via `createRequire`, expose as `__PDFJS_WORKER_SOURCE__` define constant
- `src/core/pdf-converter.ts` — wrap constant in `Blob`, set as `workerSrc`. Removed bogus `disableWorker: true` option (v5 doesn't read it)
- `src/types/esbuild-globals.d.ts` — TS declaration so the injected constant compiles without `@ts-expect-error`

### Trade-off
`main.js` grows ~2 MB (worker source inlined). Cold-load is one extra second; the alternative (separate `pdf.worker.mjs` shipped alongside main.js) needs URL resolution tricks that are brittle in the Obsidian renderer under CJS bundle (`import.meta.url` unavailable, `document.baseURI` varies between `app://` and `file://`).

### Verified
Local build installed into the user's vault (`karpathywiki` v1.28.0-local-vision) — `getDocument` no longer throws; pdfjs spawns the worker thread and parses the first page successfully. PR description text is now accurate; the previous "synchronous, no Worker" claim is updated to "spawns the worker from an inlined blob URL".

luxezreal:
## Update — fix Ollama context overflow (commit f716e8a)

The worker fix is confirmed: the next error was Ollama's context guard, not pdfjs:
`request (75681 tokens) exceeds the available context size (32768 tokens)`.

At 150 DPI, a 30-page paper produced ~75K image tokens in one multimodal request. The converter now rasterizes as before but sends at most **6 pages per LLM request**, then concatenates responses in page order. Six pages is calibrated for a 32K context model: roughly 15K image tokens + prompt + 8K output reserve stays below the limit. The cache-key vision version was bumped from 1 to 2 because the conversion strategy changed.

Changed:
- `src/constants.ts`: added `VISION_PDF_PAGES_PER_CHUNK = 6`, bumped `VISION_PDF_VERSION = 2`
- `src/core/pdf-converter.ts`: sequential per-chunk calls, page-range instructions, ordered markdown concatenation
- also made the OffscreenCanvas context cast explicit for TypeScript's DOM type mismatch

Build completed and the new main.js is installed in the local vault. Reload the plugin before retrying Ingest. A 30-page PDF will now make 5 Ollama calls instead of one oversized call.

green-dalii:
Thanks for this PR, @luxezreal — the description is thorough, the Reviewer FAQ is comprehensive, and the risk section is written honestly (150 DPI fidelity, 200-page memory peak, fragile Chinese OCR all called out). That's a high bar for community PRs. After careful review we've decided not to merge this and not to schedule it for future work (including MINOR). Here's the full reasoning.

**1. The technical premise is partly true — and the part that's true is the narrowest one.**

We verified the core premise: AI SDK's `@ai-sdk/openai-compatible` provider does produce a `{type:"file", file:{file_data}}` PDF part (see `dist/index.mjs`), but Ollama's `/v1/chat/completions` accepts only `text` and `image` parts, and LM Studio explicitly rejects a non-image `image_url` block. So "ollama/lmstudio cannot consume a PDF file part" is real, and per-page PNG is the only path available under those local servers' capability limits. The technical investigation is sound.

**2. But "feasible" doesn't mean "worth it" — three structural reasons.**

  - **Architectural mismatch.** The pipeline is essentially "screenshot the document, let a general-purpose vision model read it." OCR is a highly specialized task (layout recovery, two-column flow, table structure, formulas, CJK typography), and general vision LLMs are a weak substitute. By contrast the plugin's two existing paths use purpose-built systems: native providers (Anthropic/OpenAI/Bedrock) run server-side document parsers with industry-top quality; MinerU (built into v1.27.0) is a dedicated OCR engine tuned for academic PDFs, with formula→LaTeX and table/layout recovery. This path has the highest cost (per-page PNGs + N image parts in tokens and latency) yet the weakest quality (the PR itself acknowledges "Chinese OCR fragile on Q1_0 quant"). For wiki ingest — where quality drives all downstream analysis — it's a losing tradeoff.

  - **Dependency cost has no positive ROI.** `pdfjs-dist` is ~1.3 MB minified; current production bundle is 3.86 MB, so ~+34% before the esbuild config is changed to actually externalize the dynamic import (currently it would be bundled). On top of that: OffscreenCanvas compatibility, DPI/version cache-key contract, 200-page ≈ 120 MB memory peak (no chunking per the PR's own risk section), and Obsidian Bot compliance review. That's a permanent maintenance surface in exchange for an unvalidated niche scenario. This project is focused on the LLM-wiki engine itself; that complexity adds nothing to the core value.

  - **Better local paths already exist.** The project's `docs/PDF-OCR-GUIDE.md` recommends oMLX + Markitdown + a dedicated OCR model (Baidu Unlimited-OCR / GLM-OCR / Qwen3-VL) for Apple Silicon — the whole pipeline stays on the machine and OCR quality is far above a general vision model. On non-M-series hardware the guide also lists llama.cpp / vLLM + dedicated OCR models (`deepseek-ocr-2`, `glm-ocr`). That's the correct posture: "use a specialized tool for a specialized job."

**3. Tangential find: PDF-OCR-GUIDE Fallback section has a factual error.**

We noticed that `docs/PDF-OCR-GUIDE.md`'s "Fallback: any local multimodal LLM" section claimed Ollama/LM Studio can accept PDF file parts via Force PDF Support — directly contradicting the facts above. That's a doc bug we fixed independently (commit `2acde24` on `docs/fix-pdf-ocr-guide-fallback`); it's unrelated to your PR.

**4. Conclusion.** This is a well-written PR but in the wrong direction — a +34% dependency for a feature whose quality ceiling sits below existing free paths. We're declining the merge and not scheduling it for future work. If a genuine local-conversion need surfaces later (e.g. a real cohort of users who need offline PDF ingest and reject MinerU/oMLX), we're open to revisiting — the right shape then would likely be a platform-level image-ingest abstraction or direct integration with a local OCR engine, not a PDF-specific vision bypass. Your PR is preserved here as a reference implementation for that discussion.

---

感谢这份 PR,@luxezreal — 描述详尽、Reviewer FAQ 完备、风险区写得诚实(150 DPI 保真、200 页内存峰值、中文 OCR 脆弱性都点出来了),在社区 PR 里属于相当高的水准。我们仔细评审后决定不合并此 PR,也不列入未来工作(含 MINOR)。以下是完整理由。

**1. 技术前提部分成立,但成立的部分恰好是最窄的。**

我们核实了核心前提:AI SDK 的 `@ai-sdk/openai-compatible` provider 确实会产出 `{type:"file", file:{file_data}}` 的 PDF part(`dist/index.mjs` 可证),但 Ollama 的 `/v1/chat/completions` 只接受 `text` 和 `image` 两种 part,LM Studio 也明确拒绝非图片的 `image_url` 块。所以"ollama/lmstudio 收不了 PDF file part"是事实存在的,逐页 PNG 是本地服务器能力限制下的唯一可行路径。这部分技术调查是扎实的。

**2. 但"可行"不等于"值得" — 三个结构性理由。**

  - **架构错位。** 这条管线本质是"把文档截图,让通用视觉模型读图"。OCR 是高度专业化的任务(版式还原、双栏、表格结构、公式、CJK 排版),通用视觉 LLM 是弱解。对比插件现有的两条路径:原生 Provider(Anthropic/OpenAI/Bedrock)跑服务端专用文档解析器,质量行业顶级;MinerU(v1.27.0 内置)是专用 OCR 引擎,学术 PDF 优化、公式转 LaTeX、表格/版式还原。这条路径在三者中成本最高(逐页 PNG + N 个 image part 的 token/延迟)却质量最弱(PR 自认 "Chinese OCR fragile on Q1_0 quant")。对 wiki 摄取这种"质量直接决定下游分析"的场景,得不偿失。

  - **依赖成本对 ROI 无正面贡献。** `pdfjs-dist` 约 1.3MB minified,当前产包 3.86MB,即约 +34%;还需改 esbuild 配置才能真正 externalize 动态 import(目前会打进包)。加上 OffscreenCanvas 兼容、缓存键 DPI/版本契约、200 页约 120MB 内存峰值(PR 自曝无分块)、Obsidian Bot 合规审查,这是一块永久维护面,换取的是未证实的细分场景。本项目聚焦 LLM-wiki 引擎本身,此复杂度对核心价值没有提升。

  - **已有更成熟的本地路径。** 项目内置 `docs/PDF-OCR-GUIDE.md`,其中本地 Apple Silicon 推荐 oMLX + Markitdown + 专用 OCR 模型(Baidu Unlimited-OCR / GLM-OCR / Qwen3-VL),整条管线不出机器且 OCR 质量远高于通用视觉模型;非 M 系机器也列出了 llama.cpp / vLLM + 专用 OCR 模型(`deepseek-ocr-2`、`glm-ocr`)。这是"用专用工具做专用活"的正确姿势。

**3. 附带发现:PDF-OCR-GUIDE 的 Fallback 节有事实错误。**

评估中我们注意到 `docs/PDF-OCR-GUIDE.md` 的 "Fallback: any local multimodal LLM" 一节声称 Ollama/LM Studio 可通过 Force PDF Support 接收 PDF file part,与上述事实直接矛盾。这是文档 bug,我们已独立修复(分支 `docs/fix-pdf-ocr-guide-fallback` 上的 commit `2acde24`),与你的 PR 无关。

**4. 结论。** 这是一个写得很好的 PR,但方向不对 — 用 +34% 体积的依赖去实现一个质量天花板低于现有免费路径的功能。我们决定不合并,也不列入未来工作。如果未来出现真实的本地化转换需求(比如一批用户确实需要离线摄取 PDF 且拒绝 MinerU/oMLX),我们愿意重新讨论,届时正确形态可能是平台级的 image ingest 抽象或直接接入本地 OCR 引擎,而非专为 PDF 引入 vision 旁路。你的 PR 会保留在这里,作为该讨论的参考实现。

green-dalii (dismissed):
The rasterization itself looks careful — `OffscreenCanvas` rather than a native module, `bundle: true` + `format: 'cjs'` + no `splitting` so pdf.js and its inlined worker stay inside `main.js`, and an exact pin on `pdfjs-dist`. One thing blocks, and it is the trust boundary this file already documents.

**Blocking — the provider ID decides capability, and it overrides what the user declared.**

`providerSupportsPdf` (`pdf-converter.ts:278-283`) now reads:

```ts
if (NATIVE_PDF_PROVIDER_IDS.includes(settings.provider)) return true;
if (VISION_PDF_PROVIDER_IDS.includes(settings.provider)) return true;   // new
if (settings.forcePdfSupport === true) return true;
```

The comment directly above it states the rule this breaks: *"A pre-flight whitelist reject violates user intent (the user explicitly opted in). The trust boundary is: the user said 'try it' — so we try, and surface the truth if it fails."* And `forcePdfSupport` is rendered **only for non-native providers** (`advanced-section.ts:35`), so ollama and lmstudio users are exactly the people who hold that toggle — and their `false` is now unreachable, because the new line returns first.

Two concrete consequences:

1. A user who deliberately left `forcePdfSupport` off, because they know their local model takes text only, gets image parts sent anyway. That is the pre-flight reject the comment forbids, arrived at from the other direction.
2. Provider ID is not modality. `ollama` serves `llama3` as readily as a vision model, and nothing here consults the model. For a text-only local model this turns a working PDF ingest into a failing call.

`forcePdfSupport` already exists as the user's declaration of this exact fact. Either route the vision path through it (`forcePdfSupport === true && VISION_PDF_PROVIDER_IDS.includes(...)`), or add a modality check against the provider's model catalog — but a provider literal should not be the last word on what a model accepts.

**Three further items, in the order I would fix them:**

- `pdf-converter.ts:388` — `new OffscreenCanvas(width, height)` and `canvas.getContext('2d')!`. No capability check and a non-null assertion, so a renderer without `OffscreenCanvas` fails as an opaque type error. A up-front check with a readable message costs a few lines and saves a bug report.
- **Gate 3 / Gate 4.** `VISION_PDF_PAGES_PER_CHUNK = 1` at `VISION_PDF_DPI = 150` means a 200-page PDF becomes 200 calls carrying image payloads. That may well be forced by the 4096-ctx ceiling your last commit cites, but the PR body should carry the measurement (pages → calls → bytes) so the cost is a number a reviewer can weigh, not a surprise a user discovers. And since `ollama`/`lmstudio` users get this **without opting in**, it is a default-behaviour change, not just a new path.
- **CI has not run** on this PR (no checks reported). Given it touches `esbuild.config.mjs`, a green Five-Gate on the rebased head is a precondition for merge, not a formality.

**One question, not a request:** pdf.js is Apache-2.0 and is now bundled into the shipped `main.js`, whose NOTICE obligation is an attribution. The repo's NOTICE currently lists contributors only — no third-party section at all — so this is a pre-existing project-wide gap rather than something you introduced. Flagging it because vendoring a full PDF engine is the point at which it starts to be worth closing.

Happy to re-review as soon as the capability question is settled — the rendering half needs no changes from what I can see.

green-dalii:
A correction, so the notification you received is not left dangling: **I dismissed the `CHANGES_REQUESTED` review I just posted.**

I submitted it without reading this thread or the label first. That was wrong twice over — it re-reviewed a PR whose direction was already decided, and a `CHANGES_REQUESTED` reads as "fix these and we'll merge", which is the opposite of what @green-dalii's decision above says. Asking you to re-plumb the capability gate on a PR that is not being merged would have sent you to work nobody asked for.

The decision stands exactly as written: declined, `wontfix`, preserved as a reference implementation. Nothing in my review changes it, and the numbers I raised (bundle size, per-page cost) are the same ones already measured in the decision above. Apologies for the noise.

green-dalii:
Closing this, two weeks after the decision above.

Reason it was still open: an open pull request asserts that merging is possible, which is the one thing a decline denies. Three signals said otherwise — the `wontfix` label, the decision comment, and draft status — and a `CHANGES_REQUESTED` review still landed on it earlier today, which is the argument for closing rather than labelling harder. A closed PR cannot collect a stray review at all.

**What would reopen it**, from the decision above: a real cohort of users who need offline PDF ingest and reject MinerU/oMLX. The shape then is a platform-level image-ingest abstraction or a direct integration with a local OCR engine — not a PDF-specific vision bypass.

Nothing is lost. The branch, the diff and this thread stay readable and linkable, which is what "preserved as a reference implementation" requires. If the need surfaces, reopen or reference it in a new issue.

PR #571, opened 2026-08-29
fix(task-policy): reject duplicate __proto__ entries (#543)

### LINKED ISSUE #543: parseTaskPolicySpec: a __proto__ entry bypasses the duplicate-task guard
Found during the PR #539 review round (probe-verified against the compiled module).

`parseTaskPolicySpec`'s new duplicate-task guard (`Object.prototype.hasOwnProperty.call(policies, task)`) has exactly one hole: the key `__proto__`.

```
parseTaskPolicySpec('__proto__=text:on,__proto__=schema:off')
// returns WITHOUT throwing; result has no own properties;
// formatTaskPolicyMap(result) === ''
```

Cause: `policies['__proto__'] = {...}` invokes the inherited accessor setter instead of defining an own property, so `hasOwnProperty.call(policies, '__proto__')` stays false forever. (`constructor` behaves correctly — assignment shadows with an own property and the second occurrence throws.) Side effect: `resolveTaskPolicy` can read truthy garbage off the mutated prototype chain.

Practical impact today is nil — step labels are string literals and nothing looks up magic names — but the invariant "a task named twice is an error" has one exception, and a hand-edited `data.json` with a `__proto__` own-key renders into the settings field and silently drops on the next edit's reparse.

## Fix (one line)

```ts
const policies: Record<string, TaskPolicy> = Object.create(null);
```
(or a seen-names `Set`). Plus one regression test asserting the throw for `'__proto__=text:on,__proto__=schema:off'`.

Pre-existing nit while touching this file: the unknown-mode error says "expected one of -, schema, json, text" while `MODE_ALIASES` also accepts `default`, `json_schema`, `json_object`, `text_prompt`.

green-dalii:
## Summary

`parseTaskPolicySpec`'s duplicate-task guard used `Object.prototype.hasOwnProperty.call(policies, task)` against a plain object literal — one key defeats it: `__proto__`. `policies['__proto__'] = ...` invokes the inherited accessor setter instead of defining an own property, so a second `__proto__=...` entry passes the guard and silently mutates the object's prototype instead of throwing.

Switching the map to a null prototype (`Object.create(null)`) makes `__proto__` a normal own key: `hasOwnProperty.call` reports it, and the "a task named twice is an error" invariant (#525 review) now holds for every key the parser accepts.

## Changes

- `src/core/task-policy.ts`: one-line map construction change (+ comment)
- `src/__tests__/core/task-policy.test.ts`: new test asserting `__proto__=text:on,__proto__=schema:off` throws

## Verification

- RED: new test fails against unfixed code (no throw — the bug)
- GREEN: all 21 tests pass after fix
- Gate 1 (lint + tsc + build + test + css-lint): all green, 3691 tests pass

Closes #543

### DIFF
--- a/src/__tests__/core/task-policy.test.ts
+++ b/src/__tests__/core/task-policy.test.ts
@@ -178,6 +178,17 @@ describe('parseTaskPolicySpec — bounded thinking', () => {
       .toThrow(/"extract" is set twice/);
   });
 
+  // #543: the duplicate-task guard relied on `Object.prototype.hasOwnProperty.call`
+  // against a plain object literal, which leaves `__proto__` as an inherited
+  // accessor — `policies['__proto__'] = ...` never becomes an own property, so
+  // a second `__proto__=...` entry passed the guard and silently mutated the
+  // object's prototype instead. The contract "a task named twice is an error"
+  // must hold for every key the parser accepts as a task name.
+  it('rejects a __proto__ entry set twice (closes the hasOwnProperty.call bypass, #543)', () => {
+    expect(() => parseTaskPolicySpec('__proto__=text:on,__proto__=schema:off'))
+      .toThrow(/"__proto__" is set twice/);
+  });
+
   it('rejects a repeated wildcard too', () => {
     expect(() => parseTaskPolicySpec('*=text,*=schema')).toThrow(/set twice/);
   });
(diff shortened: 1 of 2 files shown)


green-dalii:
Reviewed: fix #543 — parseTaskPolicySpec uses Object.create(null) so __proto__ becomes an own key; duplicate-task guard now fires for every accepted key. Gate 1 green (3739 tests), RED test confirmed against unfixed code. Accept.

PR #572, opened 2026-08-29
fix(repetition-loop): require consecutive source occurrences for echo (#542)

### LINKED ISSUE #542: isSourceBorneLoop suppresses the halve-retry for common-word degeneration (echo-guard blind spot)
PR #539 introduced `isSourceBorneLoop` (src/core/repetition-loop.ts:97) to avoid spending the halve-retry on a faithful echo. The heuristic's blind spot, surfaced by the PR's independent code-review round: **it cannot distinguish a refrain from a stopword.**

## Failure mode

The guard fires when the note contains the loop's rawUnit ≥4 times anywhere (scattered, non-consecutive). Token-level degeneration — the primary mode per the project's own calibration (`repetition-loop.ts` header: runs ≥2,280 chars; `types.ts` repetitionPenalty rationale) — loops high-frequency words. Verified end-to-end during review:

- Response loops `"und die "` ×80 against an ordinary German note containing the phrase 25× scattered → guard returns true → `"echo, not damage; no retry spent"` → damaged batch parsed and merged silently.
- The premise "the retry reproduces the same batch anyway" holds only under deterministic decoding; sampling is non-deterministic by default (#423), so pre-#539 the re-roll could genuinely differ. Net: for the most common degeneration units, the only recovery attempt #524 provided is skipped.
- Related mismatch: the detector requires 4 **consecutive** repeats (`LOOP_RE` `\1{3,}`) but the guard accepts 4 **scattered** occurrences — a glossary listing `- Eintrag: Vitamin D, …` twelve times neutralizes the retry for genuine consecutive degeneration on `"Vitamin D, "` (partially intended per the CHANGELOG's "tabulated column" case, but the threshold makes any list-heavy note immune for its vocabulary).
- Multiline/quote-containing refrains are never recognized: the response side carries JSON escapes (`\n` as two chars) while the note has real newlines — safe direction (retry still spent), but the CHANGELOG's "quoted chorus" example only works single-line.

## Suggested direction

Require the response-side repeat count to be explainable by the note — e.g., unit occurs ≥4× in the note AND the response run shows few repeats beyond what those occurrences explain, or compare scattered-note-count against response-repeat-count ratio before classifying as echo. Keep the conservative bias (skip only when clearly faithful).

## Acceptance criteria

1. Exactly-4-occurrences → true, exactly-3 → false (the boundary is currently unpinned; a `>=`→`>` mutation passes the whole suite).
2. An analyzer-layer echo test whose refrain exceeds the 40-char log-trim threshold (current fixtures use 34 chars, so a `rawUnit`→`unit` swap survives every integration test).
3. A direct test of `forcedTextPromptSystem`'s prose branch (pinned `text_prompt`, no `response_format` → NO JSON prefix) on both clients.

green-dalii:
## Summary

The pre-fix guard counted any non-overlapping occurrence of the loop unit in the source, so a stopword scattered through ordinary prose (`"und die "` ×25 in a German note) suppressed the halve-retry for the very token-level degeneration the response detector (#524) was built to catch. A loop the note carries is reproduced by every retry, but a stopword is not the note's pattern to repeat — it is how the language works.

Symmetric with the response detector's `LOOP_RE` `\1{3,}`: require ≥4 *consecutive* occurrences in the source, not ≥4 scattered mentions. The asymmetry is what let the scattered check through — closing it re-establishes the response/source symmetry at the four-repeat threshold.

## Changes

- `src/core/repetition-loop.ts`: `isSourceBorneLoop` now tests a consecutive-run regex (escaped unit, `{4,}` quantifier) instead of a scattered split count
- `src/__tests__/core/repetition-loop.test.ts`: new stopword-scattered test (the bug), boundary-pin test (exactly 4 → true / 3 → false), refrain fixture updated to back-to-back shape
- `src/__tests__/wiki/source-analyzer.test.ts`: `REFRAIN_NOTE` anchored with a consecutive run at the head, preserving the echo-path assertion

## Verification

- RED: stopword-scattered test fails against unfixed code (returns true — the bug)
- GREEN: all tests pass after fix (16 unit + integration)
- Gate 1 (lint + tsc + build + test + css-lint): all green, 3692 tests pass

Closes #542

### DIFF
--- a/src/__tests__/core/repetition-loop.test.ts
+++ b/src/__tests__/core/repetition-loop.test.ts
@@ -100,12 +100,20 @@ describe('findRepetitionLoop', () => {
 
 // #525 review: halving changes how many items are asked for, never the note,
 // so a loop the note itself carries is reproduced by every retry.
+// #542: the source check has to mirror the response detector — the response
+// detector requires 4 *consecutive* repeats (LOOP_RE `\1{3,}`), so a faithful
+// echo is also a *consecutive* pattern in the note, not 4 scattered mentions.
+// A stopword scattered through ordinary prose ("und die " ×25 in a German
+// note) is not the note's pattern to repeat; spending the halve-retry on it
+// drops the only recovery attempt #524 provided for a damaged batch.
 describe('isSourceBorneLoop', () => {
   const REFRAIN = 'Und täglich grüßt das Murmeltier. ';
 
-  it('recognises a refrain the note states at least four times', () => {
+  it('recognises a refrain the note states at least four times consecutively', () => {
     const loop = findRepetitionLoop(REFRAIN.repeat(10))!;
-    const note = (REFRAIN + 'Dazwischen anderer Text. ').repeat(6);
+    // REFRAIN×4 back-to-back, then intervening prose, then REFRAIN×2: the
+    // 4-consecutive run at the start is the refrain pattern.
+    const note = `${REFRAIN.repeat(4)} Dazwischen anderer Text. ${REFRAIN.repeat(2)}`;
     expect(isSourceBorneLoop(loop, note)).toBe(true);
   });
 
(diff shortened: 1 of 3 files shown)


green-dalii:
Reviewed: fix #542 — isSourceBorneLoop now requires ≥4 consecutive occurrences (mirrors response LOOP_RE \1{3,}), escaping the unit for literal match. Stopword-scattered prose no longer suppresses the halve-retry. Gate 1 green (3740 tests). Accept.

PR #573, opened 2026-08-29
docs(pdf): correct the Fallback claim that local servers accept PDF file parts

green-dalii:
## Summary

PDF-OCR-GUIDE (EN + CN) stated that Ollama / LM Studio accept PDF file parts via Force PDF Support. Verified against the OpenAI-compatible wire format and both servers: /v1/chat/completions carries text and image content parts only — Ollama rejects a PDF file part (400), LM Studio rejects a PDF sent as an image block. The llama.cpp / vLLM row is reworded to 'verify against your server' instead of a blanket claim.

## Changes

- `docs/PDF-OCR-GUIDE.md` — Fallback section rewritten
- `docs/PDF-OCR-GUIDE_CN.md` — Fallback section rewritten (CN parity)

The Fallback section now points local-server users to the reliable pattern: convert the PDF before the LLM call (oMLX + Markitdown, MinerU extractor, markitdown CLI, marker) and ingest the resulting Markdown as a regular text source.

## Verification

- docs-only change; no Gate 1 impact
- shazam_verify PASS

Found during the PR #570 evaluation. The PR itself is not affected — this fix is independent.

### DIFF
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -2,7 +2,7 @@
 
 > Feature planning and improvement proposals
 
-**Latest shipped:** v1.27.0 MINOR (2026-08-27). See [CHANGELOG.md §1.27.0](./CHANGELOG.md#1270---2026-08-27) for the canonical composition record. | **Updated:** 2026-08-27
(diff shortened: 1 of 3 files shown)


green-dalii (commented):
**Maintainer self-merge — procedural note (audit trail per CLAUDE.md)**

GitHub blocks self-approving a PR owned by the same account; gh pr review --approve is therefore impossible here. This COMMENT review event serves as the formal review verdict landing on the PR timeline, replacing the would-be APPROVE.

Review verdict: approve. Docs-only fix from the PR #570 evaluation; corrects the Fallback claim that Ollama/LM Studio accept PDF file parts (wire format carries text+image only). EN+CN parity. Gate 1 SUCCESS on commit 2acde24. Merge to follow.

PR #589, opened 2026-08-31
fix(dedup): route cross-folder designators through the semantic call — the folder never decides (#588)

### LINKED ISSUE #588: Same-referent pages are minted as entity/concept twins — the dedup never looks across the folder
Reported with a measurement from an 83-note ingest batch (German medical vault, v1.27.0 pipeline): **18 twin pairs** — the same page name in both `wiki/entities/` and `wiki/concepts/` (Stress, ADHS, Alzheimer, BDNF, COPD, Mitochondrien, …). Checked by hand: every pair denotes the same referent. Several pairs were born seconds apart from a single note — two items for the same thing, typed differently by the extraction, each resolved against its own folder only.

## Mechanism

`resolvePagePath` is folder-blind in both directions. The exact-slug fast path probes only the item's own folder, `ConflictResolver` matches same-type only, and the semantic dedup's candidate window is built from same-type pages — so a page carrying the designator in the opposite folder is invisible to every stage. When the extraction types the same referent as `entity` in one note and `concept` in the next, the second ingest cannot find the first page, and the twin is created without any call ever seeing it.

This is the complementary case to #472. The type-scoping that resolved #472 correctly stopped the alias-bridging (a cross-folder page must never be merged into deterministically — `KHK` the enzyme vs. `KHK` the heart disease). But the extraction's `type` is a per-call classification, not a property of the referent: measured on this batch it flips between notes often enough that folder-scoped matching mints a twin roughly every fifth note.

## What would help

The direction #472 itself suggested: a cross-folder slug/alias hit routes the resolution through the **semantic dedup call** with that page in the candidate list — the model decides identity, the folder never does, in either direction. No new call for the ordinary case. And on a confirmed cross-folder match, re-decide the classification once against the user's own Classification Rules — the first ingest's draw, possibly made on a passing mention, should not own the folder forever. PR follows.

DocTpoint:
Closes #588. Builds on the #472 resolution rather than reverting it: deterministic merges stay same-type only.

## First commit — the folder never decides identity

- `ConflictResolver` now reports opposite-folder slug/alias claims as `crossFolderCandidates` on any action — detected there, decided nowhere.
- `resolvePagePath` routes any such claim through the semantic dedup call, claimants seeded at the head of the list. A direct file probe of the opposite slug path backstops the page index for a twin written moments ago in the same run (the measured pairs include ones born seconds apart).
- The candidate window draws from both wiki folders, so near-name twins (`Brainfog` next to `Brain-Fog` in the same batch) reach the model too.
- The prompt states that the folder reflects an earlier classification and does not decide identity — and that a related, broader, narrower, or component thing is not a match.
- Conservative fallbacks: on no-decision a create stays a create (the twin is the visible failure; an undecided cross-folder merge would be the silent one), and a deterministic same-type match stays the target — a failure never crosses the folder on its own.

## Second commit — the classification is re-decided, once

The matched page's folder is the first ingest's draw, made on whatever context that call had — possibly a passing mention that produced little more than a name. Letting it stand unquestioned makes the classification a function of ingest order.

- On a cross-folder-routed match the dedup call also answers the classification question, judged against the vault's own **Classification Rules** — the `index` schema context now delivers that section to its single caller, so the user's definitions are the ruler.
- The decision is ratcheted: written back as `type_confirmed: true`, asked at most once per page, so a page cannot ping-pong between the folders on later draws.
- A page whose decided classification disagrees with its address moves via `app.fileManager.renameFile` (link-safe in Obsidian); the dev-instrument shim gains an equivalent `renameWithLinkRewrite` (file move plus inbound wiki-link rewrite, bounded by `]`/`|`/`#` so `entities/Stress` never rewrites `entities/Stress-Resilienz`).
- Where the move is impossible — target address occupied by an existing twin (healing that is a merge, not a move), or no link-safe rename on the host — the page gets `type_conflict: <classification>` in its frontmatter: the disagreement becomes a greppable artifact a lint or a later pass can surface, and `type_confirmed` stays absent so the question is re-asked once the obstacle is gone.

## Deliberately out of scope

- The same-type exact-slug fast path still returns early without consulting the opposite folder: once both twins exist, re-ingests keep merging same-type. Healing existing twins is a merge over two full pages, not a resolution decision — separate pass.
- `fix-dead-link.ts` builds its page list inline and is untouched.

Both #472 directions are covered by tests (twin case, KHK case, index-lag case, move/ratchet/conflict-marker cases). Pre-checks: tsc clean (main and tools), 3703/3703 tests, eslint clean, and one real instrument start against a live vault.

### DIFF
--- a/src/__tests__/core/conflict-resolver.test.ts
+++ b/src/__tests__/core/conflict-resolver.test.ts
@@ -155,3 +155,32 @@ describe('ConflictResolver — ambiguous designators', () => {
     expect(second.candidates).toHaveLength(2);
   });
 });
+
+describe('ConflictResolver — cross-folder claims (#472 both ways)', () => {
+  it('reports an opposite-folder title claim on a create', () => {
+    const r = new ConflictResolver(WIKI_FOLDER, pages(['concepts/Stress.md']));
+    const result = r.resolve({ name: 'Stress', slug: 'Stress', pageType: 'entity' });
+    // The claim never decides: the action stays 'create', the caller routes it.
+    expect(result.action).toBe('create');
+    expect(result.targetPath).toBe('wiki/entities/Stress.md');
+    expect(result.crossFolderCandidates?.map(p => p.path)).toEqual(['wiki/concepts/Stress.md']);
+  });
+
+  it('reports an opposite-folder alias claim alongside a same-type merge', () => {
+    const r = new ConflictResolver(WIKI_FOLDER, [
+      { path: 'wiki/concepts/Koronare-Herzkrankheit.md', title: 'Koronare-Herzkrankheit', aliases: ['KHK'] },
+      { path: 'wiki/entities/Ketohexokinase.md', title: 'Ketohexokinase', aliases: ['KHK'] },
+    ]);
+    const result = r.resolve({ name: 'KHK', slug: 'KHK', pageType: 'concept' });
+    expect(result.action).toBe('merge');
+    expect(result.targetPath).toBe('wiki/concepts/Koronare-Herzkrankheit.md');
+    expect(result.crossFolderCandidates?.map(p => p.path)).toEqual(['wiki/entities/Ketohexokinase.md']);
+  });
+
+  it('reports no cross-folder field when the opposite folder is silent', () => {
+    const r = new ConflictResolver(WIKI_FOLDER, pages(['entities/foo.md', 'concepts/bar.md']));
+    const result = r.resolve({ name: 'Foo', slug: 'foo', pageType: 'entity' });
+    expect(result.action).toBe('merge');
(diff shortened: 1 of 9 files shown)


green-dalii (approved):
Approve. #588 fix — cross-folder dedup routed through semantic call, never decided by folder. ConflictResolver reports opposite-folder slug/alias claims as crossFolderCandidates; resolvePagePath routes them to semantic dedup with both sides seeded; direct file probe backstops pages born moments ago. Conservative: no-decision keeps same-type target (failure never crosses folder on its own). No new unconditional LLM calls. Conflict with #591 resolved as interface union; Gate-1 3792 tests, tools tsc clean.

PR #591, opened 2026-08-31
fix(dev-instrument): give the shim a link cache and resolver — retargetLinksToPage was a silent no-op (#590)

### LINKED ISSUE #590: Under the dev-instrument, a duplicate merge silently orphans every inbound link — the shim never fulfilled the retarget contract
Found on a real merge run (19 duplicate pairs, healing the twins from #588): every `[done]` line reported zero retargeted links although the deleted pages had 2–17 inbound files each. A control grep afterwards found **65 dead link hits across ~50 files**; nothing in the run had said so.

## Mechanism

`retargetLinksToPage` (#386) walks `metadataCache.getFileCache(file).links/embeds` and resolves each reference through `getFirstLinkpathDest`. The instrument's shim serves **frontmatter only** from `getFileCache` and has no `getFirstLinkpathDest` at all — so the module sees zero references in every file, rewrites nothing, and reports `linksRewritten: 0` as an ordinary result. The `stale` warning cannot fire either: it guards positions that stopped matching, not references that were never seen. A silent no-op, indistinguishable from "there were no links".

This is a shim-contract gap, not a bug in #386: the module behaves correctly for the inputs it was given. But `mergeDuplicatePages` deletes the losing page right after the retarget, so under the instrument the damage is real and invisible until a dead-link scan runs.

## Fix (PR follows)

Give the shim the two missing pieces, matching the app's semantics: `getFileCache` reports body wiki-links and embeds (link without display text, original verbatim, character offsets; body-only — Obsidian keeps frontmatter links in `frontmatterLinks`, which #386 deliberately does not read), and `getFirstLinkpathDest` resolves exact path → path suffix → bare basename, shortest path first with a deterministic tie-break, aliases not consulted.

DocTpoint:
Closes #590. One file: `tools/dev-instrument/src/vault-fs.ts`.

- `linkCacheOf` parses body references on demand (like `frontmatterOf`): wiki-links and embeds with `link`/`original`/offsets, scan starting after the closing `---` so quoted frontmatter links stay out, matching the app's cache shape that #386 consumes.
- `getFirstLinkpathDest` mirrors the app's resolver: exact vault path, then path-suffix form (`entities/Foo`), then bare basename — shortest path wins, tie broken by path so the answer never depends on index order; frontmatter aliases deliberately not consulted, exactly like the app.
- Verified end-to-end against a scratch vault: 5 references across two files rewritten (plain, `|display`, `#subpath`, embed, and a bare `[[Foo]]` from a note outside the wiki), a decoy sharing the prefix left alone, a quoted frontmatter link left alone, `stale: 0`.

Pre-checks: tools tsc clean, `eslint.tools-bot.config.mjs` clean.

### DIFF
--- a/tools/dev-instrument/src/vault-fs.ts
+++ b/tools/dev-instrument/src/vault-fs.ts
@@ -306,11 +306,90 @@ export class NodeVault {
     }
     return parseFrontmatter(content);
   }
+
+  /**
+   * Body wiki-links and embeds for `metadataCache.getFileCache`, in the shape
+   * `retargetLinksToPage` consumes: `link` (destination as written, incl. any
+   * `#subpath`, excl. any `|display`), `original` (the reference verbatim,
+   * embed `!` included), and character offsets. Obsidian's cache reports body
+   * references only — frontmatter wiki-links live in `frontmatterLinks`,
+   * which that module deliberately does not read — so the scan starts after
+   * the closing `---`. Parsed on demand, like `frontmatterOf`: the instrument
+   * runs one operation and exits, a cache would outlive nothing.
+   */
+  linkCacheOf(file: TFile): {
+    links: Array<{ link: string; original: string; position: { start: { offset: number }; end: { offset: number } } }>;
+    embeds: Array<{ link: string; original: string; position: { start: { offset: number }; end: { offset: number } } }>;
+  } {
+    let content: string;
+    try {
+      content = this.readSync(file.path);
+    } catch {
+      return { links: [], embeds: [] };
+    }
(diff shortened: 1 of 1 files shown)


green-dalii (approved):
Approve. #590 fix — dev-instrument vault-fs gains linkCacheOf (body wiki-links + embeds, parsed on demand, scan starts after closing frontmatter `---`) so retargetLinksToPage is no longer a silent no-op. Mirrors the app's getFirstLinkpathDest resolution (exact path → path-suffix → basename). Tools-only file; production untouched. Gate-1 green, tools tsc clean.

Issue #81, opened 2026-06-02
YAML sources field: inconsistent link formats and duplicate entries on ingest/re-ingest


DocTpoint:
---

> **Environment**
> - karpathywiki v1.14.0
> - Vault: ~2600 notes
>
> **Problem**
> The `sources` field in YAML frontmatter is populated inconsistently during ingest, resulting in three distinct issues that accumulate over time:
>
> **1. Notizen/ references included alongside sources/**
> On ingest, the plugin writes both the original note path and the wiki sources page into the `sources` field:
> ```yaml
> sources:
>   - "[[Notizen/Glycin.md]]"       # should not be here
>   - "[[sources/Glycin.md]]"
>   - "[[sources/Glycin]]"
> ```
> Only `[[sources/Glycin]]` should appear.
>
> **2. Inconsistent link format**
> Three different formats are generated for the same source depending on the code path:
> ```yaml
> "[[sources/X]]"        # correct
> "[[sources/X.md]]"     # .md extension should be stripped
> "[[sources/X|X]]"      # pipe alias serves no purpose in YAML
> ```
>
> **3. Duplicate entries on re-ingest**
> When a sources page is deleted and re-ingested, the new entry is appended without checking for existing entries. Combined with the format variations above, a single source can appear 3–4 times.
>
> **Scale**
> After running a cleanup script across the vault: **572 affected files, 1616 individual fixes** required.
>
> **Expected behavior**
> - Only `[[sources/X]]` format in YAML (no `.md`, no pipe alias, no `Notizen/` paths)
> - De-duplication check before appending a sources entry

---

green-dalii:
@DocTpoint — confirmed. Root cause: `normalizeSourcePath` in `utils.ts:537` strips `[[ ]]` but leaves `.md`, `|alias`, and the `Notizen/` prefix untouched, so the same logical source maps to 4 different Set keys and dedup is broken. Pollution source is `wiki-engine.ts:646` replacing `{{source_file}}` with the raw vault path.

I'll open a fix PR today:
- Extract `normalizeWikiLink()` shared by utils / wiki-engine / lint-fixes
- One-shot idempotent migration script (dry-run first) for the 572 files you've already got polluted
- Tests first per the repo TDD standard

Open a draft PR for the migration script if you want — your data is the real edge case generator here.

DocTpoint:
---

[...]

PR #596, opened 2026-09-01
fix(ingest): resolve ingest ownership from source_file, not sources (#595)

### LINKED ISSUE #595: Batch ingest never skips a note whose summary page was extended by a second source
## What happens

`runBatchIngest` asks `isAlreadyIngested` whether each selected file has already been ingested. Whenever the note's summary page carries a `sources:` list, the answer is always "no", and the batch ingests the note again. A re-ingest is additive, so this does not revise the page — it adds to it.

## Why

`isAlreadyIngested` ([`ingest-commands.ts`](https://github.com/green-dalii/obsidian-llm-wiki/blob/main/src/main-commands/ingest-commands.ts#L57)) finds the summary page by slug, then confirms the page belongs to this note by looking for the note's vault path in `sources:`.

`sources:` cannot hold that. It records the multi-source merge — the *other* ingests that later extended this page — and holds `[[sources/X]]` wikilinks by contract; the Issue #81 normalizer removes or remaps any external `[[Notizen/X.md]]` entry that lands there. The origin lives in the scalar `source_file:` written by the generation template.

`scanSourceDrift` (#577) already resolves origin that way, scalar first and list as a fallback. So two readers of the same question disagree, and the one guarding the ingest reads the field that by design never names the origin.

## Measured

On a 230-page vault, every summary page carries `source_file:` and every one resolves to an existing note. 103 pages also carry a `sources:` list, and in **0 of the 103** does the page's own note path appear in it.

The affected set is not arbitrary — it is exactly the pages a second ingest has extended. Of the 127 pages without a `sources:` list, **0** have ever been rewritten after creation; of the 103 with one, **49** carry a later `updated:` (the rest were extended the same day, and the stamp only has day resolution). One of them states it in the body: `Erweitert durch: [[sources/Magnesium]]`.

So the set grows with the vault, and it singles out the pages that accumulated the most content.

## The other half

When `sources:` is absent the function returns `true` without checking anything. Two notes in different folders that slugify to the same page path therefore both read as ingested — the second is skipped and never ingested at all. This is presumably the case the ownership check was meant to catch.

## Scope

`runBatchIngest` is the only caller, so this affects the multi-file picker and folder ingest. Single-file ingest and the dev-instrument CLI never consult it.

## What I did not measure

I did not run a batch ingest to observe a duplicate. The evidence is the code path, the frontmatter counts above, and a unit test that drives `isAlreadyIngested` with a real page's frontmatter and reproduces both failures.

A PR follows.

DocTpoint:
Fixes #595.

## The problem

`isAlreadyIngested` locates a note's summary page by slug and then confirms the page belongs to that note by looking for the note's vault path in `sources:`.

`sources:` records the multi-source merge — the other ingests that later extended the page — and holds `[[sources/X]]` wikilinks by contract, since the Issue #81 normalizer removes or remaps any external `[[Notizen/X.md]]` entry that lands there. The path being searched for is the one shape the field cannot hold, so the comparison could not succeed. The origin lives in the scalar `source_file:`.

## Measured

On a 230-page vault:

| | pages |
|---|---|
| summary pages | 230 |
| carry `source_file:`, resolving to an existing note | 230 |
| also carry a `sources:` list | 103 |
| of those, own note path present in `sources:` | **0** |

The affected set is exactly the pages a second ingest has extended: of the 127 pages without a `sources:` list, 0 have ever been rewritten after creation; of the 103 with one, 49 carry a later `updated:`. It grows with the vault and singles out the pages holding the most content — and a re-ingest is additive rather than revising.

## The fix

`originNoteRefs` (in `core/frontmatter.ts`) reads the canonical scalar first and falls back to the list — the resolution `scanSourceDrift` (#577) already uses. `isAlreadyIngested` now answers from it.

Behaviour is unchanged for pages with no recorded origin: they keep the existing existence-based answer, so pre-#164 pages are not suddenly re-ingested.

The fix also closes the inverse case the old `return true` fallback hid: two notes in different folders that slugify to one page path both read as ingested, so the second was skipped and never ingested at all.

## Tests

New `src/__tests__/main-commands/ingest-skip.test.ts`, four cases driving `isAlreadyIngested` against a real page's frontmatter:

- summary page with a `sources:` list → skip (failed before: reported "not ingested")
- summary page without one → skip
- no summary page → ingest
- same-slug note from another folder → ingest (failed before: reported "ingested")

Two of the four fail on `main`. Full suite 3745/3745, lint and typecheck clean, `build` + `css-lint` green.

## On the second commit

`refactor(lint): read origin notes through the shared resolver` points `scanSourceDrift` at the same helper. It is deliberately a separate commit: the duplication is what let the two readers drift apart in the first place, but it touches code you merged recently, so drop the commit if you would rather leave the scanner alone. The fix does not depend on it.

## Not included

I did not run a batch ingest to observe a duplicate in the wild — the evidence is the code path, the frontmatter counts, and the failing tests.

### DIFF
--- a/src/__tests__/main-commands/ingest-skip.test.ts
+++ b/src/__tests__/main-commands/ingest-skip.test.ts
@@ -1,0 +1,94 @@
+// The batch-ingest skip check must answer one question: has THIS note
+// already been ingested?
+//
+// `isAlreadyIngested` locates the summary page by slug and then verifies
+// ownership by looking for the note's path in the page's frontmatter. It
+// reads `sources:` for that path. But `sources:` is a list of
+// `[[sources/X]]` wikilinks by contract — the Issue #81 normalizer removes
+// or remaps any `[[Notizen/X.md]]` entry it finds there. The path being
+// searched for is therefore the one shape the field can never hold, and
+// every summary page that carries a `sources:` list reads as "not
+// ingested".
+//
+// The canonical owner is the scalar `source_file:` written by the
+// generation template. `scanSourceDrift` (#577) already resolves ownership
+// that way: scalar first, list as a fallback.
+
+import { describe, it, expect } from 'vitest';
+import { TFile } from 'obsidian';
+import { ingestCommands, type IngestHost } from '../../main-commands/ingest-commands';
+
+const NOTE_PATH = 'Notizen/ATP-Produktion.md';
+
+function note(path = NOTE_PATH): TFile {
+  const basename = path.split('/').pop()!.replace(/\.md$/, '');
+  return Object.assign(new TFile(), { path, basename, extension: 'md' });
+}
+
+/** Minimal host: `isAlreadyIngested` only touches the vault and two settings. */
+function host(pages: Record<string, string>): IngestHost {
+  return {
+    app: {
+      vault: {
+        getAbstractFileByPath: (p: string) =>
+          p in pages ? Object.assign(new TFile(), { path: p }) : null,
+        read: async (f: { path: string }) => pages[f.path],
+      },
+    },
+    settings: { wikiFolder: 'wiki', slugCase: 'preserve' },
+  } as unknown as IngestHost;
+}
+
+const call = (h: IngestHost, f: TFile) =>
+  ingestCommands.isAlreadyIngested.call(h, f);
+
+// A real page from the vault. The canonical scalar names the origin note.
+// `sources:` names the OTHER notes whose later ingest extended this page —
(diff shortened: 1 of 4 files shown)


green-dalii:
Audit note: this PR was merged without a preceding formal `gh pr review --approve` event (procedural miss, not content miss — code review passed: resolveContradictionTarget intercepts write path, originNoteRefs unified, Gate-1 green). Per AGENTS.md §Mandatory merge sequence, the approve event is a separate audit surface from the merge. Recording here so the audit trail is complete.

PR #600, opened 2026-09-01
feat(picker): show what the vault knows, not what the session remembers (#598)

### LINKED ISSUE #598: The file picker shows session memory, not what is on disk
## What happens

`MultiFileSuggestModal` decides how every row looks from `ingestQueue.getSnapshot()` alone. The queue is session state — it is built in memory and empty again after a plugin reload. So on a fresh Obsidian start every note reads as never ingested, even when its summary page has been on disk for weeks. The picker is the one place where the user decides what to ingest, and it is the one place that does not consult the result of the last ingest.

`isAlreadyIngested` does consult it, but only inside `runBatchIngest` — after the click. The user learns what the plugin already knew from a notice telling them how many of their selected files were skipped.

## The checkbox carries two meanings, and the second one wins

`updateLeftPaneSelections` runs on every queue mutation and assigns `checkbox.checked = isQueued` for every row. The checkbox is therefore not the user's selection — it is a rendering of queue membership.

The consequence is that ticking boxes while a batch is running is unstable: each `start` or `complete` transition notifies the subscriber, which resets every checkbox the queue does not know about. A selection in progress is silently discarded.

## The store already knows more than the picker asks

`IngestQueue.isIngested()` exists and is tested, and its `completedPaths` set is documented as being there "so the modal can grey out already-ingested files even after their jobs have been removed". No production code calls it. The picker rebuilds that set from the job list instead, and so loses the one property `completedPaths` was introduced to provide.

## What the picker could show

Three row states, resolved once when the modal opens:

- **never ingested** — normal row
- **ingested, note unchanged** — greyed out
- **ingested, note edited since** — marked, via the stored `contentHash` against the note's current body, the same comparison `scanSourceDrift` (#577) makes

and the checkbox decoupled from all three, so it means only "I selected this".

## Deliberately not included

No re-ingest action on the changed rows. A re-ingest is additive rather than revising (#220 Tier 1/2), so offering a button there would invite exactly the operation that damages the page. Showing that a note has drifted is useful on its own; acting on it is a separate decision and a separate issue.

## Side findings, same surface

- A failed job is never reconciled. `enqueue` adds a new job for a path in a terminal state, so after a successful retry the queue holds both the old ❌ row and the new ✅ row, and the counter keeps reporting the failure.
- An orphaned pending job cannot be restarted from the UI. `enqueue` skips paths already `pending` or `running` and returns no ids, and the picker's confirm handler only calls `onStartIngest` when it received ids. If the batch loop stopped while jobs were still pending, those rows cannot be picked up again — only cancelled.

## What I did not measure

I read this off the code rather than reproducing it in a running vault. The checkbox reset in particular is derived from the assignment in `updateLeftPaneSelections`, not observed.

## Depends on

#595 / #596. The disk-side answer the picker would display comes from `isAlreadyIngested`, which today reports "not ingested" for every page a second source has extended.

DocTpoint:
Closes #598. Stacked on #596 — the disk answer this displays comes from
`isAlreadyIngested`, which without that fix reports "not ingested" for
every page a second source has extended. Review the top two commits; the
first two are #596.

## The picker described session state

`MultiFileSuggestModal` decided how every row looked from
`ingestQueue.getSnapshot()` alone. The queue is built in memory and is
empty again after a plugin reload, so on a fresh Obsidian start every
note read as never ingested even when its summary page had been on disk
for weeks. The one place the user chooses what to ingest was the one
place that did not consult the result of the last ingest — that arrived
afterwards, as a notice counting how many of their files had been
skipped.

## Three row states, resolved when the modal opens

No page under this note's slug; a page built from this note; or that,
with the note edited since. The third is the `contentHash` comparison
`scanSourceDrift` (#577) makes.

Drift is answered conservatively, because the per-note question is not
always decidable. A page carries ONE hash, belonging to whichever note
wrote it last, so on a multi-source page a mismatch cannot be told apart
from "a different origin wrote it last". Marking drift there would flag
most origins of every merged page, and a marker that fires on pages
nobody touched is worse than no marker. Those pages report "ingested".

Measured on a 259-page vault: 134 pages have one recorded origin and are
decidable, 125 have more; all 259 carry a hash. Three of the 134 have
drifted — three notes edited since their page was written, and nothing
in the UI said so.

## Settled rows carry a greyed tick

A row that is ingested and unchanged is settled, whether the job ran in
this session or weeks ago: greyed tick, locked checkbox — the way a
completed job's row always looked. The skip check would decline it
anyway, and "I ticked it and nothing happened" is worse feedback than a
locked row.

A drifted row is deliberately left live and unticked. It is the one case
where picking the note again is an intent rather than a mistake, because
the page was built from a body that no longer exists. No re-ingest button
is offered there — a re-ingest is additive rather than revising (#220
Tier 1/2), so a button would invite the operation that damages the page
— but the row can still be selected by hand.

## The checkbox is the user's selection again

`updateLeftPaneSelections` assigned `checkbox.checked = isQueued` on
every queue mutation, so the checkbox rendered queue membership. A batch
running in the background reset every tick made while it ran. The
selection now lives in a set on the modal, which also survives the
search filter — that rebuilds the left pane, so a DOM-only selection
lost every mark outside the current query. Queue membership keeps its
signal as a row class.

Two adjacent consequences of the same assignment: the bulk "Select all"
and a click on the row body both set `.checked` directly, which bypasses
`disabled` and fires no `change` event, so either could mark a locked row
the user could not then unmark. Both now skip disabled rows.

## Structure

`pageBelongsToNote` is the ownership decision moved out of
`isAlreadyIngested` verbatim, so the picker cannot report one thing and
the ingest do another — a second implementation of that question is how
#595 happened. It stays in core/ so the direction is fixed as
note -> slug -> page; indexing `sources/` and inverting it reads more
naturally and is wrong, because an origin-less page cannot be attributed
to any note that way and would disagree with the skip check on exactly
the pages where the existence fallback applies.

`resolveRowState` is extracted for the reason
`schema-diff-modal-classes.ts` was: the Modal base class cannot be
instantiated outside Obsidian, so anything decided inside it is verified
by reading only — which is how the checkbox assignment survived. 27 new
tests, 21 of them on behaviour that had none.

`llm-wiki-multi-file-ingested-tag` needed no new CSS. It was already in
styles.css and the old comment called it "the visual 'Ingested' tag",
but no code ever rendered it — the greyed row had always been the only
cue. This is the element that rule was written for.

## Cost

Bounded by the number of `sources/` pages, not by the vault: a candidate
whose page is absent costs one path lookup and no read, pages are read
once even when several notes slug-collide, and the scan runs after the
first paint so the pane never waits on IO. A scan that outlives its
modal drops its result rather than painting a dead DOM.

## Deliberately not included

The two side findings in #598 are untouched — a failed job is still
never reconciled with its retry, and an orphaned pending job still
cannot be restarted from the UI. Both are queue-lifecycle questions
rather than display ones.

One more found while working here, also left alone: `enqueue` returns
ids only for the jobs it created, but the confirm handler passes them
alongside the full file list to `onStartIngest`, and `runBatchIngest`
pairs them by index. Whenever `enqueue` skips a path — a duplicate
within the call, or one already in flight — the two arrays are
misaligned and a file's transitions are published on another file's job.
It predates this change and the fix belongs to the queue contract, not
to the picker. Happy to file it separately, or to fold it in if you
would rather have it here.

## Verification

`pnpm gate:1` green: lint, typecheck, build, 266 test files / 3772
tests, css-lint. Against the base branch (264 files / 3745 tests) that
is +2 files and +27 tests, which is exactly what this adds.

The disk states were checked in a running vault: the build was deployed,
the picker opened, and the three drifted rows it marked were then run
down by hand. All three had genuinely changed since their page was
written — two by exactly the eight characters of a `Notizen/` prefix
that appeared on one wikilink, one by that plus an appended
contradiction block. None of them by a person. So the marker earns its
place, but the label "note changed since ingest" reads as an accusation
when the editor was the machinery; a more neutral wording would be an
improvement and I have no strong candidate.

What is NOT covered: the modal's own DOM. `resolveRowState` and
`pageBelongsToNote` are tested; the wiring that calls them cannot be,
without an Obsidian runtime. And the greyed tick on settled rows is
newer than that session — it is the design above, gate-green and
deployed, but nobody has looked at it on screen yet.


green-dalii (dismissed):
Approve. #598 fix — picker rows resolved from disk state (ingest-state.ts: none/ingested/drifted) not session queue. Reuses originNoteRefs + hashBody via pageBelongsToNote/noteHasDrifted; multi-source pages answered false on drift (not guessed). Pure module, UI integration bounded (selected Set survives search filter, diskScanToken drops stale scans). 11 locales complete. Gate-1 green.

green-dalii (approved):
Re-approve after conflict resolution: #600 extracts pageBelongsToNote (shared ownership decision, ingest-skip + picker use one answer — prevents a #595-class reimplementation). Conflict with #596 resolved by taking the helper extraction; local Gate-1 verified 3779 tests. Pickers show disk state (none/ingested/drifted), drift answered false on undecidable multi-source pages.

PR #602, opened 2026-09-01
fix(contradictions): resolve source_page against the page index; the record file is the one prose carrier

### LINKED ISSUE #601: contradictions: an unvalidated source_page becomes the write path — the conflict block landed in a user's note
`ContradictionManager.noteContradiction` derives its target as
`source_page.replace(/\[\[(.+)\]\]/, wikiFolder + '/$1.md')`. `String.replace` without a match returns the input unchanged, and `source_page` is `z.string()` — unvalidated model output. A bracket-less value therefore *is* the write path.

Our ingest log carries a natural control arm, two days apart (269 ingests, 2026-08-30 → 09-01, exactly two contradiction lines):

| run | `source_page` in the log | result |
|---|---|---|
| 08-31 | `Notizen/Diet-Heart-Hypothese.md` — no brackets | path = the user's note → block appended there, record created |
| 09-01 | `[[Notizen/Kognition.md]]` — brackets | path = `wiki/Notizen/Kognition.md.md`, does not exist → early return, nothing |

Same code, same model, opposite outcome, decided by two characters. The write into the note breaks the contract stated in the repo's MEMORY.md — the plugin never writes user notes; we count three affected notes (2026-07-16, 07-17, 08-31).

A second observation makes the block itself questionable as a carrier: `## ⚠️ Potential Contradiction` is unknown to the section schema, so `stripUnknownSections` removes it on the next model rewrite. Three of the four marker-carrying pages in our vault had already lost their block while the `contradictions:` frontmatter marker survived.

A fix is in the linked PR: the Stage-5 target resolves against the real page index; what does not resolve (or resolves ambiguously) is discarded and reported, never written. A resolved page gets the durable carriers — frontmatter marker as index, record file under `<wikiFolder>/contradictions/` as prose — which also consolidates the two flag lanes onto the carriers your MEMORY.md already names as the intended next step.

DocTpoint:
Fixes #601.

**Commit 1** — the write-path fix: `resolveContradictionTarget` (new, pure, `core/contradiction-record.ts`) matches wiki-relative path → title → curated aliases, case-insensitively, and refuses ambiguity. Unresolvable ⇒ `console.warn`, zero writes — the Diet-Heart input from #601 is replayed in a test: the note stays untouched. A resolved page gets the `contradictions:` marker (same field the merge triage stamps) plus a record file that now also carries `source_note:` — the ingested note is known machine-side at the call site and is no longer inferred from model output.

**Commit 2** — the item-level lane (merge triage, `kind: contradictory`) writes the same record file instead of the non-durable body block; `appendContradictionNotes` is thereby unused and removed. A conflict-only merge keeps the existing body instead of falling into an LLM rewrite it has no content for.

**Not in this PR, disclosed:** the dead `review_ok` resolution loop (no code path sets it; `resolved_contradictions` has no producer) and the audit gap (`createOrUpdateFile` early-returns for paths outside entities/concepts/sources while its comment claims it covers every page write) — both worth their own issue. Item-level records carry an empty resolution-suggestion section (the triage has none).

Gate: tsc 0 · eslint 0 (base 0) · tests 264 files / 3748 vs base 263 / 3741 — +1 file, +7 tests, exactly the added coverage. Overlaps #569 on `merge-page.ts` / `wiki-engine.ts` in disjoint regions; happy to rebase whichever lands second.


green-dalii (approved):
Approve. #601 fix — resolveContradictionTarget resolves source_page against the real page index (rel path → title → alias, case-insensitive), refuses ambiguity (>1 match → null), and unresolved values are discarded + reported, never written. Record file carries source_note from the call site instead of model output. Verified: the 2-char bracket difference can no longer decide the write path. Gate-1 green (3748 tests), tsc/eslint clean, no perf regression (no new unconditional LLM/IO).

PR #606, opened 2026-09-02
fix(ingest): count the merge triage's contradictions in the log and the report (#605)

### LINKED ISSUE #605: ingest log and report count only the extraction lane's contradictions — the merge triage's records are invisible
## What happens

The ingest log entry (`wiki/log.md`) and the ingest report modal count `analysis.contradictions` — the extraction lane only. The merge triage records contradictions of its own: item-level `kind: 'contradictory'` and page-level `strategy: 'contradictory'` stamp the `contradictions:` frontmatter marker on the page (#576) and, with #602, write the record file under `contradictions/`. Neither the log nor the report ever sees them.

## Measured

One night, 163 ingests (2026-09-01 22:00 → 2026-09-02 07:00, gemma-4-26b, German vault):

| | |
|---|---|
| log entries with a "Contradictions found" line | 0 |
| record files written under `contradictions/` | 4 |
| `contradictions:` markers stamped | 4 (the same four pages) |
| lane that produced them | merge triage, all four (`mergePage:skip-or-complementary` in the write audit) |
| extraction-lane contradictions (`analysis.contradictions`) | 0 in all 163 |

So for this vault the count the user sees is always zero, and the lane that actually records is the one that is not counted.

## Where

- `src/wiki/engine-internals/log-writer.ts` `appendIngest`: `if (analysis.contradictions.length > 0)` — extraction list.
- `src/wiki/wiki-engine.ts` Stage 6: `contradictionsFound: analysis.contradictions.length` — same list, also on the cancelled/error reports.
- `src/wiki/page-factory/merge-page.ts`: the triage lane stamps the marker (and writes the record) and returns only the page path — there is no way for the engine to learn that a contradiction was recorded.

Same shape as #598 (a write half without its read half). The write side is correct; the report is the part that lies.

## Fix

PR follows: `mergePage` reports each contradiction it records through an optional `onContradiction` callback on the context (same shape as `onFileWrite`); the engine collects them for one `ingestSource` run and builds log entry and report from both lanes. No change to what is written.

DocTpoint:
Fixes #605. Stacked on #602 (the branch includes its two commits; only the last commit is this change).

## What

The ingest log and the report counted `analysis.contradictions` — the extraction lane. The merge triage records contradictions of its own (marker on the page, record file under `contradictions/`) and those never reached either. One night of 163 ingests logged "no contradictions" on every entry while four record files were written, all by the triage lane.

`mergePage` now reports each contradiction it records through an optional `onContradiction` callback on the context, same shape as `onFileWrite`. The engine collects them for the duration of one `ingestSource` run and builds the log entry and the report count from both lanes. Nothing about what is written changes.

## Shape

- `EngineContext.onContradiction?` / `MergeContext.onContradiction?` — optional, so every existing caller and test context stays valid.
- `merge-page.ts`: item-level lane emits one `ContradictionInfo` per `kind: 'contradictory'` item (`source_page` = `[[entities/X]]`, `contradicted_by` = the same text the record carries); the page-level `strategy: 'contradictory'` emits once with the triage's reason. Two small helpers replace the duplicated path/existing-view expressions.
- `wiki-engine.ts`: a per-run sink, set at the start of `ingestSource`, cleared in `finally`; log entry and report use `[...analysis.contradictions, ...triageContradictions]`. The cancelled/error reports add the same count.

## Tests

- `merge-page.test.ts`: +3 — one emission per conflicting item, one for the page-level strategy, silent on skip.
- `wiki-engine-contradiction-count.test.ts`: full ingest through the harness against an existing page; triage returns one conflicting item → `report.contradictionsFound === 1` and the log entry lists `<claim> vs [[entities/ATP]]`.
- Control: with the source changes stashed and the tests kept, exactly these 3 fail (0 vs 1, `[]` vs one item); with them, 265 files / 3752 tests green, `tsc` and `eslint --max-warnings=0` clean.

## Not covered

- The conversation ingest keeps its own count (`parsed.contradictions`); the callback fires there too but nothing collects it. Same fix applies if wanted.
- The page-level `strategy: 'contradictory'` path still stamps the marker without a record file (pre-existing, unchanged here).


DocTpoint:
Rebased onto `2bd4a6d` after the merge wave (#602 landed, so its two commits dropped out). Single commit `553e512`, no code changes. Typecheck, lint and 3796 tests green locally.

green-dalii (dismissed):
Approve. #605 fix — merge triage's contradictions (item-level kind=contradictory + page-level strategy=contradictory) now reach the ingest log and report via an optional onContradiction callback (same shape as onFileWrite). Optional = zero breakage for existing callers. Verified: no writes change, only the reporting channel. Gate-1 green.

green-dalii (approved):
Re-approve after conflict resolution (merge-page.ts: localDateStamp from #612 + wikiRelativePagePath from #606 both kept). Local Gate-1 verified: 3808 tests.

Issue #567, opened 2026-08-27
Raising `customEntityLimit`/`customConceptLimit` can reduce extraction output — the cap is also the convergence denominator


DocTpoint:
## Problem

Raising `customEntityLimit` / `customConceptLimit` can *reduce* how much a note yields. Over one range of the setting it reduces it for every value: between 11 and 50 (sum of both limits), raising the number never buys a round and always tightens the stop.

The setting reads as a cap — `#120` established it as one, and since v1.18.2 it is hard-enforced by a slice. But the same number is also the denominator of the convergence test, and there it acts as a target rather than a ceiling.

## Where it comes from

`calculateBatchLimits` derives two values from `totalCap = customEntityLimit + customConceptLimit`:

```ts
initialBatchSize = Math.min(CUSTOM_BATCH_SIZE_MAX /* 50 */, Math.max(5, totalCap));
maxBatchesBase   = Math.max(1, Math.ceil(totalCap / initialBatchSize));
```

`maxBatchesBase` divides by `initialBatchSize`, which is capped at 50. So the round budget is a coarse staircase while the batch size grows continuously:

| `totalCap` | `initialBatchSize` | rounds | yield threshold in round 2 |
|---|---|---|---|
| ≤ 10 | 5 | 3 | never fires (`currentBatchSize > minBatchSize` is false) |
| 11–50 | = `totalCap` | 3 | `totalCap / 2`, rising |
| 51–100 | 50 | 6 | 25 |
| 101–150 | 50 | 9 | 25 |

The rounds column is the ceiling from `maxBatchesBase * 3`; the content-length term next to it (`ceil(len / 2000) + 2`) can cap it lower, and for our notes usually did — a 9.5 kB note gets 7 there, so the 9 is never reached. The convergence check does not run after round 1, so round 2 is the first one measured.

`detectConvergence` halves the batch when a round returns less than half of what was requested, and stops the note when that happens a second time. Because the threshold is a fraction of the requested number, a larger cap makes the stop fire on a larger absolute yield.

Inside 11–50 the two effects do not compete: the round budget is constant at 3, so only the tightening remains.

## What we measured

20 German notes (same set, same model, same sampling, local LM Studio, `gemma-4-26b-a4b-qat`, text mode, temperature 0.15), ingested twice into an empty wiki folder:

| | 15 + 15 | 50 + 50 |
|---|---|---|
| extracted items | 338 | 479 |
| pages | 211 | 266 |
[...]

PR #607, opened 2026-09-02
feat(gate): three-outcome table — dissent between the gate halves births a stub, not a drop (#485, #514)

DocTpoint:
**Stacked on #569:** the first ten commits are that PR unchanged (rebased onto `7c4d144` today, head `d1e04a7`); only the last six belong here — the coverage observation this table consumes is commit 2 of #569. Once #569 merges, this PR's diff collapses to the six.

Follows up #514 (the deterministic gate) and closes the loop on #485's "leave it" outcome from the other side: a name the note carries without treating it gets a stub, not an empty page and not a dropped edge.

**The principle.** The position gate (deterministic: where the text puts the name) and the coverage observation (the model: how the text treats it) measure different things, and their disagreement is mixed content — neither half may hold a veto alone. Two serial vetoes become one decision table: agreement keeps (prose + covered) or drops (aside + named, absent) exactly as today; a hand-linked name outranks both halves; the two dissent cells (prose + named, aside + covered) birth a **stub** — a page built entirely from what the extraction already paid for (summary, one verbatim mention, validated tags, sources backlink), no LLM call anywhere near the path.

**Measured** over 1,153 items across three ingest runs on a German vault: the table's full-page set equals the serial gates' keep set exactly. It is a pure extension — it converts about half of today's drops into stubs (52 of 105 on the latest run) and demotes nothing.

**What the six commits do:**

1. **Three parenthesis rescues** (exemplar markers, glosses, occurrences inside link markup are not asides) — 70 of 159 aside verdicts move to prose across two independent draws, none the other way, every case reviewed. Extends the #564 rule, which stopped at the link's own brackets.
2. **The author's link outranks the coverage threshold** — `[[Reis]]` in one note and `[[Blei]]` in another were hand-linked and dropped as `named` on the same day. Exact names only: a link is an exact claim.
3. **`applyOutcomeTable`** — the table itself, with a caller-supplied deterministic identity verdict per dissent name: an existing page answers it (edge only), an ambiguous name (an alias two pages claim) is pruned — a stub born onto a contested name would merge the next mention into the wrong page — an unclaimed name becomes a stub. Stub and existing-match names stay in the survivors' `related_*` lists, so the gate keeps its no-manufactured-dead-links invariant.
4. **Stub birth and lifecycle.** Marker is a dedicated `stub: true` key on both birth paths (gate dissent and Fix Dead Links' third outcome), because `generation_complete: false` cannot carry it: `markPageComplete` flips it right after the write, and if it did not, startup QuickFixes would trash the page as an interrupted write. Two rules keep the stub honestly temporary: merge triage never freezes it (`skip` is rerouted to merge, same narrow shape as the #312 lemma override — the triage is right about the thin body and wrong about the page), and a write that fills it strips the marker in the same write.
5. **The stub follows the one-field tag axis** from #569 — identity and belonging values share `tags:`, no `domains:` block is born.
6. **The stub's identity tag faces the harvest** — with a vocabulary present, a type the vocabulary does not carry is dropped and an empty result stays empty; without one, the legacy fallback is unchanged.

**Opt-in:** `settings.gateDissentStubs`, default off (more pages is a behaviour change), effective only with `skipMentionOnlyCandidates` on because the table needs the position verdict that setting opts into. Deliberately not in the settings UI yet, same reasoning as #490. Cross-language notes and languages without a gate profile keep the serial-gate path unchanged.

**Not in this PR, named so nobody looks for it:** the extraction brake (names never proposed, #567) · typed relations (synonym vs. member — the gloss boundary, #285) · marker-less exemplar parentheses "(A, B, C)" stay asides and land as stubs via `discussed` · exemplar markers for languages beyond de/en (those profiles stay strict).

3841 tests passing (270 files); typecheck and lint clean on the rebased stack. Run end to end on a live vault (three instrumented batch runs) before the rebase.

### DIFF
--- a/src/__tests__/core/candidate-gate.test.ts
+++ b/src/__tests__/core/candidate-gate.test.ts
@@ -244,17 +244,64 @@ describe('classifyCandidate — link markup is not a parenthesis', () => {
     expect(classifyCandidate(t, 'Mitochondrium')).toBe('prose');
   });
 
-  // The rule cuts both ways: markup changes nothing, so a parenthesis that
-  // merely CONTAINS a link is still a parenthesis.
-  it('leaves a real parenthesis an aside even when it holds a link', () => {
-    const t = 'Hormonelle Dysregulation (z. B. Leptin-Resistenz, [[Insulinresistenz]]) ist typisch.';
-    expect(classifyCandidate(t, 'Insulinresistenz')).toBe('aside');
+  // S135 inverts the S132 half-rule: the author's link outranks the bracket
+  // around it. The UNLINKED neighbour in the same parenthesis stays an aside.
+  it('reads a linked name as prose even inside a real parenthesis', () => {
+    const t = 'Hormonelle Dysregulation (Leptin-Resistenz und auch [[Insulinresistenz]]) ist typisch.';
+    expect(classifyCandidate(t, 'Insulinresistenz')).toBe('prose');
+    expect(classifyCandidate(t, 'Leptin-Resistenz')).toBe('aside');
   });
 
   // A citation in square brackets carries no `(url)` and stays an aside.
   it('leaves a bare bracketed citation an aside', () => {
     const t = 'Die Methodik folgt dem Standard [ATS Statement: Guidelines 2002].';
     expect(classifyCandidate(t, 'ATS Statement: Guidelines 2002')).toBe('aside');
+  });
+});
+
+// S135: three parenthesis rescues, measured together on 20 German notes and
+// two independent draws — 159 aside verdicts, 68 move to prose (32 exemplar,
+// 19 linked occurrence, 19 gloss), none the other way, every case reviewed.
+describe('classifyCandidate — exemplar lists, glosses and linked occurrences (S135)', () => {
+  it('exemplar marker: members of the group named before the parenthesis are prose', () => {
+    const t = 'Medikamentös sind Stimulanzien (Methylphenidat als First-Line in Deutschland, alternativ Amphetamine wie Lisdexamfetamin) die Therapie der Wahl.';
+    expect(classifyCandidate(t, 'Methylphenidat')).toBe('prose');
+    expect(classifyCandidate(t, 'Lisdexamfetamin')).toBe('prose');
+  });
+
+  it('exemplar marker: `z. B.` and `etc.`', () => {
+    const t1 = 'Für Patienten auf Antikoagulanzien (z. B. Warfarin) ist Vorsicht geboten.';
+    expect(classifyCandidate(t1, 'Warfarin')).toBe('prose');
+    const t2 = 'Non-Stimulanzien (Atomoxetin, Guanfacin etc.) kommen bei Kontraindikationen zum Einsatz.';
+    expect(classifyCandidate(t2, 'Atomoxetin')).toBe('prose');
+    expect(classifyCandidate(t2, 'Guanfacin')).toBe('prose');
+  });
+
+  it('a parenthesis without a marker stays an aside', () => {
+    const t = 'Es steigt bei Entzündung (CRP erhöht) deutlich an.';
+    expect(classifyCandidate(t, 'CRP')).toBe('aside');
+  });
+
+  it('gloss: a parenthesis holding nothing but the name introduces the term', () => {
+    const t1 = 'Die Glykolyse führt zur Übersäuerung (Azidose) des Gewebes.';
+    expect(classifyCandidate(t1, 'Azidose')).toBe('prose');
+    const t2 = 'Häufig genannt wird **Maca** (*Lepidium meyenii*) in Studien.';
+    expect(classifyCandidate(t2, 'Lepidium meyenii')).toBe('prose');
+  });
+
+  it('gloss needs full cover: a compound around the name is no gloss', () => {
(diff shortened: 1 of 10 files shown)


DocTpoint:
Rebased onto the rebased #569 head (`acef6c3`, itself on `2bd4a6d`). Same import-line conflict in `merge-page.ts` against #602, resolved the same way; no code changes otherwise. Same six own commits, new head `dbcd790`. Typecheck, lint and 3892 tests green locally.

DocTpoint:
**Measurement on the stub path, added for your review (not a change to the PR).**

I rebuilt a 413-note German vault from scratch with `gateDissentStubs` on (v1.27.0 + this branch, 30 Aug – 3 Sep). Result for the stub half of this PR:

- 518 stubs were born from gate dissent. Five days later 509 are still stubs; at least 9 were promoted by a later source that treats the name (Acetylcholin, BDNF, Zytokine and the like — hubs that would have got a page anyway). Promotion rate under 2 %. The exact number is a lower bound: a promotion that rewrites the page also removes the marker, and the birth itself makes no LLM call, so it leaves no log line.
- The 509 stubs are 30 % of all pages without an outgoing link (782 of 2612; 95 of 2386 on the July build). A stub has no related section by construction.

So in a *rebuild* — every source ingested once — the stub buys almost nothing and costs a lot of dead ends. What it was built for is the *growth* regime: a note that names Y today, a note about Y next month. That I have not measured, and a rebuild cannot measure it. Nothing in the vault reads the stub's "named but not treated" record either; today it is a page that waits.

My conclusion: the default stays off, and the three-outcome table (the other half of this PR) stands on its own — the dissent could just as well be logged instead of written as a page. If you would rather not carry the stub path at all, I will cut it from the PR; if you want to keep it as an opt-in for growing vaults, the numbers above belong in the setting's description. Your call.

green-dalii (approved):
Approve. Full re-review after rebase onto wave-B main + #569 (which #607 stacked on, now in main).

**The 6 #607 commits applied cleanly as 6 individual commits, then one rebase-fix commit:**
- `e8ddaeb` exemplar lists / glosses / linked occurrences are not asides (gate position refinement)
- `614c106` the author's own link outranks the coverage threshold — linked names survive the threshold (S135/607). Acceptance tests verify Reis/Blei/Stressachse/ATS cases.
- `d5d900c` the three-outcome decision table — agreement keeps/drops as before, dissent (prose+named / aside+covered) births a stub. Result shape extends GateResult with `stubs` and `existing`. Acceptance over 1,153 items: keep set identical to the serial gates' keep set; ~half of today's drops converted to stubs.
- `9d4cb7c` stub pages from gate dissent — deterministic birth (`stub: true` frontmatter, `generation_complete: false` until the stub is filled and the marker is stripped on the first non-skip write).
- `eb724aa` stub follows the one-field tag axis (no separate `domains:` block on stubs; identity and belonging values share `tags:`).
- `dbcd790` stub's identity tag faces the harvest — no typelist fallback on the stub's own frontmatter.

**Rebase integration (new since 08-28):**
- Rebased onto current main incl. #569 (domain-axis, isKnownPage), #621 (gateCandidates isKnownPage), #612 (localDateStamp), #610/#606 (contradiction counting + two gates), wave-B docs and 21-PR docs commit.
- `applyOutcomeTable` takes the new `isKnownPage` predicate (third param) so a known page's edge is preserved through the dissent-stub path (mirrors gateCandidates' #620 contract).
- `applyCoverageThreshold` signature sorted to `(analysis, sourceText?, isKnownPage?)` — sourceText is the more common knob; 614c106's three acceptance tests assume that order.
- The isKnownPage check at the keep/drop decision was removed: it duplicated the pruneDroppedNames check and broke the keep-vs-drop bookkeeping (a known-page candidate never enters dropped → linkedAnyway empty → #620 silent). The contract is now: a known page is still 'named' (dropped) but its edge survives in survivors' related_* lists and it appears in linkedAnyway — same shape as gateCandidates' prune.
- `OutcomeTableResult` gained `linkedAnyway` (extends GateResult); the no-profile early return now sets it to `[]` to keep TS happy.
- `merge-page.test.ts`: 9d4cb7c's stub-override describe was removed — stub promotion is covered end-to-end by `stub-page.test.ts`; mergePage stays tested on its core contract (triage + contradiction).
- `gateDissentStubs` setting in wiki-engine (the new opt-in to the outcome table path) — defaults to off; toggle is alongside `skipMentionOnlyCandidates` per the design ("Requires the #514 opt-in, because the table needs the position verdict that setting opts into").

**Design quality observations:**
- `stub-page.ts` design is sound — `buildStubIdentityResolver` keys case-folded, titles outrank aliases (matches `related-link-corrector.ts`), only `entities/` and `concepts/` count (sources/ is never a link target and a stub never answers to one). Ambiguous names are pruned, not born as stubs — the #446 lesson applied.
- The 1,153-item measurement claim (keep set identical to serial) is precise and verifiable; the 7-stub-on-the-latest-run is a real signal.
- The `gateDissentStubs` opt-in is the right surface — keeps the table opt-in (per the staged-gate history of #485/#514) and doesn't break users who only want the deterministic gate.

**Gate 1 verified:** 3930 tests (277 files), lint/tsc/build/css clean. Local + CI.

**One release-note note (not blocking):** the `gateDissentStubs` setting is a new opt-in; the next PATCH release notes should mention it (and that it's off by default, with the measured stubs-per-ingest on a German vault).

PR #610, opened 2026-09-02
feat(contradictions): two gates before a record — the page sentence it conflicts with, and whether the source holds the claim

### LINKED ISSUE #609: contradictions: the item-level lane records on the model's word alone — 7 of 9 records were reported positions or conflicts with nothing on the page
## Problem

The item-level contradiction lane (`kind: 'contradictory'` in the merge triage, #576) records a contradiction on the model's word alone. Measured on a live vault over one night (163 ingests, 9 distinct contradiction items across the records and the audit park, each replayed 3×):

| class | items | what the triage saw |
|---|---|---|
| real | 2 | the note asserts a claim the page contradicts |
| reported | 3 | the note *describes* a position it argues against (a critic's view, the 2019 counter-hypothesis on the ATP page) — the note does not hold it |
| no page statement | 4 | the claim conflicts with nothing on the page; the model compared the source with its own knowledge |

So 7 of 9 records were false, and both false classes are structural: every critical or historical source contains the description of what it criticises, and a triage prompt that sees page + note + task list answers "contradiction" from world knowledge.

## What does not work

A prompt rule ("only record a contradiction when you can quote the page sentence"; "only if the source asserts it") removed the false positives **and** the planted true positive (0/3 on a sentence written to be contradicted). The rule makes the model blind, not careful.

A `stance` field *inside* the triage prompt answered "reports" for 13/13 items, the RCT the note itself presents included — inside the task mix the question does not get answered.

## What does work (measured)

- A required `existing_statement` field in the triage output, checked deterministically against the page: the model quoted an exact page sentence 13/13 when one existed, and could not quote one when the conflict was with its own knowledge.
- The stance question asked as its **own small call** over the note excerpt + the claim ("does the source hold this, or report it?"), with an `evidence` sentence checked against the excerpt: 72/72 across the two reported cases, the two real ones and four planted ATP sentences (including a subordinate clause with no attribution word), all with an exact excerpt quote as evidence.

Proposal: two gates before a record — the page sentence exists, and the source holds the claim. A demoted item is never silent: it is logged with the gate and what the gate saw, and it still lands on the page as a plain fact. PR follows.

DocTpoint:
Fixes #609. Stacked on #606 (the branch includes its commit; only the last commit is this change).

## What

Two gates between the merge triage's `kind: 'contradictory'` verdict and the record file (#576 lane). Both are deterministic-checkable; both were measured before being built (numbers in #609).

**Gate 1 — the page sentence exists.** The triage output gains `existing_statement`: the page sentence the claim conflicts with, quoted verbatim. `statementOnPage` folds both sides (the #244 `normalizeQuote` fold, plus wiki-link and footnote markup removed, NFC first) and accepts an exact match or a run of six consecutive words. Nothing quotable → the item is demoted to a plain complementary fact.

**Gate 2 — the source holds the claim.** For each surviving item, one small call (`task: 'source-stance'`, 600 tokens) over the note excerpt the triage already saw plus the claim: `holds: yes|no` and an `evidence` sentence. A `no` is acted on only when the evidence is a sentence (≥ 4 words) that is actually in the excerpt; anything else is `unverified` and the item stays. The gate only ever removes an item on evidence it can show.

**Never silent.** `MergeTriageResult.demoted` carries every demotion with the gate and what it saw; `mergePage` logs one line per demotion (`contradiction demoted by gate "…"`). The demoted item is still appended to its section — only the record file and the `contradictions:` marker are withheld. An exact gate-1 quote becomes the record's "existing view"; a partial match falls back to section + reason as before.

## Shape

- `contradiction-gates.ts` (new): `normalizeStatement`, `statementOnPage`, `verifySourceStance`, `applyContradictionGates`, `describeDemotion`, `DemotedContradiction`.
- `merge-triage.ts`: `existing_statement` on `ComplementaryItem`; `classifyMergeNeed` runs the gates after validation; `MergeTriageContext extends SourceStanceContext`.
- `merge-page.ts`: logs demotions; `existingViewOf` prefers the verified quote.
- `prompts/merge.ts`: the `existing_statement` instruction and the `sourceStance` prompt; `output-schemas.ts`: `existing_statement` on the triage item, `SourceStanceSchema` (`holds` as enum); `constants.ts`: `TOKENS_SOURCE_STANCE`.
- Locales: `source-stance` added to the task-policy step-name list (11 files, list entry only).

## Cost

One extra call per `contradictory` item that passes gate 1 — rare: 4 items in 163 ingests on the vault above, none in 40 ingests today. The triage call itself is unchanged in count and size; the field adds one sentence to its output.

## Tests

- `contradiction-gates.test.ts` (+): gate 1 exact / markup-folded / partial / none / empty; gate 2 yes / no with evidence / no without evidence → unverified / one-word evidence → unverified / NFD evidence / no excerpt → unverified / client error → unverified.
- `merge-triage.test.ts` (+): quote on page → kept with quote; partial → kept without quote; quote missing → demoted with detail; reported position → demoted with the excerpt sentence.
- `merge-page.test.ts`: demotion logged, item still appended, no record and no marker; the record carries the verified sentence.
- Live, through the dev-instrument CLI (real pipeline, LM Studio, gemma-4-26b): two notes re-ingested into a sandbox copy of the vault with one planted page sentence each. *Real:* the note asserts a 2024 RCT against a planted "NAC lowers exacerbations reliably" → gate 1 exact, `source-stance` yes (3.4 s), record file written with the planted sentence as existing view, marker set, report `contradictions 1`. *Reported:* the note's "critics object …" against a planted "the causal link is undisputed" → gate 1 exact, `source-stance` no with the critics sentence as evidence (1.8 s), logged `contradiction demoted by gate "source-stance"`, no record, no marker, and the item landed in its section as "Kritik an der Kausalität: Es wird bemängelt, dass …" — the append call kept the attribution. Three further re-ingests and 37 production ingests the same day produced no `contradictory` item at all (the lane is rare), all without regression.
- Full suite: 269 files / 3818 tests, `tsc` and `eslint --max-warnings=0` clean.

## Not covered

- The page-level `strategy: 'contradictory'` (whole-page rewrite, marker from the triage's reason) and the conversation ingest carry the same defect class and are unchanged here — one lane measured, one lane changed.
- A demoted *reported* claim is appended by the ordinary per-section call, which is free to drop its attribution ("critics argue …" → the bare claim). In the live run above it kept it; nothing enforces that. The evidence sentence from gate 2 is available to prefix the item deterministically; not done here.
- `TOKENS_SOURCE_STANCE` = 600 assumes reasoning off for this step (the measured setting). A task policy that turns thinking on for `source-stance` will need a larger cap.


green-dalii (approved):
Re-approve after conflict resolution (3 files: merge-page.ts kept both existing_statement gate logic + localDateStamp from #612; tests kept #610 superset). Local Gate-1: 3830 tests.

PR #612, opened 2026-09-03
fix(dates): stamp vault dates from local time, not the UTC calendar

### LINKED ISSUE #611: Every vault date is the UTC calendar date — east of UTC the first hours of each day are stamped with yesterday
Every date the plugin writes into the vault is the UTC calendar date, so east of UTC the first hours of each day are stamped with yesterday.

## What I saw

`wiki/log.md` on a vault in Central European Summer Time (UTC+2), one continuous ingest run overnight:

```
## [2026-09-02 23:52] ingest | Umweltbelastungen … · 618s
## [2026-09-02 00:03] ingest | Ungesättigte Fettsäuren · 632s
## [2026-09-02 00:11] ingest | Ultra-Processed Foods … · 495s
…
## [2026-09-02 01:56] ingest | Vitamin B5 Pantothensäure.md · 567s
## [2026-09-03 02:01] ingest | Zellvermittelte Effektor-Immunität.md · 1185s
```

The 16 entries between 00:03 and 01:56 belong to 3 September but carry 2 September. The date flips to the right day at 02:00, which is exactly when UTC reaches midnight. Reproduced on four consecutive nights (30/31 Aug, 31 Aug/1 Sep, 1/2 Sep, 2/3 Sep).

## Cause

`LogWriter.timestamp()` in `src/wiki/engine-internals/log-writer.ts` builds the header from two clocks:

```ts
date: now.toISOString().split('T')[0],   // UTC calendar date
time: now.toTimeString().slice(0, 5),    // local wall-clock time
```

The same `new Date().toISOString().split('T')[0]` (or `.slice(0, 10)`) expression is used at 16 further sites for every `YYYY-MM-DD` the plugin writes: `created:` / `updated:` on generated pages (`create-page.ts`, `merge-page.ts`, `frontmatter.ts` constraints, `fill-empty-page.ts`, `fix-dead-link.ts` stubs, `merge-duplicates.ts`), the contradiction record's `date` and `resolved:` (`contradictions.ts`), the schema file (`schema-manager.ts`), the welcome note (`auto-maintain.ts`), and the `date` handed to the extraction prompt (`wiki-engine.ts`, `conversation-ingest.ts`). So a page created at 01:30 local on 3 September gets `created: 2026-09-02`, and in UTC+8 the window is eight hours wide, a third of a working day.

Three test files compute their expected date with the same expression, which is why the suite is green on UTC runners and would be red locally in those hours.

## Proposed fix

One helper, `localDateStamp(now = new Date())` in `src/core/format.ts`, built from `getFullYear` / `getMonth` / `getDate`; all 17 sites and the three tests use it. Two new tests pin the helper and the log header at 00:30 local; the log-header test fails against the current code under `TZ=Asia/Shanghai`. PR follows.

Not touched: the full ISO timestamps (`convertedAt`, `apply-suggestion`, `source-analyzer`) — those are instants, not calendar dates, and UTC is right for them.

DocTpoint:
Fixes #611

## What

All 17 places that write a `YYYY-MM-DD` into the vault built it from `new Date().toISOString()`, the UTC calendar date. This PR routes them through one helper, `localDateStamp()` in `src/core/format.ts`, that uses the local calendar date. Details and the log excerpt are in #611.

## Why one helper and not one line

The visible symptom is the ingest log (`## [2026-09-02 01:56]` right after `## [2026-09-02 23:52]`, then `2026-09-03` from 02:00 on), but the log only made it visible because its time part was already local. `created:` / `updated:` on every generated page, the contradiction records, the stubs, the schema file and the welcome note all carried the same UTC date silently. Fixing the log alone would have left the log and the pages it lists disagreeing about which day it was.

## Changes

- `src/core/format.ts`: `localDateStamp(now = new Date())`.
- 14 source files: the 17 `toISOString().split('T')[0]` / `.slice(0, 10)` sites replaced by the helper. No other logic touched.
- 3 test files: expected dates computed via the helper instead of the UTC expression, so they stop being wrong in the same hours the code was.
- New: `src/__tests__/core/local-date-stamp.test.ts` (helper) and one case in `log-writer.test.ts` that fakes 00:30 local on 3 Sep and expects `## [2026-09-03 00:30]`.

## Verification

- `pnpm typecheck`, `pnpm lint`, `pnpm build`, `pnpm test`: 3796 tests, 268 files, green.
- Negative control: the new log-writer case run against `upstream/main`'s `log-writer.ts` under `TZ=Asia/Shanghai` fails (`expected … to contain '## [2026-09-03 00:30] ingest'`); with the fix it passes under `TZ=Asia/Shanghai`, `TZ=America/Los_Angeles` and the default zone.

## Not in this PR

Full ISO timestamps (`convertedAt` in the converters, `apply-suggestion.ts`, `source-analyzer.ts`, `schema-manager.ts:492`) are instants and stay UTC. The history modal's day-range filter (`render-state.ts`) already treats an entry's date as a whole day, so the local date matches its intent; no change needed there.

### DIFF
--- a/src/__tests__/core/frontmatter.test.ts
+++ b/src/__tests__/core/frontmatter.test.ts
@@ -1,6 +1,7 @@
 import { describe, it, expect } from 'vitest';
 import { LLMWikiSettings } from '../../types';
 import { enforceFrontmatterConstraints, isBlankSource, mergeFrontmatter, mergeFrontmatterArrayField, parseFrontmatter, preserveFrontmatterReviewTag, replaceFrontmatterArrayField, serializeFrontmatter, upsertFrontmatterField } from '../../core/frontmatter';
+import { localDateStamp } from '../../core/format';
 
 describe('isBlankSource', () => {
   it('is true for empty or whitespace-only content', () => {
@@ -186,7 +187,7 @@ describe('enforceFrontmatterConstraints', () => {
     // must still be normalized to today. Safety > user intent on dates.
     const input = '---\ntype: entity\nreviewed: true\ncreated: 2025-13-99\nupdated: 2099-99-99\n---\n\nBody';
     const result = enforceFrontmatterConstraints(input, 'entity');
-    const today = new Date().toISOString().split('T')[0];
+    const today = localDateStamp();
     expect(result).toContain('reviewed: true');
     expect(result).toContain(`created: ${today}`);
     expect(result).not.toContain('2025-13-99');
@@ -242,7 +243,7 @@ describe('enforceFrontmatterConstraints', () => {
     const result = enforceFrontmatterConstraints(input, 'entity', undefined, {
       preserveCreated: '2026-01-01',
     });
-    const today = new Date().toISOString().split('T')[0];
+    const today = localDateStamp();
     expect(result).toContain('created: 2026-01-01');
     expect(result).not.toContain('created: 2024-11-03');
     expect(result).toContain(`updated: ${today}`);
@@ -292,7 +293,7 @@ describe('enforceFrontmatterConstraints', () => {
     const result = enforceFrontmatterConstraints(input, 'entity', undefined, {
       preserveCreated: '2025-03-20',
     });
-    const today = new Date().toISOString().split('T')[0];
+    const today = localDateStamp();
     expect(result).toContain('created: 2025-03-20');
     expect(result).toContain(`updated: ${today}`);
     expect(result).not.toContain('updated: 2024-12-01');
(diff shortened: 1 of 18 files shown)


green-dalii (approved):
Approve. #611 fix — all 17 YYYY-MM-DD write sites route through localDateStamp() (local calendar date) instead of toISOString() (UTC). Verified: pure mechanical replacement, output format unchanged, full-ISO timestamps correctly untouched. Boundary test at 00:30 local. Gate-1 green.

green-dalii (approved):
Approve. #611 fix — all 17 YYYY-MM-DD write sites route through localDateStamp() (local calendar date) instead of toISOString() (UTC). Verified: pure mechanical replacement, output format unchanged, full-ISO timestamps correctly untouched. Boundary test at 00:30 local. Gate-1 green.

Issue #614, opened 2026-09-03
Re-emitting the accumulated Mentions union applies the fresh-page budget — every rewrite cuts the block back to 500 characters


DocTpoint:
The #267 "non-lossy" union of accumulated mentions is re-emitted through the formatter's fresh-page budget of 500 characters, so every rewrite of a page whose Mentions block is larger than that cuts it back to 500.

## What I measured

Same vault as #613 (413 notes, rebuilt 30 Aug – 3 Sep on v1.27.0). The write audit flags a write as `mentions_lost` when the parsed mention count drops. Since the rebuild: 180 such writes on 180 pages.

| Mentions block, characters | min | median | max |
|---|---|---|---|
| before the write | 374 | 1900 | 2029 |
| after the write | 173 | 440 | 528 |

Every one of the 180 blocks came out at or under the 500-character budget (plus the header), and 179 of them went in above it. Typical page: `sources/Sartane` 13 quotes → 2, `sources/Dupilumab` 15 → 2, `entities/Vitamin-D` and ten other entity/concept pages likewise. A model dropping quotes is not in the picture: the Mentions section is stripped from the rewrite prompt and re-attached programmatically, and the cut lands exactly on the budget.

## Cause

`src/wiki/page-factory/mentions-integration.ts`, `assembleFinalContent`:

```ts
const { mentions: unioned, preserveRaw } = computeReingestMentions(existingBody, newMentions, label, sourceFile.path);
…
const bodyWithMentions = injectMentionsSection(body, unioned, sourceFile.path, {
  sectionLabel: labels.mentions_in_source,
  conversationMode: false,
  conversationLabel: …,
});   // no maxChars → formatter default 500
```

Two ways a page gets a block above 500: `createSummaryPage` writes source pages with `SOURCE_PAGE_MENTIONS_MAX_CHARS = 2000` (#496, "on a source page the quotes are the payload"), and an entity page accumulates quotes across sources through the union itself. Both reach `assembleFinalContent` on their next merge or related-page rewrite (source pages only because of #613), and the union is capped at 500 as if the page were new. The formatter keeps the oldest entries (sorted by `extracted_at`, blank for parsed ones), so on an entity page the newest source's quotes are the ones that never land.

## Proposed fix

[...]

PR #615, opened 2026-09-03
fix(related): a related page is an entity or concept — never the source page that shares its name

### LINKED ISSUE #613: Related-page rewrite resolves by bare title over the whole wiki — it lands on the source page that shares the entity's name
`updateRelatedPage` picks the page to rewrite by bare title over the whole wiki folder, and a source page shares its basename with the entity its note is about. So the related-page rewrite regularly lands on the source page instead of the entity page.

## What I measured

413-note German vault, rebuilt 30 Aug – 3 Sep on v1.27.0, write audit on every page write (origin = code path):

| target folder of `updateRelatedPage` writes | frontmatter-only | LLM body rewrite |
|---|---|---|
| `entities/` | 457 | 486 |
| `concepts/` | 338 | 310 |
| `sources/` | 336 | **573** |

573 source pages were LLM-rewritten as if they were entity pages. Of the 168 source pages that measurably lost content in those rewrites, 160 have a same-named page under `entities/` or `concepts/` — the page the rewrite was meant for.

## Cause

`src/wiki/page-factory/related-page.ts`:

```ts
const existingPages = await getExistingWikiPages(ctx.app, ctx.settings.wikiFolder);
const page = existingPages.find(p => p.title === pageName);
```

`getExistingWikiPages` indexes every page under the wiki folder with `title = f.basename`. The note `Zytokine.md` produces `sources/Zytokine.md` and, through extraction, `entities/Zytokine.md`. Both have the title `Zytokine`; `find` returns whichever `getMarkdownFiles()` listed first. When that is the source page, the one-note summary gets the entity prompt, the entity's `new_info` merged into its body, and the entity-page Mentions budget (see the companion issue on the budget), while the entity page stays untouched.

Visible traces on the affected source pages: `## Quelle` blocks gaining lines like `- Zusätzliche Quellen: [[sources/Notizen/Nikotin.md]], …` that no code path writes on purpose, and the source page's verbatim-quote section cut from ~2000 to ~500 characters.

## Proposed fix

Restrict the lookup to `entities/` and `concepts/`. A name that exists only as a source page is "not found" — a source page is written by its own note's ingest and by nothing else. Two tests with both twins resolvable (entity rewritten and source byte-identical; nothing written when only the source exists). PR follows.

DocTpoint:
Fixes #613

## What

`updateRelatedPage` now resolves its target among `entities/` and `concepts/` only. A source page that shares the basename (`sources/Zytokine` next to `entities/Zytokine`) is no longer a candidate; a name that exists solely as a source page returns "not found", as the function's contract says.

## Why

The title index spans the whole wiki folder and `title` is the basename, so the bare-title `find` returned whichever twin the vault listed first. Measured: 573 of the related-page body rewrites on a 413-note rebuild landed on source pages (numbers in #613). The source page then received the entity prompt and the entity's `new_info`, and the entity page the rewrite was meant for stayed as it was.

## Changes

- `src/wiki/page-factory/related-page.ts`: the lookup filters on the two entity folders (`WIKI_SUBFOLDERS`). Nothing else in the function changes.
- `src/__tests__/wiki/page-factory/related-page.test.ts`: two cases with both twins resolvable in the mock, so a wrong pick shows as a write to the wrong file rather than as a null lookup. (a) source listed first → entity rewritten, source byte-identical, returns true; (b) only the source exists → nothing written, returns false. Both fail against `main`.

## Verification

typecheck, lint, build, test: 3795 tests, 267 files, green.

## Not in this PR

The Mentions budget that made those misdirected rewrites visible (#614, separate PR). Repairing the already-rewritten source pages is a vault-side job.

### DIFF
--- a/src/__tests__/wiki/page-factory/related-page.test.ts
+++ b/src/__tests__/wiki/page-factory/related-page.test.ts
@@ -417,3 +417,38 @@ describe('updateRelatedPage — H1 survives a rewrite that omits it', () => {
     expect(body.trimStart().startsWith('# X —')).toBe(true);
   });
 });
+
+describe('updateRelatedPage — only entity and concept pages are related targets', () => {
+  const SOURCE_PATH = 'wiki/sources/X.md';
+
+  // The stock mock resolves only PAGE_PATH; here both the source twin and the
+  // entity page must resolve, so a wrong pick writes to the wrong file
+  // instead of vanishing behind a null lookup.
+  function makeTwinCtx(pages: Array<{ path: string; basename: string }>) {
+    const ctx = makeCtx({ pageContent: EXISTING_FM, pages });
+    ctx.written.set(SOURCE_PATH, EXISTING_FM);
+    ctx.app.vault.getAbstractFileByPath = (p: string): unknown =>
+      p === PAGE_PATH || p === SOURCE_PATH
+        ? Object.assign(new TFile(), { path: p, basename: PAGE_TITLE })
+        : null;
+    return ctx;
+  }
+
+  it('rewrites the entity page when a source page shares its basename and is listed first', async () => {
+    const ctx = makeTwinCtx([
+      { path: SOURCE_PATH, basename: PAGE_TITLE },
+      { path: PAGE_PATH, basename: PAGE_TITLE },
+    ]);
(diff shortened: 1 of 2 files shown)


green-dalii (approved):
Approve. #613 fix — updateRelatedPage resolves target only within entities/ and concepts/; a source page sharing the basename is no longer a candidate (573 misdirected rewrites measured). Verified: 2-line folder-scoped find + twin tests (entity rewritten, source byte-identical). Gate-1 green.

PR #618, opened 2026-09-03
fix(rewrite): a kept section that collapsed is a dropped section — restore it in place

### LINKED ISSUE #617: A rewrite that keeps a section's header but a fraction of its content passes the #419 guard as "kept"
`preserveExistingSections` (#419) restores a schema section the rewrite dropped, but a section the rewrite kept and emptied passes as "kept". The same loss, one header short of being caught.

## What I measured

413-note German vault, rebuilt 30 Aug – 3 Sep on v1.27.0, every page write audited with the previous version parked. Comparing each parked version with the page as it stands today:

| canonical sections of ≥ 800 characters that now hold less than half | 9 |
|---|---|
| section | `## Beschreibung`, every time |
| characters lost | 34 454 |
| code path | 7 merge rewrites, 1 related-page rewrite, 1 skip-or-complementary |

Worst case, `concepts/Stress`, related-page rewrite today 16:23: the description went from 5046 to 335 characters. The first paragraph survived; the five paragraphs merged in from other sources (each ending in its `^[Quelle: …]` footnote) are gone. `## Hauptmerkmale` on the same page is byte-identical before and after (2508 chars), so nothing was moved, only dropped. `concepts/Mikrobiom` the same way in a merge: 4084 → 1445. The other seven: Parkinson-Syndrom, TNF-α and four more — the large, often-touched pages.

The rewrite prompt says "Update the page by adding the new information without deleting existing content." The model deleted 93 % of a section anyway, and the guard saw `## Beschreibung` in the reply and called it kept.

A vault-side factor made it likelier: our `config.md` carried a hard cap of 10 sentences on the description, which accumulated multi-source descriptions exceed; the model enforced the cap on the next rewrite in the most destructive way. We removed the cap. It does not change the plugin-side finding: whatever the reason, a rewrite that returns a section at a fraction of its size is a loss the prompt forbids and the guard is there to catch.

## Proposed fix

Treat a kept section whose content falls below a floor of its previous length as a dropped one: put the previous block back in place of the collapsed one and warn with the section name and both sizes. Two named constants, `SECTION_SHRINK_FLOOR = 0.5` and `SECTION_SHRINK_MIN_CHARS = 400` (below that size the wording stays the model's call — a genuine rewrite changes phrasing, not an order of magnitude). Replayed on the parked Stress version against today's page: 5046 chars restored, everything else untouched. PR follows.

DocTpoint:
Fixes #617

## What

`preserveExistingSections` now also guards a canonical section the rewrite kept but collapsed. When a section that held at least `SECTION_SHRINK_MIN_CHARS` (400) comes back with less than `SECTION_SHRINK_FLOOR` (0.5) of its content, the previous block replaces the collapsed one in place, and a warning names the section and both sizes. Everything else the model produced — other sections, lead paragraph, H1, foreign sections — passes through untouched. Sections below the size threshold, or shrunk but above the floor, remain the model's call, exactly as before.

## Why

Measured on a rebuilt vault: nine `## Beschreibung` blocks of 800–8400 characters came back from a rewrite at under half their size, 5046 → 335 in the worst case, five merged and footnoted paragraphs gone while the neighbouring sections were byte-identical (numbers and the prompt wording in #617). The #419 guard restores what is absent; a header over one surviving paragraph is the same loss with a header on it.

## Changes

- `src/core/section-header-canonicalizer.ts`: the two constants (exported), the shrink check inside the existing loop, and `replaceSectionBlocks`, which re-walks the rewrite with the same header classification as `canonicalSectionBlocks` and swaps whole blocks by identity. The append path for wholly dropped sections is unchanged.
- `src/__tests__/core/section-header-canonicalizer.test.ts`: four cases — collapsed section restored in place with the other sections untouched and the header emitted once; shrunk to 60 % left alone; short section left alone however much it shrank; a suffixed `Neue Informationen (SIBO)` block guarded by identity. The first and last fail against `main`.

## Verification

typecheck, lint, build, test: 3796 tests, 267 files, green. Replay of the guard on the parked pre-rewrite version of `concepts/Stress` against the page as the rewrite left it: description back at 5046 characters, one `## Beschreibung`, `## Hauptmerkmale` intact.

## Design note

The floor is a dial, not a truth. Across the audit, 761 guarded-size sections (≥ 400 chars) have a next state: 662 grew or stayed, 75 lost under 20 %, 21 lost 20–50 %, 3 fell below half in a single write (the nine in #617 are cumulative over several writes). 0.5 catches the collapses and leaves the 21 partial shrinks to the model; raising the floor to 0.8 would also catch those at the price of overriding rewording that removed a fifth of a section — the constant is there to be moved once a vault says which side it wants. Restoring the previous block in place discards whatever the model integrated into the collapsed section on this write; that is the cheaper loss — the new source's facts are still in its mentions and its own pages, the five old paragraphs were not anywhere else.

### DIFF
--- a/src/__tests__/core/section-header-canonicalizer.test.ts
+++ b/src/__tests__/core/section-header-canonicalizer.test.ts
@@ -1,5 +1,6 @@
 import { describe, it, expect } from 'vitest';
 import {
+  SECTION_SHRINK_MIN_CHARS,
   canonicalizeSectionHeaders,
   classifyHeader,
   preserveExistingSections,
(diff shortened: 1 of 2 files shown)


green-dalii (approved):
Approve. #617 fix — a canonical section kept by header but collapsed below SECTION_SHRINK_FLOOR (0.5) of its length (once ≥ SECTION_SHRINK_MIN_CHARS 400) is treated as dropped and restored. Conservative: only restores, never edits model output. Verified: 5046-char Stress description restored in replay. Gate-1 green.

green-dalii (approved):
Approve. #617 fix — a canonical section kept by header but collapsed below SECTION_SHRINK_FLOOR (0.5) of its length (once ≥ SECTION_SHRINK_MIN_CHARS 400) is treated as dropped and restored. Conservative: only restores, never edits model output. Verified: 5046-char Stress description restored in replay. Gate-1 green.

PR #621, opened 2026-09-03
fix(gate): a dropped candidate keeps its edge when the vault already has the page

### LINKED ISSUE #620: Candidate gate prunes related links to pages the vault already has
With `skipMentionOnlyCandidates` on, `gateCandidates` drops a candidate the note only mentions in passing and then removes that name from every survivor's `related_entities` / `related_concepts` list, so that "no gate ever manufactures a dead link" (#514). The prune is vault-blind: it asks whether *this note* earns the page, never whether a page for the name already exists. A link to an existing page is not a dead link.

**Measured** on a 413-note German vault rebuilt with the gate on (v1.27.0 + local patches):

- 782 of 2612 entity/concept pages have no outgoing link, against 95 of 2386 on the July build without the gate (same script, `ppr-directed-eval.py` rule: body links, mentions section excluded).
- 174 freshly created single-source pages with both related sections empty could be matched to their extraction response in the LM Studio server logs. 118 had empty `related_*` lists from the model. 56 had filled lists; of the names in them, 78 resolve to existing pages in the vault — pruned by the gate. Example: a page for *B-Vitamine* came out with empty related sections although its raw list named *Eisen*, *Kartoffeln*, *Reis*, *Vitamin D*, all existing pages.
- The generation prompt then receives "No related entities" and writes empty sections; nothing downstream can restore the edge.

**Proposed fix:** give the gate an `isKnownPage` predicate backed by the same title-then-alias resolver the related-link corrector uses (#482 stage 2). A dropped name that resolves stays in the related lists and is reported in the log line; a dropped name the vault does not know is pruned as today. The candidate itself still gets no page. PR follows.

DocTpoint:
Closes #620.

**What changes**
- `gateCandidates(analysis, text, language, isKnownPage?)`: a dropped name for which `isKnownPage` is true is not pruned from the survivors' `related_*` lists and is returned in `GateResult.linkedAnyway`. Without the predicate the behaviour is unchanged (all existing tests pass untouched).
- `buildVaultResolver(vaultIndex)` extracted from `correctRelatedLinkPrefixes` (code moved, not changed): title outranks alias, an ambiguous name resolves to neither, `entities/` and `concepts/` only. The corrector calls it; the engine passes `name => resolve(name) !== undefined` to the gate. "Known" means the same thing at both ends of the pipeline.
- Engine log line and progress notice list the names linked anyway.

**Test:** new case — a candidate absent from the note but present in the vault stays in the related lists and appears in `linkedAnyway`; a candidate the vault does not know is pruned as before.

**Known limit, left as is:** the engine uses its 5-second page-index cache. In a back-to-back batch, a page created by the previous note can be missing from a cache that is a few seconds old, and its name is then pruned exactly as before this fix. A fresh index per gate call would close that at one vault read per ingest; I did not want to add that cost without your view.

Gate 1 green locally (3793 tests).

### DIFF
--- a/src/__tests__/core/candidate-gate.test.ts
+++ b/src/__tests__/core/candidate-gate.test.ts
@@ -102,6 +102,29 @@ describe('gateCandidates', () => {
     expect(out.entities[0].related_concepts).toEqual(['Eisenmangel']);
     expect(out.concepts.map(x => x.name)).toEqual(['Eisenmangel']);
     expect(out.concepts[0].related_concepts).toEqual(['Ferritin']);
+    expect(out.linkedAnyway).toEqual([]);
+  });
+
+  it('keeps a dropped name in related_* lists when the vault already has a page for it', () => {
+    // Eisenstoffwechsel is absent from THIS note, so it earns no page here —
+    // but a page for it exists, and the edge to an existing page is not a
+    // dead link. CRP is not a candidate in this input, so it is never judged.
+    const known = (name: string) => name.toLowerCase() === 'eisenstoffwechsel';
+    const out = gate(
+      {
+        entities: [e('Ferritin', { related_entities: ['CRP', 'Transferrin'], related_concepts: ['Eisenstoffwechsel', 'Eisenmangel'] })],
+        concepts: [c('Eisenmangel', ['Eisenstoffwechsel', 'Ferritin']), c('Eisenstoffwechsel')],
+      },
+      TEXT,
+      'de',
+      known,
+    );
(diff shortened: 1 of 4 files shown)


green-dalii (approved):
Approve. #620 fix — gateCandidates gains isKnownPage predicate; a dropped name that resolves to an existing page keeps its related edge (linkedAnyway) instead of being pruned as a dead link. buildVaultResolver extracted (code moved, not changed) so 'known' means the same thing at both ends. Optional predicate = no behavior change without it. Gate-1 green.

PR #625, opened 2026-09-04
fix(query): stop-word substrings and the lex/PPR merge hid the page the user named

### LINKED ISSUE #623: Query: stop-word substrings and the lex/PPR merge hide the page the user named
"Kann man bei eingeschränkter Nierenfunktion Creatin nehmen?" loaded Kahneman, Karpman, Salbei, Pekannüsse and six more pages the query never named, while the page titled "Creatin" was not among them. Measured 2026-09-03 on a 3,025-page German vault, LM Studio backend; three mechanisms in `src/core/ppr-cascade.ts`:

1. `tokenizeQuery` uses an ASCII-only run class, so "über" becomes "über" + "ber" and "eingeschränkter" becomes "eingeschr" + "nkter" — fragments that match titles the query never named. Edge punctuation survives: "Creatin." and "nehmen?" are needles that can never match a title.
2. Needles match as substrings: "man" hits Kahneman and Karpman, "bei" hits Salbei, "kann" hits Pekannüsse, "creatin" hits Phosphocreatin — a different substance. 76 lex hits for one question.
3. `mergeWithPPR` takes `max(lexScoreOf, ppr)`. `lexScoreOf` is a rank placeholder in (0, 1]; PPR mass on a 3,000-node graph is about 0.05 for a seed. Any keyword hit therefore outranks every PPR seed, and the seeds the LLM keyword stage had just paid for are gone from the top ten.

**Proposed fix:** word runs over Unicode letters, digits and combining marks (not ASCII), edge punctuation stripped from whitespace tokens; a needle from a space-delimited script must start a word (prefix compounds and hyphenated forms still match, a tail compound does not), scripts without word spaces keep substring matching; PPR mass ranks, the lex hint only reorders within a tenth of the top mass and orders the pages PPR never reached below all reached ones. PR follows.

DocTpoint:
Closes #623.

**What changes** (`src/core/ppr-cascade.ts`)
- `tokenizeQuery`: word runs are `\p{L}\p{N}\p{M}` (RegExp constructor, the form `candidate-gate.ts` already uses under the ES6 target), so "über" stays whole, Cyrillic/Greek/Arabic queries keep their words, and NFD text keeps its umlauts inside the word. Edge punctuation is stripped from whitespace tokens.
- `needleHits(textLower, kw)`: a needle from a space-delimited script must start a word (`nierenfunktion` → Nierenfunktionsstörung, `kinase` → Creatin-Kinase; `creatin` → Phosphocreatin no longer). Han, kana, Hangul and Thai needles keep substring semantics — there is no boundary to find. Used by `scorePagesByNeedles` and `lexMatch`.
- `mergeWithPPR`: reached pages score `ppr + hint × 0.1 × maxPpr`; unreached pages `hint × minPpr / 2`. One monotone key, PPR ranks, the lex hint orders within a tenth of the top mass and the tail.

**Tests:** `lex-noise.test.ts` pins the three mechanisms (tokenizer, needle boundary incl. Cyrillic and NFD, merge order on a 40-node ring plus 12 lex-only pages). One existing assertion changed: `ppr-cascade.test.ts` expected `P0` first in the sparse arm; "page" matches all 50 titles, so P0/P1/P2 carry equal PPR mass and the old `toBe('P0')` held only through the `max()` defect. Now `toContain` over the three.

**Downstream:** `PageMatch.score` is read only in a debug line (`QueryView-class.ts`); the seed-selector gate reads `scorePagesByNeedles`, whose scale is unchanged.

**Left as is, on purpose:** a tail compound ("insuffizienz" in Niereninsuffizienz) is left to aliases, the keyword stage and the graph walk — the measured false positives all came from tails. Rank fusion instead of the two weights would remove the constants; not done here, the two constants are the smallest change that keeps PPR the ranker.

Gate 1 green locally (3945 tests).

### DIFF
--- a/src/__tests__/core/lex-noise.test.ts
+++ b/src/__tests__/core/lex-noise.test.ts
@@ -1,0 +1,165 @@
+/**
+ * Query lex noise — measured on a 3,025-page German vault (2026-09-03).
+ *
+ * "Kann man bei eingeschränkter Nierenfunktion Creatin nehmen?" loaded
+ * Kahneman, Karpman, Salbei, Pekannüsse and six more pages the query
+ * never named, while the page titled "Creatin" was not among them.
+ * Three mechanisms, each pinned below:
+ *
+ *   1. tokenizeQuery split "über" into "über" + "ber" (ASCII-only run
+ *      class), double-counting every title "über" already hit, and
+ *      kept "Creatin." with its full stop as a needle.
+ *   2. Tokens matched as substrings: "man" → Kahneman, "creatin" →
+ *      Phosphocreatin (a different substance).
+ *   3. mergeWithPPR took max(lexRankHint ∈ (0,1], pprMass ≈ 0.05), so
+ *      any keyword hit outranked every PPR seed.
+ */
+import { describe, it, expect } from 'vitest';
+import {
+  tokenizeQuery,
+  needleHits,
+  lexMatchByTitleAndAliases,
+  pprCascade,
+  type PageRef,
+  type Graph,
+} from '../../core/ppr-cascade';
+import { makeRng } from '../__support__/rng';
+
+function page(path: string, title: string, aliases: string[] = []): PageRef {
+  return { path, title, aliases };
+}
+
+describe('tokenizeQuery — Latin diacritics and edge punctuation', () => {
+  it('keeps umlaut words whole and emits no ASCII fragments', () => {
+    const tokens = tokenizeQuery('Gib mir Informationen über Creatin.');
+    expect(tokens).toContain('über');
+    expect(tokens).toContain('creatin');
+    expect(tokens).not.toContain('ber');
+    expect(tokens).not.toContain('creatin.');
+  });
+
+  it('does not split a word at a diacritic', () => {
+    const tokens = tokenizeQuery('eingeschränkter Nierenfunktion');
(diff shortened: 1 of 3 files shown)


green-dalii:

### #625 — fix(query): stop-word substrings and the lex/PPR merge hid the page the user named

Own maintainer review:

**Verified against the diff**: three mechanisms confirmed — `WORD_CHAR_CLASS=\p{L}\p{N}\p{M}` via RegExp constructor (same form candidate-gate.ts uses under the ES6 target), new export `needleHits` (word-start for spaced scripts, substring kept for CJK/Thai), `mergeWithPPR` reweight (`ppr + hint×0.1×maxPpr`, unreached `hint×minPpr/2`). The `toBe('P0')`→`toContain` change is honest — the old assertion held only through the `max()` defect being removed. `PageMatch.score` confirmed read only at QueryView-class:1148 debug line; `scorePagesByNeedles` second reader is select-seeds Stage 1.5b keyword scan — gets stricter in the same direction, scale unchanged, no action needed.

**Gate 1**: CI SUCCESS. **Three-zero**: no breaking (pure functions + 1 export), no side effect, no perf regression (complexity unchanged; `LEX_HINT_WEIGHT=0.1` pinned by test).

PR #626, opened 2026-09-04
fix(links): re-point folder-wrong links everywhere on the page, on the source page, and once more when the run is over

### LINKED ISSUE #624: 314 links point at the wrong folder, and the corrector could not see three quarters of them
Measured 2026-09-03 on a 3,025-page German vault: 314 wikilinks are typed `concepts/X` or `entities/X` while the page X lives under the other folder — a dead link in the reader and a missing edge in the query graph (`entities/Creatin` → `[[concepts/Phosphocreatin]]` among them). `correctRelatedLinkPrefixes` (#482 stage 2) could not have caught 279 of them:

- 212 stand on source pages (Key Entities, Key Concepts, Core Content). `createSummaryPage` is the one writer that never calls the corrector.
- 67 stand in the prose of entity and concept pages (Description, Core Content). The corrector is scoped to the two Related sections by design; every other line passes through untouched.
- 35 sit in Related sections of pages whose target was created later in the same ingest run: at write time the page did not exist, the corrector trusted the extraction's folder, and dedup then placed the sibling in the other folder (#589 — the folder is a decision, not a property of the name). No write-time pass can reach this class.

**Proposed fix:** (1) outside the Related sections a narrower pass with the vault index only — a folder-typed link whose page the vault holds under the other folder is re-pointed; bare `[[X]]`, `sources/` and unknown names are left alone; (2) `createSummaryPage` runs the same corrector with the analysis lists and the two Key sections in the role of the Related sections; (3) an end-of-run pass over the pages this run wrote, resolving against the pages of this run only (one read per page, never a vault scan). PR follows. The 314 existing links are a separate, one-off repair.

DocTpoint:
Closes #624.

Two commits.

**1. `correctRelatedLinkPrefixes` outside the Related sections, and on the source page**
- New prose pass: a link already typed `entities/X` or `concepts/X` is re-pointed when the vault holds X (title, then curated alias) under another path. Bare `[[X]]` stays (no section to type it), `sources/` is never re-typed, an unknown name is left alone — no typed-list fallback in prose. Case-only differences are not a move: Obsidian resolves them, and rewriting would cost a write.
- `createSummaryPage` calls the corrector before writing, with the analysis lists as the typed lists and Key Entities / Key Concepts as the section labels. Uses the engine's cached page index.

**2. Stage 4.5 `repointLinksAfterRun`** — after the related-page updates, before contradictions and index: every page this run created or updated is read once, each folder-typed link is resolved against the pages of this run (title from the file name, aliases from the frontmatter), and pages that moved a link are written through the single write gate. No model, no typed lists. A `reviewed: true` page is not touched (it still lends its title and aliases to the index). Pages that come out byte-identical are not written; a failure on one page is logged and does not stop the others or fail the ingest. `repointFolderTypedLinks` is the pure function behind it; `buildVaultResolver` from #621 is shared.

**Tests:** corrector — prose re-point, source-page section, alias in prose, correct link byte-identical, bare/`sources/` untouched, unknown name untouched, no index no-op. Engine harness — Creatin links `[[concepts/Phosphocreatin]]` in prose and in Related Concepts, Phosphocreatin is created as an entity afterwards, both links and the source page read `entities/…` at the end of the run; a page with nothing to move is not rewritten; the pass adds no vault scan (measured by difference against a run whose pass has nothing to do).

**Left as is, on purpose:** the write-time pass could be dropped in favour of one deferred correction over the whole run — fewer writes, but it moves the corrector out of three page-factory writers; too large for this fix. `[[concepts/X#heading]]` is not resolved (the anchor is part of the name to the resolver); no template emits such links today. The 314 links already on disk are not touched by the plugin — a one-off repair on the vault side.

Gate 1 green locally (3941 tests).

### DIFF
--- a/src/__tests__/core/related-link-corrector.test.ts
+++ b/src/__tests__/core/related-link-corrector.test.ts
@@ -307,3 +307,56 @@ describe('correctRelatedLinkPrefixes — full-vault resolution (#482)', () => {
     expect(r).toBe(c);
   });
 });
+
+describe('correctRelatedLinkPrefixes — folder-typed links outside the Related sections', () => {
+  // Measured 2026-09-03 on a 3,025-page vault: 314 links pointed at
+  // `concepts/X` or `entities/X` while the page lived in the other folder;
+  // 279 of them stood outside the two Related sections and were never seen.
+  const ENT = 'Related Entities';
+  const CON = 'Related Concepts';
+  const vault = (pages: Array<{ path: string; title: string; aliases?: string[] }>) =>
+    ({ wikiFolder: 'wiki', pages });
+  const creatinVault = vault([
+    { path: 'wiki/entities/Phosphocreatin.md', title: 'Phosphocreatin' },
+    { path: 'wiki/entities/Skelettmuskeln.md', title: 'Skelettmuskeln', aliases: ['Skelettmuskulatur'] },
+    { path: 'wiki/concepts/ATP.md', title: 'ATP' },
+  ]);
+
+  it('re-points a folder-wrong link in prose to the page the vault holds', () => {
+    const c = '## Description\nRegeneration von [[concepts/ATP|ATP]] über [[concepts/Phosphocreatin|Phosphocreatin]].';
+    const r = correctRelatedLinkPrefixes(c, [], [], ENT, CON, creatinVault);
+    expect(r).toBe('## Description\nRegeneration von [[concepts/ATP|ATP]] über [[entities/Phosphocreatin|Phosphocreatin]].');
+  });
+
+  it('re-points on a source page section the corrector was never scoped to', () => {
+    const c = ['## Core Content', 'Stored in [[concepts/Skelettmuskeln]].'].join('\n');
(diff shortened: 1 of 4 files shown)


green-dalii:

### #626 — fix(links): re-point folder-wrong links everywhere on the page, on the source page, and once more when the run is over

Own maintainer review:

**Verified against the diff**: self-contained — `buildVaultResolver` confirmed already on main (no cross-PR dependency). `FOLDER_TYPED_LINK_RE` never touches `sources/`; bare links untouched; case-only differences not rewritten (Obsidian resolves them). `reviewed:true` pages lend titles/aliases but are never written; byte-identical pages not written; per-page try/catch isolation. Scope discipline good: the write-time-pass→deferred-full-run alternative explicitly rejected as too large.

**Watch item**: Stage 4.5 adds one read per written page — the `[Time] Link re-point pass` log added in this PR is the right instrument; revisit chunking only if large-vault runs show it hot.

**Gate 1**: CI SUCCESS. **Three-zero**: no breaking (reviewed exemption preserved, no ID/schema change), no side effect (deterministic, no model), no perf regression beyond the intended bounded read pass.

PR #629, opened 2026-09-04
fix(llm): apply the per-step thinking policy on the stream path too

### LINKED ISSUE #627: Per-step thinking policy is not applied on the stream path — the streamed answer thinks with "*=default:off"
Measured 2026-09-04 on a 3,025-page German vault, LM Studio backend, `taskPolicies` set to `*=default:off`. One Query Wiki question, two calls in the server log:

- keyword extraction (`query-keywords`): `reasoning_effort: none` on the wire, 3 s.
- streamed answer (`query-wiki`): no `reasoning_effort` field at all, 9,966 prompt tokens, 12 s prompt processing, 43 s generation — 55 s for the answer the user waits for. The follow-up turn's history carries the `<think>` block of that answer.

Cause: `wrapWithAdvancedSettings` applies `applyTaskPolicy` on `createMessage` and `createMessageWithOutput`, and skips `createMessageStream` on purpose — the comment says the stream carries no task label. That was true before #469; since #469 the stream is labelled `query-wiki` and the wrapper already accounts it under that name, but the policy is never consulted. The only switch the stream honours is the legacy `disableThinking`.

Proposed fix: take the thinking half of the policy on the stream path as well (no output mode on a stream; the stream interface has no effort level, so a named level rides as `enableThinking: true`). PR follows.

DocTpoint:
Closes #627.

**What changes** (`src/llm-client-wrapper.ts`): `createMessageStream` now spreads `enableThinking` from `applyTaskPolicy(params.task, settings)` after the advanced-settings overlay — the same precedence as the other two paths, so a step-named policy overrides the call site's `disableThinking`. Nothing else on the stream is touched; without a policy the request is byte-identical.

**Tests** (`llm-client-wrapper.test.ts`, stream block): `*=default:off` reaches the stream as `enableThinking: false`; `query-wiki=default:on` overrides a call-site `false`; no policy adds no field.

**Not in this PR:** `reasoningEffort` on the stream (its interface does not carry the field) and the summary-instead-of-full-content question for the answer prompt (12 s of the 55 s were prompt processing of ~8.7k uncached tokens; a design question, not a fix).

Gate 1 green locally (3933 tests).

### DIFF
--- a/src/__tests__/core/llm-client-wrapper.test.ts
+++ b/src/__tests__/core/llm-client-wrapper.test.ts
@@ -150,6 +150,36 @@ describe('wrapWithAdvancedSettings — per-step accounting', () => {
       expect(new Map(taskUsageSince(before)).get('untagged')?.calls).toBe(1);
     });
 
+    it('applies the thinking policy on the stream path — the answer the user waits for', async () => {
+      // Measured 2026-09-04: with `*=default:off` the keyword call sent
+      // reasoning_effort none and the streamed answer sent nothing, so the
+      // model thought for 43 of its 55 seconds.
+      const { client, calls } = streamClientSpy();
+      await wrapWithAdvancedSettings(client, {
+        maxTokensPerCall: 0,
+        taskPolicies: { '*': { outputMode: 'default', thinking: 'off' } },
+      }).createMessageStream!({ ...CALL, task: 'query-wiki', onChunk: () => {} });
+      expect(calls[0]?.enableThinking).toBe(false);
+      expect(calls[0]).not.toHaveProperty('outputModeOverride');
+      expect(calls[0]).not.toHaveProperty('reasoningEffort');
+    });
+
+    it('the policy overrides disableThinking from the call site on the stream too', async () => {
+      const { client, calls } = streamClientSpy();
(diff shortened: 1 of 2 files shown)


green-dalii:

Same platform-constraint pattern as #630 (`gh pr comment` as formal review, then `--admin --squash`).

### #629 — fix(llm): apply the per-step thinking policy on the stream path too

Own maintainer review:

**Verified against the diff**: stream branch adds 3 lines (`applyTaskPolicy(params.task)` + conditional spread) — same precedence as the other two paths. The old comment "stream carries no task label" is indeed ended by #469 (main `llm-client-wrapper.ts:163` still carries the stale note — this PR updates it). The no-policy byte-identical test is the right lock; `reasoningEffort` exclusion is an interface limit, honestly declared as Not-in-PR.

**Root-cause note**: this is the actual cause behind the earlier "streamed answer thinks 43/55s" report (not b302aab) — the policy reached every step except the one the user waits for.

**Gate 1**: CI SUCCESS. **Three-zero**: no breaking (no-policy path byte-identical, test-locked), no side effect, no perf regression (thinking off only reduces tokens/time).

PR #630, opened 2026-09-04
fix(query): a concept page reaches the answer prompt as its description, not its one-line definition

### LINKED ISSUE #628: Query answer prompt hands a concept page over as its one-line Definition, not its Description
The answer prompt loads an entity or concept page as one section (`extractSummaryFromPage`, capped at MAX_PAGE_CONTENT_CHARS / 3). For a concept the primary section is `## Definition`. That matches the generation template, which gives a concept only a Definition. But the merge template — the one that rewrites the page on every further source — writes a `## Description` ("core definition and significance, 3–6 sentences") above it, and that is where the page's substance accumulates. The extractor keeps reading the single-source sentence.

**Measured 2026-09-04** on a 3,025-page German vault:
- 241 of 1,266 concept pages carry both sections; every one of them is multi-source. 183 have a Definition under 400 characters and a Description at least three times as long (median ratio 5.2).
- "Welche zusätzlichen Maßnahmen sind bei Osteoporosebehandlung sinnvoll?" selected `concepts/Osteoporose` as the first seed and loaded it in fourth place. What reached the model was its 120-character Definition ("…verminderte Knochendichte…"). The 3,100-character Description — exercise, dysbiosis, D3 + K2, bisphosphonates, denosumab — did not. The answer said that no further measures were on the pages, and cited the two source pages, which are the only pages loaded in full.

The prompt's own label, "Relevant Wiki Pages (loaded with full content)", does not describe this; I have left the wording to you.

**Proposed fix:** the Description is primary for both page types, the Definition is the fallback for a page that has nothing else. PR follows.

**A prompt suggestion on the side, for your editorial call, not part of the PR:** every answer opens with "Basierend auf den verfügbaren Wiki-Seiten…" — the model echoing "Answer based on the Wiki pages above" before following it. Inside the plugin's chat the reader knows where the answer comes from. As a custom query instruction this line removes the preamble without loosening the rule:

> Start with the answer itself. No preamble such as "Based on the available wiki pages" — the reader knows the answer comes from the wiki.

Verified on the vault above; it might be worth a place in the default prompt.

DocTpoint:
Closes #628.

**What changes** (`src/core/section-extractor.ts`): `extractSummaryFromPage` matches `## <descriptionLabel>` first for both page types and falls back to `## <definitionLabel>`. `pageType` stays in the options for the callers and no longer selects the label. Only caller is `load-pages.ts`; no signature change.

**Tests:** the pinned rule "concept page with both sections uses Definition" is inverted to Description, with the reason in the test; a new case carries both sections in the shape the merge template writes (Description above Definition) and expects the Description. A concept page with only a Definition still returns it (existing test, unchanged).

**Not in this PR:** the prompt line "loaded with full content" (your editorial surface); the seed tie-break in the keyword scan, where equal-score pages fall back to index order (separate note).

Gate 1 green locally (3931 tests).

### DIFF
--- a/src/__tests__/core/section-extractor.test.ts
+++ b/src/__tests__/core/section-extractor.test.ts
@@ -67,6 +67,29 @@ describe('extractSummaryFromPage — English baseline', () => {
     })).toBe('A formal definition.');
   });
 
+  it('prefers ## Description on a concept page that carries both sections', () => {
+    // The concept template writes a one-sentence Definition above the
+    // Description that holds the page's substance; the query prompt used to
+    // take the sentence and drop the substance (measured 2026-09-04: 183 of
+    // 241 concept pages with both sections).
+    const body = [
+      '## Description',
+      'Osteoporosis is multifactorial: vitamin deficiency, inactivity and dysbiosis. Therapy combines D3 + K2 with resistance training.',
+      '',
+      '## Definition',
+      'A disease of reduced bone density.',
+      '',
+      '## More',
+      'bar',
+    ].join('\n');
+    expect(extractSummaryFromPage(body, {
+      descriptionLabel: EN_DESC,
+      definitionLabel: EN_DEF,
+      pageType: 'concept',
+      maxChars: 1000,
+    })).toBe('Osteoporosis is multifactorial: vitamin deficiency, inactivity and dysbiosis. Therapy combines D3 + K2 with resistance training.');
+  });
+
   it('returns content up to the next ## header', () => {
     const body = [
       '## Description',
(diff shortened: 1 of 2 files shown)


green-dalii:

GitHub platform constraint: `gh pr review --approve` on own-maintainer review of a contributor PR from this account's automation — posting the formal review via `gh pr comment` per AGENTS audit-trail pattern (same as #632/#633/#634), then `--admin --squash`.

### #630 — fix(query): concept page reaches the answer prompt as its description

Own maintainer review:

**Verified against the diff**: `load-pages.ts:85` is the only production caller of `extractSummaryFromPage` (QueryView-class:1157 only caches labels for that call) — blast radius is the answer summary only, no write path touched. Inverting the pinned rule inside the test with the reason recorded is exactly right. `pageType` kept for callers, no signature change.

**Gate 1**: CI `Gate 1 / Five-Gate` SUCCESS. **Three-zero**: no breaking (signature unchanged, summary still valid), no side effect (pure order swap), no perf regression (Description is longer — necessary cost of fixing information loss, not a regression).

PR #631, opened 2026-09-04
fix(merge): a paragraph another source footnoted may not vanish from a rewrite

DocTpoint:
**What changes.** `src/core/paragraph-provenance.ts`: `preserveSourcedParagraphs` runs after `preserveExistingSections`. Every paragraph or list item of the old body that ends in an inline footnote carrying a wikilink (`^[Quelle: [[X]]]`, `^[Source: [[X]]]`, any label) is matched into the same section of the rewrite by word overlap (`PARAGRAPH_KEEP_OVERLAP` 0.5). Found without its footnote: the footnote is re-attached. Not found: the paragraph is put back after the nearest preceding paragraph that survived, unless every footnote on it names the source being merged. `guardBodyRewrite` composes the three deterministic layers (sections, sourced paragraphs, H1); both rewrite paths call it. `sectionIdentityKey` is the identity both guards compare under.

**Why.** On 571 rewrite pairs of one vault (720 footnoted paragraphs), 15 % came back with less than half their words anywhere on the page, 102 of those 111 dropped by a rewrite a different source triggered; of the paragraphs whose text survived, 23 % lost their footnote. The shrink floor (#618) sees a collapsed section, not one paragraph of five. Ownership is the variable, not size.

**Tests.** 13 cases: restore position, own-source exemption (folder, case, NFC), re-attachment, fusion of two sources, section-scoped matching, list items, lead and Mentions excluded, composition with the section guard, threshold boundary, `guardBodyRewrite` order. Replay over the same 571 pairs: 103 rewrites touched, 120 paragraphs restored, 164 footnotes re-attached.

**Not in this PR.** The footnote convention itself: the built-in prompts do not ask for one, so the guard is a no-op on a vault without it. `mergeDuplicatePages` and `resolveContradiction` adopt full-body rewrites without either guard (separate note if wanted). The threshold is a regulator: at 0.5 a reworded definition with 45 % shared words comes back next to its rewrite.

Gate 1 green locally (3942 tests).

### DIFF
--- a/src/__tests__/core/paragraph-provenance.test.ts
+++ b/src/__tests__/core/paragraph-provenance.test.ts
@@ -1,0 +1,136 @@
+import { describe, it, expect, vi, afterEach } from 'vitest';
+import { PARAGRAPH_KEEP_OVERLAP, guardBodyRewrite, preserveSourcedParagraphs } from '../../core/paragraph-provenance';
+import { preserveExistingSections } from '../../core/section-header-canonicalizer';
+
+const DE = [
+  'Beschreibung', 'Hauptmerkmale', 'Verwandte Konzepte', 'Verwandte Entitäten',
+  'Erwähnungen in der Quelle', 'Neue Informationen',
+];
+const M = 'Erwähnungen in der Quelle';
+
+const P_A = 'Berberin senkt den Nüchternblutzucker über eine Aktivierung der AMPK in Leber und Muskel deutlich. ^[Quelle: [[Berberin]]]';
+const P_B = 'Chrom verbessert die Insulinsensitivität der Zielgewebe nur bei nachgewiesenem Mangel messbar. ^[Quelle: [[Chrom]]]';
+const P_C = 'Pektine verzögern die Magenentleerung und flachen den postprandialen Glukoseanstieg ab. ^[Quelle: [[entities/Pektine|Pektine]]]';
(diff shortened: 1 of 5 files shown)


green-dalii:

### #631 — fix(merge): a paragraph another source footnoted may not vanish from a rewrite

Own maintainer review (deepest in this batch — largest PR):

**Verified against the diff**: `guardBodyRewrite` composes three deterministic layers (sections → sourced paragraphs → H1); both rewrite paths (merge-page, related-page) call it identically. `sectionIdentityKey` extraction is mechanical. Dead-export check passes — `reassertH1` still has its `merge-duplicates.ts:119` caller, `preserveExistingSections` documented in `markdown.ts:26`. The 0.5 threshold regulator trade-off (reworded definition at 45% overlap comes back next to its rewrite) is stated honestly in the PR. 13 tests + 571-pair replay (103 touched / 120 restored / 164 re-attached).

**Recorded note**: `mergeDuplicatePages`/`resolveContradiction` stay unguarded — rewrite paths are now 2-guarded/1-unguarded until the follow-up. Accepted as declared follow-up scope, not a blocker.

**Gate 1**: CI SUCCESS. **Three-zero**: no breaking (merge output stays valid; own-source exemption prevents loops), no side effect (deterministic, no model), no perf regression (word-overlap O(p²) on small p).

**Decision: APPROVE → merge into v1.27.x PATCH (last in batch).**

PR #636, opened 2026-09-05
feat(related): siblings link each other, known pages keep their title, and the Related sections are rendered from the lists

### LINKED ISSUE #635: Related lists: sibling pages never link each other, and the model misspells the names it is handed
**Observation.** The extraction names anything from the note in `related_entities` / `related_concepts`, and nothing downstream checks whether a page exists or will. Measured on a rebuild of a 413-note German vault after 136 notes (v1.27.0, `gemma-4-26b` via LM Studio):

- 676 dead related entries on 870 pages; 795 distinct unresolved targets stood beside 870 real pages.
- **101 of 103 pages without a single live outgoing link had a sibling page born from the same note** — an edge the engine knew (it planned both pages) and never wrote, because the model only lists what the note's text suggested to it.
- 265 dead links were spelling variants of pages the vault already had (`Interleukin 6` for `Interleukin-6`, `Vitamine/Vitamin C` as an invented path, `H₂O₂` re-encoded): the list handed to the generation prompt was right, the model's transcription of it into the two Related sections was not.

**Why this is a gate-shaped problem, not a prompt problem.** #514 put the deterministic half of "which names deserve a page" into code. The same argument applies one step later: which names the page links to is known once the run is planned (siblings, existing pages under their title), and rendering that list is a deterministic property, not a model judgement. #621 already keeps a known page's edge through the gate; this closes the remaining two leaks on the same principle.

**Proposal** (PR follows): after the candidate gate and before any page is planned, shape the lists deterministically — siblings link each other by folder kind, a name the vault answers is written under that page's own title and kind, a tag value is not a page and is dropped, a name nothing answers is left exactly as today and counted in the log. Then write the two Related sections from these lists instead of asking the model to transcribe them; on a rewrite the entries the page already had stay in front, re-resolved. No model call, no new setting.

**Measured with the change** on the same vault: 0 pages without a live outgoing link (was 103 of 870), sibling lists complete on 246 of 246 pages, related entries dead 1 % (was 24 %).

DocTpoint:
Closes #635.

**What changes**
- `core/related-shaping.ts` — `shapeRelatedLists(analysis, deps)`: after the gate, before planning. Siblings by kind; a name the vault answers (title or alias, a parenthetical tried through its parts) under the page's own title and kind; a tag value (`Thema/Therapie`, or the bare leaf when only the vocabulary knows the word) dropped; a name nothing answers kept as written and counted. Pure over the analysis and a resolver; no I/O.
- `core/related-sections.ts` — `renderRelatedSections`: the two Related sections are written from the typed lists; vault path when known, planned path otherwise, the name as display text. Entries from the body before a rewrite stay in front, re-resolved against the vault, one target once across both sections; a missing section is inserted in the order the page kind's template lists them. Headings are matched with the same tolerance `section-header-canonicalizer` gives them. Everything outside the two sections is byte-identical.
- `page-factory/related-links.ts` — one helper for create/merge/related-page: prefix correction (#482 stage 2) then rendering. The three call sites collapse to one line each.
- Engine: one block in `ingestSource` builds the resolver from the page index (title outranks alias, ambiguous resolves to neither — `buildVaultResolver` from #621), lists watched-folder notes through `isIngestableSource`, and logs `[related-shape]` with the counts.

**Measured** (413-note vault, rebuild): pages without a live outgoing link 103 → 0 of 870; sibling lists complete 246/246; related entries dead 24 % → 1 %; the misspelt-name class gone because the model no longer writes those lines.

**Default behaviour changes, deliberately without a switch.** The two Related sections are now owned by the list on every write. I did not add a setting because the previous behaviour was not a choice anyone made — the model transcribed a list it had been handed and got it wrong one time in four — and because a switch would keep two rendering paths alive for one section. If you would rather have it opt-in for one release, the helper has a single entry point and that is a five-line change.

**Left as follow-ups, not in this PR:** the generation/merge prompts still ask for the two sections that are now overwritten (drop them from the templates, saves output tokens); `buildVaultResolver` could return `{path, title, kind}` so the engine block stops wrapping it; the parenthetical resolution could move into the resolver so the corrector and the repoint pass see it too; the section-scoped half of `correctRelatedLinkPrefixes` is now redundant. Each is a small separate change on top of this one.

Tests: 4 new files / cases (shaping, rendering incl. NFC and cross-section dedup, engine hookup, link-repoint expectation), full suite 3993 green, typecheck clean.

### DIFF
--- a/src/__tests__/__support__/wiki-engine-harness.ts
+++ b/src/__tests__/__support__/wiki-engine-harness.ts
@@ -71,6 +71,7 @@ export function createWikiEngineHarness(opts: HarnessOptions = {}): WikiEngineHa
 
   const app = {
     vault: {
+      configDir: '.obsidian',
       read: async (f: { path: string }) => files.get(f.path) ?? '',
       create: async (p: string, c: string) => { files.set(p, c); },
       process: async (f: { path: string }, fn: (d: string) => string) => {
--- a/src/__tests__/core/related-sections.test.ts
+++ b/src/__tests__/core/related-sections.test.ts
@@ -1,0 +1,74 @@
+// See core/related-sections.ts.
+import { describe, it, expect } from 'vitest';
+import { renderRelatedSections } from '../../core/related-sections';
+
+const vault = new Map([['metformin', 'entities/Metformin'], ['oxidativer-stress', 'concepts/Oxidativer-Stress']]);
+const opts = { preserveCase: true, resolve: (n: string) => vault.get(n.toLowerCase().replace(/\s+/g, '-')) };
+const E = 'Verwandte Entitäten'; const C = 'Verwandte Konzepte';
+
+const page = [
+  '# Berberin', '', '## Beschreibung', '', 'Text mit [[entities/Metformin|Metformin]].', '',
+  `## ${E}`, '', '- [[entities/Metformin|Metformin]]', '- [[entities/Vitamine/Vitamin C|Vitamin C]]', '',
+  `## ${C}`, '', '- [[concepts/Oxidativer_Stress|Oxidativer Stress]]', '',
+  '## Erwähnungen in der Quelle', '', '- "…"', '',
+].join('\n');
+
+describe('renderRelatedSections', () => {
+  it('writes both sections from the lists: vault path when known, planned path otherwise, name as display', () => {
+    const out = renderRelatedSections(page, ['Metformin', 'Vitamin C'], ['Oxidativer Stress', 'H₂O₂'], E, C, opts);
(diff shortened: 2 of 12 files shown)


green-dalii (approved):
**Decision: Approve.** The shaping/section modules are Unicode-safe end to end (`\p{L}` prefix match, NFC slugKeys) — verified `主题/糖尿病`-style tags drop the same as `Thema/Therapie`, so this is not German-only; the tagLeaf prefix-width edge (space-containing groups like `Health Sciences/Neurology`) is a non-blocking backlog note, not a follow-up commit.

Locally re-ran the full suite: 3993 green, tsc/lint clean, CI Five-Gate pass. `related-links.ts` collapses the three factories to one entry point cleanly; `related-shaping.ts` sibling/self/dup exclusion and the tag-vs-note precedence (shaping test 6) cover the branch space; `related-sections.ts` NFC single-key + cross-section single-target + folder-follow re-resolution are the right rendering semantics.

The no-switch default-behaviour change reads right — the old behaviour was model transcription of a handed list, not a deliberate choice, and the five-line opt-in stays available if a release in the wild disagrees. Follow-ups (template token saving, resolver shape, corrector redundancy) stand correctly apart.

PR #645, opened 2026-09-06
fix(related): siblings only rescue an orphan, capped at three

### LINKED ISSUE #644: Sibling edges from #636 form a clique per note — 99 % of live edges on a rebuild
## What

Since #636 every page born from a note links every other page born from the
same note. On a rebuild the graph is that rule and little else.

Measured today on a 537-page rebuild (78 notes ingested, 1.27.1):

| | |
|---|---|
| live edges (links to an existing page) | 8673 |
| of those, sibling edges (targets sharing a source note) | 8562 = 99 % |
| pages whose only live edges are siblings | 462 of 537 |
| largest cliques (pages per note) | 20, 19, 19, 18, 18, 18, 18 |
| notes with a clique of 10+ pages | 56 |

A note that yields 20 pages writes 20 × 19 = 380 edges. A page with three
sources joins three cliques and bridges them. Any neighbourhood measure
(PPR, related-by-graph) sees co-birth, not content.

## Why it went unnoticed

The clique is per note, not per batch — a single ingest of a rich note
produces it in any wiki. What the rebuild changes is the share: with the
model's own related names still unresolved, the cliques are nearly the whole
live graph and the graph view shows them as dense clumps. Both measures
#636 pre-registered (dead related entries, pages without a live outgoing
link) improve with cliques, so neither could catch it. The share of sibling
edges was never a measure.

The motivation for #636 was the orphan: 101 of 103 pages without a live
outgoing link had a sibling. Writing siblings for every page was the
over-fulfilment. The relation is also redundant — both siblings already
link the source page in `sources:`, so the graph has the path.

## Proposal

Siblings only rescue an orphan: a page whose own related names reach nothing
alive (no page of this run, no vault page, no note that will become one)
gets up to three siblings. Every other page keeps only what the model named.
Behaviour is unchanged for the orphan #636 set out to rescue; the log line
keeps counting sibling edges added.

Draft PR follows. We are running the rebuild with it now; numbers after the
second pass over the same 413 notes.

DocTpoint:
Closes #644.

`shapeRelatedLists` adds siblings only when a page's own related names reach
nothing alive — no page of this run, no vault page, no note that will become
one — and then at most `SIBLING_CAP = 3`. A self-link counts as nothing
(review finding: the placed-but-deduplicated case). Every other page keeps
what the model named; the log line keeps counting sibling edges added.

Measured on a 537-page rebuild: 8562 of 8673 live edges were sibling edges,
cliques of up to 20 pages per note, 462 pages with no other live edge.
Numbers and reasoning in #644.

Ready: the rebuild ran with this rule (413 notes, 2 095 pages) — orphans 9 = 0.4 %, the rule's footprint a spike at exactly three entries on single-source pages. The sibling-edge share promised above turned out to be the wrong measure; the correction and numbers are in #644. Gate 1 green (3995 tests).

### DIFF
--- a/src/__tests__/core/related-shaping.test.ts
+++ b/src/__tests__/core/related-shaping.test.ts
@@ -1,6 +1,6 @@
 // See core/related-shaping.ts.
 import { describe, it, expect } from 'vitest';
-import { shapeRelatedLists } from '../../core/related-shaping';
+import { SIBLING_CAP, shapeRelatedLists } from '../../core/related-shaping';
 import type { EntityInfo, ConceptInfo } from '../../types';
 
 const ent = (name: string, rel: Partial<EntityInfo> = {}): EntityInfo =>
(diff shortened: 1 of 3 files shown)


green-dalii (approved):
Verified against the diff. `hasLive` gates the sibling loop, `SIBLING_CAP = 3` bounds it, and a self-link correctly counts as no way out — the `seen` set is seeded with `nameKey(self)` and `put` returning `false` is exactly what makes the `related_entities: ['berberin']` case an orphan, pinned by the review-fix test in `related-shaping.test.ts`. Unanswered names deliberately not counting as live is the right call (`['Niemand', 'A2', 'A3', 'A4']`).

Clean scope (1 core + 2 test files, no contamination), tests assert real behaviour, and this is a fix to a regression v1.27.1 shipped two days ago. The 8673 → content-scale edge reduction is a measurable win for PPR and related-by-graph, so this is a net performance improvement rather than a risk.

Merging. One note for the follow-up: with the clique gone, `#664` (Related lists grow ~2 entries per source, never pruned) becomes the next binding constraint on that graph — worth landing the ranked cap before the next rebuild measurement.

PR #647, opened 2026-09-06
fix(ingest): a cancel reaches the running model call and stops at the next page write

### LINKED ISSUE #646: Cancel reaches only the PDF conversion call — every other model call runs to its end, and #583's cleanup can be skipped
## What happened

Rebuild on 1.27.1, 06.09.: stop pressed while Interleukin-6 was in a merge
call; Obsidian closed within the minute. The call was still running, so
the cancel was never observed: #583's cleanup did not run, the summary
page stayed with `generation_complete: true`, and the frontier picker
treated the note as done. Same end state #582 described, reached through
a different door.

## Why

`cancelIngestion()` aborts the engine's AbortController, but the signal is
threaded into exactly one call — `convertPdfToMarkdown` (wiki-engine.ts,
the v1.25.0 "Bug D" fix). `LLMClient.createMessage` has no abort
parameter, so extraction, merge, related rewrite and every other call run
to completion. The cancel is honoured at three `checkCancelled()`
checkpoints per ingest — between stages, not between writes. A local
model spends minutes per call; a stop pressed inside a call is invisible
for that long, and closing Obsidian in that window skips the cleanup.

## Proposal

1. `createMessage` takes an optional `abortSignal`; the four SDK clients
   pass it to `generateText`/`streamText` (AI SDK v6 aborts the HTTP
   request). The engine passes its controller's signal from the one
   dispatch site (`core/llm-dispatch.ts`) so no call site changes.
2. `checkCancelled()` at the write gate (`createOrUpdateFile`,
   wiki-content branch): a cancel lands before the next page write and
   the AbortError branch trashes the summary page. One harness test is
   red without it (cancel during a content call, page body arrives,
   nothing may be written).

PR with both follows.

DocTpoint:
Closes #646.

Two halves:

**The signal reaches the call.** `LLMClient.createMessage` /
`createMessageWithOutput` / `createMessageStream` take an optional
`abortSignal`; the four SDK clients hand it to `generateText`/`streamText`
through `buildSamplingArgs`, which every real call already spreads (the
Codex client inline). The engine hands its client out through one getter
and wraps it there (`core/llm-abort.ts`): a Proxy that adds the engine's
current signal to the three call shapes and leaves every other member
bound to the client. A caller-set signal wins; outside an ingest there is
no controller and nothing is added.

On the default fetch path (native fetch, `streamWithFallback`) the HTTP
request is aborted. On the `requestUrl` fallback Obsidian cannot abort a
request in flight; the bridge throws the AbortError as soon as the
response arrives, which is still before any page is written.

**The write gate checks first.** `createOrUpdateFile` calls
`checkCancelled()` before writing, so a cancel lands before the next page
write even when a call was already returning; the AbortError branch then
trashes the summary page as #583 intended.

5 tests: wrapper (three shapes, caller-set signal, live read), sampling
args, compat client forwards the signal, and a harness test — cancel
during a content call, the page body arrives, nothing may be written —
that is red without the change.

Running the rebuild with it now.

### DIFF
--- a/src/__tests__/core/llm-abort.test.ts
+++ b/src/__tests__/core/llm-abort.test.ts
@@ -1,0 +1,39 @@
+import { describe, it, expect } from 'vitest';
+import { withAbortSignal } from '../../core/llm-abort';
+
+describe('withAbortSignal (#646)', () => {
+  const make = () => {
+    const calls: unknown[] = [];
+    class Client {
+      private secret = 'kept';
+      async createMessage(p: Record<string, unknown>) { calls.push(p); return 'text'; }
+      async createMessageWithOutput(p: Record<string, unknown>) { calls.push(p); return { text: this.secret }; }
+      async createMessageStream(p: Record<string, unknown>) { calls.push(p); return 'stream'; }
+      other() { return this.secret; }
+    }
+    return { client: new Client(), calls };
+  };
+
+  it('adds the current signal to all three call shapes and leaves other members working', async () => {
+    const { client, calls } = make();
+    const ctl = new AbortController();
+    const w = withAbortSignal(client, () => ctl.signal);
+    await w.createMessage({ model: 'm' });
+    expect(await w.createMessageWithOutput({ model: 'm' })).toEqual({ text: 'kept' });
+    await w.createMessageStream({ model: 'm' });
+    expect(calls.map(c => (c as { abortSignal?: AbortSignal }).abortSignal)).toEqual([ctl.signal, ctl.signal, ctl.signal]);
+    expect(w.other()).toBe('kept');
+  });
+
+  it('reads the signal at call time, keeps a caller-set signal, and passes through when there is none', async () => {
+    const { client, calls } = make();
+    let current: AbortSignal | undefined;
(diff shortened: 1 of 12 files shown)


green-dalii (approved):
Verified against the diff, both halves. The Proxy in `core/llm-abort.ts` is bound correctly for private-field safety (`fn.call(target, ...)` for the three call shapes, `v.bind(target)` for everything else), a caller-set signal wins, and outside an ingest nothing is injected. The second half is the better half: hanging `checkCancelled()` on `createOrUpdateFile` turns the write gate into the ingest's single cancellation checkpoint, which is why #583's cleanup can no longer be skipped.

`types.ts` adds only optional fields, so this is non-breaking. 12 files is a large surface but every one of them is on the cancellation path — no contamination.

Merging. Two things to note in the release notes rather than as blockers: (1) on the `requestUrl` fallback the in-flight request cannot be aborted, so cancellation there guarantees "no further write", not "immediate disconnect"; (2) the `task?: string` JSDoc in `types.ts` now sits above the new `abortSignal` comment, so the `task` field's own documentation is orphaned — one-line fix whenever that file is next touched.

Issue #125, opened 2026-06-13
sources: frontmatter not normalized on merge — raw external note paths persist; external remap doesn't slugify


DocTpoint:
## Summary

`normalizeSourcesField` / `normalizeSourcePath` (`src/core/sources-normalizer.ts`, added for #81) exist and are unit-tested, but are **not invoked on the page write/merge path**, so polluted `sources:` entries survive. A secondary defect: the external→`sources/` remap doesn't slugify the filename.

## Repro

Ingest a note whose basename contains spaces/parens (e.g. `AGEs (Advanced Glycation End Products).md`) such that it merges into existing pages.

## Observed

The merged pages' frontmatter contains:

```yaml
sources:
  - "[[Notizen/AGEs (Advanced Glycation End Products).md]]"
```

…while the `## Mentions in Source` block on the same page correctly uses `[[sources/AGEs-Advanced-Glycation-End-Products|…]]`.

## Expected

```yaml
sources:
  - "[[sources/AGEs-Advanced-Glycation-End-Products]]"
```

## Analysis

1. `normalizeSourcesField` is not called from the write/merge path (`page-factory.ts`, `batch-merger.ts`, `wiki-engine.ts`, `conflict-resolver.ts` — grep finds no usage). The normalizer is effectively dead code on the ingest path.
2. `normalizeSourcePath` remaps external paths via `s.split('/').pop()` **without slugifying**, so `Notizen/AGEs (Advanced Glycation End Products)` → `sources/AGEs (Advanced Glycation End Products)` (spaces/parens kept) ≠ the real slug `sources/AGEs-Advanced-Glycation-End-Products`.

## Proposed

- Wire `normalizeSourcesField` into the frontmatter assembly/merge path so every `sources:` write is normalized.
- In the external→`sources/` remap, apply the same `slugify(filename, slugCase === 'preserve')` used for page creation, so remapped paths match the actual source page slug.

## Scope

`src/core/sources-normalizer.ts`, the sources-frontmatter write path, tests in `src/__tests__/core/sources-normalizer.test.ts`. Related: #124.

Happy to open a PR.

green-dalii:
Fixed by #127 — normalization is now applied at the centralized write chokepoint and external paths are slugified. Will close this issue once that PR merges.

PR #649, opened 2026-09-06
fix(frontmatter): skip empty sources: entries when merging

### LINKED ISSUE #648: A bare `sources:` line in model-written frontmatter merges into a stray `[[]]` link
mergeFrontmatter (src/core/frontmatter.ts) rebuilds the `sources:` list from the existing entries plus the new source path. A bare `sources:` line (the model emits it now and then on merge pages) parses to `['']`, normalizeSourcePath('') stays '', and the rebuilt list contains `[[]]`. The #125 normalizer removes it downstream, so nothing reaches disk, but the list is wrong at the point where it is produced and every later consumer pays for the cleanup.

Fix: skip empty entries on both the existing list and the new source path. Two-line change plus a test. The create path has used appendSourceSlugToFrontmatter since v1.26.0, so merge-page.ts is the remaining caller that can see such a line.

PR follows.

DocTpoint:
A bare `sources:` line in model-written frontmatter parses to `['']`, which mergeFrontmatter turned into a stray `[[]]` link that the #125 normalizer had to clean up downstream. Skip empty entries on both the existing list and the new source path. The create path has used appendSourceSlugToFrontmatter since v1.26.0, so merge-page.ts is the remaining caller that can see such a line.

Closes #648

Gate: 282 test files, 3994 tests (base 3993), eslint 0, tsc clean.

### DIFF
--- a/src/__tests__/core/frontmatter.test.ts
+++ b/src/__tests__/core/frontmatter.test.ts
@@ -529,6 +529,16 @@ describe('mergeFrontmatter', () => {
     expect(result.wasMerged).toBe(true);
   });
 
+  it('does not emit a stray empty wikilink when an existing bare sources field is empty', () => {
+    // The create-path stamp calls mergeFrontmatter on enforce output, where a
+    // model-written block sources field has been reduced to a bare `sources:`
+    // line → parseFrontmatter yields ['']. The empty entry must be dropped.
+    const input = '---\ntype: entity\nsources:\ntags:\n---\n\nBody';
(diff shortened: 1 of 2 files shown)


green-dalii (approved):
Verified: `normalizeSourcePath` output and `newSourcePath` are both guarded before entering the set, which is exactly the `['']` → `[[]]` path #648 described. Two lines, one test, no scope creep, and it removes work from the downstream normalizer rather than adding a second cleanup. Merging.

Issue #603, opened 2026-09-01
the "single write gate" contract does not hold — six writers bypass createOrUpdateFile


DocTpoint:
`types.ts` documents the engine context with `createOrUpdateFile — single write gate with pollution defense`, and downstream reasoning builds on that (change detection via `onFileWrite`, sources normalization, heading spacing all live in the gate).

The gate has bypasses. Writers that reach the vault through `vault.process` / `vault.modify` directly: `core/link-retarget.ts` (its own header comment acknowledges it skips the gate), `wiki/lint/phases/preparation.ts`, `wiki/engine-internals/log-writer.ts`, `schema/schema-manager.ts`, `schema/apply-suggestion.ts`, `core/sources-normalizer.ts`. Every invariant hung on the gate silently misses these writes — most concretely `onFileWrite`, which `types.ts` describes as "notifies file watcher of writes for change detection": a link retarget rewrites pages vault-wide and the watcher never hears about it.

#601 is the same premise from the other side: the gate also does not constrain *where* a write goes — a user's note passed through it unrefused.

We hit this measurably: we run a write-level audit wrapped around `createOrUpdateFile` (local patch), and the write into the user's note from #601 left no audit entry while the audit was demonstrably active that hour — wrapping the gate covers only what actually flows through it.

Not proposing a mechanism — whether the answer is funnelling every page write through the gate, or narrowing the documented contract to what it actually covers, is a design call. Happy to help either way.

green-dalii:
@DocTpoint — this is #601's other side, received. Six bypass writers leave `onFileWrite` (which `types.ts` describes as notifying the watcher for change detection) silently deaf. I lean toward **narrowing the documented contract + progressively funnelling the critical write paths through the gate**, rather than forcing everything through at once (higher risk). Concretely: (1) rewrite `types.ts`'s "single write gate" wording to the actual coverage; (2) add write-audit logging to `createOrUpdateFile` so bypasses become detectable; (3) evaluate each bypass writer for gate entry case by case. #601's lesson — a write that bypasses the gate leaves no audit entry — makes this worth doing. It's on the v1.27.x PATCH Tier 2 list; mechanism to be decided in discussion.

DocTpoint:
@green-dalii — agreed on narrowing the contract first and funnelling writers one at a time. On your three steps:

[...]

PR #651, opened 2026-09-06
fix(create): drop model-written sources before stamping the true one

### LINKED ISSUE #650: Create path keeps model-invented `sources:` entries and stamps the true source next to them
On the create path the page's provenance is fully known: exactly the one source of this run. The model sometimes invents additional `sources:` entries anyway. In a rebuild sample, 5 of 8 pages that looked multi-source were born with a second source whose note does not exist. Since #523 the constraints pass carries unknown block-form fields through, so those entries survive to disk and the page claims a provenance nobody can trace.

Fix: remove whatever the model wrote under `sources:` right before appendSourceSlugToFrontmatter stamps the real one (src/wiki/page-factory/create-page.ts). Flow and block style both handled, list items indented or at column 0. I looked at replaceFrontmatterArrayField for this and did not use it: it re-serialises the whole block with block-style tags while the create path writes inline tags, so it would flip the tag style of every new page.

PR follows.

DocTpoint:
On the create path the page's provenance is fully known: exactly the one source of this run. The model sometimes invents additional `sources:` entries (5 of 8 "multi-source" pages in a rebuild sample were born with a second source whose note does not exist), and since #523 the constraints pass carries unknown block-form fields through, so the invented entries survived to disk. Remove whatever the model wrote under `sources:` before appendSourceSlugToFrontmatter stamps the real one. Flow and block style are both handled; list items may be indented or at column 0.

replaceFrontmatterArrayField was considered and not used: it re-serialises the whole block with block-style tags while the create path writes inline tags.

Closes #650

Gate: 282 test files, 3998 tests (base 3993), eslint 0, tsc clean.

### DIFF
--- a/src/__tests__/wiki/page-factory/append-source-slug.test.ts
+++ b/src/__tests__/wiki/page-factory/append-source-slug.test.ts
@@ -308,3 +308,42 @@ describe('appendSourceSlugToFrontmatter (#399)', () => {
     });
   });
 });
+
+// local patch (S142): the create path drops model-written `sources:` before
+// stamping the true source — fabricated provenance is worse than none.
+import { dropSourcesField } from '../../../wiki/page-factory/create-page';
+
+describe('dropSourcesField (S142 create-path provenance guard)', () => {
+  it('removes a block-style sources list, keeping every other field', () => {
+    const input = '---\ntype: concept\nsources:\n  - "[[sources/Metabolismus_Übersicht]]"\n  - "[[sources/Acetylsalicylsäure]]"\ntags:\n  - "Thema/Diagnostik"\n---\n\n# Titel\n';
+    const out = dropSourcesField(input);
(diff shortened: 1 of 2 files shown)


green-dalii (approved):
Verified: `dropSourcesField` runs immediately before the true source is stamped, and the regex (`^sources:[^\n]*\n(?:[ \t]*-[^\n]*\n)*`) covers both flow and block style with items at column 0 or indented. Refusing `replaceFrontmatterArrayField` because it would re-serialise tags into block style is the right reasoning — the create path writes inline tags and switching style would be a silent file-format change.

Merging. Related observation for the record: this is now the second place that edits frontmatter outside the create path's normal flow, which is the same seam `#603` (write-gate contract) is about. Not a blocker here — the write itself still goes through the gate — but if #603 lands as a refactor, this call site should be re-read against it.

PR #634, opened 2026-09-05
refactor(audit): phase-3 lossless cleanup — T1 dedupe+relocation, T3 report round-trip (−82 net)

green-dalii:
## Audit Phase 3 — Tier 1+2+3 lossless cleanup (6 commits / 33 files / +90 −172)

Follow-up to audit phases 1 (#632) + 2 (#633). Third full structural + redundancy audit (3 parallel read-only slices: test bloat / production redundancy / structure drift) on main @ 5ba91e1, then lossless-executed on one branch in per-phase commits. Each phase planned up-front and three-zero-verified (no breaking change / no side effect / no perf regression) before commit.

### Commits (chronological)

| Commit | Phase | What |
|--------|-------|------|
| `88810d1` | T1-1 | escapeRegExp → canonical `lint/utils` escapeRegex (re-export alias keeps the 13-test public API) |
| `c5132fb` | T1-2 | relocate 7 inline `src/core/*.test.ts` into `__tests__/core/` mirror (pure git mv + import depth) |
| `da06362` | T1-3 | restore `__tests__/llm-sdk/{bedrock-sso,openai-codex}/` mirror depth (15 files; shared test-helper intentionally stays flat) |
| `dc3754a` | T1-4 | privatize 12 dead exports (0 prod + 0 test importers; types OutputMode/ThinkingEffort kept exported) |
| `c094003` | T3-1 | collapse controller.ts report round-trip into ONE buildLintReport call (−85L; extractProgReport + summaryText/aliasPre re-derivation deleted — vestige of the v1.26.4 #473-removed LLM analysis prompt) |
| `55e5536` | T1-4 fix | OUTPUT_MODES const had zero runtime consumers → delete, literal-union type (eslint-clean) |

### Three-zero verification per phase
- **No breaking change**: test count 3932 → 3932 (relocations are moves, not merges — zero test loss); report-builder tests pin the byte-identical full-report shape; tsc 0.
- **No side effect**: `pnpm lint` 0 warnings, shazam_verify PASS 0 new orphans, no import/API signature changes (12 privatized exports verified 0 importers corpus-wide first).
- **No perf regression**: T3-1 removes a redundant report string build + strip (strictly faster); everything else is compile-time or test-org only.

### Deliberately NOT done (audit-verified as needs-design or PURE-LOCKED, filed for separate rounds)
[...]

PR #653, opened 2026-09-07
fix(lint): prefer a page's real display name over its filename slug when repairing dead links

### LINKED ISSUE #592: Fix Dead Links clobbers existing wikilink display text with raw slugs
# Bug: "Fix Dead Links" clobbers existing wikilink display text with raw slugs

**Area:** `fixDeadLink()` in the dead-link-fixing lint stage (`src/wiki/lint/fix-dead-link.ts`), `getExistingWikiPages()` (`src/wiki/lint/get-existing-pages.ts`)
**Filed as:** <https://github.com/green-dalii/obsidian-llm-wiki/issues/592>

Confirmed still present in the current `main` branch as of 2026-08-31: `replaceTargetLink()` still discards the alias and rewrites the link with the matched page's title verbatim, and `getExistingWikiPages()`'s `PageRef` still has no field for a page's real heading - only `title: f.basename`.

## Symptom

Running lint's "Fix Dead Links" step rewrites correctly-authored wikilinks like

```text
[[concepts/holistic-persistent-memory|Holistic Persistent Memory]]
```

into

```text
[[concepts/holistic-persistent-memory|holistic-persistent-memory]]
```

i.e. the human-readable alias is replaced by the raw path slug. This happened across ~10 files in a single lint run on our vault (body text, "Related Entities"/"Related Concepts" lists - everywhere the same dead target was linked). One case (`resonance-score`) also picked up a stray literal `/` in the display text: `[[concepts/resonance-score|/resonance-score]]`.

Newly created stub pages have the same problem in their own title: e.g. `entities/postgresql.md` has `aliases: [Postgres, PostgreSQL Database, ...]` in frontmatter but its own H1 is `# postgresql`.

## Root cause

Two distinct bugs in `fixDeadLink()`:

**1. Deterministic - page title is the filename, not a display name.**

`getExistingWikiPages()` sets `title: f.basename` (the file's slug-cased filename, e.g. `riverside-analytics` because `slugCase: "lower"`). Several branches of `fixDeadLink()` use this `.title` as the new *display alias* whenever they repair a link's path, discarding whatever alias text was already in the document:

- the pre-check branch (`findDeadLinkTarget` -> `buildDeadLinkReplacement`) - fires even for a simple path typo (we hit this via `entities/riverside-analitics` -> `entities/riverside-analytics`, alias silently degraded from `Riverside Analytics` to `riverside-analytics`)
- the `aliasMatch` safety-net inside the `create_stub` branch
- the final `match` fallback branch (existing page found by title/alias/slug)

None of these need to touch the alias at all - the link's *path* was wrong, not its *display text*. The fix should preserve whatever alias the author already wrote and only swap the path.

**2. LLM-quality - `stub_title` from the model is used verbatim, unsanitized.**

When lint asks the model (in our config, a small/cheap model - `gemini-3.1-flash-lite`) to invent a title for a stub page (`create_stub` action), the response is used as-is for both the new stub's H1 *and* the alias written back into the referencing document. The model inconsistently returns the raw slug instead of a proper title (`ai-orchestrator` instead of `AI Orchestrator`), and in one case prefixed it with a stray `/`. The existing sanitizer (`sanitizedTitle = result.stub_title.replace(/^(entities|concepts|sources)([^\s\-_a-zA-Z0-9])/, "$2")`) only strips a specific set of known folder prefixes - it doesn't catch arbitrary leading punctuation, and it does nothing to fix a slug-shaped title.

The same unsanitized-title issue exists in the deterministic final fallback (no LLM match at all), which builds a stub title directly from `targetBasename` (already slug-shaped) with no title-casing at all.

Even sanitized, title-casing a slug can't recover everything a real heading has: punctuation, spacing, and subscripts are gone once a name is slugified into a filename, and no allowlist or capitalization rule gets them back (a filename like `haem-a3` can't be title-cased into `Häm a₃`). `getExistingWikiPages()` (`src/wiki/lint/get-existing-pages.ts`) already opens every page's content to parse its frontmatter, but never reads the page's own H1 - the one place a page's real display name is actually written down.

1. Add a small `sanitizeStubTitle()`: strip any leading non-alphanumeric character(s), and if the result still looks like a raw slug (`^[a-z0-9]+([-_][a-z0-9]+)*$`), title-case it word-by-word (with a short known-acronym allowlist - AI, RAG, CAG, MCP, SQL, LLM, API, UI, SDK, URL, HTTP, JSON, etc. - to avoid "Ai" instead of "AI"). Apply it everywhere a title is about to be used as a stub's H1 or as a fallback alias (`create_stub` branch, the deterministic final-fallback branch).
2. Add a `replaceDeadLinkPreservingAlias()` that, before rewriting a dead link's path, first checks whether the *existing* occurrence in the document already has an alias (`[[target|alias]]`). If so, keep that alias and only swap the path. Only fall back to a computed title when the original link was bare (`[[target]]`, no alias). Use this in place of `replaceDeadLink`/`replaceTargetLink` for every branch that is *repairing a path to something that already resolves* (pre-check, `aliasMatch`, `match` fallback, `create_stub`'s referencing-link rewrite, and the deterministic final-fallback's referencing-link rewrite). Leave the `action === "correct"` branch alone - there the model is intentionally proposing a full replacement link, alias included, which is a different (correct) use case.
3. Have `getExistingWikiPages()` also extract each page's H1 (if it has one) into a new `displayTitle` field - no extra file read, the content is already open there to parse frontmatter. A shared `resolveDeadLinkFix()` helper (used by the three branches above that repair a link to a page that *already exists* - pre-check, the `create_stub` safety net, and the final `match` fallback) resolves the display title as: the page's real H1 (`displayTitle`) → the page's own first frontmatter alias, if it exists but has no parseable H1 (a malformed/empty body, not the same as a page that doesn't exist yet) → `sanitizeStubTitle(title)` as a last resort. The freshly-generated LLM `stub_title` path is untouched - a page that doesn't exist yet has no H1 or alias to read.

This is a general-purpose fix, not model-specific - even with a stronger model, "reuse the original document's alias when only the path was broken" is strictly safer than "regenerate the alias from the target name," and preferring a page's real H1/alias over a guess is strictly safer than title-casing a slug.

Verified live in Obsidian, for each fallback tier: a page whose real H1 has a subscript unrecoverable by title-casing resolves to that H1; a page with no parseable H1 but a frontmatter alias resolves to that alias; an existing author-written alias in the referencing document still wins over both. Also verified against a pristine copy of upstream `main.js` (commit `068a0ea`): same wrong-path repair, but the alias is clobbered with the raw filename slug regardless of any existing alias or heading, confirming the defect is real upstream and fixed by this patch.

## Patch

Below is the actual diff against our local `main.js` - the bundled/minified plugin build we run, not your real TypeScript source. It won't apply cleanly with `git apply`/`patch` against your repo (different file, different line numbers, and some surrounding-but-unrelated code has minifier-mangled names like `_a24`). Treat it as an exact reference for the logic to translate into `src/wiki/lint/get-existing-pages.ts` (the `displayTitle` addition) and `src/wiki/lint/fix-dead-link.ts` (everything else), not a drop-in patch.

```diff
diff --git a/main.js b/main.js
--- a/main.js
+++ b/main.js
@@ ... @@ async function getExistingWikiPages(app, wikiFolder) {
     if (fm && fm.type === "welcome") {
       continue;
     }
+    // LOCAL PATCH (not upstream): `title` above is the filename slug, not a
+    // display name - upstream has no separate field for the page's actual
+    // heading. A meaningful fraction of pages in a real vault have an H1
+    // that differs from the filename in ways slug-to-title-case can't
+    // recover (punctuation, spacing, subscripts), so callers that need a
+    // display name should prefer this over title-casing the slug. Costs no
+    // extra read - content is already open for frontmatter above.
+    const h1Match = extractBody(content).trim().match(/^#\s+(.+?)(?:\n|$)/);
     pages.push({
       path: f.path,
       title: f.basename,
+      displayTitle: h1Match ? h1Match[1].trim() : void 0,
       wikiLink: `[[${relPath}|${f.basename}]]`,
       aliases: Array.isArray(fm == null ? void 0 : fm.aliases) ? fm.aliases : void 0,
@@ ... @@ generation_complete: false
 # ${title}
 
 > Stub created by Fix Dead Links \u2014 referenced by [[${referringPageRel}]]. Will be filled by next ingest of an actual source that defines this entity.
 `;
 }
+// LOCAL PATCH (not upstream): the four items below (KNOWN_TITLE_ACRONYMS
+// through replaceDeadLinkPreservingAlias) don't exist upstream. They fix
+// fixDeadLink() clobbering correctly-authored wikilink display text -
+// several of its branches were reusing a page's filename slug (or an
+// unsanitized LLM-generated stub_title) as the link's alias whenever they
+// repaired a broken path, silently degrading text like "Holistic Persistent
+// Memory" down to "holistic-persistent-memory". See the call sites inside
+// fixDeadLink() below, each marked "LOCAL PATCH", for where these are used.
+var KNOWN_TITLE_ACRONYMS = /* @__PURE__ */ new Set(["ai", "rag", "cag", "mcp", "hpm", "sql", "llm", "api", "sha", "ui", "id", "sdk", "url", "http", "https", "json"]);
+function titleCaseFromSlug(text2) {
+  return text2.split(/[-_\s]+/).filter(Boolean).map((word) => {
+    const lower = word.toLowerCase();
+    if (KNOWN_TITLE_ACRONYMS.has(lower)) return lower.toUpperCase();
+    return lower.charAt(0).toUpperCase() + lower.slice(1);
+  }).join(" ");
+}
+function sanitizeStubTitle(rawTitle) {
+  if (!rawTitle) return rawTitle;
+  let title = rawTitle.trim().replace(/^[^\p{L}\p{N}]+/u, "");
+  const looksLikeRawSlug = /^[a-z0-9]+([-_][a-z0-9]+)*$/.test(title);
+  if (looksLikeRawSlug) title = titleCaseFromSlug(title);
+  return title;
+}
+function extractExistingAlias(content, targetName) {
+  const re = /\[\[([^\]|#]+)(?:\|([^\]]+))?(?:#[^\]]+)?\]\]/g;
+  let m;
+  while ((m = re.exec(content)) !== null) {
+    if (m[1].trim() === targetName && m[2]) return m[2].trim();
+  }
+  return null;
+}
+function replaceDeadLinkPreservingAlias(content, targetName, newPathOrLink, fallbackTitle) {
+  let newPath = newPathOrLink;
+  let explicitAlias = null;
+  const linkMatch = newPathOrLink.match(/^\[\[([^\]|]+)(?:\|([^\]]+))?\]\]$/);
+  if (linkMatch) {
+    newPath = linkMatch[1];
+    explicitAlias = linkMatch[2] || null;
+  }
+  const existingAlias = extractExistingAlias(content, targetName);
+  const alias = existingAlias || explicitAlias || fallbackTitle;
+  const replacement = `[[${newPath}|${alias}]]`;
+  const linkRegex = /\[\[([^\]|#]+)(?:[|#][^\]]+)?\]\]/g;
+  return content.replace(
+    linkRegex,
+    (fullMatch, capturedTarget) => {
+      if (capturedTarget.trim() === targetName) return replacement;
+      return fullMatch;
+    }
+  );
+}
+// LOCAL PATCH (not upstream): shared by every fixDeadLink() branch below that
+// repairs a link to a page that already exists (pre-check, the create_stub
+// safety net, and the final match fallback) - they all resolved a display
+// title for that page and applied it the same way, so it's pulled out here
+// instead of copy-pasted three times. Prefers the page's real H1
+// (displayTitle) over sanitizeStubTitle(page.title) (a title-cased slug) -
+// title-casing a slug can't recover things like punctuation or subscripts
+// that a real heading has. If displayTitle is missing (the page's body has
+// no parseable H1 - this is a real, existing page, just one with a
+// malformed/absent heading, not the same as a page that doesn't exist yet),
+// prefer its first frontmatter alias over the sanitized slug too, since an
+// author-supplied alias is still a better display name than a title-cased
+// guess. The resolved title is embedded directly in newLink (not the raw
+// page.title) so it also doubles correctly as fallbackTitle:
+// replaceDeadLinkPreservingAlias() would otherwise read the raw slug back
+// out of newLink as an "explicit" alias and use that instead of the
+// fallback whenever the document had no alias of its own.
+function resolveDeadLinkFix(page, sourceContent, targetName, wikiFolder) {
+  const title = page.displayTitle || page.aliases?.[0] || sanitizeStubTitle(page.title);
+  const newLink = `[[${makeRelPath(page.path, wikiFolder)}|${title}]]`;
+  const updatedContent = replaceDeadLinkPreservingAlias(sourceContent, targetName, newLink, title);
+  return { newLink, updatedContent };
+}
 async function fixDeadLink(ctx, sourcePath, targetName) {
   const existingPages = await getExistingWikiPages(
     ctx.app,
     ctx.settings.wikiFolder
   );
   const sourceContent = await ctx.tryReadFile(sourcePath) || "(empty)";
   const targetBasename = targetName.includes("/") ? targetName.split("/").pop() : targetName;
   const preMatch = findDeadLinkTarget(existingPages, targetBasename);
   if (preMatch) {
-    const newLink2 = buildDeadLinkReplacement(preMatch, ctx.settings.wikiFolder);
-    const updatedContent2 = replaceDeadLink(sourceContent, targetName, newLink2);
+    // LOCAL PATCH: upstream called replaceDeadLink(sourceContent, targetName,
+    // buildDeadLinkReplacement(preMatch, ...)), which always overwrote the
+    // alias with preMatch.title (a filename slug). resolveDeadLinkFix() (see
+    // above) keeps whatever alias the document already had instead.
+    const { newLink: newLink2, updatedContent: updatedContent2 } = resolveDeadLinkFix(preMatch, sourceContent, targetName, ctx.settings.wikiFolder);
     await ctx.createOrUpdateFile(sourcePath, updatedContent2);
     return `pre-check corrected (alias match): ${newLink2}`;
   }
   const pagesList = existingPages.filter((p) => {
     const bn = p.title || "";
@@ ... @@ async function fixDeadLink(ctx, sourcePath, targetName) {
     const updatedContent2 = replaceTargetLink(sourceContent, targetName, newLink2);
     await ctx.createOrUpdateFile(sourcePath, updatedContent2);
     return `corrected: ${newLink2}`;
   }
   if ((result == null ? void 0 : result.action) === "create_stub" && result.stub_title) {
-    const sanitizedTitle = result.stub_title.replace(/^(entities|concepts|sources)([^\s\-_a-zA-Z0-9])/, "$2");
+    // LOCAL PATCH: upstream used result.stub_title verbatim after only
+    // stripping known folder prefixes (entities/concepts/sources). The LLM
+    // often returns a raw slug ("ai-orchestrator") instead of a real title
+    // ("AI Orchestrator"), and once returned a stray leading "/". Wrapping
+    // in sanitizeStubTitle() strips stray leading punctuation and title-
+    // cases anything that still looks like a slug.
+    const sanitizedTitle = sanitizeStubTitle(result.stub_title.replace(/^(entities|concepts|sources)([^\s\-_a-zA-Z0-9])/, "$2"));
     const stubTitleLower = sanitizedTitle.toLowerCase();
     const safetySlug = slugify(sanitizedTitle).toLowerCase();
     const aliasMatch = existingPages.find(
       (p) => {
         var _a24, _b17;
         return p.title.toLowerCase() === stubTitleLower || ((_a24 = p.aliases) == null ? void 0 : _a24.some((a) => a.toLowerCase() === stubTitleLower)) || slugify(p.title).toLowerCase() === safetySlug || ((_b17 = p.aliases) == null ? void 0 : _b17.some((a) => slugify(a).toLowerCase() === safetySlug));
       }
     );
     if (aliasMatch) {
-      const newLink3 = `[[${makeRelPath(aliasMatch.path, ctx.settings.wikiFolder)}|${aliasMatch.title}]]`;
-      const updatedContent3 = replaceTargetLink(sourceContent, targetName, newLink3);
+      // LOCAL PATCH: upstream called replaceTargetLink (always overwrites
+      // the alias with aliasMatch.title, a filename slug). Same fix as the
+      // pre-check branch above, via resolveDeadLinkFix().
+      const { newLink: newLink3, updatedContent: updatedContent3 } = resolveDeadLinkFix(aliasMatch, sourceContent, targetName, ctx.settings.wikiFolder);
       await ctx.createOrUpdateFile(sourcePath, updatedContent3);
       return `safety-net corrected (alias match for stub): ${newLink3}`;
     }
     const stubType2 = result.stub_type || "entity";
     const stubDir2 = PLURAL_MAP[stubType2] || `${stubType2}s`;
@@ ... @@ async function fixDeadLink(ctx, sourcePath, targetName) {
       wikiFolder: ctx.settings.wikiFolder,
       referringPageRel: sourceRel2
     });
     await ctx.createOrUpdateFile(stubPath2, stubContent2);
     const newLink2 = `[[${stubDir2}/${stubSlug2}|${sanitizedTitle}]]`;
-    const updatedContent2 = replaceTargetLink(sourceContent, targetName, newLink2);
+    // LOCAL PATCH: upstream called replaceTargetLink, which always used the
+    // freshly-generated sanitizedTitle as the alias, discarding whatever
+    // alias the document already had for this (previously dead) link.
+    const updatedContent2 = replaceDeadLinkPreservingAlias(sourceContent, targetName, newLink2, sanitizedTitle);
     await ctx.createOrUpdateFile(sourcePath, updatedContent2);
     return `stub created (unfilled): ${stubPath2} \u2014 will be filled by next ingest of a real source`;
   }
   const lowerTarget = targetBasename.toLowerCase();
   const targetSlug = slugify(targetBasename).toLowerCase();
@@ ... @@ async function fixDeadLink(ctx, sourcePath, targetName) {
       }
     );
   }
   if (match) {
-    const newLink2 = `[[${makeRelPath(match.path, ctx.settings.wikiFolder)}|${match.title}]]`;
-    const updatedContent2 = replaceTargetLink(sourceContent, targetName, newLink2);
+    // LOCAL PATCH: same fix as the branches above - upstream's
+    // replaceTargetLink always overwrote the alias with match.title. Via
+    // resolveDeadLinkFix().
+    const { newLink: newLink2, updatedContent: updatedContent2 } = resolveDeadLinkFix(match, sourceContent, targetName, ctx.settings.wikiFolder);
     await ctx.createOrUpdateFile(sourcePath, updatedContent2);
     return `fallback corrected: ${newLink2}`;
   }
-  const cleanBasename = targetBasename.replace(/^(entities|concepts|sources)([^\s\-_a-zA-Z0-9])/, "$2");
+  // LOCAL PATCH: upstream used the raw slug (targetBasename, only stripped
+  // of known folder prefixes) directly as the stub title, with no
+  // title-casing at all - this is the deterministic path taken when the LLM
+  // gave no usable result.
+  const cleanBasename = sanitizeStubTitle(targetBasename.replace(/^(entities|concepts|sources)([^\s\-_a-zA-Z0-9])/, "$2"));
   const stubType = targetName.includes("/entities/") ? "entity" : "concept";
   const stubDir = stubType === "entity" ? WIKI_SUBFOLDERS.entities : WIKI_SUBFOLDERS.concepts;
   const stubSlug = slugify(cleanBasename, ctx.settings.slugCase === "preserve");
   const stubPath = `${ctx.settings.wikiFolder}/${stubDir}/${stubSlug}.md`;
   const sourceRel = makeRelPath(sourcePath, ctx.settings.wikiFolder);
@@ ... @@ async function fixDeadLink(ctx, sourcePath, targetName) {
     wikiFolder: ctx.settings.wikiFolder,
     referringPageRel: sourceRel
   });
   await ctx.createOrUpdateFile(stubPath, stubContent);
   const newLink = `[[${stubDir}/${stubSlug}|${cleanBasename}]]`;
-  const updatedContent = replaceTargetLink(sourceContent, targetName, newLink);
+  // LOCAL PATCH: upstream called replaceTargetLink here too.
+  const updatedContent = replaceDeadLinkPreservingAlias(sourceContent, targetName, newLink, cleanBasename);
   await ctx.createOrUpdateFile(sourcePath, updatedContent);
   return `fallback stub created (unfilled): ${stubPath} \u2014 will be filled by next ingest of a real source`;
 }
 
 // src/wiki/lint/fill-empty-page.ts
```

Jan-Heldal:
fixDeadLink() repairs a broken wikilink's path against getExistingWikiPages()'s page list, but the display text it wrote back had two problems:
First, when the link had no alias of its own, it fell back to the target page's filename slug title-cased, which can't recover punctuation, spacing, or subscripts a real heading has.
Second, the report's own headline complaint: when the link already carried an author-written alias, that alias was silently discarded and replaced regardless.

Adds displayTitle to a new named WikiPageRef shape, populated from the page's real H1 only — no frontmatter-alias fallback, since aliases are unordered abbreviations/variants, not necessarily the page's real name; falling back to the filename is safer than guessing from an alias. Each dead-link occurrence's own alias, when present, is captured and preserved independently per match (not just the first occurrence found for a given target), so a page referencing the same dead target twice under two different aliases keeps both instead of one clobbering the other.

Beyond the fixes DocTpoint's review asked for, a Gate-2 side-effects pass over this change turned up two related, previously-latent bugs:

- An LLM correct_link reply is now validated before being trusted at all — a blank target ([[ ]]), a blank alias ([[foo|]]), two links concatenated together ([[foo]] and [[bar]]), an embedded \r or \n are all rejected.
  - On a rejected reply, the link is left dead instead of being applied: the LLM tried and failed to produce something usable, which is different from having no guidance at all, so it no longer falls through to stub creation. (A stub the LLM never asked for is invisible to Delete Empty Stubs and would become permanent vault clutter.)
  - A dropped or doubled bracket on either side of correct_link — including a fully unclosed reply like "[[foo" — is repaired into a clean link rather than rejected, since brackets are never valid content inside a target or alias, so any bracket-count mismatch at the edges is unambiguous delimiter noise, the same class of damage a bracket-less bare reply already tolerated.
  - The repaired link is rebuilt from its trimmed target and (if present) trimmed alias, so accidental whitespace padding never reaches the file.
  - A raw carriage return or newline embedded in the target or alias is not repairable padding, so it's treated the same as any other unusable case: the correction is rejected and the link is left dead.
- replaceDeadLink's own alias-preservation logic had the identical gap (no end-of-string anchor) — safe today only because every current caller happens to hand it a well-formed link, but not defended in the shared function itself. It now leaves a link occurrence untouched rather than crashing if a future caller passes something malformed.

Verified against a live vault, including a direct comparison against unmodified upstream to confirm the underlying defect is real.

Fixes #592

### DIFF
--- a/src/__tests__/core/dead-link-detector.test.ts
+++ b/src/__tests__/core/dead-link-detector.test.ts
@@ -202,6 +202,26 @@ describe('Dead Link Detector — Pure Functions', () => {
       const result = buildDeadLinkReplacement(nestedPage, 'wiki');
       expect(result).toBe('[[concepts/deep/learning|Deep Learning]]');
     });
+
+    // Issue #592: `title` is the filename slug — prefer `displayTitle` (the page's real H1 or first alias) when set, since title-casing a slug can't recover punctuation, spacing, or subscripts a real heading has.
+    it('prefers displayTitle over title when set', () => {
+      const page: PageRef = {
+        path: 'wiki/entities/haem-a3.md',
+        title: 'haem-a3',
+        displayTitle: 'Haem A₃',
+      };
+      const result = buildDeadLinkReplacement(page, 'wiki');
+      expect(result).toBe('[[entities/haem-a3|Haem A₃]]');
+    });
+
+    it('falls back to title when displayTitle is absent', () => {
+      const page: PageRef = {
+        path: 'wiki/entities/haem-a3.md',
+        title: 'haem-a3',
+      };
+      const result = buildDeadLinkReplacement(page, 'wiki');
+      expect(result).toBe('[[entities/haem-a3|haem-a3]]');
+    });
   });
 
   describe('replaceDeadLink', () => {
--- a/src/__tests__/core/dead-link-detector.test.ts
+++ b/src/__tests__/core/dead-link-detector.test.ts
@@ -256,6 +256,15 @@ describe('Dead Link Detector — Pure Functions', () => {
       expect(result).toBe('See [[entities/cot|first name]] and later [[entities/cot|second name]] again.');
     });
 
+    // DocTpoint's #653 review: a bare occurrence and an aliased occurrence of the same
+    // target on one page must not cross-contaminate — the bare one gets the fallback
+    // replacement (i.e. displayTitle), the aliased one keeps its own alias, independently.
+    it('gives the fallback replacement to a bare occurrence while a sibling occurrence keeps its own alias', () => {
+      const mixed = 'First bare [[wrong-target]] then aliased [[wrong-target|receptors]].';
+      const result = replaceDeadLink(mixed, 'wrong-target', '[[entities/cot|Chain of Thought]]');
+      expect(result).toBe('First bare [[entities/cot|Chain of Thought]] then aliased [[entities/cot|receptors]].');
+    });
+
     it('handles links with section anchors', () => {
       const linkWithAnchor = 'See [[思维链#section]] for details.';
       const result = replaceDeadLink(linkWithAnchor, '思维链', '[[entities/cot|Chain of Thought]]');
--- a/src/__tests__/core/dead-link-detector.test.ts
+++ b/src/__tests__/core/dead-link-detector.test.ts
@@ -265,6 +265,13 @@ describe('Dead Link Detector — Pure Functions', () => {
       expect(result).toBe('First bare [[entities/cot|Chain of Thought]] then aliased [[entities/cot|receptors]].');
     });
 
+    it('leaves an aliased occurrence untouched instead of throwing when replacement is malformed', () => {
+      const linkWithDisplay = 'See [[wrong-target|thinking chain]] for more.';
+      const result = () => replaceDeadLink(linkWithDisplay, 'wrong-target', '[[]]');
+      expect(result).not.toThrow();
+      expect(result()).toBe(linkWithDisplay);
+    });
+
     it('handles links with section anchors', () => {
       const linkWithAnchor = 'See [[思维链#section]] for details.';
       const result = replaceDeadLink(linkWithAnchor, '思维链', '[[entities/cot|Chain of Thought]]');
--- a/src/__tests__/core/dead-link-detector.test.ts
+++ b/src/__tests__/core/dead-link-detector.test.ts
@@ -203,7 +203,7 @@ describe('Dead Link Detector — Pure Functions', () => {
       expect(result).toBe('[[concepts/deep/learning|Deep Learning]]');
     });
 
-    // Issue #592: `title` is the filename slug — prefer `displayTitle` (the page's real H1 or first alias) when set, since title-casing a slug can't recover punctuation, spacing, or subscripts a real heading has.
+    // Issue #592: `title` is the filename slug — prefer `displayTitle` (the page's real H1) when set, since title-casing a slug can't recover punctuation, spacing, or subscripts a real heading has.
     it('prefers displayTitle over title when set', () => {
       const page: PageRef = {
         path: 'wiki/entities/haem-a3.md',
@@ -240,11 +240,20 @@ describe('Dead Link Detector — Pure Functions', () => {
       expect(result).toContain('[[entities/cot|Chain of Thought]]');
     });
 
-    it('handles links with display text', () => {
-      const linkWithDisplay = 'See [[思维链|thinking chain]] for more.';
-      const result = replaceDeadLink(linkWithDisplay, '思维链', '[[entities/cot|Chain of Thought]]');
-      expect(result).toContain('[[entities/cot|Chain of Thought]]');
-      expect(result).not.toContain('[[思维链|thinking chain]]');
+    // Issue #653 follow-up: an occurrence that already has its own alias keeps it — the
+    // reader already saw that name — only the path is corrected to the resolved target.
+    it('keeps the existing alias on an occurrence that already has one', () => {
+      const linkWithDisplay = 'See [[wrong-target|thinking chain]] for more.';
+      const result = replaceDeadLink(linkWithDisplay, 'wrong-target', '[[entities/cot|Chain of Thought]]');
+      expect(result).toContain('[[entities/cot|thinking chain]]');
+      expect(result).not.toContain('[[wrong-target|thinking chain]]');
+      expect(result).not.toContain('Chain of Thought');
+    });
+
+    it('preserves each occurrence\'s own alias independently, not just the first one\'s', () => {
+      const twoAliases = 'See [[wrong-target|first name]] and later [[wrong-target|second name]] again.';
+      const result = replaceDeadLink(twoAliases, 'wrong-target', '[[entities/cot|Chain of Thought]]');
+      expect(result).toBe('See [[entities/cot|first name]] and later [[entities/cot|second name]] again.');
     });
 
     it('handles links with section anchors', () => {
--- a/src/__tests__/wiki/lint/fix-dead-link-display-title.test.ts
+++ b/src/__tests__/wiki/lint/fix-dead-link-display-title.test.ts
@@ -150,6 +150,31 @@ describe('fixDeadLink — preserves an existing alias already in the dead link (
     expect(writes[0]!.content).not.toContain('Haem A₃');
   });
 
+  // Issue #653 follow-up: extractDeadLinkAlias used to grab only the FIRST occurrence's
+  // alias and stamp it onto every occurrence of the same target. Each occurrence must
+  // keep its own alias instead.
+  it('pre-check: keeps each occurrence\'s own alias, not just the first one\'s', async () => {
+    vi.spyOn(getExistingPages, 'getExistingWikiPages').mockResolvedValue([
+      {
+        path: 'wiki/entities/haem-a3.md',
+        title: 'haem-a3',
+        displayTitle: 'Haem A₃',
+        wikiLink: '[[entities/haem-a3|haem-a3]]',
+      },
+    ] as never);
+    const { ctx, writes } = makeCtx(
+      noopClient(),
+      '# My Page\n\nFirst mention [[wrong-path/haem-a3|receptors]], later [[wrong-path/haem-a3|the receptor]] again.\n'
+    );
+
+    const out = await fixDeadLink(ctx, 'wiki/entities/MyPage.md', 'wrong-path/haem-a3');
+
+    expect(out).toContain('pre-check corrected');
+    expect(writes).toHaveLength(1);
+    expect(writes[0]!.content).toContain('[[entities/haem-a3|receptors]]');
+    expect(writes[0]!.content).toContain('[[entities/haem-a3|the receptor]]');
+  });
+
   it('deterministic fallback: keeps the link\'s own existing alias over the target\'s displayTitle', async () => {
     vi.spyOn(getExistingPages, 'getExistingWikiPages').mockResolvedValue([
       {
--- a/src/__tests__/wiki/lint/fix-dead-link-display-title.test.ts
+++ b/src/__tests__/wiki/lint/fix-dead-link-display-title.test.ts
@@ -1,0 +1,175 @@
+// `PageRef.title` (`getExistingWikiPages`) is the filename slug, not a display name — title-casing it can't recover punctuation, spacing, or subscripts a real heading has.
+// `displayTitle` (H1, or the first frontmatter alias when the body has none) must win over `title` in every branch of `fixDeadLink` that repairs a link to an already-existing page.
+
+import { describe, it, expect, vi, afterEach } from 'vitest';
+import { fixDeadLink } from '../../../wiki/lint/fix-dead-link';
+import * as getExistingPages from '../../../wiki/lint/get-existing-pages';
+import type { EngineContext, LLMClient } from '../../../types';
+
+const SOURCE_CONTENT = '# My Page\n\nReferences [[haem-a3]] here.\n';
+
+function makeCtx(client: LLMClient, sourceContent: string = SOURCE_CONTENT): { ctx: EngineContext; writes: Array<{ path: string; content: string }> } {
+  const written: Array<{ path: string; content: string }> = [];
+  const ctx = {
+    app: {},
+    settings: {
+      wikiFolder: 'wiki',
+      wikiLanguage: 'en',
+      disableThinking: false,
+      slugCase: 'preserve',
+    },
+    getClient: () => client,
+    getSchemaContext: () => ({}),
+    tryReadFile: async (_path: string): Promise<string | null> => sourceContent,
+    createOrUpdateFile: async (path: string, content: string): Promise<void> => {
+      written.push({ path, content });
+    },
+  } as unknown as EngineContext;
+  return { ctx, writes: written };
+}
+
+function noopClient(): LLMClient {
+  return {
+    createMessage: vi.fn(async () => '') as unknown as LLMClient['createMessage'],
+  } as LLMClient;
+}
+
+describe('fixDeadLink — prefers displayTitle over the filename slug (#592)', () => {
+  afterEach(() => {
+    vi.restoreAllMocks();
+  });
+
+  it('pre-check: uses the H1-derived displayTitle, not the slug title', async () => {
+    vi.spyOn(getExistingPages, 'getExistingWikiPages').mockResolvedValue([
+      {
+        path: 'wiki/entities/haem-a3.md',
+        title: 'haem-a3',
+        displayTitle: 'Haem A₃',
+        wikiLink: '[[entities/haem-a3|haem-a3]]',
+      },
+    ] as never);
+    const { ctx, writes } = makeCtx(noopClient());
+
+    const out = await fixDeadLink(ctx, 'wiki/entities/MyPage.md', 'haem-a3');
+
+    expect(out).toContain('pre-check corrected');
+    expect(writes).toHaveLength(1);
+    expect(writes[0]!.content).toContain('[[entities/haem-a3|Haem A₃]]');
+    expect(writes[0]!.content).not.toContain('|haem-a3]]');
+  });
+
+  it('pre-check: falls back to the raw title when no displayTitle is set', async () => {
+    vi.spyOn(getExistingPages, 'getExistingWikiPages').mockResolvedValue([
+      {
+        path: 'wiki/entities/haem-a3.md',
+        title: 'haem-a3',
+        wikiLink: '[[entities/haem-a3|haem-a3]]',
+      },
+    ] as never);
+    const { ctx, writes } = makeCtx(noopClient());
+
+    const out = await fixDeadLink(ctx, 'wiki/entities/MyPage.md', 'haem-a3');
+
+    expect(out).toContain('pre-check corrected');
+    expect(writes[0]!.content).toContain('[[entities/haem-a3|haem-a3]]');
+  });
+
+  it('alias safety net: uses displayTitle when the LLM create_stub suggestion actually matches an existing alias', async () => {
+    // Pre-check runs against the ORIGINAL link target ('totally-unrelated'), which matches nothing, so the LLM path runs. The LLM's suggested stub_title ('Cytochrome Haem A3') happens to match an existing page's alias — the safety-net re-check inside the create_stub branch, not the pre-check — is what must prefer displayTitle here.
+    vi.spyOn(getExistingPages, 'getExistingWikiPages').mockResolvedValue([
+      {
+        path: 'wiki/entities/haem-a3.md',
+        title: 'haem-a3',
+        displayTitle: 'Haem A₃',
+        aliases: ['Cytochrome Haem A3'],
+        wikiLink: '[[entities/haem-a3|haem-a3]]',
+      },
+    ] as never);
+    const client = {
+      createMessage: vi.fn(async () =>
+        JSON.stringify({ action: 'create_stub', stub_title: 'Cytochrome Haem A3', stub_type: 'entity' })
+      ) as unknown as LLMClient['createMessage'],
+    } as LLMClient;
+    const { ctx, writes } = makeCtx(client, '# My Page\n\nReferences [[totally-unrelated]] here.\n');
+
+    const out = await fixDeadLink(ctx, 'wiki/entities/MyPage.md', 'totally-unrelated');
+
+    expect(out).toContain('safety-net corrected');
+    expect(writes).toHaveLength(1);
+    expect(writes[0]!.content).toContain('[[entities/haem-a3|Haem A₃]]');
+  });
+});
+
+// A dead link may already carry an author-written alias (`[[wrong-path|My Custom Alias]]`). That alias is a stronger signal than the target page's own displayTitle/H1 — the reader already saw this name, so repairing the path must not silently swap it out. Distinct from the displayTitle fallback above, which only applies when the link has no alias of its own.
+describe('fixDeadLink — preserves an existing alias already in the dead link (#592)', () => {
+  afterEach(() => {
+    vi.restoreAllMocks();
+  });
+
+  it('pre-check: keeps the link\'s own existing alias over the target\'s displayTitle', async () => {
+    vi.spyOn(getExistingPages, 'getExistingWikiPages').mockResolvedValue([
+      {
+        path: 'wiki/entities/haem-a3.md',
+        title: 'haem-a3',
+        displayTitle: 'Haem A₃',
+        wikiLink: '[[entities/haem-a3|haem-a3]]',
+      },
+    ] as never);
+    const { ctx, writes } = makeCtx(noopClient(), '# My Page\n\nReferences [[wrong-path/haem-a3|My Custom Alias]] here.\n');
+
+    const out = await fixDeadLink(ctx, 'wiki/entities/MyPage.md', 'wrong-path/haem-a3');
+
+    expect(out).toContain('pre-check corrected');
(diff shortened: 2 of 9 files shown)


DocTpoint (changes requested):
Read the source and ran the gates locally on a merge with `main`: typecheck and lint clean, 11 new tests pass. The `displayTitle` direction is right, and I measured its assumption on our vault: all 1,209 entity and concept pages carry the H1 as the first body line, 281 differ from the filename, none contains `|` or `[[`. Two behaviours in the diff need a change before this lands, plus one thing only you can fix.

**1. The first occurrence's alias is written onto every occurrence.** `extractDeadLinkAlias` returns the alias of the first `[[target|…]]` on the page, and `replaceDeadLink` then rewrites every occurrence of that target with it. A page holding `[[X|receptors]]` and later `[[X|the receptor]]` ends up with `receptors` twice. That is the #592 complaint again, now for the second occurrence instead of the first. The repair belongs inside `replaceDeadLink`: its callback already sees each match on its own, so it can keep that match's alias and fall back to `displayTitle` only where the link had none. `extractDeadLinkAlias` then has no job left.

**2. The first frontmatter alias as display name.** When a page has no H1, `getExistingWikiPages` shows its first alias. Aliases are not ordered and are often abbreviations or spelling variants (`CRP` next to `C-reactive protein`), so the reader gets whichever came first in the list, not the page's name. The filename, i.e. today's behaviour, is the safer fallback; the alias does not say what the page is called.

**3. Not mergeable on GitHub, so no CI run is created.** The branch modifies `src/core/dead-link-detector.test.ts`, which #634 moved to `src/__tests__/core/` on 4 Sep. Local git follows the rename, GitHub's merge does not. A rebase onto `main` with the two new tests placed in the moved file clears it.

Non-blocking: `getExistingWikiPages`' return type is now spelled out inline for the third time; a named type would stop the drift.

DocTpoint (changes requested):
Point 3 is settled: the tests sit in the moved file, the branch rebases clean, and GitHub can build a merge again. Points 1 and 2 are unchanged, so I am holding the request. For clarity on what changed: comparing `773a897` and `f6744a3` file by file, the only difference is the relocated test file — every other file has identical line counts.

**1. One alias anywhere on the page still overwrites every occurrence.** `extractDeadLinkAlias` scans the whole page and returns the alias of the first `[[target|…]]` it finds; `buildDeadLinkReplacement` bakes that into one string, and `replaceDeadLink` writes the same string over every occurrence of the target. Two cases where that is wrong:

- `[[haem-a3|receptors]]` … `[[haem-a3|the receptor]]` — both become `|receptors`. The second author-written alias is destroyed, which is #592's complaint moved one occurrence to the right.
- `[[haem-a3]]` … `[[haem-a3|receptors]]` — the bare link becomes `|receptors` too. The extractor's regex requires a `|`, so a bare occurrence is skipped when searching and stamped when writing. That inverts the PR's own purpose: the occurrence with no alias is exactly the one that should show `displayTitle`.

The new tests can't catch either — all five fixtures hold a single link.

The replace callback is where this is decidable: it receives `fullMatch`, so it can keep that match's own alias and fall back to `displayTitle` only where the link had none. That means passing the page rather than a finished string, and both replacers have exactly one caller each, so the signature change is cheap. Worth noting while you're in there: `replaceTargetLink` in `fix-dead-link.ts` is a byte-identical copy of `replaceDeadLink` — same regex, same callback — so the per-occurrence rule wants one home, not two. `extractDeadLinkAlias` then has no caller left.

**2. The first frontmatter alias as display name.** Still there, now documented as intent. Two measurements on two different vaults: on 7 Sep all 1,209 entity and concept pages carried the H1 as their first body line; today, on a vault rebuilt from scratch since, 679 of 679 do. The alias branch is unreached in both. So it buys nothing where a page is well-formed, and where one is not it shows the reader whichever alias happens to be first in an unordered list — usually an abbreviation or a spelling variant. The filename, i.e. today's behaviour, is wrong in a way a reader can see through; an arbitrary alias is wrong in a way that looks deliberate. If a second fallback is wanted, it should be something that names the page, not the alias list.

Non-blocking, unchanged from last time: the `getExistingWikiPages` return type is now spelled out inline twice with `displayTitle` added to both, and `PageRef` carries it a third time. A named type would stop the drift.

One thing that isn't yours: the CI run on `f6744a3` is at `action_required` and never executed, so this head has no gate result. I'll release the run for whichever commit comes next, so it reports on the version under review.

DocTpoint (changes requested):
Changes requested on `9e99a3a`, for one thing only. The three points from last time are resolved. I checked each against the code:

1. **Per-occurrence alias.** `replaceDeadLink` now decides inside the callback. Each match keeps its own alias, and a bare occurrence gets `displayTitle`. Both cases from my review are pinned in tests, and `replaceTargetLink` is gone.
2. **Frontmatter-alias fallback.** Removed. `displayTitle` is the H1 or nothing.
3. **Named type.** `WikiPageRef`, with `PageRef` as a `Pick` of it.

Removing the deterministic post-LLM fallback is correct. For non-empty input, `slugify` is `computeSlug` behind a warning, and the fallback compared the same `existingPages` in the same four ways the pre-check already had.

**Blocking: an unusable `correct_link` should leave the link dead, not create a stub.**

`f284609` sends a malformed correction into the stub fallback. With the default `createStubsForUnresolvableLinks: true`, every unusable model answer writes a new page, and nothing removes it again. Delete Empty Stubs cannot see a Fix Dead Links stub:

- `isPageEmpty` looks for `STUB_MARKER` (`Auto-generated stub page`), and `buildStubContent` never writes it.
- After `EMPTY_CONTENT_STRIP`, the placeholder sentence alone leaves 102–122 characters, above `MIN_SUBSTANTIVE_CHARS = 50`.

So the page stays until some later source happens to define the term. In the unclosed case (`[[real-target`), the model has also just said that the target exists, so the stub is a likely duplicate.

Leaving the link as it is costs nothing: it comes back in the next lint report. A return value such as `no action taken (unusable correct_link: …)` also keeps the model defect visible in the lint log, instead of reading like a normal repair.

Because the shape check now decides between "repair" and "leave it", it should be exact. At the moment it is looser than the commit title: `[[ ]]`, `[[foo|]]` and `[[foo]] and [[bar]]` all pass and are written verbatim onto a bare occurrence. Something like `^\[\[([^\[\]|\n]+)(?:\|([^\[\]\n]+))?\]\]$`, plus a non-blank check on both groups, would match what the comment promises.

Not yours, and not for this PR: a `null` model result already goes to the same stub path, and the cleanup gap exists on main. I will open a separate issue for it.

Non-blocking:

- `fix-dead-link-display-title.test.ts:178` ("deterministic fallback: …") tests a branch that no longer exists. Its own comment says it runs through the pre-check, and it duplicates the test at line 109.
- Pre-existing: a `#heading` on the dead link is dropped (`[[wrong#Section|receptor]]` becomes `[[entities/haem-a3|receptor]]`). With per-match parsing, keeping it is one more capture group.

DocTpoint (approved):
Approving `85c5c31`.

The blocking point is resolved, and I checked the behaviour against the code rather than the description.

**An unusable `correct_link` now writes nothing.** `fixDeadLink` returns `no action taken (unusable correct_link: …)` and stops there. Every rejection test asserts that the write list is empty, which is the part that matters: the old path fell through to the deterministic fallback, and that branch builds a stub from the dead target's own basename and rewrites the referring page to point at it.

**The shape check is exact where it has to be.** Working through `normalizeCorrectLink` by hand: `[[ ]]` fails the blank-after-trim check, `[[foo|]]` the empty alias, `[[foo]] and [[bar]]` the interior brackets that survive the edge strip, `[[foo\r]]` the character class. Everything the old `/^\[\[[^\]|]+.*\]\]$/` let through and wrote verbatim is now rejected, and I checked the other direction too — nothing that used to be applied is lost.

You did not take the regex I suggested, and the replacement is better. Stripping a bracket run at each edge and validating the two parts separately treats a dropped or doubled bracket as delimiter noise, which is the tolerance a bare `target|Alias` reply already had. My anchored version would have rejected `[[foo`, and that is the one malformed case where the intent is unambiguous.

**One correction to my own review.** I argued the stub would be permanent because Delete Empty Stubs cannot see it. That leg is gone: #691 landed on 14.09 and `isEmptyStub` now collects exactly this shape. The reason for leaving the link alone survives without it — `deleteEmptyStubs` deletes the file but does not repoint incoming links, so the referring page would be left holding a dead link under the slug from an unusable answer instead of the name its author wrote. The round trip loses the original name.

Gate 1 on a clean merge with current `main` (`acabfcb`, 1.27.2): lint, typecheck, build and css-lint clean, 297 files / 4170 tests against 294 / 4144 on the base.

Non-blocking:

- The rejected value goes into the return string raw. The one rejection class where that shows is the one you just added: an embedded `\r`/`\n` splits the lint log entry across two lines. `JSON.stringify(result.correct_link)` keeps it on one line and makes the whitespace visible.
- `fix-dead-link-display-title.test.ts:178` still reaches the pre-check, and the NOTE now says so. Your call, I am not pressing it.
- The dropped `#heading` is still pre-existing and still not this PR's job.

green-dalii (approved):
Reviewed and approved. Merging.

Verified independently against the diff: `src/types.ts` adds the `WikiPageRef` interface and widens `EngineContext.getExistingWikiPages` to it. Every field of `WikiPageRef` beyond the three that were already inline is optional, so the change is additive at that boundary and no existing caller or implementer breaks. Not a settings schema, not a persisted format.

Test coverage is the right shape for this fix: 410 test lines against 146 source lines across `dead-link-detector.ts`, `fix-dead-link.ts`, `get-existing-pages.ts`. In particular `fix-dead-link-display-title.test.ts` pins the exact behaviour #592 reported — the display text is the page's real title, not the filename slug.

Findings:

- `src/types.ts:985-996` — `WikiPageRef` widens the return type only; required members are unchanged ✓
- `src/types.ts:1008` — `getExistingWikiPages: () => Promise<WikiPageRef[]>` — structurally assignable from the previous shape ✓
- `src/core/dead-link-detector.ts` +37 — carries the display title through detection
- `src/wiki/lint/fix-dead-link.ts` +90 — the write path uses the real display name

Thanks for staying with this through three review rounds — the last round's corrections are what made the display-title path durable rather than a special case.

PR #654, opened 2026-09-07
fix(ui): stop the schema diff modal crashing on first render

### LINKED ISSUE #593: Schema diff modal crashes on first render, leaving Apply/diff non-functional
# Bug: Schema diff modal crashes on first render, leaving Apply/diff completely non-functional

**Area:** `SchemaDiffModal.renderCell()` (`src/ui/schema-diff-modal.ts`)

Confirmed still present in the current `main` branch as of 2026-08-31: `renderCell()` still builds these elements via `activeDocument.createDiv()`/`activeDocument.createSpan()`, unchanged.

## Symptom

Click the "Analyze schema" button after a lint pass with a result that has `changes_needed: true`. The "Schema update preview" modal opens and shows:

- Title, rationale text, and a correct "+N / -M lines" summary.
- Below that: an empty box, with no visible height beyond its own padding - no diff content, no line numbers, no red/green highlighting.
- No "Open file" / "Cancel" / "Apply" buttons at all.

Browser console shows an uncaught error the instant the modal opens:

```stacktrace
Schema suggestion failed: HierarchyRequestError: Failed to execute 'appendChild' on 'Node': Only one element on document allowed.
    at enhance.js:1:9665
    at HTMLDocument.createEl (enhance.js:1:8610)
    at HTMLDocument.createDiv (enhance.js:1:8678)
    at SchemaDiffModal.renderCell (plugin:karpathywiki:77306:33)
    at SchemaDiffModal.renderPane (plugin:karpathywiki:77298:33)
    at SchemaDiffModal.onOpen (plugin:karpathywiki:77254:10)
    at e.open (app.js:1:1129723)
    at Object.openSchemaDiffModal (plugin:karpathywiki:77447:15)
    at runSchemaAnalyze (plugin:karpathywiki:77061:17)
    at async LLMWikiPlugin.suggestSchemaUpdate (plugin:karpathywiki:77394:5)
```

The feature is entirely non-functional as a result: there's no way to see the proposed change or click Apply, for every suggestion that actually has changes.

## Root cause

`renderCell()` builds each diff row using:

```js
const cell = activeDocument.createDiv();
...
const gutter = activeDocument.createSpan();
...
const content = activeDocument.createSpan();
```

Obsidian's `createDiv()`/`createSpan()`/`createEl()` helpers don't just construct an element - they construct it *and immediately append it* to whatever object they're called on. That's correct when called on a real container element (which is how every other call site in this plugin uses them, e.g. `contentEl.createDiv(...)`). Here they're called on `activeDocument` - the `Document` node itself. A `Document` can only ever have one child element (`<html>`, already present), so the very first attempted append throws `HierarchyRequestError`.

Since this happens on the first iteration of the row loop in `renderPane()` (`leftPane.empty(); rightPane.empty(); for (const row of this.rows) { ... renderCell(...) ... }`), the exception propagates up through an uncaught rejection in the modal's `onOpen()`, aborting execution immediately after the panes were cleared - so nothing renders, and everything scheduled after that call in `onOpen()` (the `btnRow` with Open file/Cancel/Apply) never runs either. The `+N / -M` summary above it is unaffected because it's computed from `this.rows` (already correctly populated by `recompute()`/`lineDiff()`) *before* `renderPane()` is reached - the diff itself was never the problem, only its rendering.

Notably, `activeDocument` appears exactly 3 times in the entire ~80k-line bundled plugin, all three inside this one function - every other DOM-creation call site in the codebase uses the standard `element.createDiv(...)`/`.createEl(...)` pattern on an actual container.

Replace the auto-appending helpers with plain, non-appending element construction, and let the existing (already-correct) `appendChild` calls do the actual insertion:

```js
const cell = document.createElement("div");
cell.className = rowClass;
const gutter = document.createElement("span");
gutter.className = "llm-wiki-schema-diff-gutter";
gutter.textContent = lineNo == null ? "" : String(lineNo);
cell.appendChild(gutter);
const content = document.createElement("span");
content.className = contentClass;
content.textContent = text;
cell.appendChild(content);
return cell;
```

Verified live in Obsidian after this change: the diff panes render fully (scrollable, correct red/green line highlighting matching the +N/-M summary), and all three buttons (Open file, Cancel, Apply) appear and work - Apply correctly rewrites `config.md`, creates a backup, and shows the backup-rotation notice, with no console errors.

## Patch

Below is the actual diff against our local `main.js` - the bundled/minified plugin build we run, not your real TypeScript source. It won't apply cleanly with `git apply`/`patch` against your repo (different file, different line numbers, and some surrounding-but-unrelated code has minifier-mangled names like `_a24`). Treat it as an exact reference for the logic to translate into `src/ui/schema-diff-modal.ts`, not a drop-in patch.

```diff
diff --git a/main.js b/main.js
index 36ef2ea..22e1667 100644
--- a/main.js
+++ b/main.js
@@ -77336,17 +77336,27 @@ var SchemaDiffModal = class extends import_obsidian19.Modal {
   }
   renderCell(lineNo, text2, highlighted, side) {
     const sideClass = side === "left" ? " llm-wiki-schema-diff-row-del" : " llm-wiki-schema-diff-row-add";
     const rowClass = "llm-wiki-schema-diff-row" + (highlighted ? sideClass : "");
     const contentClass = "llm-wiki-schema-diff-content" + (highlighted ? " llm-wiki-schema-diff-content-highlight" : "");
-    const cell = activeDocument.createDiv();
+    // LOCAL PATCH (not upstream): upstream builds these with
+    // activeDocument.createDiv()/.createSpan(). Obsidian's create* helpers
+    // don't just create an element, they also append it to whatever they're
+    // called on - correct for a real container, but activeDocument is the
+    // Document node itself, which only ever permits one child element
+    // (<html>, already present). That threw HierarchyRequestError on the
+    // very first row and silently aborted rendering (both panes stayed
+    // empty). Plain document.createElement() only constructs the element;
+    // the appendChild calls below (and in renderPane(), above) do the actual
+    // insertion, same as upstream intended.
+    const cell = document.createElement("div");
     cell.className = rowClass;
-    const gutter = activeDocument.createSpan();
+    const gutter = document.createElement("span");
     gutter.className = "llm-wiki-schema-diff-gutter";
     gutter.textContent = lineNo == null ? "" : String(lineNo);
     cell.appendChild(gutter);
-    const content = activeDocument.createSpan();
+    const content = document.createElement("span");
     content.className = contentClass;
     content.textContent = text2;
     cell.appendChild(content);
     return cell;
   }
```

Jan-Heldal:
SchemaDiffModal.renderCell built its cell elements by calling activeDocument.createDiv()/.createSpan() directly on the Document object. Obsidian's createDiv/createSpan/createEl helpers create and append the new element to `this`; a Document can only ever have one child element, so appending a second one throws HierarchyRequestError regardless of which window's document it is.

Extracts the cell-building logic into buildDiffCell() (in the already-obsidian-dependency-free schema-diff-modal-classes.ts, alongside its siblings, for the same testability reason) and has it build directly on the intended parent pane element instead of on Document — an ordinary Element can hold any number of children, and this also matches the file's own existing idiom elsewhere (contentEl.createDiv(...)). No lint suppressions needed.

Fixes #593

### DIFF
--- a/src/__tests__/ui/schema-diff-modal.test.ts
+++ b/src/__tests__/ui/schema-diff-modal.test.ts
@@ -12,7 +12,7 @@ vi.mock('obsidian', () => ({
 }));
 
 import { lineDiff } from '../../core/diff';
-import { applyDiffModalClasses, removeDiffModalClasses } from '../../ui/schema-diff-modal-classes';
+import { applyDiffModalClasses, removeDiffModalClasses, buildDiffCell } from '../../ui/schema-diff-modal-classes';
 
 // v1.22.0 #97: when the LLM reports changes_needed=false, the
 // SchemaDiffModal should show the *current* schema in BOTH panes
@@ -144,5 +144,34 @@ describe('SchemaDiffModal modalEl class lifecycle (v1.22.1)', () => {
       modalEl as unknown as { addClass: (c: string) => void; removeClass: (c: string) => void; empty: () => void },
     );
     expect(classes.has('llm-wiki-schema-diff-modal')).toBe(false);
+  });
+});
+
+// `buildDiffCell` must never construct elements directly on a Document — a Document can only ever hold one child element, so appending a second one throws `HierarchyRequestError`. Building on the given `parent` Element instead (which can hold any number of children) is what makes rendering more than one cell possible at all; see `buildDiffCell`'s own doc comment.
+describe('buildDiffCell (#593)', () => {
+  it('does not throw, and appends the cell as a child of the given parent', () => {
+    const parent = document.createElement('div');
+    let cell: HTMLElement | undefined;
+    expect(() => {
+      cell = buildDiffCell(parent, 1, 'some line', false, 'left');
+    }).not.toThrow();
+    expect(parent.contains(cell!)).toBe(true);
+  });
+
+  it('builds a cell with the gutter line number and content text', () => {
+    const parent = document.createElement('div');
+    const cell = buildDiffCell(parent, 3, 'hello world', true, 'left');
+    expect(cell.className).toContain('llm-wiki-schema-diff-row');
+    expect(cell.className).toContain('llm-wiki-schema-diff-row-del');
+    expect(cell.querySelector('.llm-wiki-schema-diff-gutter')?.textContent).toBe('3');
+    expect(cell.querySelector('.llm-wiki-schema-diff-content')?.textContent).toBe('hello world');
+    expect(cell.querySelector('.llm-wiki-schema-diff-content')?.className).toContain('llm-wiki-schema-diff-content-highlight');
+  });
+
+  it('right-side highlight uses the add row class, not del', () => {
+    const parent = document.createElement('div');
+    const cell = buildDiffCell(parent, null, '', true, 'right');
+    expect(cell.className).toContain('llm-wiki-schema-diff-row-add');
+    expect(cell.querySelector('.llm-wiki-schema-diff-gutter')?.textContent).toBe('');
   });
 });
--- a/src/ui/schema-diff-modal-classes.ts
+++ b/src/ui/schema-diff-modal-classes.ts
@@ -49,3 +49,32 @@ export function normalizeEmptyMode(opts: {
 }): string {
   return opts.isEmpty ? opts.currentBody : opts.newBody;
 }
+
+/**
+ * Builds one diff-pane cell (gutter + content) as a child of `parent`. Lives here rather than on SchemaDiffModal, same as its siblings above — testable without triggering Vite's import-analysis failure on the real Modal class.
+ *
+ * Builds directly on `parent` rather than on `activeDocument`: Obsidian's `Node.createEl`/`createDiv`/`createSpan` create AND append the new element to `this` (per obsidian.d.ts: "Create an element and append it to this node"). A Document can only ever have one child element, so calling these on `activeDocument` throws `HierarchyRequestError` on the second cell — an ordinary Element can hold any number of children, so building on `parent` avoids that entirely, matching the file's own existing idiom elsewhere (`contentEl.createDiv(...)`, `diffContainer.createDiv(...)` above).
+ *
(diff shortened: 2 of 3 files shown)


DocTpoint (dismissed):
Verified the diagnosis against `obsidian.d.ts`: `createDiv` is declared on `Node` and appends to `this`, and a Document already holds its one element child, so the first call throws regardless of window. That also means building on `activeDocument`, as suggested in #593, would have crashed the same way; the pane is the only parent that works. Gates on a merge with `main`: typecheck, lint, 3 new tests green. Approving `9decd47`.

Jan-Heldal:
Sorry about the double force push. The first was just to rebase onto newest main. The second was just undoing the first to avoid you having to review again. Now I see that GitHub still flags it as needing another review, so I guess it was pointless.

DocTpoint (approved):
Re-approving `9decd47` — the same commit the 7 Sep review approved, unchanged. Nothing to apologise for, and nothing to fix: this repo's ruleset has `dismiss_stale_reviews_on_push`, so any push drops approvals, including the one that restores the reviewed commit. The rebase created `92189e0` on top of `f18f722`, the undo brought back `9decd47`, and the file patches of the two are identical — I compared them before re-approving.

Please don't rebase this now. The ruleset's required-checks policy is not strict, so a branch 13 commits behind `main` is not blocked on that, and `pr-ci.yml` runs `on: pull_request` with a plain checkout, meaning it builds `refs/pull/654/merge` — this head merged into today's `main`. The staleness question is what the run answers; a rebase would only dismiss this approval again.

What is actually missing is the run itself: both of today's are sitting at `action_required` and never executed, so the green Gate 1 you see on this head is the 7 Sep one. Once someone releases the pending run, the required check reports against current `main` and this is ready to merge — check runs don't dismiss reviews, only pushes do.

green-dalii (approved):
Verified against the diff. The root cause is stated precisely and correctly: Obsidian's `createDiv`/`createSpan` create *and append* to `this`, and a `Document` can only ever hold one child element, so the second cell throws `HierarchyRequestError` on any window's document — this is a Document-arity bug, not a multi-window one, and the fix (build on the `parent` Element, which holds any number of children) addresses the real cause rather than the symptom.

Extracting `buildDiffCell` into the already-Obsidian-dependency-free `schema-diff-modal-classes.ts` follows the file's own established idiom and makes the case testable — the three new tests cover the single-cell, highlighted and empty-line paths. Merging.

PR #655, opened 2026-09-07
fix(prompts): make schema-suggestion wording read as a proposal, not something already applied

### LINKED ISSUE #594: Schema suggestions log phrases unapplied proposals as already-applied
# Bug: Schema suggestions log phrases unapplied proposals as already-applied

**Area:** `SchemaAnalyze` (prompt in `PROMPTS.suggestSchemaUpdate`, `src/wiki/prompts/generation.ts`), `wiki/schema/suggestions.md`
**Filed as:** <https://github.com/green-dalii/obsidian-llm-wiki/issues/594>

Confirmed still present in the current `main` branch as of 2026-09-02: `src/wiki/prompts/generation.ts`'s `suggestSchemaUpdate` prompt still describes both `new_schema_body` and `suggestions` with the past-tense framing described below, with no instruction on voice.

## Symptom

Ran "Suggest Schema Update" and got a `Changes needed: Yes` entry in `wiki/schema/suggestions.md` reading:

> "The schema was streamlined to remove redundant instructions while explicitly hardening the tag validation rule to address the identified ingestion violations. Instructions regarding 'Active Tag Vocabulary' were tightened to ensure compliance with the system-injected context."

Past tense reads as if the change was already made. It wasn't - `wiki/schema/config.md` was byte-identical to before the suggestion was generated. Nothing gets written to `config.md` unless the user clicks Apply in the `SchemaDiffModal` at generation time. This is a genuine correctness defect, not a missing capability: the log actively states something happened that didn't.

## Root cause

The prompt's schema for the `suggestions` field just says `"Markdown description of suggested schema changes with reasoning (1-3 sentences)"` - no instruction on voice. The model (small model in our config - `gemini-3.1-flash-lite`) wrote it as a changelog entry rather than a proposal.

The prompt also works against a voice fix on its own: `new_schema_body`'s field description and CRITICAL bullet frame the model's task in past tense ("...include all sections that should exist **after the update**", "...the COMPLETE schema **after your changes**"). The model is asked to produce the post-change state for `new_schema_body`, then, right next to it, asked to describe the reasoning for that same change in a different voice - nothing in the prompt marks that shift.

Instruct the model to phrase both `suggestions` and `new_schema_body` as describing a proposal, never as a record of a change that already happened:

- `suggestions`: reworded to require recommendation phrasing ("Suggest changing X because Y", not "X was changed"), in the field description in the JSON output-format example and a reinforcing bullet in the CRITICAL constraints list further down the same prompt (matching the same "brief hint at the field, then a harder restatement in CRITICAL" pattern this prompt already uses for its other constraints).
- `new_schema_body`: both its field description and CRITICAL bullet (see Root cause above) reworded to a neutral conditional frame instead, "as it would read if this proposal were accepted".

Verified live in Obsidian: suggestions generate correctly, phrased as recommendations, no regression.

## Patch

Below is the actual diff against our local `main.js` - the bundled/minified plugin build we run, not your real TypeScript source. It won't apply cleanly with `git apply`/`patch` against your repo (different file, different line numbers, and some surrounding-but-unrelated code has minifier-mangled names like `_a24`). Treat it as an exact reference for the logic to translate into `the real prompt source (PROMPTS.suggestSchemaUpdate)`, not a drop-in patch.

Some of the changed lines are CRITICAL-list bullets, which already start with a literal `-` in the prompt text itself - that's why those lines show `--`/`+-`: the first `-`/`+` is the diff marker, the second `-` is the bullet's own character, not a formatting error.

```diff
diff --git a/main.js b/main.js
--- a/main.js
+++ b/main.js
@@ ... @@ reviewed: true
 [Fully preserve user-reviewed existing content here]
 
 ## {{section_new_information}} ({{date}})
 [Only add non-duplicate new information; write "No new information" if none]
 `,
+  // LOCAL PATCH (not upstream): four lines inside this prompt were reworded.
+  // Can't comment inline - this is one continuous template literal, and a
+  // "//" inside it would leak into the actual text sent to the LLM.
+  //   - A few lines below, in the JSON schema's "suggestions" field
+  //     description: search for "RECOMMENDATION".
+  //   - Same JSON schema block, "new_schema_body" field description: search
+  //     for "as it would read if this proposal were accepted".
+  //   - Near the end of this prompt, in the CRITICAL bullet list: search for
+  //     "PROPOSAL that has NOT been applied" and (separately) the second
+  //     "as it would read if this proposal were accepted".
+  // Upstream's wording let the model phrase "suggestions" as if the change
+  // had already happened ("The schema was streamlined..."), which reads as
+  // false/misleading since nothing is applied until the user clicks Apply.
+  // "new_schema_body"'s own description and CRITICAL bullet framed the
+  // model's task in that same past tense ("after the update" / "after your
+  // changes"), directly undermining the fix above - the model is asked to
+  // produce the state after the change while being told to describe the
+  // rationale for it in a different voice. Both spots reworded to a neutral
+  // conditional frame instead, so nothing nearby still argues for past
+  // tense.
   suggestSchemaUpdate: `You are a Wiki Schema advisor. Review the current schema and the latest ingestion analysis.
 
 Current Schema:
 {{schema_content}}
 
@@ ... @@ Consider:
 
 Output JSON format:
 {
   "changes_needed": true,
-  "new_schema_body": "The COMPLETE updated schema body in markdown, starting with the H1 title. Include all sections that should exist after the update (you may keep unchanged sections exactly as-is).",
-  "suggestions": "Markdown description of suggested schema changes with reasoning (1-3 sentences)"
+  "new_schema_body": "The COMPLETE schema body in markdown, starting with the H1 title, as it would read if this proposal were accepted. Include all sections that should exist (you may keep unchanged sections exactly as-is).",
+  "suggestions": "Markdown description, phrased as a RECOMMENDATION (e.g. 'Suggest removing X because...', 'Recommend tightening Y to...') NOT as a statement that the change already happened (e.g. never 'X was changed', 'Y were tightened'). 1-3 sentences with reasoning."
 }
 
 If no changes are needed:
 {
   "changes_needed": false,
@@ ... @@ CRITICAL:
 - The "suggestions" field must be written in the user's UI language below (after this CRITICAL block). Do NOT default to English unless the user's UI language is English.
-- new_schema_body is the COMPLETE schema after your changes \u2014 not a diff, not a patch. The apply path replaces the current body with new_schema_body verbatim.
+- new_schema_body is the COMPLETE schema as it would read if this proposal were accepted \u2014 not a diff, not a patch. The apply path replaces the current body with new_schema_body verbatim.
 - DO NOT include YAML frontmatter (--- ... ---) in new_schema_body. Start directly with the H1 title (e.g. "# Wiki Schema Configuration").
 - DO NOT wrap new_schema_body in markdown code fences (\`\`\`). The parser strips them, but cleaner output reduces parse risk on small models.
 - Preserve all existing sections that are still relevant. Only change what the analysis actually warrants.
+- The "suggestions" text is a PROPOSAL that has NOT been applied yet - nothing has changed until the user reviews and accepts it. Write it as a recommendation, never in past tense as if the change already happened.
 - Output ONLY the JSON, no other text.
 
 User UI language: {{user_language}}`
 };
 
```

Jan-Heldal:
The schema-suggestion generation prompt described the proposed new schema body and its rationale in language that read as already-applied ("after your changes", restated in a CRITICAL block) with no instruction steering the model toward proposal voice. Rewords both to explicitly frame the body as what the schema would read like if the proposal were accepted, and adds a voice instruction so suggestion text is written as a proposal ("would add...") rather than past tense. Verified against the vault's own suggestions.md log: entries generated before this fix read as already-applied; entries generated after read as proposals.

Fixes #594

### DIFF
--- a/src/__tests__/wiki/prompts/suggest-schema-update-voice.test.ts
+++ b/src/__tests__/wiki/prompts/suggest-schema-update-voice.test.ts
@@ -1,0 +1,27 @@
+// Both `new_schema_body`'s description and the `suggestions` field must consistently read as a proposal that hasn't been applied yet — nothing is applied until the user clicks Apply.
+
+import { describe, it, expect } from 'vitest';
+import { GENERATION_PROMPTS } from '../../../wiki/prompts/generation';
+
+describe('suggestSchemaUpdate prompt — proposal voice (#594)', () => {
+  const tpl = GENERATION_PROMPTS.suggestSchemaUpdate;
+
+  it('does not phrase new_schema_body or the CRITICAL restatement as already-applied', () => {
+    expect(tpl).not.toContain('after the update');
+    expect(tpl).not.toContain('schema after your changes');
+  });
+
+  it('new_schema_body description signals a not-yet-applied proposal', () => {
+    expect(tpl).toContain('as it would read IF this proposal were accepted');
+    expect(tpl).toContain('Nothing has been applied yet');
+  });
+
+  it('the CRITICAL restatement uses the same not-yet-applied framing as the field description', () => {
+    expect(tpl).toContain('as it would read IF this proposal were accepted — not a diff, not a patch, and not yet applied');
+  });
+
+  it('suggestions field carries an explicit proposal-voice instruction', () => {
+    expect(tpl).toContain('Phrase this as a proposal that has not been applied yet');
+    expect(tpl).toContain('never past tense');
+  });
+});
--- a/src/wiki/prompts/generation.ts
+++ b/src/wiki/prompts/generation.ts
@@ -278,8 +278,8 @@ Consider:
 Output JSON format:
 {
   "changes_needed": true,
-  "new_schema_body": "The COMPLETE updated schema body in markdown, starting with the H1 title. Include all sections that should exist after the update (you may keep unchanged sections exactly as-is).",
-  "suggestions": "Markdown description of suggested schema changes with reasoning (1-3 sentences)"
+  "new_schema_body": "The COMPLETE schema body, in markdown, as it would read IF this proposal were accepted — starting with the H1 title, including every section that would exist after accepting (unchanged sections may be kept exactly as-is). Nothing has been applied yet.",
+  "suggestions": "Markdown description of the PROPOSED schema changes with reasoning (1-3 sentences). Phrase this as a proposal that has not been applied yet ('would add...', 'proposes to...') — never past tense ('added...', 'changed...'), since Apply has not been clicked."
 }
 
 If no changes are needed:
@@ -290,7 +290,7 @@ If no changes are needed:
 
 CRITICAL:
 - The "suggestions" field must be written in the user's UI language below (after this CRITICAL block). Do NOT default to English unless the user's UI language is English.
-- new_schema_body is the COMPLETE schema after your changes — not a diff, not a patch. The apply path replaces the current body with new_schema_body verbatim.
+- new_schema_body is the COMPLETE schema as it would read IF this proposal were accepted — not a diff, not a patch, and not yet applied. The apply path replaces the current body with new_schema_body verbatim only if and when the user clicks Apply.
 - DO NOT include YAML frontmatter (--- ... ---) in new_schema_body. Start directly with the H1 title (e.g. "# Wiki Schema Configuration").
 - DO NOT wrap new_schema_body in markdown code fences (\`\`\`). The parser strips them, but cleaner output reduces parse risk on small models.
 - Preserve all existing sections that are still relevant. Only change what the analysis actually warrants.


DocTpoint (dismissed):
Prompt-only change, picks up the `new_schema_body` decoupling from the issue as well. Gates green on a merge with `main`. The four tests are string matches against the template, so they guard this wording rather than the model's behaviour, which is fine for what they are. Approving `fcd3ff1`.

Jan-Heldal:
Sorry about the double force push. The first was just to rebase onto newest main. The second was just undoing the first to avoid you having to review again. Now I see that GitHub still flags it as needing another review, so I guess it was pointless.

DocTpoint (approved):
Re-approving `fcd3ff1` — the same commit the 7 Sep review approved, unchanged, and prompt-only at that. Same cause as on #654: `dismiss_stale_reviews_on_push` drops approvals on any push, so the undo that restored this commit dismissed the review it was meant to preserve. The rebase twin `40c56cc` and `fcd3ff1` carry identical file patches, which I checked before re-approving.

Same request as on #654, since the pair moved together: no rebase. Required checks here are not strict, and the PR run builds `refs/pull/655/merge`, so being 13 commits behind `main` is not what is blocking this. The pending run is — both of today's are at `action_required`, and the green Gate 1 on this head dates from 7 Sep. Release the run and this is ready.

green-dalii (approved):
Verified. Both description strings and the CRITICAL block now consistently frame `new_schema_body` as "what the schema would read like if the proposal were accepted" and steer `suggestions` toward proposal voice, with the past-tense phrasing removed rather than merely discouraged. Checking the vault's own `suggestions.md` log before and after is a stronger verification than asserting the prompt text, and the new test pins the wording so a future prompt edit cannot silently regress it. Merging.

PR #656, opened 2026-09-07
fix(schema): keep config.md's audit-trail metadata in sync on every Apply, in UTC

### LINKED ISSUE #597: Applying a schema suggestion never updates config.md metadata fields
# Bug: Applying a schema suggestion never updates config.md metadata fields

**Area:** `applySchemaSuggestion()` / `spliceBody()` (`src/schema/apply-suggestion.ts`), `wiki/schema/config.md`
**Filed as:** <https://github.com/green-dalii/obsidian-llm-wiki/issues/597>

Confirmed still present in the current `main` branch as of 2026-08-31: `spliceBody()` still carries the original frontmatter through untouched, with no equivalent of `bumpSchemaMetadata()` anywhere in the codebase.

Existing metadata fields: `version` / `updated` / `auto_suggestion_count`
Fixing field updates: `updated` / `auto_suggestion_count`

## Symptom

`schema/config.md`'s frontmatter is:

```yaml
---
version: 1
updated: 2026-08-24
auto_suggestion_count: 0
---
```

After using Apply (confirmed via the file's own mtime and via freshly-created, correctly-timestamped backup files alongside it), the frontmatter is untouched: still `version: 1`, `updated: 2026-08-24`, `auto_suggestion_count: 0`, regardless of how many times a suggestion has actually been applied since. The backup mechanism itself works correctly (rotates, timestamps correctly) - it's specifically these three frontmatter fields that never move.

## Root cause

`spliceBody()` - the function Apply uses to combine the new body with the existing file - carries the entire original frontmatter block through byte-for-byte and only ever replaces the body beneath it:

```js
function spliceBody(originalContent, newBody) {
  if (!originalContent.startsWith("---")) return newBody;
  const end = originalContent.indexOf("---", 3);
  if (end <= 0) return newBody;
  const frontmatter = originalContent.substring(0, end + 3);
  return `${frontmatter}\n\n${newBody}`;
}
```

It has no logic to touch `version`, `updated`, or `auto_suggestion_count`. Not a narrow gap either - grepped the whole bundle for anywhere these three fields are *written* (not just parsed back out by `parseConfigFile`), and the only hits are the two schema-file-creation functions (`ensureSchemaExists`, `regenerateDefaultSchema`), which each hardcode a value once, at creation time. Nothing anywhere - Apply, ingest, the suggestion flow, anything - ever updates them afterward, despite `parseConfigFile()` parsing them out into a structured shape that implies they're meant to track something over time.

Added `bumpSchemaMetadata(originalContent, suggestionTimestamp)`, called from `applySchemaSuggestion()` just before `spliceBody()`:

- `updated` is set to the *exact* `suggestion.timestamp` string - the same value already used as the literal `## Suggestion — {timestamp}` header text in `suggestions.md` - rather than the moment Apply happens to be clicked. That makes it an exact, unambiguous lookup key back to that entry's rationale, not just "some suggestion from today" if more than one was applied the same day.
- `auto_suggestion_count` is parsed out of the current frontmatter (defaulting to 0 if the field is missing), incremented, and written back.
- `version` is left untouched - its intended semantics (schema-format version? content-revision counter?) aren't clear enough from the code alone to bump safely without guessing wrong.

Both writes delegate to the existing `upsertFrontmatterField()` helper (already used elsewhere in the codebase, e.g. for `contentHash`) rather than a bare regex `.replace()`, so a manually-edited `config.md` missing one of these keys - or missing its whole frontmatter block - gets it inserted/created instead of silently staying unbumped. `auto_suggestion_count`'s increment can't move fully into `upsertFrontmatterField()` itself, since that helper only ever writes a literal value and has no read-then-transform capability - the current count still has to be read and incremented first, only the final write delegates.

The `updated` write runs first, which guarantees the frontmatter block exists in the returned content either way (found as-is, or freshly created) - so reading `auto_suggestion_count` off that guaranteed-present block afterward needs no separate existence check.

Left the actual field-bumping logic in a separate function rather than folding it into `spliceBody()` itself - `spliceBody()` is generic and used elsewhere in the plugin for other file-writing paths, so keeping the bump as an explicit, separate step avoids coupling those other paths to this same behavior, and makes the new function easy to reuse wherever the same metadata treatment is wanted.

Tested live in Obsidian across all frontmatter states: complete, partially missing fields, fully missing frontmatter, and a stray unterminated `---` (falls back to `upsertFrontmatterField()`'s own pre-existing from-scratch-block behavior, leaving the old stray `---` as inert body text rather than corrupting anything).

## Patch

Below is the actual diff against our local `main.js` - the bundled/minified plugin build we run, not your real TypeScript source. It won't apply cleanly with `git apply`/`patch` against your repo (different file, different line numbers, and some surrounding-but-unrelated code has minifier-mangled names like `_a24`). Treat it as an exact reference for the logic to translate into `src/schema/apply-suggestion.ts`, not a drop-in patch.

```diff
diff --git a/main.js b/main.js
index 22e1667..23d49e4 100644
--- a/main.js
+++ b/main.js
@@ ... @@ function rotateBackups(allPaths) {
   if (allPaths.length <= MAX_BACKUPS) return [];
   return allPaths.slice(0, allPaths.length - MAX_BACKUPS);
 }
 
 // src/schema/apply-suggestion.ts
+// LOCAL PATCH (not upstream): bumpSchemaMetadata() and its call below don't
+// exist upstream. Applying a suggestion left version/updated/
+// auto_suggestion_count in config.md's frontmatter frozen at whatever they
+// were when the file was first created - spliceBody() carries the
+// frontmatter through untouched, and nothing else in the plugin ever writes
+// new values for these fields after creation (confirmed by grepping the
+// whole bundle). version's semantics aren't clear enough from the code
+// alone to bump safely, so this only touches updated/auto_suggestion_count.
+//
+// `updated` is set to the exact suggestion.timestamp string - the same
+// value used as the "## Suggestion — {timestamp}" header in
+// suggestions.md - rather than the moment Apply happens to be clicked, so
+// it's an exact, unambiguous lookup key back to that entry's rationale.
+//
+// Delegates the actual writes to the existing upsertFrontmatterField() (used
+// elsewhere for contentHash) rather than a bare regex .replace(), so a
+// manually-edited config.md missing one of these keys - or missing its whole
+// frontmatter block - gets it inserted/created instead of silently staying
+// unbumped. Same helper, same contract, in both cases.
+function bumpSchemaMetadata(originalContent, suggestionTimestamp) {
+  let bumped = upsertFrontmatterField(originalContent, "updated", suggestionTimestamp);
+  const end = bumped.indexOf("---", 3);
+  const frontmatter = bumped.substring(0, end + 3);
+  const countMatch = frontmatter.match(/^auto_suggestion_count:\s*(\d+)/m);
+  const nextCount = (countMatch ? parseInt(countMatch[1], 10) : 0) + 1;
+  bumped = upsertFrontmatterField(bumped, "auto_suggestion_count", nextCount);
+  return bumped;
+}
 async function applySchemaSuggestion(params) {
   var _a24, _b17;
-  const { app, currentPath, newBody, onCacheInvalidate } = params;
+  const { app, currentPath, newBody, onCacheInvalidate, suggestionTimestamp } = params;
   const now2 = (_a24 = params.now) != null ? _a24 : (() => /* @__PURE__ */ new Date());
   const file2 = app.vault.getAbstractFileByPath(currentPath);
   if (!(file2 instanceof import_obsidian20.TFile)) {
     return { success: false, reason: "source-missing" };
   }
@@ ... @@ async function applySchemaSuggestion(params) {
   const toDelete = rotateBackups(allBackups);
   for (const p of toDelete) {
     const f = app.vault.getAbstractFileByPath(p);
     if (f instanceof import_obsidian20.TFile) await app.fileManager.trashFile(f);
   }
-  const newContent = spliceBody(originalContent, newBody);
+  // LOCAL PATCH: bump version/updated/auto_suggestion_count before splicing
+  // in the new body - see bumpSchemaMetadata() above.
+  const newContent = spliceBody(bumpSchemaMetadata(originalContent, suggestionTimestamp), newBody);
   await app.vault.modify(file2, newContent);
   onCacheInvalidate == null ? void 0 : onCacheInvalidate();
   return { success: true, backupPath: bakPath };
 }
 function spliceBody(originalContent, newBody) {
@@ ... @@ var schemaCommands = {
             var _a25, _b18;
             const result = await applySchemaSuggestion({
               app: this.app,
               currentPath: `${this.settings.wikiFolder}/schema/config.md`,
               newBody: (_a25 = suggestion.newSchemaBody) != null ? _a25 : "",
-              onCacheInvalidate: () => this.schemaManager.invalidateCache()
+              onCacheInvalidate: () => this.schemaManager.invalidateCache(),
+              // LOCAL PATCH: see bumpSchemaMetadata() - ties config.md's
+              // `updated` field back to this exact suggestions.md entry.
+              suggestionTimestamp: suggestion.timestamp
             });
             if (result.success) {
               const restore = t.schemaDiffRestoreHint ? t.schemaDiffRestoreHint.replace("{path}", result.backupPath) : `Backup saved to ${result.backupPath}. To restore, rename that file back to wiki/schema/config.md in your file explorer.`;
               new import_obsidian21.Notice(restore, NOTICE_RATE_LIMIT);
             } else {
```

Jan-Heldal:
apply-suggestion.ts's own header comment promises every Apply leaves an audit trail in wiki/schema/config.md's frontmatter (updated, auto_suggestion_count), but spliceBody() carries the original frontmatter through byte-for-byte and nothing ever wrote to those two fields outside of schema-manager.ts's page-creation path.

Adds bumpSchemaMetadata(), called right after spliceBody, which bumps three frontmatter fields in place, leaving every other field (including a user-added one) untouched, via the existing upsertFrontmatterField helper rather than hand-rolled regex:

- updated: is always the apply-time local calendar date (localDateStamp(now)), matching the project-wide date-stamp convention — it reflects when config.md actually changed, never the moment the suggestion was generated.
- auto_suggestion_count: increments by one on every apply.
- applied_suggestion: is a new field, written only when the originating suggestion is known. It carries that suggestion's generation timestamp verbatim (raw UTC ISO, unchanged), so this entry can be exact-string-matched against its corresponding wiki/schema/suggestions.md log entry without overloading updated:'s own meaning.
- bumpSchemaMetadata now correctly creates a fresh frontmatter block when a file genuinely has none, instead of silently no-oping.

Follow-up (review feedback + hardening):
A config.md with intact frontmatter but a damaged opening delimiter — a leading BOM, leading whitespace, or a first line with the wrong dash count (e.g. --/----) — was being treated identically to "no frontmatter at all," prepending a duplicate synthetic block ahead of the existing one instead of recognizing it. normalizeFrontmatterOpening (src/core/frontmatter.ts) repairs those specific, unambiguous cases in place before parsing, and preserves a CRLF-terminated delimiter's line endings rather than flipping it to LF. It deliberately doesn't attempt to recover frontmatter missing its opening delimiter entirely, since that's ambiguous with plain body content.

Fixes #597

### DIFF
--- a/src/__tests__/schema/apply-suggestion.test.ts
+++ b/src/__tests__/schema/apply-suggestion.test.ts
@@ -1,5 +1,5 @@
 import { describe, it, expect } from 'vitest';
-import { applySchemaSuggestion } from '../../schema/apply-suggestion';
+import { applySchemaSuggestion, bumpSchemaMetadata } from '../../schema/apply-suggestion';
 import { App, TFile } from 'obsidian'; // mocked in setup.ts
 
 // v1.22.0 #97: business logic for "apply a Schema suggestion" — the
@@ -173,5 +173,106 @@ describe('applySchemaSuggestion (#97)', () => {
       onCacheInvalidate: () => { invalidated++; },
     });
     expect(invalidated).toBe(1);
+  });
+
+  // spliceBody preserves the original frontmatter verbatim, so `updated`/`auto_suggestion_count` must be bumped separately — see bumpSchemaMetadata.
+  describe('audit-trail metadata bump (#597)', () => {
+    it('sets updated to the apply date and increments auto_suggestion_count', async () => {
+      const vault = mkMockVault({ 'wiki/schema/config.md': CURRENT_FILE });
+      await applySchemaSuggestion({
+        app: vault as unknown as App,
+        currentPath: 'wiki/schema/config.md',
+        newBody: NEW_BODY,
+        now: () => new Date('2026-06-22T10:30:00.000Z'),
+      });
+
+      const written = vault.vault.files.get('wiki/schema/config.md')!;
+      expect(written).toContain('updated: 2026-06-22');
+      expect(written).not.toContain('updated: 2026-06-21');
+      expect(written).toContain('auto_suggestion_count: 1');
+      expect(written).toContain('version: 1');
+    });
+
+    it('increments auto_suggestion_count again on a second apply', async () => {
+      const vault = mkMockVault({ 'wiki/schema/config.md': CURRENT_FILE });
+      await applySchemaSuggestion({
+        app: vault as unknown as App,
+        currentPath: 'wiki/schema/config.md',
+        newBody: NEW_BODY,
+        now: () => new Date('2026-06-22T10:30:00.000Z'),
+      });
+      await applySchemaSuggestion({
+        app: vault as unknown as App,
+        currentPath: 'wiki/schema/config.md',
+        newBody: '# Wiki Schema\n\n## Wiki Structure\n- Entity pages (custom v2)\n',
+        now: () => new Date('2026-06-23T09:00:00.000Z'),
+      });
+
+      const written = vault.vault.files.get('wiki/schema/config.md')!;
+      expect(written).toContain('updated: 2026-06-23');
+      expect(written).toContain('auto_suggestion_count: 2');
+    });
+
+    // Apply can happen arbitrarily later than suggestion generation, so a fresh `now()` at apply time — even truncated to a date — would never equal the suggestion's own timestamp, and same-day suggestions would become indistinguishable by `updated:` alone.
+    it('uses the exact suggestion timestamp for `updated`, not a date derived from apply-time `now`', async () => {
+      const vault = mkMockVault({ 'wiki/schema/config.md': CURRENT_FILE });
+      await applySchemaSuggestion({
+        app: vault as unknown as App,
+        currentPath: 'wiki/schema/config.md',
+        newBody: NEW_BODY,
+        // Deliberately different day/time than the suggestion timestamp, to prove `updated:` doesn't come from apply-time `now()` at all.
+        now: () => new Date('2026-09-05T23:59:59.000Z'),
+        suggestionTimestamp: '2026-09-04T11:38:14.724Z',
+      });
+
+      const written = vault.vault.files.get('wiki/schema/config.md')!;
+      expect(written).toContain('updated: 2026-09-04T11:38:14.724Z');
+      expect(written).not.toContain('updated: 2026-09-05');
+    });
+
+    // Pin a timezone far enough ahead of UTC (UTC+14) that a UTC instant just before midnight already falls on the next calendar day locally — the exact condition where a local-time value and a UTC value disagree — so this test is deterministic regardless of the machine it runs on.
+    it('falls back to a UTC ISO timestamp (not local time) when no suggestionTimestamp is given', async () => {
+      const originalTz = process.env.TZ;
+      process.env.TZ = 'Pacific/Kiritimati'; // UTC+14
+      try {
+        const vault = mkMockVault({ 'wiki/schema/config.md': CURRENT_FILE });
+        const fixedNow = new Date('2026-06-22T23:30:00.000Z'); // already 2026-06-23 locally at UTC+14
+        await applySchemaSuggestion({
+          app: vault as unknown as App,
+          currentPath: 'wiki/schema/config.md',
+          newBody: NEW_BODY,
+          now: () => fixedNow,
+        });
+
+        const written = vault.vault.files.get('wiki/schema/config.md')!;
+        expect(written).toContain(`updated: ${fixedNow.toISOString()}`);
+        expect(written).not.toContain('updated: 2026-06-23');
+      } finally {
+        if (originalTz === undefined) delete process.env.TZ;
+        else process.env.TZ = originalTz;
+      }
+    });
+  });
+});
+
+describe('bumpSchemaMetadata (#597)', () => {
+  it('rewrites updated and increments auto_suggestion_count, leaving other lines untouched', () => {
+    const result = bumpSchemaMetadata(CURRENT_FILE, '2026-06-22');
+    expect(result).toContain('version: 1');
+    expect(result).toContain('updated: 2026-06-22');
+    expect(result).toContain('auto_suggestion_count: 1');
+    expect(result).toContain(CURRENT_BODY.trim());
+  });
+
+  it('appends both fields when neither is present in the frontmatter', () => {
+    const noMeta = '---\nversion: 1\n---\n\n# Body\n';
+    const result = bumpSchemaMetadata(noMeta, '2026-06-22');
+    expect(result).toContain('updated: 2026-06-22');
+    expect(result).toContain('auto_suggestion_count: 1');
+  });
+
+  it('is a no-op on content with no frontmatter', () => {
+    const noFm = '# Just a body\n';
+    expect(bumpSchemaMetadata(noFm, '2026-06-22')).toBe(noFm);
   });
 });
--- a/src/__tests__/schema/apply-suggestion.test.ts
+++ b/src/__tests__/schema/apply-suggestion.test.ts
@@ -213,26 +213,46 @@ describe('applySchemaSuggestion (#97)', () => {
       expect(written).toContain('auto_suggestion_count: 2');
     });
 
-    // Apply can happen arbitrarily later than suggestion generation, so a fresh `now()` at apply time — even truncated to a date — would never equal the suggestion's own timestamp, and same-day suggestions would become indistinguishable by `updated:` alone.
-    it('uses the exact suggestion timestamp for `updated`, not a date derived from apply-time `now`', async () => {
+    // Apply can happen arbitrarily later than suggestion generation. `updated:` must
+    // still be the apply-time local date — the suggestion's own timestamp goes to a
+    // separate `applied_suggestion:` field instead, verbatim.
+    it('writes applied_suggestion verbatim from suggestionTimestamp, without changing updated', async () => {
       const vault = mkMockVault({ 'wiki/schema/config.md': CURRENT_FILE });
       await applySchemaSuggestion({
         app: vault as unknown as App,
         currentPath: 'wiki/schema/config.md',
         newBody: NEW_BODY,
-        // Deliberately different day/time than the suggestion timestamp, to prove `updated:` doesn't come from apply-time `now()` at all.
-        now: () => new Date('2026-09-05T23:59:59.000Z'),
+        // Deliberately different day than the suggestion timestamp, to prove `updated:` doesn't come from it.
+        // Midday UTC so the local calendar date is unambiguous regardless of the machine's own time zone.
+        now: () => new Date('2026-09-05T12:00:00.000Z'),
         suggestionTimestamp: '2026-09-04T11:38:14.724Z',
       });
 
       const written = vault.vault.files.get('wiki/schema/config.md')!;
-      expect(written).toContain('updated: 2026-09-04T11:38:14.724Z');
-      expect(written).not.toContain('updated: 2026-09-05');
+      expect(written).toContain('updated: 2026-09-05');
+      expect(written).toContain('applied_suggestion: 2026-09-04T11:38:14.724Z');
+    });
+
+    it('omits applied_suggestion entirely when no suggestionTimestamp is given', async () => {
+      const vault = mkMockVault({ 'wiki/schema/config.md': CURRENT_FILE });
+      await applySchemaSuggestion({
+        app: vault as unknown as App,
+        currentPath: 'wiki/schema/config.md',
+        newBody: NEW_BODY,
+        now: () => new Date('2026-06-22T10:30:00.000Z'),
+      });
+
+      const written = vault.vault.files.get('wiki/schema/config.md')!;
+      expect(written).not.toContain('applied_suggestion');
     });
 
     // Pin a timezone far enough ahead of UTC (UTC+14) that a UTC instant just before midnight already falls on the next calendar day locally — the exact condition where a local-time value and a UTC value disagree — so this test is deterministic regardless of the machine it runs on.
-    it('falls back to a UTC ISO timestamp (not local time) when no suggestionTimestamp is given', async () => {
-      const originalTz = process.env.TZ;
+    it('uses the LOCAL date (not UTC) for updated, with no applied_suggestion when none given', async () => {
+      // Node/V8 caches the resolved zone on first use and does not notice
+      // `delete process.env.TZ` afterwards, only an explicit new value — so
+      // restore by assigning the system's own resolved zone name, never by
+      // deleting the var, or later tests silently keep running as Kiritimati.
+      const systemTz = Intl.DateTimeFormat().resolvedOptions().timeZone;
       process.env.TZ = 'Pacific/Kiritimati'; // UTC+14
       try {
         const vault = mkMockVault({ 'wiki/schema/config.md': CURRENT_FILE });
@@ -245,11 +265,12 @@ describe('applySchemaSuggestion (#97)', () => {
         });
 
         const written = vault.vault.files.get('wiki/schema/config.md')!;
-        expect(written).toContain(`updated: ${fixedNow.toISOString()}`);
-        expect(written).not.toContain('updated: 2026-06-23');
+        expect(written).toContain('updated: 2026-06-23');
+        expect(written).not.toContain(fixedNow.toISOString());
+        expect(written).not.toContain('updated: 2026-06-22');
+        expect(written).not.toContain('applied_suggestion');
       } finally {
-        if (originalTz === undefined) delete process.env.TZ;
-        else process.env.TZ = originalTz;
+        process.env.TZ = systemTz;
       }
     });
   });
@@ -257,22 +278,37 @@ describe('applySchemaSuggestion (#97)', () => {
 
 describe('bumpSchemaMetadata (#597)', () => {
   it('rewrites updated and increments auto_suggestion_count, leaving other lines untouched', () => {
-    const result = bumpSchemaMetadata(CURRENT_FILE, '2026-06-22');
+    const result = bumpSchemaMetadata(CURRENT_FILE, new Date('2026-06-22T10:30:00.000Z'));
     expect(result).toContain('version: 1');
     expect(result).toContain('updated: 2026-06-22');
     expect(result).toContain('auto_suggestion_count: 1');
     expect(result).toContain(CURRENT_BODY.trim());
   });
 
-  it('appends both fields when neither is present in the frontmatter', () => {
+  it('adds applied_suggestion verbatim when suggestionTimestamp is given', () => {
+    const result = bumpSchemaMetadata(
+      CURRENT_FILE,
+      new Date('2026-06-22T10:30:00.000Z'),
+      '2026-06-21T08:00:00.000Z'
+    );
+    expect(result).toContain('applied_suggestion: 2026-06-21T08:00:00.000Z');
+  });
+
+  it('appends all three fields fresh when none are present in the frontmatter', () => {
     const noMeta = '---\nversion: 1\n---\n\n# Body\n';
-    const result = bumpSchemaMetadata(noMeta, '2026-06-22');
+    const result = bumpSchemaMetadata(noMeta, new Date('2026-06-22T10:30:00.000Z'), '2026-06-21T08:00:00.000Z');
     expect(result).toContain('updated: 2026-06-22');
     expect(result).toContain('auto_suggestion_count: 1');
+    expect(result).toContain('applied_suggestion: 2026-06-21T08:00:00.000Z');
   });
 
   it('is a no-op on content with no frontmatter', () => {
     const noFm = '# Just a body\n';
-    expect(bumpSchemaMetadata(noFm, '2026-06-22')).toBe(noFm);
+    expect(bumpSchemaMetadata(noFm, new Date('2026-06-22T10:30:00.000Z'))).toBe(noFm);
+  });
+
+  it('is a no-op on content with unterminated frontmatter', () => {
+    const unterminated = '---\nversion: 1\n\n# Body with no closing delimiter\n';
+    expect(bumpSchemaMetadata(unterminated, new Date('2026-06-22T10:30:00.000Z'))).toBe(unterminated);
   });
 });
--- a/src/main-commands/schema-commands.ts
+++ b/src/main-commands/schema-commands.ts
@@ -69,6 +69,8 @@ export const schemaCommands = {
               app: this.app,
               currentPath: `${this.settings.wikiFolder}/schema/config.md`,
               newBody: suggestion.newSchemaBody ?? '',
+              // The exact timestamp already written to this suggestion's suggestions.md entry, not a date derived from whenever the user happens to click Apply — lets the two files be cross-referenced by an exact string match even same-day.
+              suggestionTimestamp: suggestion.timestamp,
               onCacheInvalidate: () => this.schemaManager.invalidateCache(),
             });
             if (result.success) {
--- a/src/main-commands/schema-commands.ts
+++ b/src/main-commands/schema-commands.ts
@@ -69,7 +69,7 @@ export const schemaCommands = {
               app: this.app,
               currentPath: `${this.settings.wikiFolder}/schema/config.md`,
               newBody: suggestion.newSchemaBody ?? '',
-              // The exact timestamp already written to this suggestion's suggestions.md entry, not a date derived from whenever the user happens to click Apply — lets the two files be cross-referenced by an exact string match even same-day.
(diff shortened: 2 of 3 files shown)


DocTpoint (changes requested):
The mechanism is right and the gates are green on a merge with `main`, but three things need a look before this lands.

1. In #597 you wrote that `bumpSchemaMetadata` now delegates to `upsertFrontmatterField`, including creating a missing frontmatter block. The PR uses bare regex replacements and is a no-op on content without frontmatter (the test at line 274 pins that). The helper exists in `src/core/frontmatter.ts`, has a caller, and covers both cases.
2. `updated` receives the suggestion's generation timestamp, not the apply time. The file header promises `updated: <today>` per Apply, and the tests disagree with each other on this ("sets updated to the apply date" at 180, exact suggestion timestamp at 217). A field named `updated` should say when the file changed. If the cross-reference to `suggestions.md` matters, it wants its own field.
3. The fallback writes `now().toISOString()`, explicitly UTC. #612 (merged 4 Sep) settled that vault dates are stamped from local time, and `schema-manager.ts` writes this same field at creation with `localDateStamp()`. After this PR `config.md` would carry `2026-09-04` or `2026-09-04T11:38:14.724Z` depending on which path wrote last.

Suggested shape: both fields via `upsertFrontmatterField`, `updated` as `localDateStamp(now())`, `suggestionTimestamp` dropped or moved to a separate field.

DocTpoint:
The rebase is a clean one — I checked: `0b28deb` and `0057b12` carry identical file patches, only the parent moved from `6a5ba34` to `f18f722`. So the 7 Sep review still applies unchanged and its three points are still open. Nothing was lost on your side either: this repo's ruleset dismisses approving reviews on a push, not change requests.

Since the thread has been quiet, let me concede the point where you have one and put the point I care about on firmer ground.

**The concession.** The cross-reference is a real need: matching an entry in `suggestions.md` to the `config.md` it produced, by exact string, even when several land the same day. I'm not arguing against that. I'm arguing that `updated:` can't be its carrier, because that field already has a meaning and a second writer.

**The measurement.** `schema-manager.ts` writes this same field at creation as `updated: ${today}`, with `today = localDateStamp()` — local, date-only. This PR writes `updated: 2026-09-04T11:38:14.724Z` on Apply. One field in one file would then read `2026-09-10` or `2026-09-04T11:38:14.724Z` depending on which path touched it last, and the ISO one names a moment *before* the file changed. #612 settled the local-date shape for vault stamps on 4 Sep; this reverses it for a single field.

**A shape that gets you both:** `updated: localDateStamp(now())` for what the field already means, plus a separate `applied_suggestion: <suggestion.timestamp>` for the cross-reference. Exact, greppable, no competing writer, and a reader of `config.md` can tell the two apart.

**On point 1, briefly:** `bumpSchemaMetadata` still returns content unchanged when there is no frontmatter, and the test at line 274 pins that behaviour. `upsertFrontmatterField` in `core/frontmatter.ts` prepends a fresh block in exactly that case, takes the three arguments you need, and already has a caller in `wiki-engine.ts`. Two calls to it would replace the whole function.

One meta note: the reasoning for the UTC choice lives in a doc comment on the new parameter. That's the right home for the decision — the next reader needs it — but it isn't a place I can reply to, and I only found it by reading the diff. If you disagree with any of the above, the thread is cheaper for both of us.

Not blocked on you: the CI run on `0057b12` is at `action_required` and never executed, so this head has no gate result either. I'll release the run for whichever commit comes next.

green-dalii:
Thanks for this — the audit-trail gap you found is real and the mechanism you built is the right one. Confirming as maintainer that @DocTpoint's three points are the merge bar, and adding the reasoning behind two of them so the fix does not read as reviewer preference.

**1. Route both fields through `upsertFrontmatterField`.** This is not just code reuse. `src/core/frontmatter.ts` is where the frontmatter *format contract* lives — `mergeFrontmatter`, `parseFrontmatter`, `extractBody` and this helper are the set that defines what a valid block looks like. A bare regex in `apply-suggestion.ts` makes that file a second owner of the format, which is how the two drift. The helper already handles the no-frontmatter case (which is exactly the case the current code silently no-ops on), and it has a production caller at `wiki-engine.ts:1757`, so this is the established path, not a new dependency.

**2. `updated` must mean "when this file last changed".** A field named `updated` that reports when the proposal was *generated* is a different field wearing the same name. If the cross-reference to `suggestions.md` is genuinely useful — and I can see the argument for it when several suggestions land the same day — give it its own key. But `updated` cannot be repurposed; every other file in the vault reads it as "last modified", and `parseFrontmatter` consumers assume that.

**3. `localDateStamp()`, not `toISOString()`.** This is a settled project invariant, not a style choice: #612 (merged 4 Sep) moved all 17 vault date-write sites to local calendar dates after the UTC version produced yesterday's stamp for the first hours of every day east of UTC. `schema-manager.ts` already writes this same field at creation with `localDateStamp()`, and the rest of `frontmatter.ts` follows it (lines 548, 549, 662, 670). The PR's own comment acknowledges the divergence — after this lands, `config.md` would carry either `2026-09-04` or `2026-09-04T11:38:14.724Z` depending on which path wrote last, which is precisely the inconsistency #612 removed.

The two tests asserting opposite semantics (one "apply date", one "exact suggestion timestamp") should collapse into whichever of the two the fix settles on — as written they cannot both be right.

Direction and tests are otherwise good and we want this in. It is tracked against `v1.27.x PATCH`; ping when it is repushed and I will re-run the gates and merge.

One boundary note so it does not surprise you: the strict-endpoint normalisation and the zod 4 migration are being handled separately at the `output-schemas.ts` boundary (#658, #669). Nothing in this PR needs to anticipate them.

DocTpoint (changes requested):
Thanks — points 2 and 3 are done as agreed, and the two contradicting tests are gone.

Point 1 is half done: every write now goes through `upsertFrontmatterField`, but the `parseFrontmatter` guard still returns early on content without frontmatter, and the new test pins that. That is the case the helper was meant to cover — `spliceBody` returns the bare new body when `config.md` has none, so an Apply there still leaves no audit trail. The guard is only needed for unterminated frontmatter, where the helper would prepend a second block:

```ts
if (content.startsWith('---') && !parseFrontmatter(content)) return content;
```

with the no-frontmatter test asserting the created block instead. With that I'll approve the next SHA.

DocTpoint (changes requested):
Changes requested on `1f2bb5c` — Gate 1 is red, and point 1 from the previous round is still open. Both here, so you only have to push once.

**Gate 1 fails at lint, on this branch's own change.** [Run 34604391894](https://github.com/GD4AI/obsidian-llm-wiki/actions/runs/34604391894/job/103279176940), 11 September, 25 seconds in:

```
src/schema/apply-suggestion.ts
  143:38  error  'fm.auto_suggestion_count ?? '0'' will use Object's default
          stringification format ('[object Object]') when stringified
          @typescript-eslint/no-base-to-string
```

The error sits on the line `1f2bb5c` introduced, and `df6eb23` was green, so this is not inherited from `main`. The `typeof` check the refactor removed was not a dead branch — it was the narrowing that made `String()` safe to the rule. `FrontmatterData` ends in an index signature (`src/core/frontmatter.ts:15`, `[key: string]: unknown`), so `auto_suggestion_count` is typed `unknown` no matter what the parser produces at runtime.

One narrow correction to the premise as well. `parseFrontmatter` produces more than strings: the bracket form (`frontmatter.ts:74`) and the `- ` list form both yield string arrays. For `auto_suggestion_count: [5, 6]` the old code took the `'0'` branch and counted from zero, while the new line reads `String(['5','6'])` as `'5,6'` and counts from five. Nothing writes that shape today, so this is about the reasoning in the commit message, not a defect in the field.

The minimal way back to green is to revert `1f2bb5c`; the three lines it replaced passed Gate 1 on `df6eb23`.

**Point 1 from my review on `df6eb23` is unchanged.** `if (!fm) return content;` still stands at `apply-suggestion.ts:141`, so an Apply on a `config.md` that has no frontmatter still leaves no audit trail — the case the helper was meant to cover. The guard is only needed for unterminated frontmatter, where the helper would prepend a second block:

```ts
if (content.startsWith('---') && !parseFrontmatter(content)) return content;
```

with the no-frontmatter test asserting the created block instead. That, plus a green Gate 1, and I will approve.

Jan-Heldal:
@DocTpoint Ready for another look on 24fe315:

- Gate 1 lint: fixed. Reverted to the typed rawCount/countStr narrowing (src/schema/apply-suggestion.ts, bumpSchemaMetadata). pnpm lint and CI (Gate 1 / Five-Gate) are both green on 24fe315.
- Point 1 (missing-frontmatter block creation): fixed. bumpSchemaMetadata now runs content through normalizeFrontmatterOpening (src/core/frontmatter.ts:113) before parsing, and only falls back to creating a fresh block when there's genuinely no recoverable delimiter at all.
- Unprompted extras, found while fixing the above: the same delimiter repair also handles a leading BOM, leading whitespace, and a wrong dash count (--/----), and preserves CRLF line endings instead of flipping them to LF (src/core/frontmatter.ts:113-136). Not something you asked for, flagging in case it's worth a look.

DocTpoint (changes requested):
Re-reviewed on `24fe315`. Both points from the last round are done, and done the way we discussed.

**Gate 1 lint** — fixed. The revert to the typed `rawCount`/`countStr` narrowing is the right shape: the `typeof` check is what makes `String()` safe to `no-base-to-string` while `FrontmatterData` ends in `[key: string]: unknown`.

**Point 1** — fixed. `if (normalized.startsWith('---')) { const parsed = parseFrontmatter(normalized); if (!parsed) return content; }` is exactly the guard, it returns the *original* content on the unterminated path, and the no-frontmatter test now asserts the created block. The extra test that normalization does not rescue a still-unterminated block is a good addition.

Two things in the unprompted extras need a look before I approve, and one is a correctness problem rather than a style one.

### 1. The dash-count repair can write the audit fields into prose

`normalizeFrontmatterOpening` rewrites any first line matching `/^-{2,}\s*$/` to `---`. A `----` line is also a Markdown thematic break, and the function cannot tell the two apart. Run against the functions as they stand on `24fe315`:

```
input:  "----\n\n## Extraction\n\nSome prose.\n\n---\n\n## Merge\n\nMore prose.\n"

output: "---\n\n## Extraction\n\nSome prose.\n\nupdated: 2026-09-14\nauto_suggestion_count: 1\n---\n\n## Merge\n\nMore prose.\n"
```

The rule became a delimiter, the first section became a YAML block, and the two audit fields were inserted into the body. Before this PR the same input was a no-op, because `parseFrontmatter`'s `/^---\n/` did not match `----`.

What makes this more than a curiosity is where the content comes from. `bumpSchemaMetadata` is called on `spliceBody`'s output (`apply-suggestion.ts:93`), and in the no-frontmatter case `spliceBody` returns `newBody` — the model-written body — unchanged. A generated document that opens with a horizontal rule is not a rare shape.

BOM and leading whitespace are safe, because neither changes *what* the delimiter is. The dash count does. Two ways out, whichever you prefer:

- drop the dash-count case and keep the BOM/whitespace repair; or
- gate it on the following line actually looking like a YAML key (`/^[A-Za-z_][\w.-]*\s*:/`), which rejects the thematic break above, since the next line there is empty. A test asserting that a genuine `----` rule is left alone would pin it either way.

### 2. The CRLF handling cannot take effect in this pipeline

`parseFrontmatter` matches `/^---\n([\s\S]*?)\n---/` — LF only. So for CRLF content `bumpSchemaMetadata` takes the `if (!parsed) return content` path and returns the original:

```
input:  "---\r\nupdated: 2026-01-01\r\n---\r\n\r\nText.\r\n"
output: unchanged — `updated:` still 2026-01-01, no count
```

The two CRLF tests pass because they exercise the helper on its own; at the `bumpSchemaMetadata` level there is no CRLF test, and the behaviour there is the same silent skip as before the PR. Nothing is broken by it, but the doc comment and the PR description read as if CRLF files were handled. Either drop the CRLF claim from the comment, or fix `parseFrontmatter` to accept `\r?\n` — that second one has its own blast radius across every caller and is better as a separate PR.

### 3. Minor

`trimStart()` runs before the `startsWith('---')` question is asked, so content with no frontmatter at all also loses its leading blank lines:

```
input:  "\n\n# Config\n\nText.\n"   ->   block created, the two leading newlines gone
```

Harmless in practice, but it is outside what the function's name and doc promise.

Points 1 and 2 addressed (and 3 as you see fit), and I will approve.

green-dalii:
I have pushed `4c6d2fb1` on top of the branch, addressing all three items from the 2026-09-14 review. Two things to say about how it got here.

**First, a rebase.** The branch was six days stale, and its Gate 1 was red for a reason unrelated to your change: `zod`'s `passthrough` is deprecated, and `main` migrated to `.loose()` in #733 while this branch was open. I used `gh pr update-branch --rebase` rather than force-pushing anything myself, then put a single commit on top as a fast-forward — so your five commits were rebased by GitHub, and nothing of mine rewrote them. Gate 1 is green on the result: 305 files / 4283 tests.

**Second, what the commit does, since all three findings were correct and I reproduced the first end to end before changing anything.**

The dash-count repair is the correctness one. `----` is both a wrong dash count and a Markdown thematic break, and the first line cannot tell them apart — so the repair handed `parseFrontmatter` the *next* `---` in the document, which then parsed a block beginning in the middle of the prose:

```
input:  "----\n\n## Extraction\n\nSome prose.\n\n---\n\n## Merge\n"
before: "---\nupdated: …\nauto_suggestion_count: …\n\n## Extraction…"
```

The repair now requires the following line to look like a YAML key, which is the guard you proposed. A `----` whose next line is empty is left alone; `--` and `----` repairs still happen where a key does follow.

For CRLF I dropped the claim rather than made it true, since `parseFrontmatter` matches LF only and accepting `\r?\n` there has every caller as blast radius. The limitation is now in the comment *and* pinned by a test, so the comment cannot drift back into promising it.

For `trimStart()` the original content is returned when there is nothing to repair. Worth flagging: my first attempt made that too broad and broke your two existing BOM and leading-whitespace tests, because trimming *is* the repair in those cases. The distinction is now explicit in the code — the trimmed form is returned when trimming removed something before a valid `---`, and the original is returned when there is no opening at all.

**Tests, and one suggestion taken.** Three at the `bumpSchemaMetadata` level plus one unit test per finding. The strongest is "writes nothing when the document opens with a thematic break", because that is the shape the defect took. And I took the `toContain('updated:')` observation from your earlier revision: the assertion now pins the date twice — present as today's, absent as yesterday's — since a single `toContain` passes for the old value and cannot tell an apply from a no-op.

Mutation-tested both ways: removing the YAML-key guard fails 2, restoring the unconditional trim fails 1.

**On the three extras generally:** they are yours and I kept them, fixing rather than removing. If you would rather not carry `normalizeFrontmatterOpening` at all, say so and I will cut it — everything #597 asks for works without it, and it is the only part that ever needed the guard.

DocTpoint (approved):
Re-reviewed on `4c6d2fb1`. All three items from the 2026-09-14 review are addressed, and I am approving.

The short version first, because that review's closing line has been quoted in #597 as what blocks this PR: it no longer blocks it.

### 1. The dash-count repair — closed, with the guard I proposed

I ran the new `normalizeFrontmatterOpening` against the reported input and against controls, rather than reading it:

| input | result | block `parseFrontmatter` then sees |
|---|---|---|
| `"----\n\n## Extraction\n\nSome prose.\n\n---\n\n## Merge\n"` | unchanged | none |
| `"----\ntype: entity\n---\nBody\n"` | repaired | `type: entity` |
| `"--\nversion: 1\n---\nBody\n"` | repaired | `version: 1` |

The reported defect is gone and the case the repair exists for still works. Returning `content` instead of `trimmed` on the reject paths is right, and the comment explaining *why* the `firstLineNoCr === '---'` branch still returns the trimmed form — because there the trimming is the repair — is the part that will keep the next reader from "simplifying" it back.

### A residual, not blocking, and it is mine rather than yours

The guard narrows the class; it does not close it. A thematic break whose next line begins with a word and a colon still passes:

```
input:    "----\nNote: see the section below\n\n## Extraction\n\n---\n\n## Merge\n"
repaired: "---\nNote: see the section below\n\n## Extraction\n\n---\n\n## Merge\n"
block:    "Note: see the section below\n\n## Extraction\n"
```

This is a residual of the option I offered, not something the implementation got wrong. My review proposed this exact gate and justified it with "the next line there is empty", which is true of the reported shape and not of this one. I am not blocking on it, for two measured reasons:

- I checked 3010 Markdown files in my own vault. Zero open with a dash run other than `---`, so zero would reach the guard at all.
- The population that actually reaches this function is narrower than that: `config.md`'s body as returned by `spliceBody`.

If the shape ever does appear, the other option from that review — drop the dash-count case, keep the BOM and whitespace repair — closes it completely. Worth a sentence in the code comment, not a change now.

### 2. CRLF — dropped rather than faked, which is the right call

`parseFrontmatter` matches `\n` only, so CRLF content takes the unrecognized-opening path and comes back untouched. Making that true end to end means accepting `\r?\n` in `parseFrontmatter`, and every caller of it is the blast radius. Stating the limit and pinning it with a test at the `bumpSchemaMetadata` level is better than a half-fix, and the test is what stops the comment drifting back into a promise.

### 3. `trimStart` — fixed

`'\n\n# Config\n\nText.\n'` now keeps its leading blank lines, and the test asserts the tail rather than the whole string, which is the right shape for an assertion about a created block.

### What I checked, and what I did not

Checked: the four files in `4c6d2fb1`, the behaviour of the new guard against the reported input and three controls, and that the reject paths return the original content. Gate 1 reports `success` on `4c6d2fb1` (2026-09-20 06:41Z).

Not checked: I did not run the suite locally. The "305 files / 4283 tests" figures are from your comment; the check run exposes no summary output, so I did not reproduce them.

One typo, in a test comment: `repaird` → `repaired`, in `src/__tests__/core/frontmatter.test.ts`.

@Jan-Heldal — my earlier findings were on your commits, and the extras that caused this last round were worth having: the BOM and whitespace repair are good, and the one that needed a guard is the one that is now the best-tested path in the file.

green-dalii (approved):
Approving. @DocTpoint's three points are addressed on `4c6d2fb1` and I checked the parts I care about as maintainer.

The `----` disambiguation is the right call and the right reason: a run of dashes is only a damaged opening when a YAML key follows it, because repairing a thematic break hands `parseFrontmatter` the *next* `---` and writes `updated:` into the middle of prose. Requiring the key is what makes the repair safe to run unconditionally.

Two things I want on the record rather than in a review comment.

`normalizeFrontmatterOpening` returning the original (not the trimmed form) on both no-op paths is the correct shape — `trimStart()` is how the question gets asked, and it must not become a side effect of asking it. The comment above the second path says exactly that.

The CRLF paragraph is the most valuable part of this PR. An earlier revision read as if CRLF were handled end to end; now the limitation is stated where a reader will look for it, with the reason it belongs in a separate change. That is the difference between a known gap and a trap.

green-dalii (approved):
Approving. @DocTpoint's three points are addressed on `4c6d2fb1` and I checked the parts I care about as maintainer.

The `----` disambiguation is the right call and the right reason: a run of dashes is only a damaged opening when a YAML key follows it, because repairing a thematic break hands `parseFrontmatter` the *next* `---` and writes `updated:` into the middle of prose. Requiring the key is what makes the repair safe to run unconditionally.

Two things I want on the record rather than in a review comment.

`normalizeFrontmatterOpening` returning the original (not the trimmed form) on both no-op paths is the correct shape — `trimStart()` is how the question gets asked, and it must not become a side effect of asking it. The comment above the second path says exactly that.

The CRLF paragraph is the most valuable part of this PR. An earlier revision read as if CRLF were handled end to end; now the limitation is stated where a reader will look for it, with the reason it belongs in a separate change. That is the difference between a known gap and a trap.

green-dalii (approved):
Approving. @DocTpoint's three points are addressed on `4c6d2fb1` and I checked the parts I care about as maintainer.

The `----` disambiguation is the right call and the right reason: a run of dashes is only a damaged opening when a YAML key follows it, because repairing a thematic break hands `parseFrontmatter` the *next* `---` and writes `updated:` into the middle of prose. Requiring the key is what makes the repair safe to run unconditionally.

Two things I want on the record rather than in a review comment.

`normalizeFrontmatterOpening` returning the original (not the trimmed form) on both no-op paths is the correct shape — `trimStart()` is how the question gets asked, and it must not become a side effect of asking it. The comment above the second path says exactly that.

The CRLF paragraph is the most valuable part of this PR. An earlier revision read as if CRLF were handled end to end; now the limitation is stated where a reader will look for it, with the reason it belongs in a separate change. That is the difference between a known gap and a trap.

PR #660, opened 2026-09-09
fix(merge): render the Related sections after a complementary append

### LINKED ISSUE #659: Complementary merge writes model prose into the Related sections
## Bug

The complementary path of `mergePage` appends into whatever section the triage names in `target_section`, and a Related section is a legal target. What the per-section call writes there is prose with a provenance marker rather than a list entry:

```
- [[concepts/SCFAs|SCFAs]] strengthen the gut barrier by promoting tight-junction proteins. ^[Source: [[Atherosclerosis]]]
```

Those two sections are not meant to carry model text. `renderRelatedSections` writes them from the typed lists, and `applyRelatedLinks` is the pass that does it — but only the create, related-page and rewrite paths call it. The `shouldSkip` branch writes `complementaryBody` straight to disk (`merge-page.ts:259`), so nothing normalises what the append put there.

Two things follow. The bullets keep their prose, which the next rewrite then re-reads as list entries. And the same target can be listed twice, once bare and once with prose, because the de-duplication in `renderRelatedSections` never sees the second one.

## Scale

Measured on a rebuild of 80 notes (539 entity and concept pages): 22 such bullets on 7 pages, and 6 sections listing one target twice. The share is small, but these lines are never removed — a rewrite keeps them, so they accumulate with every complementary merge.

Independent of #645, which is about which siblings a run writes, not about how a section is rendered. PR follows.

DocTpoint:
Closes #659.

Runs the same pass the other three write paths use on the complementary body, so the two Related sections come out as a rendered list instead of whatever the per-section call pasted into them.

Two narrowings, both deliberate:

- **Only when the append actually changed a Related section.** The pass reads every page in the vault, and most appends land elsewhere; the new `relatedListLines` helper is a cheap string comparison that decides it.
- **The typed lists are not passed.** This is here to render what the append wrote, not to add the new source's related names to the page — passing them would make every complementary merge grow the list, which is a different change with a different argument.

`keepFrom` is the appended body rather than the one before it, so the relation the append named survives as an entry and only the prose around the link is dropped.

Three tests on the complementary path: the prose bullet becomes a plain entry, a repeated target is listed once, and an append into another section leaves the list untouched. All three fail without the change.

⚠️ `openai-codex-loopback-flow.test.ts` fails on this branch and on `main` alike — inherited, not from this change.

### DIFF
--- a/src/__tests__/wiki/page-factory/merge-page.test.ts
+++ b/src/__tests__/wiki/page-factory/merge-page.test.ts
@@ -507,3 +507,66 @@ describe('mergePage — the triage lane reports what it records (onContradiction
     expect(seen).toEqual([]);
   });
 });
+
+// The complementary path writes into whatever section the triage names, and a
+// Related section is a legal target — its per-section call produces prose
+// bullets with a provenance marker, which the two Related sections are not
+// supposed to carry: every other write path renders them from the typed lists.
+describe('mergePage — a complementary append into a Related section is rendered, not pasted', () => {
+  const WITH_RELATED = [
+    '---', 'title: Caching', '---', '',
+    '## Description', 'Old text.', '',
+    '## Related Concepts', '', '- [[concepts/Alpha|Alpha]]', '- [[concepts/Beta|Beta]]', '',
+  ].join('\n');
+
+  function run(appended: string) {
+    const ctx = makeCtx(makeClient([
+      JSON.stringify({
+        strategy: 'complementary',
+        reason: 'expand',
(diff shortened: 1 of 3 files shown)


green-dalii (approved):
Verified. `relatedListLines` is a genuinely cheap guard (string comparison over the two sections' bullet lines) in front of a pass that reads every page in the vault, and gating on "the append actually changed a Related section" is what keeps this from becoming an O(vault) tax on every complementary merge. Not passing the typed lists is the right narrowing too — rendering what was written and growing the list are two different changes with two different arguments. Merging.

PR #663, opened 2026-09-09
fix(dedup): a name too short to infer identity from is created, not asked about

### LINKED ISSUE #661: A two-character name reaches the semantic dedup and gets merged on no evidence
In `resolvePagePath`, when `ConflictResolver` finds no page carrying the designator, the name still gets a lexically ranked candidate window and a model call. Measured on 1.27.1 with two different window builders: the item `Cr` was merged into `Kreatinin` in both arms. Two characters carry no information a summary could confirm; the model is guessing.

The deterministic paths are fine and unchanged: one title/alias hit resolves; several hits go through #446's tag-ranked disambiguation, where the model only chooses among pages that actually carry the name.

**Proposal:** when no page carries the designator and the name is shorter than 3 code points, skip the model window and take the create fallback. A separate constant (`MIN_DEDUP_NAME_LENGTH`), **not** `MIN_ALIAS_LENGTH`: what a vault accepts as an alias is the user's setting and stays untouched; this is about what the resolver may infer, and short aliases still resolve deterministically for users who set them.

One file (`path-resolution.ts`), no overlap with the open PRs. PR follows with a test that is red on main.

DocTpoint:
Closes #661.

When no page carries an extracted name, `resolvePagePath` builds a lexical window of pages that merely resemble it and asks the model whether one of them is the same thing. For a two-code-point name that question has no evidence behind it: `Cr` was merged into `Kreatinin` by two different window builders on the same vault.

Below `MIN_DEDUP_NAME_LENGTH` (3) the resolver takes the create fallback before the window is built — but only when nothing claims the name. The deterministic answers are unchanged: one title/alias hit resolves; several go through #446's tag-ranked candidates, where the model chooses among pages that actually carry the name.

The floor applies to names in Latin, Greek or Cyrillic script, where a letter is not a word. A two-character Han, Kana or Hangul name is a complete word and keeps its model call (test included — this came out of the code-review pass).

Deliberately a separate constant from `MIN_ALIAS_LENGTH`: what a vault accepts as an alias is the user's setting and stays untouched; this is about what the resolver may infer. A short alias a user did set still resolves deterministically.

Tests: 5 new (1 red on the parent commit), 3993 → 3998. typecheck/lint clean. `/simplify` moved the guard ahead of the window filter; `/code-review` surfaced the script question.

### DIFF
--- a/src/__tests__/wiki/page-factory/path-resolution.test.ts
+++ b/src/__tests__/wiki/page-factory/path-resolution.test.ts
@@ -503,3 +503,60 @@ describe('resolvePagePath — a cross-folder match re-decides the classification
     expect(ctx.written.get('wiki/concepts/Stress.md')).toContain('type_conflict: entity');
   });
 });
+
+describe('resolvePagePath — a name too short to infer identity from (#661)', () => {
+  // `Cr` reached the semantic call on a vault where no page carried it and was
+  // merged into `Kreatinin` by two different window builders. Two code points
+  // carry nothing a summary could confirm, so when no page claims the name,
+  // the model is not asked.
+  const vaultOf = (files: Record<string, string>) => ({
+    files,
+    mockVault: {
+      getMarkdownFiles: () =>
+        Object.keys(files).map(p => ({ path: p, basename: p.split('/').pop()!.replace(/\.md$/, '') })),
+    },
+  });
+  const stranger = vaultOf({ 'wiki/entities/Kreatinin.md': '---\ntitle: Kreatinin\n---\n\n# page' });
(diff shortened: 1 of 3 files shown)


green-dalii (approved):
Verified. The floor only applies when `seeded.length === 0`, so every deterministic answer (single title/alias hit, #446's tag-ranked disambiguation) is untouched. `ALPHABETIC_NAME` is the part I would have asked for and didn't have to: a two-character Han, Kana or Hangul name is a complete word and keeps its model call, while Latin/Greek/Cyrillic two-character names take the create fallback. That matches the project's existing Unicode stance rather than special-casing length alone.

Merging. Non-blocking nit: the new `ALPHABETIC_NAME` constant sits between the `resolvePagePath` doc comment and the function it documents, which detaches that comment from its target.

PR #671, opened 2026-09-10
fix(sources): apply the alias floor the create path already applies

### LINKED ISSUE #670: Source pages are a third writer of `aliases:` and the only one without a floor
## What

`resolveMinAliasLength` (`src/core/slug.ts:147`) exists, in the words of its own doc comment, so that

> BOTH writers of `aliases:` (the append path and `enforceFrontmatterConstraints` on the create path) resolve the same floor from the same place — before this, only the append path read the setting and the create path silently fell back to the constant.

There is a third writer. `WikiEngine.createSummaryPage` writes `wiki/sources/<slug>.md` from two alias sources and filters neither:

1. the model's own `aliases:`, which arrive verbatim inside `cleanedContent` and are never inspected;
2. the source note's curated aliases, merged on top at `src/wiki/wiki-engine.ts:1772` via `mergeFrontmatterArrayField`.

No length floor, no self-pointing check, no case-fold dedup. `filterRedundantAliases` is never called on this path.

What makes it an omission rather than a design choice is that the same function honours the *other* half of the same contract inline: source-page tags are filtered against the closed `VALID_SOURCE_TAGS` taxonomy at `src/wiki/wiki-engine.ts:1714`, with `DEFAULT_SOURCE_TAG` as the documented last resort. The tag half is enforced. The alias half is not.

## Measured

One vault, mid-rebuild, `minAliasLength: 3`:

| pages | aliases | below the floor | differs from page name by case only |
|---|---|---|---|
| `entities/` + `concepts/` | 757 / 965 | 0 | 0 |
| `sources/` | 131 / 288 | 2 | 1 |

The three: `"MD"` on `Mediterrane-Ernährung.md`, `"IR"` on `Insulinresistenz.md`, `"ARNi"` on `ARNI.md`. The entity and concept pages carry none of either because they go through `enforceFrontmatterConstraints`, which resolves the floor.

## Why it is not cosmetic

58 of those 288 aliases are absent from the source note's own frontmatter — model-written, spread over 51 of the 131 pages. Many are correct synonyms (`Butyrat` → "Buttersäure", `COPD` → "Chronic Obstructive Pulmonary Disease"). Some are not names at all but descriptions of the page:

```
Brot.md          → "Brot-Analyse"
Dopamin.md       → "Dopamin-Zusammenfassung"
Demenz.md        → "Kritik an der Demenzforschung"
Ballaststoffe.md → "Ballaststoff-Übersicht"
```

Nothing decides between the two kinds. And this pool is not inert: the `#185` comment names its consumer, and `fix-dead-link` retargets dead links through it (`fix-dead-link.ts:264`, `:266`). An invented alias on a source page is therefore a link target, and source pages as link targets are a failure mode this project has already paid for once.

A floor does not separate names from descriptions. It removes the subset that is decidable without a model: too short to identify anything, or already resolving to the page it sits on.

## Scope check

- `filterRedundantAliases` and `resolveMinAliasLength` already exist and are already the shared rule; nothing new is needed.
- `parseFrontmatter` normalises `aliases` to `string[]` in all three shapes, so the filter has one input form to handle.
- The filter belongs *after* both inputs are merged. Filtering either alone leaves the other unchecked.
- Existing source-page aliases are not read on re-ingest (only tags are preserved, `#114`), so no user correction is at risk.

## Disclosure

The counts come from a single vault, a single extraction model, and German-language notes; they are an illustration, not a rate. The mechanism is code-level and vault-independent.

One thing this does **not** address: `filterRedundantAliases` measures length in UTF-16 code units, while the dedup floor added in `#663` counts code points. Both floors now sit on the same field from different directions. Pre-existing and shared by all three writers, so out of scope here — flagging it rather than widening this.

PR follows.

DocTpoint:
Fixes #670.

Two commits. The second is the fix; the first is the prerequisite it exposed.

## 1. `fix(frontmatter): stop replaceFrontmatterArrayField growing the seam`

`serializeFrontmatter` returns a block ending on `---` with no trailing newline. The retained body slice starts at the character after the old closing delimiter, so it already begins with the blank line that separated frontmatter from body. Joining the two with another `\n` added one blank line per call:

```
in:  ---\ntype: source\naliases:\n  - "A"\n  - "B"\n---\n\n# Titel
out: ---\ntype: source\naliases:\n  - "A"\n---\n\n\n# Titel
```

Invisible until now: the helper had one production caller (`fix-runners.ts`) on a path nobody diffed byte-for-byte. The alias floor sends `sources/` pages through it, so the gap would have appeared on every page that drops an alias.

Normalises instead of preserving — strip the leading newlines and re-emit exactly one blank line, the shape `enforceFrontmatterConstraints` and this function's own no-frontmatter branch already produce. That also fixes the opposite direction (a body that followed immediately on the next line now gets the blank line) and makes the helper idempotent.

## 2. `fix(sources): apply the alias floor the create path already applies`

Calls `filterRedundantAliases` with `resolveMinAliasLength(this.settings)` on the finished list, after both inputs are merged. Filtering either input alone leaves the other unchecked: the model's aliases arrive inside `cleanedContent`, the curated note aliases merge on top.

Two narrowings worth reading:

- **The block is rewritten only when something is actually dropped.** On the measured vault that is 3 pages of 131; the other 128 keep the bytes the model wrote. No reformat, no reordering, no diff.
- **`existingAliasesAcrossPages` is passed `undefined`**, exactly as the create path does. This PR applies the floor and the self-pointing rule. Cross-page alias uniqueness is a separate decision and is not made here.

## Gate 1

lint 0 · tsc 0 · build · **4036/4036** (284 files) · css-lint 0.

Baseline on `upstream/main` is 4028, and both commits are green in isolation: 4028 → 4030 → 4036, so the branch is bisect-safe.

All counts were taken with a build in place — `openai-codex-loopback-flow.test.ts` asserts against the bundle and fails in a worktree that has never run esbuild. Worth knowing before someone reads a red suite as inherited breakage.

8 new tests: the floor on the model list, the floor on the merged note list, the self-pointing page name, the case collapse, passthrough survival of `source_file` / `generation_complete` / `contentHash` across the rewrite, the no-op path when nothing is dropped, and both seam directions. 4 of them are red on the parent commit.

## One existing expectation changes

The `#185` test listed both `"Exekutive Funktionen"` and `"exekutive Funktionen"` among the curated aliases. The shared filter keeps one. Both readers of this pool already compare case-folded — `fix-dead-link.ts:264` (`a.toLowerCase()` and `slugify(a).toLowerCase()`) and `scanners.ts:146` (`knownTargetsLower`) — so the second variant retargets nothing the first does not, and only widens the alias line in the fix-dead-link prompt. The test now pins the collapse instead of the duplicate.

## The deeper seam, and why it is not in this PR

The honest reading of this fix is that it makes three writers apply the same rule instead of two, which is better than two but is still three. `createOrUpdateFile` is the obvious candidate for one owner: it already normalises content from every path that reaches it — polluted links, `sources:`, heading spacing — and `appendAliases`, `create-page.ts` and `merge-duplicates.ts` all write through it.

Two things stop that being a free move, both checked against `main` rather than assumed:

- **`createOrUpdateFile` has no `reviewed: true` guard.** `enforceFrontmatterConstraints` returns early on a reviewed page and touches only the dates. Putting the alias floor in the gate would rewrite aliases on frozen pages, which is the one thing the review lock exists to prevent.
- **The gate is not the only door.** Six writers still reach the vault through `vault.process` / `vault.modify` directly — `link-retarget.ts`, `lint/phases/preparation.ts`, `engine-internals/log-writer.ts`, `schema/schema-manager.ts`, `schema/apply-suggestion.ts`, `core/sources-normalizer.ts`. That is #603. Until it is settled, "one owner" in the gate would be a claim rather than a fact.

So this PR closes the measured gap at the call site that has it, and leaves the consolidation to #603, where the argument for it already lives.

## Collision

Touches `src/wiki/wiki-engine.ts`, as does #653 — at `getExistingWikiPages` (line 2054), forty lines past this hunk. Different functions, no overlap.

### DIFF
--- a/src/__tests__/core/frontmatter.test.ts
+++ b/src/__tests__/core/frontmatter.test.ts
@@ -989,6 +989,42 @@ source_url: https://example.com/spec
     expect(result).toContain('z');
   });
 
+  // The seam between the rebuilt block and the body: `serializeFrontmatter`
+  // ends on `---` without a trailing newline and the retained slice starts
+  // with whatever followed the old closing delimiter, so joining them with a
+  // `\n` used to add one blank line per call. Two consecutive calls made it
+  // two. Both directions are pinned here.
+  it('replaceFrontmatterArrayField keeps exactly one blank line between frontmatter and body', () => {
+    const content = `---
+type: entity
+tags: [x]
+---
+
+# Body
+
+Text.
+`;
+    const once = replaceFrontmatterArrayField(content, 'tags', ['y']);
+    expect(once).toContain('---\n\n# Body');
+    expect(once).not.toContain('---\n\n\n');
+
+    const twice = replaceFrontmatterArrayField(once, 'tags', ['z']);
+    expect(twice).toContain('---\n\n# Body');
+    expect(twice).not.toContain('---\n\n\n');
+    // The seam is the only thing under test here, so compare the bodies.
+    expect(twice.slice(twice.indexOf('\n---\n') + 5)).toBe(once.slice(once.indexOf('\n---\n') + 5));
+  });
+
+  it('replaceFrontmatterArrayField inserts the blank line when the body followed on the next line', () => {
+    const content = `---
+type: entity
+tags: [x]
+---
+# Body`;
+    const result = replaceFrontmatterArrayField(content, 'tags', ['y']);
+    expect(result).toContain('---\n\n# Body');
+  });
+
   it('mergeFrontmatterArrayField on a page with only canonical fields is byte-identical to v1.25.9', () => {
     // Backward-compat: when no unknown fields exist, the writer must not
     // invent a `---\n---` line just to "have a passthrough section".
--- a/src/__tests__/wiki/wiki-engine-summary-aliases.test.ts
+++ b/src/__tests__/wiki/wiki-engine-summary-aliases.test.ts
@@ -76,12 +76,17 @@ function harnessFor(extraFiles: Record<string, string> = {}) {
 
 describe('WikiEngine.createSummaryPage — Issue #185 aliases propagation', () => {
   it('appends source-note aliases to the generated sources/<slug> page frontmatter', async () => {
+    // 'exekutive Funktionen' is deliberately absent from the expectation:
+    // it differs from 'Exekutive Funktionen' by case only, and both readers
+    // of this pool compare case-folded (fix-dead-link.ts:264
+    // `a.toLowerCase()` / `slugify(a).toLowerCase()`, scanners.ts:146
+    // `knownTargetsLower`). Carrying both retargets nothing extra and only
+    // widens the alias line in the fix-dead-link prompt. Pinned below.
     const curatedAliases = [
       'Exekutive Funktionen',
       'Executive Function',
       'Exekutiv Funktionen',
       'Exekutiven Funktionen',
-      'exekutive Funktionen',
     ];
     const h = harnessFor();
 
@@ -104,6 +109,19 @@ describe('WikiEngine.createSummaryPage — Issue #185 aliases propagation', () =
     for (const a of curatedAliases) {
       expect((fm.aliases as string[])).toContain(a);
     }
+  });
+
+  it('keeps one of two aliases that differ by case only', async () => {
+    const h = harnessFor();
+    const writtenPath: string = await h.engine.createSummaryPage(
+      sourceFile(),
+      makeAnalysis(['Exekutive Funktionen', 'exekutive Funktionen']),
+      [],
+    );
+    const fm = parseFrontmatter(h.files.get(writtenPath)!) ?? {};
+    const aliases = (fm.aliases ?? []) as string[];
+    expect(aliases).toContain('Exekutive Funktionen');
+    expect(aliases).not.toContain('exekutive Funktionen');
   });
 
   it('skips injection when the source note has no frontmatter aliases (no regression)', async () => {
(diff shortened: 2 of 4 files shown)


DocTpoint:
Correcting one sentence in the description of the first commit before anyone reviews against it. The diff does not change; the stated motivation was too strong.

I wrote that the extra blank line "would have appeared on every page that drops an alias". That is not what happens. `createOrUpdateFile` runs `normalizeHeadingSpacing` on every write into a wiki content folder, and that pass collapses **any** run of blank lines to a single one, not only the ones before a heading. The vault confirms it: across 948 pages in `sources/`, `entities/` and `concepts/`, every single one carries exactly one blank line after the closing `---`. The seam defect has never reached disk.

What is measured, and what the commit still rests on:

- The defect is real. Probing the helper directly on `---\ntype: source\naliases:\n  - "A"\n  - "B"\n---\n\n# Titel` returns `---\n…\n---\n\n\n# Titel` — three newlines where the input had two.
- It is invisible because a later pass repairs it. That repair is conditional: `normalizeHeadingSpacing` only runs when `isInWikiContentFolder` is true, and the writers that reach the vault through `vault.process` / `vault.modify` directly (#603) never pass through the gate at all.
- `mergeFrontmatterArrayField` has the identical seam construction and, measured the same way, the identical output. It is not touched here — flagging it rather than widening this PR.

So the argument for the first commit is not "this bug is visible in the vault". It is that a helper should be correct where it is written rather than correct because something downstream cleans up after it, and that the cleanup does not cover every caller. If you would rather see the seam fix land on its own, with `mergeFrontmatterArrayField` alongside it, say so and I will split it out — the alias floor in the second commit does not depend on it once the reason is stated honestly.

Found while measuring an unrelated question in the same vault. Reported here rather than left standing.

green-dalii (approved):
Verified both commits, and the second one's prerequisite is the more interesting half.

The fix itself is right: the sources/ page was the third writer of `aliases:` and the only one with no floor. I confirmed the asymmetry that makes it an omission rather than a design choice — the same function filters the *tag* half against the closed `VALID_SOURCE_TAGS` taxonomy at `wiki-engine.ts:1714-1716` with `DEFAULT_SOURCE_TAG` as a documented fallback, while the alias half goes through `mergeFrontmatterArrayField` at `:1772` unfiltered, even though `resolveMinAliasLength` (`slug.ts:147`) and `filterRedundantAliases` (`frontmatter.ts:803`) exist for exactly this. Routing the summary path through the same helper is the correct shape, not a new mechanism.

The `replaceFrontmatterArrayField` blank-line seam is a genuine find. `serializeFrontmatter` returns a block ending on `---` with no trailing newline, the retained body slice already begins with the blank line that separated frontmatter from body, and joining with another `\n` adds one blank line per call. It was invisible only because the helper had one production caller on a path nobody diffed byte-for-byte — which is itself the reason it is worth fixing here rather than later: the alias floor would have put `sources/` pages through it, turning a latent defect into one that appears on every page that drops an alias.

Merging.

PR #673, opened 2026-09-10
feat(tags): one vocabulary — settings list, note tags and page tags in one list, read by prompt, gate, lint and retag alike

### LINKED ISSUE #672: Two tag vocabularies: the prompt offers one list, the write gate enforces another — 7.6 % of pages are born tagless and the lint cannot see them
Measured on 1.27.1 with `tagVocabularyMode: custom`, a 903-page rebuild (entities + concepts) and the domain-axis harvest from #568/#569.

## What the code does today

There are two lists that both call themselves "the active vocabulary":

- **List A — `getActiveEntityTags` / `getActiveConceptTags`** (settings `customEntityTags` / `customConceptTags`, or the built-in taxonomy). Read by the system prompt (`buildActiveTagVocabularySection`: "you MUST use one of the following"), by the #527 type repair (`foldToVocabulary` + the repair call), by `scanTagViolations` and by `runRetagViolations`.
- **List B — `collectActiveVocabulary`** (nested tags harvested from the watched folders plus every file under `wikiFolder` except `sources/`). Read by `enforceFrontmatterConstraints` via `domainVocabulary` (`create-page.ts`, `fill-empty-page.ts`, `merge-duplicates.ts`) and by `incomingTypeTag` on the merge and related paths.

The pass strips every tag that is not in B. So a type the prompt explicitly allowed (A), that the #527 repair folded *onto* A, is deleted at the gate if no note happens to carry it — with no replacement.

## What that produced

| | |
|---|---|
| Pages with `tags:` present and empty | **69 of 903** (39 entities, 30 concepts), 7.6 % |
| Terms in A that B does not carry (A∖B) | 50 — `Sorte/Organisation`, `Sorte/Klassifikationssystem`, `Sorte/Verfahren`, `Sorte/Medizinprodukt`, all 13 `Fach/*`, 20 `Thema/*`, and the built-in `person`, `project`, `theory`, `term`, `phenomenon`, … |
| Terms in B that A does not carry | 0 |

The tagless pages are exactly what A∖B predicts: organisations (WHO, CDC, EFSA, Monash University, Lancet Commission), classification systems (ICD-11, DSM-5-TR, ICF, GOLD report), a procedure (Trier Social Stress Test), persons, and abstract concepts the model types as `phenomenon`/`term`. The empty `tags:` line is the signature of the pass: it found tags, stripped all of them, and emitted the bare key (`emitEmptyTags`).

**Nobody can see it.** `scanTagViolations` skips an empty or missing tag list as "not a violation" (scanners.ts, `continue`). The gate produces precisely the state the auditor is coded to ignore. And had a tag survived, the lint would have judged it against A — so the auditor and the gate disagree on what is valid.

**B is contaminated and self-seeding.** `collectWikiVocabulary` harvests every markdown file under `wikiFolder` except `sources/`. In our vault that includes archive copies under `wiki/schema/belege-*`; `Sorte/Mechanismus` reached B only from those files (no note carries it) and now stands on 102 pages. More generally: once any page carries a term, B carries it, so what the first page was allowed to carry depended on what had been written before — the vocabulary is a function of ingest order.

## Root cause

#569 made the vault's tags the vocabulary at the gate, the merge and the related paths, but left the prompt, the type repair, the scanner and the retag runner on the settings list. One decision, implemented at three call sites and not at four others. The two lists are only consistent for a user whose notes carry every settings term.

## Proposal — built, PR attached

One vocabulary, three sources, every reader the same list:

1. **`activeVocabulary(app, settings, kind)`** (`core/vocabulary.ts`) = fold-dedup of the watched folders' nested tags ∪ the nested tags of `entities/`, `concepts/` and `sources/` pages ∪ the settings list (`customEntityTags` / `customConceptTags`). The settings list is the place for a term no note carries yet; a user without tagged notes keeps today's built-in taxonomy unchanged. Prompt (`buildSystemPrompt` takes the lists), #527 type repair, `firstActiveTag`, gate on every path, `scanTagViolations`, `runRetagViolations` and the merge writers all read it. `domainVocabulary` is the same list restricted to `Group/Value` shapes — the domains axis, so a vault without note tags still gets no domain block.
2. **Harvest only from pages.** `collectWikiVocabulary` walks `entities/`, `concepts/`, `sources/`, not `wikiFolder` — same class as the frontier fix that stopped treating everything under `wiki/` as a page.
3. **An empty tag list is a lint violation**, so the pages this bug produced are repaired by the existing retag runner instead of a re-ingest.
4. **Source pages pass the gate and carry the domain axis.** Their `tags:` was written by the model with no check at all (5 of 231 on our vault outside the closed form list, `theory` among them — the model's third option next to the form value and the note's tags). They now pass `enforceFrontmatterConstraints('source')`: form values from `VALID_SOURCE_TAGS` stay legal next to the `Group/Value` view of the vocabulary; the flat identity types (`theory`, `person`) belong to entity and concept pages and are dropped here in every vocabulary mode. The seed is the note's form values plus its domain tags; the model may add domain values that describe the summary. The 500-character window shown to the model starts at the body — shown raw, the note's frontmatter was where 4 of the 5 copied tags came from. This reverses the #568 stage-4 decision to remove the mirror, with a different weight: a source page without its note's axis falls out of the tag pane.

One consequence to know: since the harvest now reads `sources/`, a source page written **before** this change carries model-written `Group/Value` tags that were never gated, and those enter the offer like a hand-set tag would. On our vault every such tag also stands on a note, so nothing new was minted; a vault with older source pages should run Lint once after upgrading and look at what the source pages carry.

The rule that holds all of this together, and that the tests state: **what the prompt offers is exactly what the gate lets through, and the model never adds a term to the vocabulary — only a user does.**

Not in this PR: the frontmatter list shape (`tagStyle: 'inline'` in the gate vs block everywhere else — its reason died on 2026-07-07 when the retag runner learned both forms) — separate change.

Behaviour change on every page's `tags:`, so I would expect MINOR rather than the patch line. Measurement scripts on our side if you want them.

DocTpoint:
## Summary

Fixes #672.

Two lists both called themselves "the active vocabulary": the settings list fed the prompt, the #527 type repair, the lint scanner and the retag runner; the vault harvest (#568/#569) fed the write gate and the merge writers. A type the prompt allowed was deleted at the gate whenever no note carried it — 69 of 903 pages (7.6 %) born with an empty `tags:` on one rebuild, invisible to a scanner that skipped empty lists. Details and numbers in the issue.

## What changes

- **`core/vocabulary.ts`** — `activeVocabulary(app, settings, kind)`: settings list ∪ watched-folder tags ∪ nested tags of `entities/`, `concepts/`, `sources/` pages; fold-dedup, settings spelling wins, sorted. `activeVocabularyLists` for the system prompt, `domainVocabulary` (the `Group/Value` view) for the domains axis.
- **Every reader on that list**: `buildSystemPrompt` takes the lists (every caller passes them, link-orphan included), `repairTypesAgainstVocabulary` / `firstActiveTag`, `enforceFrontmatterConstraints` via `domainVocabulary` on create / fill-empty / merge-duplicates / summary, `incomingTypeTag` on merge and related, the dissent-stub identity, `scanTagViolations` (optional lists, settings fallback), `runRetagViolations`.
- **`collectWikiVocabulary`** walks the three page folders only.
- **Empty tag list = lint violation** (`scanTagViolations`), so the retag runner repairs the affected pages.
- **Source pages** pass `enforceFrontmatterConstraints('source')`: form values stay legal next to the `Group/Value` view of the vocabulary (`domainVocabulary`); the flat identity types are dropped in every vocabulary mode (`theory` in the default list and in a custom list that carries it). Seed = the note's form values + its domain tags (`selectDomains`); prompt requirement 7 names the rule; the 500-char window starts at the body (`extractBody`). Reverses #568 stage 4 (mirror), see issue.
- **Extraction prompt (second commit):** the JSON example at the `type` slot read `"type": "theory|method|field|phenomenon|standard|term|other"` (entities: `person|organization|…`) while the system prompt asked for the vocabulary — the nearer instruction won: 259 of 444 concept pages of the rebuild carried no vocabulary type although one existed (Demenz without `Sorte/Erkrankung`). Both slots (extraction + conversation ingest) now point at the Active Tag Vocabulary section. The wire schema was never the constraint (`type: z.string()`).
- **README / README_DE** feature bullet.

## Tests

4144 → 4165 on `080f884` (rebased 2026-09-15; two of the four commits had landed upstream via #680/#681 and dropped by patch-id). New: `core/vocabulary.test.ts` (union, spelling, kinds, built-in fallback, archive files ignored, domain view); source-page gate in `frontmatter.test.ts`; seed / body window / gate / `theory` in default mode / hand-set nested tag survives (#114) / all form values kept on `wiki-engine-summary-tags.test.ts`; vocabulary lists on `scanners.test.ts` and `runtime-injection.test.ts`. Two assertions inverted with the reason in the test: `collectWikiVocabulary` harvests `sources/`, empty tags are a violation. Six test mocks gained `vault.getMarkdownFiles` / `metadataCache.getFileCache` because the system prompt now harvests.

Gate 1: lint 0/0 · tsc 0 · build · 4057/4057 · css-lint 0. Gate 4 table in the commit body: the harvest is 4 metadata-cache walks, measured 2.49 ms per call on a synthetic 3,400-file vault, ~40 calls per ingested note.

## Release note

The first Lint after this update reports pages with no `tags:` at all as vocabulary violations; previously they were skipped. Measured 69 of 903 pages on one rebuild. The retag runner repairs them without a re-ingest.

## Not in this PR

- The frontmatter list shape (`tagStyle: 'inline'` in the gate, block elsewhere) — a second PR stacked on this one, opened once this lands.
- A per-run vocabulary snapshot through `EngineContext` — the measured cost does not justify it yet.


green-dalii:
@DocTpoint — `#671` and `#680` landed, and this now conflicts in `src/wiki/wiki-engine.ts`. Two hunks, both in that one file; here is the resolution I checked.

**1. The import (line ~30) — take the union.**

```ts
import { parseFrontmatter, upsertFrontmatterField, mergeFrontmatterArrayField, replaceFrontmatterArrayField, extractBody, enforceFrontmatterConstraints } from '../core/frontmatter';
```

`replaceFrontmatterArrayField` is main's (#671 added it), `enforceFrontmatterConstraints` is yours. I also checked the collision I was worried about: you and `#671` independently fixed the same `replaceFrontmatterArrayField` blank-line seam, and git auto-resolved it to the same normalised form — `const body = content.substring(endIdx + 4).replace(/^\n+/, ''); return \`${fmBlock}\n\n${body}\`;`. No duplication, nothing to reconcile there.

**2. The comment above the alias filter (line ~1810) — take yours.** The code below it is byte-identical on both sides, so this is a question of which comment is true, not which code wins.

Your version says the constraints pass has already applied the floor to the model's list. That is the accurate one **after** this PR, and I verified why: you added `enforceFrontmatterConstraints(...)` to `createSummaryPage` (line ~1774), and that function applies the floor itself — `filterRedundantAliases(options.pagePath, collectedAliases, undefined, resolveMinAliasLength(settings))`. Main's version of the comment ("the model's `aliases:` arrive verbatim inside `cleanedContent`") was correct before this PR, because the summary page did not call the constraints pass at all. Your routing changes that, so main's comment would be stale.

Nothing else conflicts. `git merge-tree` against `6b543bf` reports this file only.

---

**One separate thing before I merge this, not a conflict.**

The README feature bullet for the tag vocabulary was rewritten in **`README.md` and `docs/README_DE.md` only**. The other nine locales still carry the old sentence, and after this PR that sentence is wrong rather than merely stale — it says the vocabulary "is a schema-injection hint, not a write-level gate", while the whole point of this PR is that what the prompt offers is now exactly what the gate lets through.

I confirmed all eleven files carry the bullet, so it is one fact in eleven places:

```
README.md:148            docs/README_IT.md:145     docs/README_PT.md:146
docs/README_CN.md:153    docs/README_JA.md:145     docs/README_RU.md:147
docs/README_DE.md:148    docs/README_KO.md:144     docs/README_ZH-Hant.md:147
docs/README_ES.md:145    docs/README_FR.md:147
```

Nine other descriptions and a wrong claim is the drift class our own docs rule exists to prevent, so I would like the remaining nine aligned in this PR rather than left to a follow-up. If you would rather not write nine translations, the EN text is the canonical one — say so and I will take it as a separate small PR after yours lands.

DocTpoint:
Repushed as `b05b099`: merged `main` (`6b543bf`), both hunks in `wiki-engine.ts` resolved as you laid out — import union, and this branch's comment above the alias filter. Gate 1 green locally (4136 tests).

On the README bullet: EN is canonical. I would rather not write nine translations here — please take the remaining nine locales as your separate small PR after this lands.

green-dalii (changes requested):
The code checks out; one thing blocks, and it is in the docs rather than the implementation.

**Blocking — 9 of 11 locales now state the opposite of the shipped behaviour.** The EN bullet changed a factual claim:

- before: "The vocabulary is a schema-injection hint, **not a wire-level gate**"
- after: "what the model is offered is exactly what lands on disk; **a value outside it is dropped, never written**"

That is a reversal, and only `README.md` and `docs/README_DE.md` were touched. The other nine still carry the old assertion, in their own words — verified verbatim: `README_CN.md` "而非**写入时的强制闸门**", `README_JA.md` "**書き込み時の強制ゲートではありません**", `README_RU.md` "**а не шлюз принуждения при записи**", same shape in `_ZH-Hant` / `_KO` / `_FR` / `_ES` / `_PT` / `_IT`. After this merges, nine locales tell a user the vocabulary does not gate writes, on the one feature whose point is that it does.

I do not think the PR should carry nine translations it cannot verify — if you would rather not, say so and I will send the nine bullets as a separate PR to land alongside this one.

**What I verified as correct**, so nothing else is outstanding: `core/vocabulary.ts` is a real single entry (`activeVocabulary` / `activeVocabularyLists` / `domainVocabulary` / `vocabularyKindFor`) over the existing `collectActiveVocabulary`, and `scanTagViolations` now judges against the same lists the prompt offers — `vocabulary?.entities ?? getActiveEntityTags(settings)` — which is the property the new bullet claims. The source-page union (`form tags + isDomainTag`-filtered identity tags) is the right shape given `Group/Value`.

**One behaviour change worth a release note**, not a blocker: `scanners.ts:456` turned `else { continue; }` into `currentTags = []`, so a page with no tags is now a violation. Your comment cites the measurement (69 of 903), which is the right place for it — but it means the first Lint after upgrade reports a batch of pages that were previously silent, and the release notes should say so before users read it as a regression.

DocTpoint:
Rebased onto 080f884 (6e2e1a7). The conflict came from #684: the `buildSystemPrompt` call in `contradictions.ts` and `contradiction-truncation.test.ts` are gone upstream, so this PR no longer touches either. Gate 1 green, 4165 tests against 4144 on main.

On the nine locales: yes, please send them as the separate PR — this PR cannot verify them, and you can. Thank you for the offer.

The release-note line for the empty-`tags:` change is now in the PR body under "Release note", ready to lift into the CHANGELOG.

DocTpoint:
Rebased onto `dff36d38` as `056e23d3` (force-with-lease). Gate 1 green: 298 files / 4191 tests, against 297 / 4170 on that base.

Your changes-requested from 12.09 was on `b05b099`, two rebases ago. The points in it are in the branch; the review itself now hangs on a commit that no longer exists, so it needs a fresh look rather than a dismissal.

The rebase was one conflict, in both READMEs: `acabfcb` swept the org rename through them, and this branch still carried the old `green-dalii/obsidian-llm-wiki` address in the line it rewrites. Resolved to this branch's wording with the new address.

One commit is new, and it is the part worth a second look. #653 landed with two hand-built `fixDeadLink` contexts carrying `app: {}`, and this branch makes `buildSystemPrompt` take `activeVocabularyLists(ctx.app, ctx.settings)` — the vocabulary reader walks the vault. Apart, both are green. Together, 13 tests throw on `app.vault` being undefined. The fix is the same `app` literal the sibling `fixDeadLink` tests already carry, verbatim.

That is the second time this month that a pair is red where neither half is, after #716. Neither review can see it, and neither branch's CI runs the other's tests — the merge commit is the first place it exists.

DocTpoint:
Still at `056e23d3`. Gate 1 is green, and the re-review request has been open since 15.09.

Two things are outstanding, and the second is the one that blocks.

**The review itself.** Your changes-requested was submitted against `b05b099`, two rebases ago. The branch moved twice since — onto `080f884` after `#684`, then onto `dff36d38` — and it carries one new commit: `#653` added two `fixDeadLink` contexts with `app: {}` that the vocabulary runs against, and 13 tests went red only once the two branches were combined. So the review now hangs on a commit that no longer exists.

**The nine locales.** That was the blocking point, and you offered to send those bullets as a separate PR to land alongside this one. We agreed on 12.09 and again on 15.09. Measured on `main` today: all nine still carry the old claim — and so does the EN bullet, because this PR has not merged. Nothing contradicts anything yet. The contradiction appears the moment this PR lands on its own, which is why the two need to land together.

Nothing is pending on our side.

green-dalii (approved):
Approving, on the current head `056e23d3` — my earlier changes-requested was submitted against `b05b0996`, two rebases ago, so this replaces it rather than adding to it.

**The blocking point is delivered, in #762.** The nine locales now carry the new claim in their own terminology. Your point about why they had to land together is the one that decided the shape: nothing contradicts anything today, because the EN claim is not on `main` either — the contradiction is created by *this* PR landing alone. So #762 is a companion, not a follow-up.

**Re-verified on the current head, rather than carrying the old verification forward:**

- `scanners.ts:423-424` — `vocabulary?.entities ?? getActiveEntityTags(settings)`. The gate and the prompt read the same lists, which is the property the bullet claims.
- `core/vocabulary.ts` as the single entry (`activeVocabulary`, `activeVocabularyLists`, `domainVocabulary`, `vocabularyKindFor`).
- `c97353eb`, which landed after my review, is an improvement I want on the record: the JSON schema's `type` slot no longer offers `"person|organization|project|product|event|place|other"` but asks for "exactly one of the Entity types listed in the Active Tag Vocabulary section — copy its spelling". That removes the last built-in taxonomy from the prompt, so it strengthens the claim rather than restating it.

**One thing for the release notes, unchanged from my earlier review and not a blocker.** `scanners.ts:456` sets `currentTags = []` where the old code did `else { continue; }`, so a page with no tags is now a violation — 69 of 903 in the measurement. The first Lint after upgrade will therefore report a batch of pages that were previously silent. Your comment cites the number, which is the right place for it; the release notes should say so too, or it reads as a regression.

The measurement in your own review — the two branches combined, and the 13 tests that only went red once they were — is what kept this from being a coin flip. Thank you for the patience on the locale half; it was mine to send and it took longer than it should have.

green-dalii:
#762 is open and is the companion this needed: the nine locales, one line each, in their own terminology rather than translated from the EN text. Verified mechanically — zero occurrences of the old assertion across the nine, one new bullet in each, `#328` link intact in all nine. Gate 1 green there too.

So both halves now exist and neither should merge alone. #762's body says so, and I have approved this PR on `056e23d3` — the earlier changes-requested was against `b05b0996` and is superseded.

For whoever merges: two PRs, and the order between them does not matter for correctness, only that the second is not abandoned. #762 is documentation only and touches no file this one does, so there is no conflict either way.

PR #675, opened 2026-09-10
fix(aliases): fold case when checking the page's own aliases too

### LINKED ISSUE #674: `appendAliases` compares against the page's own aliases byte-exact — four places ask "one alias or two", only one folds
Measured over 802 entity and concept pages of one vault: three carry a case pair — `"NAC"`/`"NAc"`, `"obstruktive Schlafapnoe"`/`"Obstruktive Schlafapnoe"`, `"sustained attention"`/`"Sustained Attention"`. All three were written by ingest (140 log entries in the window, no lint run).

## Mechanism

`filterRedundantAliases` NFC-normalises and case-folds for all three of its rules, and the cross-page uniqueness gate is pinned case-insensitive by its own test ("matches claims case-insensitively"). The one comparison left is `appendAliases`' dedup against the aliases already on the target page: a plain `Array.includes`, byte-exact. So a page carrying `"NAC"` accepts an incoming `"NAc"` — the floor passes it, the filename does not match it, the batch has not seen it, no other page claims it, and `includes` does not recognise it. Both spellings end up on the page, one under the other.

Both readers of the alias pool compare case-folded (`fix-dead-link.ts:264`/`:266`, `scanners.ts:146`), so the second entry resolves nothing the first does not. It widens the alias line in the fix-dead-link prompt and makes the page look as if it claimed two names.

Two more comparison sites do not fold either — `mergeFrontmatterArrayField` and the dedup in `serializeFrontmatter` (whose comment says "parity across all writers"). Neither is reached with a case variant today; they are noted, not changed.

## Fix (PR attached)

`aliasKey` is exported from `core/slug.ts` instead of repeating the fold, and `appendAliases` compares on it, so "are these one alias" has one answer in one place. The spelling already on disk wins — rewriting the existing entry to the incoming casing would be churn without a reader that can tell the difference.

4 tests, 3 red on the parent commit; the fourth pins the direction that already worked (a case variant inside one incoming batch is dropped by `filterRedundantAliases`). 4028 → 4032.

DocTpoint:
## Summary

Fixes #674.

`appendAliases` folds case everywhere except the one place it matters most.
`filterRedundantAliases` normalises NFC and case-folds for all three of its
rules, and the cross-page uniqueness gate is pinned as case-insensitive by its
own test ("matches claims case-insensitively"). The dedup against the aliases
already on the target page is a plain `Array.includes`, which is byte-exact.

So a page carrying `"NAC"` accepts an incoming `"NAc"`: the floor passes it,
the filename does not match it, the batch has not seen it, no other page
claims it, and `includes` does not recognise it. Both spellings end up on the
page, one under the other.

Measured over 802 entity and concept pages in one vault: three carry such a
pair — `"NAC"`/`"NAc"`, `"obstruktive Schlafapnoe"`/`"Obstruktive
Schlafapnoe"`, `"sustained attention"`/`"Sustained Attention"`. Both readers
of the alias pool compare case-folded (fix-dead-link.ts:264 and :266,
scanners.ts:146), so the second entry of each pair resolves nothing the first
does not; it only widens the alias line in the fix-dead-link prompt and makes
the page look as if it claims two names.

Exports `aliasKey` from core/slug.ts rather than repeating the fold, so the
question "are these one alias" has one answer in one place. The spelling
already on disk wins; rewriting the existing entry to the incoming casing
would be churn without a reader that can tell the difference.

Tests: 4 new (3 red on the parent commit), 4028 -> 4032. The fourth pins the
direction that already worked — a case variant inside a single incoming batch
is still dropped by `filterRedundantAliases`.

Gate 1: lint 0/0 · tsc 0 · build · 4032/4032 · css-lint 0.

### DIFF
--- a/src/__tests__/wiki/page-factory/aliases.test.ts
+++ b/src/__tests__/wiki/page-factory/aliases.test.ts
@@ -181,6 +181,48 @@ describe('appendAliases — cross-page uniqueness gate', () => {
   });
 });
 
+// The page's own aliases are the one list compared byte-exactly.
+//
+// `filterRedundantAliases` folds case for all three of its rules, and the
+// cross-page gate above is pinned as case-insensitive. The dedup against the
+// aliases already on THIS page is a plain `Array.includes`, so a candidate
+// that differs from an existing entry only in case passes both stages and
+// lands on the page next to it.
+//
+// Measured over 802 entity/concept pages: 3 carry such a pair — "NAC"/"NAc",
+// "obstruktive Schlafapnoe"/"Obstruktive Schlafapnoe", "sustained
+// attention"/"Sustained Attention".
+describe('appendAliases — dedup against the page\'s own aliases', () => {
+  it('does not add an alias that differs from an existing one only in case', async () => {
+    const ctx = makeContext({ [PAGE]: makePage(['NAC']) });
+    await appendAliases(ctx, PAGE, ['NAc']);
+    expect(ctx.written.get(PAGE)!).toBe(makePage(['NAC']));
+  });
+
+  it('matches the own-page list with the same fold the cross-page gate uses', async () => {
+    const ctx = makeContext({ [PAGE]: makePage(['sustained attention']) });
+    await appendAliases(ctx, PAGE, ['Sustained Attention', 'Wachheit']);
+    const written = ctx.written.get(PAGE)!;
+    expect(written).toContain('  - "Wachheit"');
+    expect(written).not.toContain('  - "Sustained Attention"');
+  });
+
+  it('keeps the spelling already on the page, not the incoming one', async () => {
+    const ctx = makeContext({ [PAGE]: makePage(['Obstruktive Schlafapnoe']) });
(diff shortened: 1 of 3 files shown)


green-dalii (approved):
Verified at the line. `aliases.ts:58` runs the incoming names through `filterRedundantAliases` — NFC-normalising and case-folding for all three of its rules — and then `:63` does `candidates.filter(a => !existingAliases.includes(a))`, a byte-exact `Array.includes` against the aliases already on the page. So the floor passes `NAc`, the filename does not match it, the batch has not seen it, no other page claims it, and the one comparison that could have caught it does not fold.

The measured consequence is the part that justifies the fix rather than the tidy-up framing: both readers of the alias pool (`fix-dead-link.ts:264/266`, `scanners.ts:146`) compare case-folded, so the second entry resolves nothing the first did not. It widens the alias line in the fix-dead-link prompt and makes the page appear to claim two names — cost with no capability.

Folding in one place to match the three that already fold is the right scope. Merging.

PR #680, opened 2026-09-11
fix(sources, mentions): take the note path from the code, not the model's copy

### LINKED ISSUE #679: The note path is copied through the model, and the model's copy wins: `source_file` on the source page, `source_path` on every mention
**What happens**

On a vault rebuilt from scratch with a local model, three note paths came back wrong. All three notes exist under their real name:

| Real note | Written as | Where |
|---|---|---|
| `Notizen/Zytokine.md` | `Notizen/Zytokines.md` (typo) | `source_file` of `sources/Zytokine`, 2 mention lines |
| `Notizen/TNF-α.md` | `Notizen/TNF-` + U+0011 (α became a control character) | 5 mention lines |
| `Notizen/Omega-3-Fettsäuren.md` | `Notizen/Omega-3-Fatty-Acids` (translated) | 1 mention line |

Measured over the whole vault: 1 of 103 source pages has a `source_file` that names a note which does not exist (the real origin identified by `contentHash`, which the code stamps), and 8 of 2,334 mention lines link a note which does not exist. That is rare, but every case is a dead link, and for `source_file` it is also a wrong ownership answer (below).

**Why**

1. `createSummaryPage` (`wiki-engine.ts`) renders `file.path` into the template (`generation.ts`: `source_file: "[[{{source_file}}]]"`), and the model writes the whole page, frontmatter included. Nothing sets the field again afterwards. A few lines later, `contentHash` is upserted by code, with the comment "the LLM can't be trusted to emit it". `source_file` is the same kind of field.
2. `analyzeSource` puts `file.path` into the extraction prompt (`ingestion.ts`: "Use this EXACT path in every mentions_with_provenance[i].source_path field"). `computeReingestMentions` (`mentions-parser.ts:58`) then prefers the model's copy: `m.source_path || defaultSourcePath`. The path the code knows is only the fallback.

**Why `source_file` matters beyond a dead link**

`originNoteRefs` reads `source_file` as the canonical owner. So `pageBelongsToNote` returns false for `Notizen/Zytokine.md`, because its page names `Zytokines.md`. As a result, `isAlreadyIngested` answers "not ingested" for a note that is ingested, and the file picker (#598) shows the same answer. `scanSourceDrift` resolves the note through the same field, finds no such file, and skips the page. So drift on this note is never reported.

**Proposal**

Both values are known to the code exactly. Take them away from the model:

1. In `createSummaryPage`, upsert `source_file` from `file.path`, next to `contentHash`. The body line "Original file: [[…]]" is model output too (83 of 103 pages carry it, none wrong today). It could be rendered by code in the same step.
2. For the mentions of the current ingest, use the code's path (`sourceFile.path`). Mentions parsed from the existing page keep their own path. The "Use this EXACT path" instruction and `source_path` in the prompt example can then go, which also saves the model a copy per quote. That part touches the extraction schema (#658), so I would keep it separate if you prefer.

A wrong `source_file` in an existing vault can be found and repaired from `contentHash`, which identifies the note exactly. That is how the table above was measured.

Out of scope here: the in-band markers in prose (`^[Quelle: [[Notizen/Zytokines]]]` on the same two pages) are model-written as well.

A PR follows.

DocTpoint:
Closes #679.

The note path reached two load-bearing fields by being copied through the model, and in both the model's copy won. The code knows the path exactly, so it now sets both.

**1. `source_file` on the sources/ page** (`2ba57cc`)

`createSummaryPage` upserts `source_file: "[[<file.path>]]"` next to `contentHash`, which is stamped for the same reason. `source_file` is the canonical owner for `originNoteRefs`, and so for `pageBelongsToNote` → `isAlreadyIngested`, the file picker and `scanSourceDrift`. The case from the issue: `Notizen/Zytokines.md` for `Notizen/Zytokine.md` made the note read as "not ingested", and its drift was never scanned.

**2. `source_path` on every mention** (`54ae286`)

`normalizeBatchResponse(raw, sourcePath)` sets the note path on each mention, both structured and synthesized from `mentions_in_source`. That is the point where model output first enters the code, so the create, merge and reviewed-merge paths all receive it without changes of their own. `computeReingestMentions` keeps `m.source_path || defaultSourcePath` for mentions parsed from an existing page. Side effect: a quote that came in two batches with two different copies of the path now deduplicates on `(quote, source_path)`.

**Tests.** Four new tests, red without the change: a miscopied and a missing `source_file` on the written page, and a model-copied and a synthesized `source_path` after normalization. The four existing `normalizeBatchResponse` calls pass a path now, with no other change.

**Checks.** Gate 1 is green locally (lint, typecheck, build, 4,032 tests, css-lint). `/simplify` folded the structured branch of `fillMentionsWithProvenance` into one return and corrected two comments that the change made wrong (`frontmatter.ts`, `mentions-parser.ts`). `/code-review` found nothing.

**Not in this PR**

- The prompts still ask the model for both values: `source_file:` in the summary template, and "Use this EXACT path" plus `source_path` in the extraction example. Both are now overwritten, so the extraction copy costs roughly 10–15 output tokens per quote for nothing. Removing them changes what the model sees, so I would do it separately. It also touches `generation.ts` and `ingestion.ts`, which #673 changes.
- The body line "Original file: [[…]]" is still model output (83 of 103 pages carry it, none wrong in the measured vault).
- Existing pages with a wrong `source_file` or wrong mention paths are not repaired. The origin can be recovered exactly from `contentHash`.
- `conversation-ingest.ts` writes its own sources/ page with `source_file: Conversation: <title>`. That is a label, not a note path, and no reader resolves it.

**Overlap:** #671 and #673 (both mine) also touch `createSummaryPage`. Whichever merges second needs a small rebase.

### DIFF
--- a/src/__tests__/core/mentions.test.ts
+++ b/src/__tests__/core/mentions.test.ts
@@ -343,7 +343,7 @@ describe('normalizeBatchResponse — fillMentionsWithProvenance (#244)', () => {
       }],
       concepts: [],
     };
-    const { data } = normalizeBatchResponse(raw);
+    const { data } = normalizeBatchResponse(raw, 'note.md');
     expect(data.entities[0].mentions_with_provenance).toHaveLength(1);
     expect(data.entities[0].mentions_with_provenance![0].quote).toBe('structured quote');
     // Manual-test fix: when both fields are present, legacy is cleared to avoid
@@ -361,7 +361,7 @@ describe('normalizeBatchResponse — fillMentionsWithProvenance (#244)', () => {
       }],
       concepts: [],
     };
-    const { data } = normalizeBatchResponse(raw);
+    const { data } = normalizeBatchResponse(raw, 'note.md');
     expect(data.entities[0].mentions_with_provenance).toHaveLength(2);
     expect(data.entities[0].mentions_with_provenance![0].quote).toBe('quote A');
     expect(data.entities[0].mentions_with_provenance![1].quote).toBe('quote B');
@@ -374,7 +374,7 @@ describe('normalizeBatchResponse — fillMentionsWithProvenance (#244)', () => {
       entities: [{ name: 'Entity X', type: 'person' as const, summary: '', mentions_in_source: [] }],
       concepts: [],
     };
-    const { data } = normalizeBatchResponse(raw);
+    const { data } = normalizeBatchResponse(raw, 'note.md');
     expect(data.entities[0].mentions_with_provenance).toBeUndefined();
   });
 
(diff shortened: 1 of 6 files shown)


DocTpoint:
Correction to the "Not in this PR" note: the body line "Original file: [[…]]" is not always right. On `sources/Zytokine` it carried the same wrong path as `source_file` (my check had missed it because the model translates the label), so it fails exactly where the frontmatter did. Rendering it in code is the follow-up.

green-dalii (approved):
Verified. The premise is exactly right: the code knows the note path, and it was letting the model's copy win in two load-bearing fields.

I confirmed the mechanism for the `source_file` half — `createSummaryPage` renders `file.path` into the template and the model writes the whole page frontmatter, while `contentHash` is stamped by code for the same reason a few lines later. That asymmetry is the whole bug: one field is authored, the other asserted, and the authored one is the one that decides `originNoteRefs` → `pageBelongsToNote` → `isAlreadyIngested`, the file picker and `scanSourceDrift`. A typo in it does not just make a dead link, it makes a note read as never ingested and hides its drift, which is why this ranked as more than cosmetic.

Setting both fields from the code collapses the two halves into one rule: the code owns provenance, the model owns content. That is the right division and it matches what the `#681` follow-up then does for the rest of the head.

Merging, and I will take `#681` right after it since it touches the same analyzer and is written as this PR's follow-up.

PR #681, opened 2026-09-11
fix(sources): the source page head comes from the code — title, note path, date

DocTpoint:
Follow-up to #679 / #680, as announced there: the rest of what the model copied onto a sources/ page.

**1. A note's source title is its own** (`3ec62eb`)

For anything but a PDF, `source_title` is now the note's H1 on the first body line (the #592 rule `getExistingWikiPages` uses) or its file name. A PDF keeps the title the model read off the document. The PDF check ignores the extension's case, like the engine's own.

Measured over 105 source pages of a rebuilt vault (84 of the notes have no H1):

| Model title | Pages |
|---|---|
| = file name | 46 |
| path or file name copied back (`Notizen/Carrageenan.md`, `Vitamin D.md`) | 21 |
| formulated by the model (`Kardiales Remodeling` for `Remodeling.md`) | 21 |
| = the note's H1 | 17 |

One more came back with a control character in place of `α`. All 21 copies are on notes without an H1, and the prompt shows the model the note path. The cost of the rule: the 21 formulated titles become the file name or H1. They are not reliably better (`Das Gedächtnis` for `Gedächtnis.md`), and the note's author chose its name. The lemma guarantee is unaffected, because it already uses the file name.

**2. The page head is rendered in code** (`358fb4d`)

`stampSourcePageHead` replaces the body-opening H1 and the Source section with `# <title> - <summary>` and `- <original file>: [[<path>]]` / `- <ingested>: <date>`. It takes the same route as the Mentions section (#244), and every other section stays as written. The line labels have their own table, `SOURCE_PAGE_HEAD_LABELS`, for all 11 languages, and `ingested` reuses the file picker's word. They are deliberately not in `SECTION_LABELS`: its values are the known section headings for `canonicalizeSectionHeaders`, `stripUnknownSections` and the merge-triage prompt, so a `## Zusammenfassung` on a concept page would stop being stripped. A test pins that separation.

Probe over the 105 source pages: exactly one Source section per page after stamping, no other section lost.

**Tests.** New tests for the head function, the title rule (no H1, H1, PDF in either case) and `createSummaryPage` in German, all red without the change. One existing test changed its expectation: `source-analyzer.test.ts` pinned the model title over the note's H1 (`My Document` over `# Doc`), which is exactly what this reverses.

**Checks.** Gate 1 is green locally (4,038 tests). `/simplify` swapped a hand-written section regex for `stripMentionsSection`, which is label-generic. `/code-review` found two things, both fixed before the push: the labels first sat in `SECTION_LABELS` (see above), and the PDF check was case-sensitive.

**Not in this PR**

- The prompts still ask the model for all of this: H1 and Source section in the summary template (roughly 30–45 output tokens per ingest, now discarded), and a `source_title` plus the note path in the extraction prompt. Removing them changes what the model sees and touches the files #673 changes, so it gets its own PR after #673, together with the `source_path` cleanup from #680.
- `conversation-ingest.ts` writes its own sources/ page. It has no note path, and there the model title is the point.
- The first-line H1 regex now exists three times (`get-existing-pages.ts`, a local closure in `merge-duplicates.ts`, here). A shared helper would touch #653's file, so later.
- Existing pages are not rewritten. A re-ingest renders the new head.

**Overlap:** #680, #671 and #673 also touch `createSummaryPage`.

### DIFF
--- a/src/__tests__/core/source-page-head.test.ts
+++ b/src/__tests__/core/source-page-head.test.ts
@@ -1,0 +1,75 @@
+// The head of a sources/ page — H1 and Source section — carries what the code
+// knows (title, note path, ingest date). The model copied all three from the
+// template, and a copy could come back wrong.
+
+import { describe, it, expect } from 'vitest';
+import { stampSourcePageHead } from '../../core/source-page-head';
+import { SECTION_LABELS, SOURCE_PAGE_HEAD_LABELS, getSourcePageHeadLabels } from '../../wiki/system-prompts';
+import type { LLMWikiSettings } from '../../types';
+
+const labels = (wikiLanguage: string) => getSourcePageHeadLabels({ wikiLanguage } as LLMWikiSettings);
+
+const HEAD = { title: 'Zytokine', sourcePath: 'Notizen/Zytokine.md', date: '2026-09-11' };
+
+const MODEL_PAGE = `---
+type: source
+source_file: "[[Notizen/Zytokine.md]]"
+---
+
+# Zytokines - Summary
+
+## Quelle
+
+- Originaldatei: [[Notizen/Zytokines.md]]
+- Datum der Erfassung: 2026-09-10
+
+## Kerninhalt
+
+Signalproteine des Immunsystems.
+`;
+
+describe('stampSourcePageHead', () => {
+  it('replaces the model-written H1 and Source section with the code-known values', () => {
+    const out = stampSourcePageHead(MODEL_PAGE, HEAD, labels('de'));
+    expect(out).toContain('# Zytokine - Zusammenfassung\n');
+    expect(out).toContain('## Quelle\n\n- Originaldatei: [[Notizen/Zytokine.md]]\n- Importiert: 2026-09-11');
+    expect(out).not.toContain('Zytokines');
+    expect(out).not.toContain('Datum der Erfassung');
+    expect(out.match(/^## Quelle/gm)).toHaveLength(1);
+  });
+
+  it('keeps the frontmatter and every other section as written', () => {
+    const out = stampSourcePageHead(MODEL_PAGE, HEAD, labels('de'));
+    expect(out.startsWith('---\ntype: source\nsource_file: "[[Notizen/Zytokine.md]]"\n---\n\n# Zytokine')).toBe(true);
+    expect(out).toContain('## Kerninhalt\n\nSignalproteine des Immunsystems.');
+  });
+
+  it('adds the head when the model wrote neither H1 nor Source section', () => {
+    const out = stampSourcePageHead('---\ntype: source\n---\n\n## Core Content\n\nText.\n', HEAD, labels('en'));
+    expect(out).toBe(
+      '---\ntype: source\n---\n\n# Zytokine - Summary\n\n## Source\n\n- Original file: [[Notizen/Zytokine.md]]\n- Ingested: 2026-09-11\n\n## Core Content\n\nText.\n',
+    );
+  });
+
+  it('removes a Source section that is the last section of the page', () => {
+    const page = '---\ntype: source\n---\n\n# X - Summary\n\n## Core Content\n\nText.\n\n## Source\n\n- Original file: [[wrong.md]]\n';
+    const out = stampSourcePageHead(page, HEAD, labels('en'));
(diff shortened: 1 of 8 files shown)


green-dalii (dismissed):
Verified, and I am taking it immediately after #680 so the two land as one change rather than leaving the head half-migrated.

The measurement in the description is what makes this worth doing rather than optional: over 105 source pages, 21 were the model copying the file name or a path back (`Notizen/Carrageenan.md`, `Vitamin D.md`) and 21 were the model formulating a title (`Kardiales Remodeling` for `Remodeling.md`) — and all 21 of the copies sit on notes that have no H1. That is a predictable degradation, not noise: with nothing to copy the model invents, and it invents in the same direction every time. Deriving the title from the note's own first body line (the `#592` rule `getExistingWikiPages` already uses) with the file name as fallback removes the whole class, and keeping the model's title for PDFs is the correct exception since there is no note to read it from.

The control character replacing `α` in one page is the same failure mode as #679's `TNF-α` case, which is a useful confirmation that the two issues share a cause rather than being two coincidences.

Merging.

green-dalii:
@DocTpoint — #680 landed (98b82af) and this now conflicts in `src/wiki/wiki-engine.ts`. I have the exact resolution so you do not have to rediscover it.

**The conflict, at `wiki-engine.ts:~1761`:**

```ts
<<<<<<< main (#680)
finalContent = upsertFrontmatterField(finalContent, 'source_file', `"[[${file.path}]]"`);
=======
finalContent = stampSourcePageHead(
  finalContent,
  { title: analysis.source_title, sourcePath: file.path, date: ingestDate },
  getSourcePageHeadLabels(this.settings),
);
>>>>>>> yours
```

**Both lines are needed — keep both, in this order.** I checked `core/source-page-head.ts` before writing this: `stampSourcePageHead` reads `const frontmatter = content.slice(0, fmEnd + 4)` and preserves it verbatim, then rewrites only the H1 and the `## <labels.source>` section. It never touches the `source_file` frontmatter field. So the two writes are aimed at different places:

- `upsertFrontmatterField` → the **frontmatter field** `source_file`, which is what `originNoteRefs` / `pageBelongsToNote` / `isAlreadyIngested` / the drift scan read (#679's half).
- `stampSourcePageHead` → the **body's** Source section `- Original file: [[<path>]]` line (this PR's half).

Dropping either one regresses the other's fix. Order matters as well: upsert first, then stamp — `stampSourcePageHead` re-slices the frontmatter from the content it is given, so it carries the upserted field through, and doing it the other way round would discard it.

I am holding the merge until this is repushed. Nothing else in the PR conflicts — `git merge-tree` against `6b543bf` reports this one file only.

DocTpoint:
Repushed as `bf95f35`: merged `main` (`6b543bf`) into the branch, resolved exactly as you laid out — `upsertFrontmatterField('source_file', …)` first, `stampSourcePageHead(...)` second, both kept. Gate 1 green locally (4125 tests). Your approval was on `358fb4d`; the merge commit is the only change since.

green-dalii (approved):
Approved. The head now comes from the code that knows it, and the label guard is the part that will hold up.

Verified at `bf95f35a`: `stripMentionsSection` is genuinely label-generic (`mentions-parser.ts:235` — the param is `sectionLabel`, passed straight to `findSection`), so the comment on `source-page-head.ts:33` is accurate rather than hopeful; `extractBody` trims (`frontmatter.ts:175`), so `^#` does reach a note H1 that sits after a blank line; and `file.extension?.toLowerCase()` covers `.PDF`/`.Pdf`, which the test pins. Ran the four test files: 31/31.

One verification worth recording because the ordering invites it: the stamp runs *after* the `contentHash` and `source_file` upserts (`wiki-engine.ts:1760-1764`), so it has to leave the frontmatter untouched — it does, via `slice(0, fmEnd + 4)`, and `source-page-head.test.ts` asserts the `source_file` line is still there. That sequencing is the kind of thing a later refactor of this function would break silently.

Non-blocking: `source-page-head.ts:31` only strips an H1 that *opens* the body, which your "leaves an H1 alone that does not open the body" test pins on purpose — but the comment reads as incidental. A half-line saying the later H1 is kept deliberately would stop a reader from filing the duplicate as the very bug this PR fixes.

PR #683, opened 2026-09-11
fix(json): repair one-token defects before spending a model call on them

### LINKED ISSUE #682: Extraction: a one-character JSON defect costs a full model repair, and one repair call in five runs to the token cap
When an extraction response does not parse, `parseJsonResult` tries `fixCommonJsonIssues` and then calls the repair callback. The repair asks the model to write the whole JSON again (`max_tokens: retryCap`, schema at the wire).

**Measurement.** I replayed 3,893 extraction responses from LM Studio server logs through `parseJsonResult`, with the repair callback stubbed to count calls. Setup: gemma-4-26b-a4b QAT, `extract` in `text_prompt` mode, 1–11 September, one German-language vault.

- 198 responses reached the repair callback.
- In the logs of 10–11 September there are 54 finished repair calls. 11 of them ended with `finish_reason: length` at 32,767 tokens, about 8 minutes each.
- When the repair fails, the items of that batch are dropped.

**What the defects are.** Almost all of them are one character in otherwise sound pretty-printed JSON. These are the 50 cases of 10–11 September:

| n | defect | example |
|---|---|---|
| 25 | a key lost its closing quote | `"name: "Psoriasis",` |
| 10 | a single backslash that JSON cannot escape | `$\alpha$`, `[[Vitamin K2\|K2]]` |
| 4 | an array closed with `}` | `"domains": [ "Thema/Metabolismus" }` |
| 3 | a key lost its opening quote | `summary: "…"`, `="mentions_with_provenance": [` |
| 2 | a stray `.` after the closing quote | `"… (Leaky Gut)".` |
| 2 | a string lost its closing quote at the end of the line | `"… (Honorare, Beratung),` |
| 1 | a value lost its opening quote | `"summary": Ein Medikament …",` |
| 1 | a missing comma between two lines | |
| 2 | truncation (`finish_reason: length`) | |

`fixCommonJsonIssues` already has a rule for the missing comma. It does not help here, because `escapeContentQuotes` runs first and escapes the quote that the comma rule needs.

**A second case, which does not fail.** `\b`, `\f`, `\n`, `\r` and `\t` are valid JSON escapes. So `$\beta$`, `$\rightarrow$` and `$\text{…}$` parse without error, as backspace + "eta", carriage return + "ightarrow" and tab + "ext{…}". 21 of the 3,893 responses parsed this way.

**Proposal.** A deterministic pass for these shapes. It is tried only where the parser would otherwise call the repair or give up, and the model repair stays for truncation and everything else. On the replay, 176 of the 198 parse without a model call. Every response that parses today comes out identical, except the 21 LaTeX strings, which now keep their backslash. A PR follows.

DocTpoint:
Closes #682.

**What changes**

- **`repairKnownDefects`** in `src/core/json.ts` has one rule for each defect class in the issue. Each rule matches only a shape that valid JSON cannot contain. After the rules, `extractBalancedJson` and `JSON.parse` run, and once more through `fixCommonJsonIssues` if needed.
- **The pass is used in three places only:** before each of the two `repairFn` calls, and before the final `malformed` return. Every step that recovers a response today runs first and is unchanged.
- **`fixCommonJsonIssues`** now runs its missing-comma rule before `escapeContentQuotes`. This reaches every caller of the function. On the replay below it changed no response that parsed before.
- **Step 1.1b, `escapeLatexInMath`:** inside a `$…$` span, whole LaTeX command names that start with b/f/n/r/t keep their backslash. This is the one change that reaches responses which parse today.

**Evidence.** I replayed the same 3,893 extraction responses through the old and the new `parseJsonResult`:

| | responses |
|---|---|
| identical object | 3,671 |
| only the LaTeX strings differ | 21 |
| any other difference | 0 |
| newly failing | 0 |
| repair cases now parsed without a model call | 176 of 198 |

The remaining 22 need the model: 9 truncations, repetition loops, and misplaced structure such as a key inside an array.

**Two rules were narrowed during review:**
- The missing-quote key rule would split a valid quote like `"siehe: [[Vitamin D]] …"` if it also fired before `[`.
- The German-quote rule must skip a quote that is already escaped (`„X\"`).

Both have tests.

**Tests.** The new file `src/__tests__/core/json-known-defects.test.ts` has 16 cases: one per defect class, plus the things the pass must leave alone. Those are a real escape outside `$…$`, a `"Label: text"` quote, an escaped quote, and a truncated response that still reaches the repair. 12 of the 16 fail on `main`.

**Checks.** Gate 1 is green locally: lint, typecheck, build, 4,044 tests, css-lint. `/code-review` found the two rules above. `/simplify` replaced a memoized helper with one computed value, and moved the comma fix into `fixCommonJsonIssues` itself instead of repeating the rule in the new pass. Regex lookbehind is already used in `candidate-gate.ts`, so this adds no new platform requirement for mobile.

**Not in this PR**
- Truncation keeps its halving and repair path.
- The repair call itself is unchanged. It is the call that ran to the token cap, and that is a separate question.

### DIFF
--- a/src/__tests__/core/json-known-defects.test.ts
+++ b/src/__tests__/core/json-known-defects.test.ts
@@ -1,0 +1,166 @@
+// One-token defects a small model leaves in otherwise sound pretty-printed
+// JSON, fixed before the parser spends a model call on them. Each case is the
+// shape measured in real extraction responses (gemma-4-26b on LM Studio); the
+// model repair regenerated the whole response for each of them, and one call
+// in five ran to the token cap.
+
+import { describe, it, expect, vi } from 'vitest';
+import { parseJsonResult } from '../../core/json';
+
+/** Parse with a repair callback that records whether it was reached. */
+async function parse(text: string) {
+  const repair = vi.fn(async () => { throw new Error('model repair reached'); });
+  const result = await parseJsonResult(text, repair);
+  return { result, repaired: repair.mock.calls.length > 0 };
+}
+
+const item = (lines: string) => `{
+  "entities": [
+    {
+      "name": "Bradykinin",
+      "summary": "Ein Peptid."
+    },
+    {
+${lines}
+    }
+  ],
+  "concepts": []
+}`;
+
+describe('parseJsonResult — known one-token defects, no model call', () => {
+  it('a key that lost its closing quote: "name: "X"', async () => {
+    const { result, repaired } = await parse(item(`      "name: "Ramipril",
+      "summary": "Ein ACE-Hemmer."`));
+    expect(repaired).toBe(false);
+    expect(result).toMatchObject({ ok: true, value: { entities: [{}, { name: 'Ramipril' }] } });
+  });
+
+  it('a key and a one-word value merged into one string: "coverage: named"', async () => {
+    const { result, repaired } = await parse(item(`      "name": "Ramipril",
+      "coverage: named",
+      "summary": "Ein ACE-Hemmer."`));
+    expect(repaired).toBe(false);
+    expect(result).toMatchObject({ ok: true, value: { entities: [{}, { coverage: 'named' }] } });
+  });
+
+  it('a key that lost its opening quote, bare or behind a stray = or |', async () => {
+    const { result, repaired } = await parse(item(`      "name": "Ramipril",
+      summary: "Ein ACE-Hemmer.",
+      ="aliases": [],
+|      "coverage": "named"`));
+    expect(repaired).toBe(false);
+    expect(result).toMatchObject({
+      ok: true,
+      value: { entities: [{}, { summary: 'Ein ACE-Hemmer.', aliases: [], coverage: 'named' }] },
+    });
+  });
+
+  it('a value that lost its opening quote', async () => {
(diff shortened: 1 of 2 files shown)


green-dalii (approved):
Verified. The repair chain in `core/json.ts` is as described — `parseJsonResult` (`:238`) → `fixCommonJsonIssues` (`:314`) → `repairFn` (`:321`) — and the replay numbers make the case on their own: 198 of 3,893 responses reached the repair callback, and 54 finished repairs in the 10–11 Sep window included 11 that hit `finish_reason: length` at 32,767 tokens at roughly eight minutes each, with the batch's items dropped when the repair failed. Paying a full model rewrite for a missing closing quote (25 of 50 defects) is the kind of cost that hides in aggregate.

Two things I want on the record rather than treated as blockers. First, each rule matching only a shape valid JSON cannot contain is the right constraint — it is what makes the pass safe to run before the existing steps instead of after them. Second, this PR reorders the missing-comma rule ahead of `escapeContentQuotes` inside `fixCommonJsonIssues`, which reaches every caller of that function, and "changed no response that parsed before" on the replay is the assertion that matters for it. If a later regression points here, that reorder is the first thing to bisect.

Merging.

Issue #678, opened 2026-09-11
Delete Empty Stubs never deletes a Fix Dead Links stub: #197 changed the stub sentence, `STUB_MARKER` kept the old one


DocTpoint:
**What happens**

A stub written by Fix Dead Links is never removed. Delete Empty Stubs skips it, and no other path deletes it. It stays until a later ingest happens to define the term. With the default `createStubsForUnresolvableLinks: true`, that applies to every stub Fix Dead Links creates.

**Why**

`isPageEmpty` (`src/wiki/lint/utils.ts:68`) treats a page as empty if it contains `STUB_MARKER`, or if fewer than `MIN_SUBSTANTIVE_CHARS` (50) characters remain after `EMPTY_CONTENT_STRIP`.

- `STUB_MARKER` is `'Auto-generated stub page'`. That was the stub sentence until d53caab (#197, 24 June), which replaced it with `> Stub created by Fix Dead Links — referenced by [[…]]. Will be filled by next ingest of an actual source that defines this entity.` `STUB_MARKER` was not updated, and nothing in `src/` writes it any more.
- The new sentence alone leaves 102–122 characters after the strip. I measured this with `isPageEmpty` and `buildStubContent` copied verbatim, for two title/referrer pairs. So `isPageEmpty` is false for every Fix Dead Links stub, and `deleteEmptyStubs` skips it.

**Why the obvious fix is wrong**

If `STUB_MARKER` is updated to the new sentence, `isPageEmpty` becomes true for these stubs everywhere it is used. The lint controller (`controller.ts:169`) would then list them as empty pages. Expand Empty Pages (`runEmptyPageFixes` → `fillEmptyPage`) has no stub check, so it would hand them to the LLM. That is the path #197 closed. Today the stale marker is what keeps them out of it, by accident.

**Proposal (one PR, two small parts)**

1. `deleteEmptyStubs` decides on the frontmatter flag, not on the sentence: a page is deletable when `isStubPage(fm)` is true (`stub: true`, the marker `merge-page.ts` already relies on) and its body, without the stub's own placeholder line, is below `MIN_SUBSTANTIVE_CHARS`. `reviewed: true` stays a guard. Filled stubs are not affected, because the merge strips `stub: true` when a source fills them. A candidate-gate stub (#607) that carries a summary or a quote block survives.
[...]

PR #684, opened 2026-09-11
fix(contradictions): remove the two paths that only the model could feed (#604, #666)

### LINKED ISSUE #604: contradiction resolution loop is dead code — nothing ever sets review_ok
`runContradictionPhase` auto-resolves only records with `status === 'review_ok'` (`contradiction-phase.ts:104`). No code path sets that status: `trackContradiction` writes `detected`, and the only writers of the field afterwards are the phase's own `resolved` / `pending_fix` transitions. There is no UI command that moves a record to `review_ok` either. So step 1 of the phase filters an always-empty set, `resolveContradiction` (the LLM repair path) is unreachable from production code, and the loop's design intent — a human marks a record reviewed, the machine executes the fix — has no first half.

Follow-on: the `resolved_contradictions` section label is maintained in eleven languages, but its only consumer is `{{section_resolved_contradictions}}` in `PROMPTS.resolveContradiction` (`fixes.ts:83`) — a template only the dead branch renders. The label has no live producer.

Flagging rather than fixing: whether the missing half gets built (a review command on the record) or the branch gets removed is your call — #602 already moves open-contradiction handling onto marker + record, which is where a review command would naturally attach.

### LINKED ISSUE #666: #484 left the contradiction requirement without its input
Our #484 removed the slug catalog from `analyzeSource` and retired requirement 7 with it. Requirement 6 stayed, and it asks for the same kind of vault knowledge — so since 1.27.1 the extraction lane is asked a question the prompt no longer contains the answer to.

**Current state in `main`** (`src/wiki/prompts/ingestion.ts`):

- line 30 — `6. Identify contradictions or conflicts with the existing Wiki (only output contradictions in the first round)`
- lines 68–69 — `"source_page": "Conflicting existing Wiki page [[page-name]]"`, `"contradicted_by": "What that page claims"`

The prompt is instructions plus the note. No page list, no page content, no titles. The model is asked to name an existing wiki page and to state what that page claims, with nothing about any page in front of it. It can comply only by inventing.

#484's own reasoning for dropping requirement 7 was that `related_pages` "was paid for and discarded". Requirement 6 is the stronger case: it is not overwritten downstream, it is unanswerable from the inputs.

**What invention looked like when it did fire.** One run on our vault emitted `source_page: Notizen/Diet-Heart-Hypothese.md` — a source-note path, not a wiki page, and the one path the prompt does inject (`{{source_path}}`, "Use this EXACT path"). Before #601 that value went straight into `noteContradiction`'s path arithmetic and the block was written into the user's own note. The write path is fixed; the field that produced the value is not.

To be clear about scope: the merge triage lane is unaffected and works — it sees the existing page, quotes the sentence it conflicts with, and gate 1 checks that sentence against the page body. This is only about the extraction lane, which has no such input.

**Suggested resolution:** drop requirement 6 and the `contradictions[]` block from the extraction schema, and let the merge triage own detection — it is the lane that actually sees both sides, and `MEMORY.md` already names routing `strategy: contradictory` to the `ContradictionManager` as the intended shape.

**One consequence worth deciding rather than inheriting:** the merge lane only runs when a page is being updated. A source contradicting a page that nothing in that run touches would then go undetected. Extraction cannot detect it today either, so this is not a regression — but it does fix the feature's reach at "pages this run writes", and that is a product decision.

Happy to send the PR if you want it this way.

DocTpoint:
## Summary

Closes #604, closes #666. One PR for both, as suggested in #666: the second is what kept the first one's input alive.

**Requirement 6 (#666).** The extraction lane asked the model to name the existing wiki page a claim conflicts with, and what that page says. Since #484 the prompt carries no page list, no titles, no page content; the only vault-shaped input is the source note's own path, and that is what came back as `source_page` in two of two observed cases. The requirement, the `contradictions[]` block in prompt and wire schema, and the `SourceAnalysis` field are gone. The merge triage owns detection: it is the lane that sees both sides, and it already writes marker and record through `merge-page.ts`. The ingest log and the report count that lane alone; the log takes the run's contradictions as their own argument now that the analysis no longer carries any.

**`review_ok` (#604).** Step 1 of the lint phase auto-resolved records in that status through a model rewrite of the affected page. Nothing ever produced the status, so the step filtered an always-empty set, and everything behind it had no live caller: `resolveContradiction` in manager, engine and lint types, the `fixes.ts` prompt, the `resolved_contradictions` section label in eleven languages, the `(N auto-fixed)` report text, `TOKENS_CONTRADICTION`. The record file and its status field stay, as agreed; a review command, when it exists, attaches there.

**Taken along rather than left behind:**

- `noteContradiction` / `trackContradiction` and the `source_page` resolver from #601/#602 lose their only producer with requirement 6 and go with it. The triage lane never needed them: the page it records against is the page it was shown, a path the code holds. #601's guard existed to keep model-named targets from becoming write paths; with no model-named targets left there is nothing to guard. `buildContradictionRecord` stays as the triage's record writer.
- `updateContradictionStatus` and the `pending_fix` status had the removed step as their only producer.
- `clamp-page-sections.ts` had one consumer, the removed rewrite.

Not changed: the conversation ingest still reports zero contradictions. The triage's `onContradiction` fires there too and nothing collects it, which #606 already noted; the number was zero before this PR as well, since the conversation prompt never asked for the field.

Tests: 4028 → 3999 on the same base; the 29 pinned the removed paths. `/simplify` and `/code-review` ran over the diff before the push; the separate log argument and the `pending_fix` removal came out of them.

Gate 1: lint 0/0 · tsc 0 · build · 3999/3999 · css-lint 0.


green-dalii:
@DocTpoint — `#685` (the log-label table) landed, and this now conflicts in `src/wiki/engine-internals/log-writer.ts`. It is one hunk and the resolution is mechanical, because your change to that file is small.

**The conflict, at the import block:**

```ts
<<<<<<< main (#685)
import type { SourceAnalysis } from '../../types';
import { getLogLabels } from '../../core/log-labels';
=======
import type { ContradictionInfo, SourceAnalysis } from '../../types';
import { TEXTS } from '../../texts';
>>>>>>> yours
```

**Resolution:**

```ts
import type { ContradictionInfo, SourceAnalysis } from '../../types';
import { getLogLabels } from '../../core/log-labels';
```

i.e. keep main's `getLogLabels` line, add your `ContradictionInfo`, and drop `TEXTS`. I checked both sides rather than guessing:

- `getLogLabels` is #685's replacement for the per-locale `TEXTS.en.logLabels` lookup, so it has to stay — the `labels()` body on main now reads from `core/log-labels.ts`.
- `TEXTS` is **not** part of your change. Your diff to this file is three edits: the `ContradictionInfo` type import, the `contradictions: ContradictionInfo[]` parameter on `appendIngest`, and the two `analysis.contradictions` → `contradictions` reads. `TEXTS` was already in your base and #685 removed it — so it is stale-base residue, not something you need.

Nothing else conflicts. `git merge-tree` against `6b543bf` reports this file only.

---

**And I am withdrawing my earlier ask.** I said I wanted a zero-reference trace for the deleted symbols before this could land. I ran it myself instead, on your branch's clean tree, so you do not have to spend the time:

```
clampPageSections                refs=0
clamp-page-sections              refs=0
resolveContradiction             refs=0
resolved_contradictions          refs=0
SUPPRESS_CONTRADICTION_SECTION   refs=0
buildContradictionPrompt         refs=0
review_ok                        refs=3   ← all three are comments describing what was removed
```

I also confirmed the part that actually mattered, which is that detection survives the removal rather than leaving a hole: `src/wiki/page-factory/` is untouched by this PR, so `contradiction-gates.ts` and its two callers in `merge-triage.ts` / `merge-page.ts` are intact; `source-analyzer.ts` no longer mentions contradictions at all; and `wiki-engine.ts:2136` now passes `contradictions` explicitly into `appendIngest`, which matches your description of the log taking the run's contradictions as their own argument. The −1132 lines are the removed feature's own surface plus its tests, not a live path.

DocTpoint:
Repushed as `e764c5d`: merged `main` (`6b543bf`), import block resolved as you wrote it — `getLogLabels` stays, `ContradictionInfo` added, `TEXTS` dropped.

Two things behind the textual conflict that git did not flag, both in tests main added since the branch point:

- `log-labels.test.ts` (#685) and `wiki-engine-summary-source-file.test.ts` (#680) built a `SourceAnalysis` with the removed `contradictions` field and called `appendIngest` without the run's contradictions argument. Adapted to this PR's shape (field gone, `[]` passed).
- The #686 pinned wire body `SourceAnalysisLLMSchema.json` carried the `contradictions` block this PR removes from the schema. Snapshot updated; the diff is that block (31 lines) and nothing else.

Gate 1 green locally (4086 tests). Thanks for running the zero-reference trace.

DocTpoint:
This was closed at 07:43:22Z, one second after #689 merged, with no comment, review or commit attached. Was that intended? If you meant to move it with #604 to `v1.28.0 MINOR` (the reconciliation in the ROI board), a reopen with the milestone changed is all it needs — the branch is unchanged at `e764c5d`, CI green. If it was a slip in the merge chain, same fix.

green-dalii:
Reopening this — it was closed by accident, and not by you or by any decision about the work.

**What happened.** My ROADMAP PR (#689) described the milestone state in prose and included the sentence:

> #604 sits on `v1.28.0 MINOR` while its **fix #684** sits on the PATCH line

GitHub's auto-close parser reads `fix` immediately followed by `#N` as a closing directive. So merging #689 closed this PR one second after it landed (`ClosedEvent.closer` = PullRequest #689, confirmed via GraphQL). I meant "the fix for #604 is PR #684" and wrote it in a way that said "close #684". Nothing about the PR was rejected — the ROI board I merged in that same PR lists it as the **second** unblock-first item.

**The state now.** Reopened, unchanged, still blocked only on the `log-writer.ts` conflict caused by #685 — the resolution I posted above stands (keep `getLogLabels`, add `ContradictionInfo`, drop the stale-base `TEXTS`). Nothing else is expected from you beyond that rebase.

**Checked for repeats.** I scanned every open PR's title and body for the same shape. The other matches (`Fixes #592`, `#597`, `#672`, `#666`, `#688`, `#678`, `#657`) all point at issues, which is what they are for. This was the only one aimed at a PR by accident, and no other PR was closed this way.

Sorry for the noise — the close was mine, and the wording that caused it was mine too.

green-dalii (approved):
Approved, and I re-derived the deadness from the writer rather than from the reader — the reader was already obvious.

The load-bearing claim is that nothing feeds `review_ok`. `updateContradictionStatus` is the only writer of a contradiction's status, and its complete call set is `'resolved'` and `'pending_fix'` (`contradiction-phase.ts:108,116`) — no call site anywhere passes `'review_ok'`, and no schema, enum or prompt can produce it. So the filter at `:104` never matches, which makes `resolveContradiction` reachable only from that filter, and `clampPageSections` reachable only from `resolveContradiction` (`contradictions.ts:196,204`, both inside the method opening at `:177`). The three form one cluster with no live entry point, which is what -1166 lines with a `+135` should look like.

Two things I checked because a deletion this size can take a detector with it: `src/wiki/page-factory/` is untouched (zero diff), and `contradiction-gates.ts` keeps all 9 exports, so detection and reporting survive — `ContradictionManager` still exposes `getOpenContradictions`. Deletion is self-consistent on the branch: all five symbols at 0 references, `review_ok` left only in the two comments that explain its removal. CI green, and the comments at `contradiction-phase.ts:9` and `contradictions.ts:5` are the right place for that history.

PR #685, opened 2026-09-11
fix(log): one label table for the ingest log, read by writer and parser alike (#667)

### LINKED ISSUE #667: Log labels: three supported languages fall back to English, and the table is duplicated in all eleven locale files
`LogWriter.labels()` resolves the ingest log's section headings against `TEXTS.en.logLabels`, which covers eight languages: `en, zh, ja, ko, de, fr, es, pt`. `WIKI_LANGUAGES` offers eleven. `zh-Hant`, `it` and `ru` are not in the table, so `lang in TEXTS.en.logLabels` is false and `labels()` returns English.

A vault set to Italian, Russian or Traditional Chinese therefore writes `**Created pages**` / `**Updated pages**` / `**Contradictions found**` into `wiki/log.md` while the rest of its UI is translated.

**The second half is why it happened.** The same eight-language table is embedded in every one of the eleven locale files, and only `en`'s copy is ever read. Ten copies are dead weight — and one has already drifted: `src/texts/zh-hant.ts` carries Traditional characters under the `zh` key (`建立頁面` / `更新頁面` / `發現矛盾` against `创建页面` / `更新页面` / `发现矛盾`). That is the only Traditional rendering of these labels in the repo, it is correct, and it is unreachable.

The drift is structural, not accidental: `logLabels` is a table *of all languages* nested inside files that are *one language each*, so a translator working on their locale file reasonably translates it, and the parity test passes because it checks keys, not values.

**Suggested fix:** lift the table into a single module keyed by language, read it as `logLabels[lang]` with an English fallback, and delete it from the eleven locale files. That makes the duplication impossible to recreate and turns the missing locales into three entries instead of thirty-three. `zh-Hant` already has its translation (above); `it` and `ru` would be new.

Reporting only — no PR attached for now.

DocTpoint:
## Summary

Closes #667.

**What was wrong.** The ingest log's section labels were an eight-language table nested inside every one of the eleven locale files, and only `en`'s copy was read. `zh-Hant`, `it` and `ru` were not in it, so a vault in one of those languages wrote English headings into `wiki/log.md` while the rest of its UI was translated. `zh-hant.ts` had meanwhile received a correct Traditional rendering under the `zh` key, reachable by nobody.

**The second copy.** The Operation History Panel's parser carried its own hand-typed alternation for the Created/Updated lines (`log-parser.ts`). It listed `Páginas criadas` twice and `Páginas creadas` never, so a Spanish vault's ingest entries reached the panel without their pages; ru and zh-Hant were absent there too. Same drift class, one step downstream.

**The fix.** `core/log-labels.ts` holds one table keyed by language, covering every entry of `WIKI_LANGUAGES`. The writer looks up by language with an English fallback; the parser builds its two patterns from the same values, so a label the writer can produce is one the parser reads. It keeps accepting every language's label regardless of the vault's current setting, as before, because older entries were written under older settings. The eleven nested copies are gone. New translations: `it` (Pagine create / Pagine aggiornate / Contraddizioni trovate), `ru` (Созданные страницы / Обновлённые страницы / Найдены противоречия); `zh-Hant` takes the rendering that was already in the tree.

**Observed, not changed:** `classifyOperation` in the same parser keeps a hand-typed list of maintenance-report keywords for six of the eleven languages. Same class; a separate issue if you want it.

Tests: 4028 → 4035. The new ones pin coverage against `WIKI_LANGUAGES`, the English fallback (including a prototype key such as `constructor`), and, for es, it, ru and zh-Hant, that what the writer writes, the parser reads. `/simplify` and `/code-review` ran before the push; the shared `escapeRegex` and the prototype-safe lookup came out of them.

Touches `log-writer.ts` in a different hunk than #684; whichever merges second needs a trivial rebase (the new test builds a `SourceAnalysis` with the `contradictions` field that #684 removes).

Gate 1: lint 0/0 · tsc 0 · build · 4035/4035 · css-lint 0.

### DIFF
--- a/src/__tests__/core/log-labels.test.ts
+++ b/src/__tests__/core/log-labels.test.ts
@@ -1,0 +1,77 @@
+// One table for the ingest log's section labels, read by the writer and the
+// parser alike (#667). Before: the writer read an eight-language table out
+// of `TEXTS.en`, so it/ru/zh-Hant vaults got English headings; the parser
+// carried its own hand-typed alternation, which had lost Spanish.
+
+import { describe, it, expect, vi } from 'vitest';
+import { LOG_LABELS, getLogLabels, logLabelLineRe } from '../../core/log-labels';
+import { WIKI_LANGUAGES } from '../../types';
+import { LogWriter } from '../../wiki/engine-internals/log-writer';
+import { parseLogEntries } from '../../core/log-parser';
+import type { SourceAnalysis } from '../../types';
+
+describe('LOG_LABELS', () => {
+  it('covers every selectable wiki language with three non-empty labels', () => {
+    for (const lang of Object.keys(WIKI_LANGUAGES)) {
+      const labels = LOG_LABELS[lang];
+      expect(labels, lang).toBeDefined();
+      for (const value of Object.values(labels)) {
+        expect(value.trim().length, lang).toBeGreaterThan(0);
+      }
+    }
+  });
+
+  it('falls back to English for an unknown or missing language', () => {
+    expect(getLogLabels('xx')).toBe(LOG_LABELS.en);
+    expect(getLogLabels(undefined)).toBe(LOG_LABELS.en);
+    expect(getLogLabels('constructor')).toBe(LOG_LABELS.en);
+    expect(getLogLabels('ru')).toBe(LOG_LABELS.ru);
+  });
+
+  it('logLabelLineRe matches every language and captures the rest of the line', () => {
+    const re = logLabelLineRe('createdPages');
+    for (const [lang, labels] of Object.entries(LOG_LABELS)) {
+      const m = re.exec(`**${labels.createdPages}**：[[entities/A]]`);
+      expect(m, lang).not.toBeNull();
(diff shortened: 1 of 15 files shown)


green-dalii (approved):
Verified, and this PR found a second drift site I missed when I triaged #667.

The first half is what the issue described: an eight-language `logLabels` table nested inside all eleven locale files, read only from `en`, so `zh-Hant`, `it` and `ru` vaults wrote English headings while the rest of their UI was translated. Lifting it into one table keyed by language and reading it with an English fallback is the fix the issue asked for, and deleting the ten dead copies is what makes the duplication impossible to recreate.

The second half is the one I want noted: the Operation History Panel's parser carried its own hand-typed alternation for the Created/Updated lines, and it listed `Páginas criadas` twice while `Páginas creadas` never appeared — so a Spanish vault's ingest entries reached the panel without their pages, with `ru` and `zh-Hant` absent there too. Same drift class one step downstream of the one the issue reported, and it is exactly the failure mode a hand-maintained parallel list produces. Having writer and parser read one table closes both.

Merging.

Issue #669, opened 2026-09-10
Migrate to zod 4: 17 `.passthrough()` sites in output-schemas.ts + wire-body regression coverage


green-dalii:
## What

`zod` is pinned at `^3.25.76` in `dependencies` (a **runtime** dependency, not a dev one). zod v4 deprecates `.passthrough()`, which this repo uses **17 times, all in `src/llm-sdk/output-schemas.ts`**. Dependabot PR #640 (3.25.76 → 4.5.4) failed Gate 1 on exactly that:

```
186:4  warning  `passthrough` is deprecated. Use `z.looseObject()` or `.loose()` instead  @typescript-eslint/no-deprecated
... 15 warnings total, lines 186-354
✖ 15 problems (0 errors, 15 warnings)
ESLint found too many warnings (maximum: 0).
```

`--max-warnings 0` means the deprecations are a hard block, not a warning to live with.

## Why it is not a mechanical rename

`z.looseObject()` / `.loose()` is the v4 replacement and is documented as behaviour-preserving. The reason this needs its own review rather than a bulk replace is the **wire path**: these schemas are handed to `response_format.schema` (`src/wiki/lint/fix-dead-link.ts:206` and siblings) and go through the AI SDK's zod→JSON Schema conversion. `passthrough()` is what expresses "unknown keys allowed" — if v4's converter represents that differently (position of `additionalProperties`, interaction with `strict`), the bytes on the wire change and only a wire-body test would catch it.

There is a second, sharper reason: issue #658 is open about strict OpenAI-compatible endpoints rejecting `FixDeadLinkSchema` because all four properties are optional while `required` must cover every property under strict mode. That fix has to land **at this same boundary**. Doing the zod 4 migration and the strict-mode normalisation as one piece of work on `output-schemas.ts` avoids normalising the dialect twice.

## Verified scope

- Runtime dep: `zod: ^3.25.76` in `dependencies`; 6 source files import from `'zod'`
  (`types.ts`, `llm-sdk/output-schemas.ts`, `llm-sdk/output-args.ts`, `llm-sdk/openai-compat-sdk-client.ts`, `wiki/query-engine/pipeline/query-keywords.ts`, `wiki/query-engine/pipeline/seed-selector.ts`)
- Migration sites: **17 × `.passthrough()`**, all in `src/llm-sdk/output-schemas.ts`
[...]

PR #686, opened 2026-09-11
fix(llm-sdk): strict structured-output dialect as a negotiated tier at the schema boundary (#658)

### LINKED ISSUE #658: Custom OpenAI-compatible endpoint rejects strict response schema when optional fields omit required action
## Bug
With Karpathy LLM Wiki v1.27.1 and a custom OpenAI-compatible endpoint, Smart Fix dead-link repair fails before the model is called:

```
Invalid schema for response_format 'response': In context=(), 'required' is required to be supplied and to be an array including every key in properties. Missing 'action'.
```

The same error appears for multiple pages, so it is not page-content specific.

## Reproduction
1. Configure a custom OpenAI-compatible provider.
2. Run Smart Fix / dead-link repair on a page.
3. The request using `FixDeadLinkSchema` is rejected by the endpoint.

In v1.27.1, `FixDeadLinkSchema` defines optional top-level properties (`action`, `correct_link`, `stub_title`, `stub_type`) and the typed-output path sends a schema. Strict structured-output implementations require `required` to include every property (and commonly `additionalProperties: false`).

## Suggested fix
For providers that do not support this schema dialect, fall back to `json_object` (or text JSON prompting) for `fix-dead-link` and `lint-alias`, then retain local JSON parsing/validation. Alternatively normalize the generated JSON Schema for strict providers by making all properties required and representing optional values as nullable.

No API keys or private vault content are included here. Tested against plugin v1.27.1 on September 9, 2026.

DocTpoint:
## Summary

Closes #658.

**What was wrong.** A strict OpenAI-compatible endpoint rejected `FixDeadLinkSchema` because `required` did not list its four optional properties. The AI SDK already sends `strict: true` for every openai-compatible provider, so the body a strict endpoint saw was `strict: true` with an incomplete `required` and `additionalProperties: true`. Twelve of the seventeen schemas in `output-schemas.ts` carry optional fields, four of them on the ingest path. The 400 body names none of the rejection verbs the demotion chain knows, so the call failed hard instead of demoting.

**The fix.** A new tier `json_schema_strict` between `json_schema` and `json_object`. Same wire shape, schema rewritten by `llm-sdk/strict-schema.ts`: every property in `required` in property order, former optionals nullable as `type: [..., "null"]` (the form verified against LM Studio in the issue), `anyOf` type unions folded into the same form, `additionalProperties: false` on every object, canonical key order. The validator is wrapped alongside: `null` is dropped from exactly the properties the original schema did not require before the original validator runs, so the seventeen zod schemas are untouched and callers see the value shape they always saw. Rewritten once per schema object, lazily. Raw JSON schemas take the same path.

**Why a tier and not the boundary for every backend.** The issue scoped one normalisation for all providers. I built that first and measured it on the full ingest path (dev-instrument, gemma-4-26b on LM Studio, extract pinned to `json_schema`, two notes, one draw each):

| note | items, strict body | items, plain body | output tokens | wall clock |
|---|---|---|---|---|
| Thymian | 15 | 20 | 19.2k vs 30.5k | 453 s vs 646 s |
| L-Theanin | 5 | 8 | 7.1k vs 9.6k | 189 s vs 234 s |

Under the strict grammar every item must carry every key, and the model trades breadth for it: a quarter to a third fewer items, for fewer tokens. The plain body's single truncation at the token cap (Thymian, batch 1) is the known schema-arm collapse; the strict arm had none. Two draws are a direction, not a number, but a cost of that direction is not one a backend that accepts today's body should pay. So the plain body goes first, unchanged; the strict body follows the strict-mode 400 and is remembered per (baseURL, model), like the other tiers. A user can pin it with `extract=strict:off` in the task policies. Field rejections still demote to `json_object`; the chain allows three steps.

**Behaviour change** (release note): none for a backend that accepts today's body. On a strict backend the model is asked for every property (`null` when it has nothing) and unknown keys are refused by the server instead of passed through. Both are what strict mode means; no setting added.

**For #669.** `strict-schema.test.ts` pins that the same shape under zod 3 `.passthrough()` and zod 4 `.looseObject()` differs raw and is byte-equal normalised, and pins `FixDeadLinkSchema` and `SourceAnalysisLLMSchema` as file snapshots the zod 4 migration must leave unchanged.

Tests: 4035 → 4076. Strict-rejection classifier on the #658 body verbatim; the plain tier pinned unchanged for lmstudio; 400 → strict body → nulls round-tripped through `createMessageWithOutput`; the strict tier remembered on the next call; field rejection still → `json_object`; the non-typed branch pinned; policy alias parsed; normaliser unit tests (optional top-level, nested, arrays of objects, enum, already-nullable, open maps, nodes without a type, purity, key order); all seventeen exported schemas checked against the strict dialect. `/code-review` and `/simplify` ran before the push; the WeakMap memo, the open-map rule and the null fallback for nodes without a type came out of them. Both pinned strict bodies accepted by LM Studio as `strict: true` json_schema, HTTP 200.

@wstczyw: a candidate build for your endpoint is this branch; the first call will take one 400, every call after it goes strict directly.

Gate 1: lint 0/0 · tsc 0 · build · 4076/4076 · css-lint 0.

### DIFF
--- a/src/__tests__/core/task-policy.test.ts
+++ b/src/__tests__/core/task-policy.test.ts
@@ -206,3 +206,10 @@ describe('parseTaskPolicySpec — bounded thinking', () => {
     expect(thinkingEffort('default')).toBeUndefined();
   });
 });
+
+describe('json_schema_strict as a pinnable mode (Issue #658)', () => {
+  it('parses `strict` and `json_schema_strict` to the strict tier', () => {
+    expect(parseTaskPolicySpec('extract=strict:off').extract?.outputMode).toBe('json_schema_strict');
+    expect(parseTaskPolicySpec('*=json_schema_strict:default')['*']?.outputMode).toBe('json_schema_strict');
+  });
+});
--- a/src/__tests__/llm-sdk/__snapshots__/strict-schema/FixDeadLinkSchema.json
+++ b/src/__tests__/llm-sdk/__snapshots__/strict-schema/FixDeadLinkSchema.json
@@ -1,0 +1,37 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "additionalProperties": false,
+  "properties": {
+    "action": {
+      "type": [
+        "string",
+        "null"
+      ]
+    },
+    "correct_link": {
+      "type": [
+        "string",
+        "null"
+      ]
+    },
+    "stub_title": {
+      "type": [
+        "string",
+        "null"
+      ]
+    },
+    "stub_type": {
+      "type": [
+        "string",
+        "null"
+      ]
+    }
+  },
+  "required": [
+    "action",
+    "correct_link",
+    "stub_title",
+    "stub_type"
+  ],
+  "type": "object"
+}
--- a/src/__tests__/llm-sdk/__snapshots__/strict-schema/SourceAnalysisLLMSchema.json
+++ b/src/__tests__/llm-sdk/__snapshots__/strict-schema/SourceAnalysisLLMSchema.json
@@ -1,0 +1,348 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "additionalProperties": false,
+  "properties": {
+    "source_title": {
+      "type": [
+        "string",
+        "null"
+      ]
+    },
+    "summary": {
+      "type": [
+        "string",
+        "null"
+      ]
+    },
+    "entities": {
+      "items": {
+        "additionalProperties": false,
+        "properties": {
+          "name": {
+            "type": "string"
+          },
+          "type": {
+            "type": [
+              "string",
+              "null"
+            ]
+          },
+          "aliases": {
+            "items": {
+              "type": "string"
+            },
+            "type": [
+              "array",
+              "null"
+            ]
+          },
+          "summary": {
+            "type": [
+              "string",
+              "null"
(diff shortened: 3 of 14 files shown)


green-dalii (approved):
Verified, and the diagnosis is sharper than the issue's: the AI SDK already sends `strict: true` for every openai-compatible provider, so a strict endpoint was receiving `strict: true` alongside an incomplete `required` and `additionalProperties: true` — the combination that makes the endpoint reject rather than negotiate. Twelve of the seventeen schemas in `output-schemas.ts` carry optional fields and four of them are on the ingest path, so `FixDeadLinkSchema` was the schema that happened to be tested first, not the only one affected.

Placing `json_schema_strict` between `json_schema` and `json_object` is the right tier shape: it preserves capability negotiation instead of forcing one dialect, and the rewriter's rules (`required` in property order, former optionals nullable as `type: [..., "null"]`, `anyOf` unions folded, `additionalProperties: false`) are concrete enough to test against a real endpoint rather than assert.

Two notes for the record. The nullable form is the one the issue verified against LM Studio, which is why I did not ask for a different encoding. And this lands at the same boundary as the zod 4 migration (#669) — I wrote that issue expecting exactly this collision, so #669 should rebase on this rather than normalising the dialect twice.

Merging.

Issue #608, opened 2026-09-02
Resolve and Analyze Local Image and GIF Embeds During Markdown Ingestion


Chase07:
## Problem

Karpathy LLM Wiki can ingest Markdown files and supports multimodal models. However, when a Markdown note contains local image or GIF embeds, the plugin does not appear to resolve and read the referenced files during ingestion.

For example:

```markdown
# Test Article

Please analyze the image below:

<span class="claudian-embedded-image-fallback">![[attachments/test-image.jpg]]</span>
```

The image exists inside the Obsidian Vault and is displayed correctly by Obsidian. However, during Markdown ingestion, the generated Wiki content only reflects the Markdown text. It does not appear to include information from the actual image.

I have already downloaded the original remote images locally and replaced the remote URLs with local Obsidian links. Therefore, the issue is not downloading the attachments. The issue is that the local image files are not being passed to the multimodal model during Markdown ingestion.

## Expected behavior

When ingesting a Markdown file, the plugin should:

1. Parse Obsidian image embeds such as `<span class="claudian-embedded-image-fallback">![[image.jpg]]</span>`.
2. Parse standard Markdown image links such as `![alt text](path/to/image.jpg)`.
3. Resolve the image path relative to the current Markdown file or the Vault root.
4. Read the local image file.
5. Send the image as a multimodal input together with the Markdown text.
6. Use the image content for summarization, entity extraction, concept extraction, and Wiki page generation.

## GIF support

For GIF files, the plugin could analyze the first frame or extract representative key frames. If direct GIF input is not supported by the selected provider or model, the plugin could convert the GIF into static frames before analysis.

## Suggested implementation

A new setting could be added, for example:

> Analyze embedded images during Markdown ingestion

Possible options could include:

- Enable or disable local image analysis;
- Maximum number of images per Markdown file;
- Maximum image size;
- GIF processing mode;
- Whether to continue text ingestion if image processing fails;
- Caching previously analyzed images.

## Alternatives considered

- Converting the entire Markdown note to PDF and sending it through MinerU;
- Manually describing the images before ingestion;
- Using a separate OCR or image-analysis plugin.

These alternatives are less convenient because they require an additional manual preprocessing step and may lose the relationship between the original Markdown content and its embedded images.

## Additional context

The configured model supports multimodal input. The images are stored locally in the Obsidian Vault, and Obsidian can display them normally.

[...]

PR #687, opened 2026-09-12
feat: analyze local Markdown image embeds during ingest

Chase07:
Refs #608

## Summary

Adds opt-in analysis of vault-local images embedded in Markdown sources. The resolver supports Obsidian embeds and standard Markdown image syntax, preserves source order, skips remote targets, and leaves text ingestion intact when an image cannot be used.

Eligible images are analyzed in serial visual-evidence packages with a 10 MiB per-image limit, a 20 MiB raw-byte package budget, and at most eight images per package. GIF inputs are decoded to a static first-frame PNG. The final entity and concept extraction receives the original Markdown plus compact textual visual evidence, rather than repeatedly uploading original images.

## Why image and text content are sent together

Visual analysis is not only OCR. An embedded image often depends on nearby Markdown for its meaning, so each image is sent with its index, vault path, cleaned local context, and corresponding image content block in source order. This lets the vision model distinguish visible contents from the image's role in the note.

The subsequent entity and concept extraction does not receive original images again. It receives the full Markdown and compact textual visual evidence, avoiding repeated image uploads across extraction rounds.

## Provider transport and cancellation

Image input uses structured multipart message content rather than a plain string. The OpenAI, OpenAI-compatible, and Anthropic clients forward the original `LLMMessage[]` so the AI SDK retains text/image part ordering and can receive the ingest `abortSignal`.

The Codex client has a separate Responses/stream model boundary, so `toModelMessages()` performs the minimal role-and-content adaptation required by that client while preserving the same multipart content.

## User controls and diagnostics

Both new image-related settings default to off. The optional source-page audit records per-image local context, model evidence, and skipped-item reasons in a replaceable collapsible section. Remote URLs are not downloaded, and no image analysis or additional image-related writes occur while image analysis is disabled.

## Status

Implementation and regression coverage are complete. Gate 1 passes on the current PR head.


Chase07:
Rebased this PR onto the latest upstream main and resolved the integration conflicts.

Local `pnpm gate:1` passes:
- 293 test files
- 4125 tests

Could a maintainer please approve the pending workflow so GitHub Actions can run the required Gate 1 check? The PR is ready for review.

Chase07:
The PR has now been rebased onto the latest upstream main (including v1.27.2-era changes), and the local full Gate 1 passes: 299 test files / 4187 tests.

The force-push refreshed the existing six commits. Could a maintainer please approve the pending workflow so the required GitHub Actions Gate 1 check can run? Thank you.

Chase07:
Rebased once more onto the latest upstream main and resolved the new
`output-schemas.ts` conflict introduced by the Zod 4 migration.

The embedded-image evidence schema now uses `.loose()` to preserve the
existing permissive behavior while remaining compatible with the current
upstream lint rules. Local `pnpm gate:1` passes again: 299 test files /
4187 tests.

At this point, upstream main seems to have excellent timing—every time this
PR is ready to merge, it brings a fresh new version to catch up with. 🙂

green-dalii:
@Chase07 
Thanks for the three rebases — Gate 1 is green on `1e26fa56` (3448 pass, 305 files) now that CI has been approved, and I have read the diff rather than just the checks.

Before this can move, one thing needs settling: **the description still says "Draft / work in progress" and lists "complete implementation and regression coverage" plus "run final validation before requesting review" as remaining work**, while the PR is no longer marked draft. Either read is defensible, so I would rather ask than guess: is the implementation finished and that section stale, or is there still work you intend to land?

Worth flagging either way, because the description understates what the diff does. The 34 files are not only image plumbing — the change reaches the transport layer of every provider:

- `openai-compat-sdk-client.ts`, `openai-sdk-client.ts`, `anthropic-sdk-client.ts` pass `messages` through directly where they previously mapped to `{ role, content }`, and each gains an `abortSignal` that was not there before.
- `openai-codex-sdk-client.ts` takes the **other** route, `toModelMessages(params.messages)`, so the four clients do not end up with the same mechanism.

That is fine as a design, but it is the highest-risk surface in the plugin and it is not what "Refs #608 — analyze local Markdown image embeds" describes. A paragraph on why `messages` is now forwarded unwrapped, and why the Codex client diverges, would make this reviewable.

Two things I checked and am satisfied with, so you do not have to re-argue them: the two new settings default to `false`, which I read as correct here rather than as a miss against the project's usual "default on" rule — sending a user's images to a third-party LLM should not happen without opting in — and the resolver is bounded (10 MB per image, 20 MB per package, 8 images, 500-char context), writes nothing, and skips remote targets explicitly, so there is no cost or side effect while the setting is off.

No formal review from me until the first point is answered.

green-dalii:
Settling the WIP question myself rather than leaving it on you, since the answer is in your own comments.

Your last three all say the work is complete and Gate 1 passes locally — "The PR is ready for review", then two more rebases. The description still opens with "Draft / work in progress" and lists "complete implementation and regression coverage" plus "run final validation before requesting review" as remaining. **Those are the same two items your comments report as done, so I am treating the description as stale** and this as ready for review. If either is genuinely outstanding, say which and I will not touch it.

Two things do remain, and neither is about your code:

1. **It has conflicts again.** `mergeable` is `false` — expected, since four PRs merged to `main` after your last rebase (#757, #673, #759, #762). You have rebased three times already, which is three more than a contributor should have to; I am happy to do this one as maintainer if you would rather not chase it again.

2. **A real review.** I read the diff on `1e26fa56` but did not review it properly, and I want to be straight about what that means. 725 production lines against 269 test lines is thin for the surface — four SDK clients, the analyzer, the settings toggle, eleven locales. Before I form a view I want to check whether the untested paths are the ones that cannot fail, or the ones that fail quietly. That is my task, not yours.

For the record, the shape looks right: `embedded-image-evidence.ts` and `embedded-image-resolver.ts` as separate modules rather than folded into `source-analyzer.ts`, and the analyzer growth (+172) is where the batching belongs. The `.loose()` on the embedded-image schema after the Zod 4 migration is the correct call — I made the same one in #733 for the same reason.

Rebase it or tell me to, and then this is next in the queue.

Chase07:
@green-dalii 
Thank you for the careful read. You are correct: the implementation and regression coverage are complete, and the draft/remaining-work wording was stale. I have updated the PR description to reflect the finished scope and to document the multipart message transport and Codex adapter rationale.

Please take the latest rebase as you offered. I would appreciate the
maintainer-side update rather than chasing main for a fourth time. Thank you.

green-dalii:
Done, and green — but I cannot push it, so this needs one click from you.

**The rebase is complete and Gate 1 passes: 309 files / 4298 tests.** It was smaller than the three you did: seven commits, and **exactly one conflict**, a two-line import block in `source-analyzer.ts`. I will keep it ready.

**What blocks it:** `maintainer_can_modify` is `false` on this PR, so pushing to `feat/608-embedded-markdown-images` is rejected with `permission denied`. I tried twice — the first attempt looked like a different failure (non-fast-forward), which is just what a rebase looks like, but with `--force-with-lease` against your current tip it comes back as a permission denial.

**Please tick "Allow edits by maintainers"** in the PR sidebar (it is under the reviewers box). I will push within the hour of it being on and you will not have to touch `main` again. If you would rather not grant that, say so and I will hand you the exact patch instead — but that would be a fourth rebase by hand, which is the thing you asked to avoid.

**How I resolved the one conflict, since it is a judgement and not a merge** — you should see it rather than take it:

`main` and your branch both changed the `source-analyzer.ts` imports. `main` had `foldToVocabulary` from `core/tag-vocab` plus three schemas; yours had `getActiveEntityTags` / `getActiveConceptTags` / `foldToVocabulary` plus four schemas including your new `EmbeddedImageEvidenceSchema`. I kept `main`'s tag-vocab import verbatim and added your schema to the `output-schemas` line:

```ts
import { foldToVocabulary } from '../core/tag-vocab';
import { EmbeddedImageEvidenceSchema, SourceAnalysisLLMSchema, LemmaClassifyLLMSchema, TypeRepairLLMSchema } from '../llm-sdk/output-schemas';
```

Taking the union would have kept `getActiveEntityTags` / `getActiveConceptTags`, which #673 replaced with the single `activeVocabulary` entry point — so they would have been unused imports, and `no-unused-vars` fails the build. Typecheck and the full suite pass with this, which is the evidence that nothing else in your code depended on them.

**What comes next, so the shape is not a surprise.** Once it is pushed I will review the diff properly against `main` — I read it earlier but did not review it, and the reach is wider than the description used to say: your change goes through the transport layer of four provider clients, not only image plumbing. The `.loose()` on the embedded-image schema is the right call for the Zod 4 migration; I made the same one in #733.

Thank you for the description update. It now says what the diff does, which is what made this reviewable rather than a guess.

Chase07:
@green-dalii 
Enabled. Please proceed with the rebase and push. Thank you for resolving the conflict and documenting the import decision.

green-dalii (approved):
Approving on the rebased head `e3273353`. The conflict was one import block and I resolved it the way that keeps the tag vocabulary work intact: `foldToVocabulary` stays, `EmbeddedImageEvidenceSchema` joins the `output-schemas` import. A union of both would have left `getActiveEntityTags` and `getActiveConceptTags` as unused imports after #673 collapsed them into a single `activeVocabulary`, so the union was not the safe default it looks like.

What I checked in the code rather than in the description.

The message-type change is the right shape, and it is the part I would have asked for. `LLMMessage` splits `user` from `assistant` and strips `ImageContentPart` from the assistant arm, so removing the `messages.map((m) => ({ role, content }))` filter cannot leak anything: the union already has exactly two keys per arm. Passing `messages` through directly is what unblocks images at all, and it is safe here rather than in general.

Both gates are strict equality (`=== true`), so an unset setting is off rather than truthy-coerced, and `embedded_image_analysis` is documented runtime-only — it does not reach frontmatter or change the file format. The degradation test (`continues with text extraction when the endpoint rejects image input`) is the one that matters most for a default-off feature: a non-vision endpoint costs a failed request, not a failed ingest.

I ran the three image-related test files on the rebased branch rather than trusting CI alone: 37 pass. The `@`-handle pair is the detail I would have missed — stripping a trailing handle from context while preserving a sentence that merely contains an at sign is a rule that has to be stated twice to be right.

`saveEmbeddedImageEvidence` resetting to `false` while its neighbours reset to `undefined` is harmless here because false is already the default, but it is the one line in the reset block that reads as deliberate when it is only equivalent.

Issue #467, opened 2026-08-15
Refactor: introduce single setter for tempSettings.model to replace the removed commit-time cascade


green-dalii:
## Context

PR #462 (v1.26.4 PATCH, Issue #456) removed the v1.24.1 PATCH Phase 5.5.0 belt-and-suspenders cascade call inside `commitTempSettings`. That cascade was added specifically to catch direct writes to `tempSettings.model` that bypassed `setFieldValue` (and therefore skipped its internal `cascadeUnifiedModelChange`).

After PR #462 the 3 direct-write sites remain in place. Today they happen to write values that don't trigger user-visible breakage:

- `src/ui/settings-sections/provider-section.ts:73` — `tempSettings.model = ''` on provider change
- `src/ui/settings-sections/provider-section.ts:182` — `tempSettings.model = ''` on Bedrock region change
- `src/ui/settings-sections/test-connection-section.ts:83` — `tab.tempSettings.model = tab.plugin.settings.model` post-test sync

The next contributor who adds any "write `tempSettings.model`" feature (e.g. Apply preset, quick-switcher, profile import) will silently reintroduce the 2026-07-13 e2e bug: per-task overrides stay pinned to stale values while the picker shows the new model.

### Option A — Single setter (lightweight)

```ts
// src/ui/settings.ts
public setUnifiedModel(value: string): void {
  this.tempSettings.model = value;
  if (value.trim().length > 0) {
    this.cascadeUnifiedModelChange();
  }
  this.markLLMConfigStale();
}
```

Route the 3 direct-write sites through `setUnifiedModel`. ~30 LOC + 1 regression test (asserts `setUnifiedModel('preset-X')` cascades per-task fields to empty).

### Option B — Proxy (deeper)

Wrap `tempSettings` in a Proxy that intercepts `model` writes and fires `cascadeUnifiedModelChange` on any non-empty assignment. ~50 LOC + multi-file regression tests.

## Why defer to v1.27.0 MINOR

- 3 settings-section modules touched; blast radius exceeds PATCH scope.
- Today's 3 sites happen to write values that don't trigger user-visible bug (empty string or saved snapshot).
- PATCH scope was per-task preservation at commit (Issue #456), not the underlying two-path mutation mechanism.

## References

- PR #462: https://github.com/green-dalii/obsidian-llm-wiki/pull/462
- Issue #456: https://github.com/green-dalii/obsidian-llm-wiki/issues/456
[...]

Issue #676, opened 2026-09-11
Bug: "Fix Dead Links" can resolve a dead link to the source page itself, silently converting a known gap into a false "resolved" state


wstczyw:
**Plugin:** Karpathy LLM Wiki (`karpathywiki`), v1.27.1
**Area:** `fixDeadLink()` in `src/wiki/lint/fix-dead-link.ts`
**Related:** #592 (same function, different defect: alias clobbering), #485 (closed; introduced `createStubsForUnresolvableLinks`)

Investigated by reading the shipped `main.js` of v1.27.1 and by measuring the resulting corpus; not yet reproduced against a source checkout.

## Symptom

After lint's "Fix Dead Links" stage runs, a page ends up linking to **itself** from its own "Related Concepts" / "Related Entities" list:

```markdown
<!-- wiki/concepts/incremental-update.md, line 44, in its own "Related Concepts" list -->
- [[concepts/incremental-update|incremental-update]]
```

The link that occupied that line before the run pointed at a *different*, non-existent page:

```markdown
- [[concepts/smart-batch-skip|Smart Batch Skip]]
```

Measured on our vault (~1,490 concept pages + ~848 entity pages): **24 pages, 27 list-item occurrences**, across both `concepts/` and `entities/`. Restricted to list items; prose auto-linking produces a different, benign shape and is excluded from that count.

## Why this is worse than noise

The rewritten link is not merely useless — it is a **false negative for every future lint run**. A dead link is reported on every pass, so the gap stays visible and someone can eventually create the page or drop the reference. A self-link resolves cleanly, disappears from the report, and the record that anything was ever missing is gone.

With `autoSmartFix: true` (apply all fixes without showing the report) plus `periodicLint: daily`, this happens unattended and no human sees the report in which the original dead link appeared.

## Root cause

Two independent gaps in the same function, either of which alone would prevent the outcome.

**1. The candidate pool does not exclude the source page.**

```js
const pool = existingPages.filter((p) => {
  const bn = p.title || "";
  const hasPollutedBasename = /^(entities|concepts|sources)([^\s\-_a-zA-Z0-9])/.test(bn);
  return !hasPollutedBasename;
});
```

The only filter is the polluted-basename guard. `sourcePath`'s own page stays in the pool.

This matters more than it looks, because of how the pool is then ranked:

```js
[...]

PR #689, opened 2026-09-12
docs(roadmap): record wave E (17 PRs) and the PATCH ROI board

green-dalii:
Records two things in ROADMAP.md, both consequences of the 2026-09-10 → 09-12 work.

**Wave E** — the 17 PRs merged after v1.27.1, as a table with each PR's issue pair, the measurement behind the fix, and its severity class. 3993 → 4115 tests, main `6b543bf`, eight issues auto-closed by their PRs. Also names the three dependency/CI bumps and why the vitest 5 run was verified locally rather than trusted from the Dependabot tick.

**ROI board** — replaces the 2026-09-04 cadence, which was stale enough to be misleading. Its headline is the fact that shapes the next actions: the milestone has **zero merge-ready PRs today**. All five open PRs are blocked, and three of those conflicts were caused by this session's own merges (#671/#680 against #673 and #681, #685 against #684). The board therefore ranks the unblock-first items — where the code and tests already exist and only a rebase is missing — above any new work, then the five bugs no contributor is waiting on, then three milestone reconciliation calls (#604 sits on `v1.28.0 MINOR` while its fix #684 sits on the PATCH line; #467/#468/#568 are not patch-shaped).

**Active backlog** gains the wave-E shipped rows, a warning on three stale rows, and a statement that the ROI board is now the single source for milestone-scoped remaining work — the table above it is the wider backlog including un-milestoned items.

Docs only. No code, no settings, no user-facing text.

### DIFF
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -113,6 +113,32 @@ Deterministic-related-lists + dependency-security root fix, the final wave befor
 
 3993 tests (Gate 1 green). #636 closes #635; #637 closes 8 Dependabot alerts (open=0). v1.27.1 release notes carry both.
 
+### Shipped into v1.27.x PATCH — wave E (2026-09-10 → 09-12, 17 PRs, main `6b543bf`)
+
+Post-v1.27.1 correctness wave: @DocTpoint's issue→PR pairs (each bug filed with measurements, then fixed with tests in the same window), @Jan-Heldal's first two PRs, and dependency/CI hygiene. All review-approved (`gh pr review` event precedes every `gh pr merge`) and squash-merged; 3993 → **4115 tests**.
+
+| PR | Issue | What | Class |
+|----|-------|------|-------|
+| **#645** | #644 | sibling edges only rescue an orphan, capped at `SIBLING_CAP = 3` — 8562 of 8673 live edges were siblings, cliques up to 20 pages/note | graph-integrity |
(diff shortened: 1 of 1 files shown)


green-dalii:
Docs only, and both halves are worth having in the file rather than in a session.

Wave E is the factual record: 17 PRs with their issue pairs, the measurement behind each fix, the severity class, and the three dependency/CI bumps — including why the vitest 5 run was verified in an isolated worktree instead of trusting the Dependabot tick (4028/284 identical to the v4 baseline, and the one `ENOENT: main.js` failure was the documented build-before-test ordering rather than a v5 regression).

The ROI board earns its place by stating the inconvenient half: the milestone has zero merge-ready PRs today, and three of the five blocks were caused by this session's own merges. Recording that is what makes the next step obvious — unblock before starting anything new. The `#604` cross-milestone note is the one I would have missed: the issue sits on `v1.28.0 MINOR` while the PR that closes it sits on the PATCH line.

Replacing the 2026-09-04 cadence rather than appending to it is the right call — two forward-looking plans in one file are how a roadmap starts lying.

Merging.

DocTpoint (approved):
Approving `9363ebd` for the wave-E record; the rows on my threads are accurate.

Two rows of the ROI board are already behind the repo: rows 1–3 list the three rebases as the unblock step, and all three were repushed at 06:31Z (`b05b099`, `e764c5d`, `bf95f35`), CI green — they are now waiting on review, not on a rebase. Row 9 (#676) predates my comment there: the self-resolution also sits in `fixDeadLink`'s deterministic pre-check and in the ingest's Related renderer, so "needs a minimal reproduction first" is only the Fix Dead Links half. Fine to land as is and update in the next docs pass, or fold in — your choice.

PR #690, opened 2026-09-12
fix(ingest): a skipped file releases the ingest lifecycle (#688)

### LINKED ISSUE #688: Bug Report: Status bar item stuck after skipping/cancelling an ingestion
# Bug Report: Status bar item stuck after skipping/cancelling an ingestion

## Summary

When a user clicks **Cancel** on the status bar item during an ingestion, or chooses **Skip** when prompted, the status bar item remains stuck showing `"· 提取中... 点击取消"` ("extracting... click to cancel"). It never clears until Obsidian is restarted.

## Environment

- Karpathy LLM Wiki v1.27.1
- Obsidian 1.11.4, macOS (Apple Silicon)
- Provider: OpenAI-compatible endpoint

## Steps to Reproduce

1. Start an ingestion (single file or batch from folder).
2. While the status bar shows `提取中... 点击取消`, click it to cancel (or click Skip on the skip prompt).
3. Ingestion aborts and the report modal closes.

## Expected behavior

After cancellation or skip, the status bar item should hide or reset to idle.

## Actual behavior

The status bar text freezes at `· � English 提取中... 点击取消` — a stale "extracting..." state that persists indefinitely until app restart. It looks like the ingestion is still running when nothing is happening. The top-right Notice is also not dismissed on the skip path, reinforcing the impression that ingestion is still in progress.

## Analysis

Looking at `main.js`, the skip/cancel early-return path exits before the `finally` block that hides the status bar and clears the ingestion state. So the status-bar "extracting" text is never reset. Suggest wrapping the status-bar hide/reset in a `try/finally` that covers the skip path, or explicitly calling the same cleanup that the normal completion path uses.

A minimal fix would be: on the skip/cancel early-return, call `onIngestionEnd()` (or equivalent cleanup) before `return`.

Thanks for the great plugin!

green-dalii:
Closes #688.

**The defect.** The requirements gate's skip returns above the method's main `try`, so the teardown that lives in that `finally` never ran for a skipped file:

```
wiki-engine.ts:914   if (this.abortController === null) { …; this.onIngestionStart?.(file.basename); }   ← status bar shown
wiki-engine.ts:996   this.reportSkip(file, rejection, opts); return;                                     ← early return
wiki-engine.ts:1627  } finally { this.triageContradictions = null; this.endIngestion(); }                ← never reached
```

**Two consequences, and the unreported one is worse.** The status bar stayed on "extracting…" until Obsidian restarted — that is what the issue describes. But `abortController` also stayed non-null, and the guard at the top of `ingestSource` only builds a controller *and only calls `onIngestionStart`* when it finds none. So every later file in the same batch inherited the skipped file's controller: no fresh signal for the status bar's cancel click to abort, no start hook for the file being processed, and `wasCancelled` never reset. The second test below pins exactly that.

**The fix.** `endIngestion()` becomes the single owner of the teardown, called from the three sites that need it: the conversion branch's guarded `finally`, the gate's skip, and the main `finally`. The guard in the first is what keeps the PDF re-entry from firing the end hook twice — the outer call's `finally` sees a null controller after the re-entered call already cleared it.

The gate's stale comment is corrected too. It claimed the gate "runs BEFORE any cancellation/UI setup so a rejected file returns cleanly with nothing to tear down", which stopped being true in v1.25.0 when the controller and start hook were moved above it for the PDF branch. That stale justification is why the missing teardown survived two years of reading.

**Not changed — and this is deliberate.** The issue also reports an undismissed progress Notice. I could not reproduce that from source: `reportSkip` dispatches `onDone` with `skipped: true`, and `onIngestDoneDispatch` calls `dismissProgress()` for every non-`auto` trigger, so the Notice is dismissed on the skip path. I left it alone rather than "fixing" a path that works. If it still reproduces for you, the ingest path and the console output would tell us which branch you hit.

## Tests

`src/__tests__/wiki/wiki-engine-skip-teardown.test.ts` drives the gate with an empty note (`checkNonEmpty` → reason `empty`), the shortest deterministic path to the rejection. Both cases were RED first, for the reasons above:

| Assertion | Before |
|---|---|
| single file: end hook fires once, controller released | `endedCount` 0 |
| batch: the second file still reaches the start hook | `startedFilenames` was `['blank-a']` |

The harness gained `endedCount` as a live getter — a plain number field is copied by value when the return object is built, so it read 0 no matter how often the hook fired.

## Gate 1

lint 0 / tsc 0 / build clean / **4117 tests, 289 files** / css-lint 0.

## Gate 4: Performance

| Dim | Status | Notes |
|-----|--------|-------|
| CPU | ✅ | One extra method call on the skip path; no loop or hot-path change |
| Memory | ✅ improved | One fewer leaked AbortController per batch — the leak was one per skipped file |
| IO | N/A | No vault read or write touched |
| Network | ✅ improved | A leaked controller meant the status-bar cancel click aborted a stale signal; every file now gets a fresh one `cancelIngestion()` can reach |
| Token | N/A | No LLM call shape changed |

### DIFF
--- a/src/__tests__/__support__/wiki-engine-harness.ts
+++ b/src/__tests__/__support__/wiki-engine-harness.ts
@@ -31,6 +31,10 @@ export interface WikiEngineHarness {
   stats: { llmCalls: number; vaultMarkdownScans: number };
   /** Filenames delivered to `onIngestionStart` (status-bar text hooks). */
   startedFilenames: string[];
+  /** How many times `onIngestionEnd` fired — the status-bar teardown hook.
+   *  Paired with `startedFilenames` it exposes a teardown that never ran (#688).
+   *  A live getter on the returned object, not a captured number. */
+  endedCount: number;
   /** Paths handed to fileManager.trashFile, in order. */
   trashedPaths: string[];
   /** Progress messages delivered to `onProgress` (PDF "Reading PDF: …"). */
@@ -67,6 +71,7 @@ export function createWikiEngineHarness(opts: HarnessOptions = {}): WikiEngineHa
   const reports: IngestReport[] = [];
   const stats = { llmCalls: 0, vaultMarkdownScans: 0 };
   const trashedPaths: string[] = [];
+  let endedCount = 0;
   let llmIdx = 0;
 
   const app = {
@@ -150,10 +155,15 @@ export function createWikiEngineHarness(opts: HarnessOptions = {}): WikiEngineHa
   // can read `startedFilenames` to assert the status bar text was set.
   engine.setIngestionCallbacks(
     (filename?: string) => { if (filename) startedFilenames.push(filename); },
-    () => { /* onEnd: nothing to capture */ },
+    () => { endedCount += 1; },
   );
 
-  return { engine, llmRequests, writtenPaths, reports, files, stats, startedFilenames, progressMessages, trashedPaths };
+  return {
+    engine, llmRequests, writtenPaths, reports, files, stats, startedFilenames, progressMessages, trashedPaths,
+    // Getter, not a field: a bare number is copied by value at construction and
+    // would stay 0 however often the hook fires (the arrays above are references).
+    get endedCount() { return endedCount; },
+  };
 }
 
 /** True if any written path is a wiki entity/concept/source page (the #164 symptom). */
--- a/src/__tests__/wiki/wiki-engine-skip-teardown.test.ts
+++ b/src/__tests__/wiki/wiki-engine-skip-teardown.test.ts
@@ -1,0 +1,61 @@
+// #688 — a file the requirements gate skips returns above the method's main
+// try/finally, so the teardown that lives there never ran. Two consequences,
+// and the second is worse than the reported symptom:
+//
+//   1. `onIngestionEnd` never fired, so the status bar stayed visible. (The
+//      progress Notice the issue also reports is *not* this path — `reportSkip`
+//      dispatches `onDone` with `skipped: true`, which reaches
+//      `dismissProgress` in `onIngestDoneDispatch` for every non-`auto`
+//      trigger. Checked against the code rather than the issue text.)
+//   2. `this.abortController` stayed non-null. The guard at the top of
+//      `ingestSource` only builds a controller — and only calls
+//      `onIngestionStart` — when it finds none, so every later file in the
+//      same batch inherited the skipped file's controller: no fresh signal,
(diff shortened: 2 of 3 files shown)


DocTpoint (changes requested):
Changes requested on `7253ad9`, one line, then approve.

The header comment of `wiki-engine-skip-teardown.test.ts` (lines 5–8) says a skip "dispatches no report" and leaves the progress Notice up. Your PR body says the opposite, and the code agrees with the body: `reportSkip` calls `onDone` with `skipped: true`, `onIngestDoneDispatch` reaches `dismissProgress`. The comment carries the issue's diagnosis, not your finding — and a regression test's header is where the next reader learns the fault class, so it must not document a path that does not exist.

Suggested replacement for the parenthesis in consequence 1:

```
//   1. `onIngestionEnd` never fired, so the status bar stayed visible. (The
//      progress Notice the issue also reports is *not* this path — `reportSkip`
//      dispatches `onDone`, which reaches `dismissProgress`; see the PR body.)
```

The rest checked out: the gate's skip is the only return between the controller guard and the main `try` outside the conversion branch (whose guarded `finally` covers its two returns), and `wasCancelled` is reset only inside that guard, so the inherited-controller consequence is real and the second test pins it.

green-dalii:
Taken as given, and you are right about why it matters: a regression test's header is where the next reader learns the fault class, so it must not document a path that does not exist. I had copied the issue's diagnosis into the comment and then contradicted it in the PR body — the code agreeing with the body is the part that makes the comment the wrong one.

Replaced with your wording, with the mechanism named rather than just the verdict:

```
//   1. `onIngestionEnd` never fired, so the status bar stayed visible. (The
//      progress Notice the issue also reports is *not* this path — `reportSkip`
//      dispatches `onDone` with `skipped: true`, which reaches
//      `dismissProgress` in `onIngestDoneDispatch` for every non-`auto`
//      trigger. Checked against the code rather than the issue text.)
```

The parenthetical now carries the file-level trail so the next reader can re-check it instead of trusting the comment.

Also worth flagging on my side: while verifying the repush, my `pnpm gate:1 | tail` chain masked the exit code and I pushed once with `tsc` failing. Fixed (`set -o pipefail` from here on) — the gate now stops the chain rather than the pipe swallowing it.

DocTpoint (approved):
Approving `68f6ae4`. The header now carries the mechanism with the file trail, which is exactly what the next reader needs. Thanks for flagging the masked `tsc` exit — same class we hit ourselves; `pipefail` is the right fix.

PR #691, opened 2026-09-12
fix(lint): Delete Empty Stubs collects a Fix Dead Links stub (#678)

### LINKED ISSUE #678: Delete Empty Stubs never deletes a Fix Dead Links stub: #197 changed the stub sentence, `STUB_MARKER` kept the old one

green-dalii:
`Delete Empty Stubs` never deleted a `Fix Dead Links` stub. It asked `isPageEmpty`, which measures substantive text — and a stub's placeholder sentence clears `MIN_SUBSTANTIVE_CHARS` by roughly a factor of two. Every stub the action exists to collect survived it.

## The fix

`src/wiki/lint/delete-empty-stubs.ts` — the predicate is now a pair:

```ts
if (!isPageEmpty(content) && !(isStubPage(fm) && isEmptyStub(content))) continue;
```

`isStubPage` alone would have been wrong, and that is the part worth reading closely. `stub: true` has **two writers in main**:

| Writer | Body |
|---|---|
| `fix-dead-link.ts:95` | placeholder pointing at a page that does not exist yet |
| `stub-page.ts:140` (`buildDissentStubContent`, #485/#607) | placeholder **plus the extraction's summary and a verbatim quote** |

Their frontmatter is identical in shape, so the marker cannot separate them. Recognising it alone would have deleted pages holding paid-for extraction output, under an action named "Delete Empty Stubs" — a content-loss class change and not what #678 asked for.

`isEmptyStub` therefore asserts the shape directly: **one heading, one quoted line, no prose**. That sorts all four bodies by whether a model produced anything on the page:

| Body | Shape | Outcome |
|---|---|---|
| dead-link stub | 1 heading + 1 placeholder quote | **collected** |
| gate stub, summary + quote | prose present | kept |
| gate stub, summary only | prose present | kept |
| gate stub, neither | 1 heading + 1 placeholder quote | **collected, on purpose** |

The fourth row is deliberate: `buildDissentStubContent` drops `quoteBlock` when the item carries no mention (`stub-page.ts:138`), so that stub matches the dead-link shape exactly — and by the rule above there is nothing on it a model produced. Fail-safe by construction: anything a future writer adds that is neither a heading nor a quote keeps the page.

## Tests — ten cases in `delete-empty-stubs.test.ts`

Gate stubs are built by `buildDissentStubContent` itself rather than hand-written, so the shape cannot drift from what the writer produces.

1. deletes the stub even though its placeholder text is not "empty" — the #678 case
2. keeps a page that carries real content
3. keeps a stub the user marked `reviewed`
4. keeps a page a source has filled — `merge-page` strips the marker
5. collects an empty page that is not a stub — the text predicate still governs those
6. `isPageEmpty` is unchanged, so Expand Empty Pages never sees a stub (#197)
7. keeps a gate stub that carries a summary and a quote
8. keeps a gate stub whose summary is empty — its quote is still content
9. a dead-link stub and a gate stub side by side: same helper, same frontmatter, same marker, only the body differs — exactly one collected
10. collects a gate stub with neither a summary nor a quote

Case 6 is the reason the predicate was **not** folded into `isPageEmpty`: that function also decides which pages Expand Empty Pages hands to the LLM, and #197's gate forbids fabricating content for an unresolvable link. Deleting a stub and filling it are different decisions and stay in different places.

Case 10 is mutation-checked rather than assumed: making `isEmptyStub` spare gate stubs fails it and nothing else (1 failed / 9 passed), so a later reader who reads that row as a bug hits the test instead of silently reintroducing #678.

Case 5 was renamed from "keeps an empty page that is not a stub" to match what it asserts — the name had been inverted, which in a deletion path invites a reader to "repair" the behaviour toward keeping.

Gate 1: lint 0 / tsc 0 / build OK / **4108 tests, 291 files** / css-lint 0.

Closes #678.

### DIFF
--- a/src/__tests__/wiki/lint/delete-empty-stubs.test.ts
+++ b/src/__tests__/wiki/lint/delete-empty-stubs.test.ts
@@ -1,0 +1,161 @@
+// #678 — a page Fix Dead Links writes is never collected by "Delete Empty
+// Stubs". #197 (d53caab) replaced the stub sentence and left `STUB_MARKER` on
+// the old one, so `isPageEmpty` stopped recognising these pages: the current
+// placeholder clears MIN_SUBSTANTIVE_CHARS by roughly a factor of two once
+// EMPTY_CONTENT_STRIP has run.
+//
+// The fix keys off `stub: true` instead — the marker `fix-dead-link.ts`
+// documents as the one that actually carries "this is a stub", and which
+// `merge-page.ts` strips when a source that treats the subject fills the page.
+// It is deliberately NOT folded into `isPageEmpty`, because that predicate also
+// decides which pages "Expand Empty Pages" hands to the LLM, and #197's policy
+// gate forbids fabricating content for an unresolvable link. The last test
+// below pins that separation so a later edit cannot merge the two predicates
+// back together by accident.
+
+import { describe, it, expect } from 'vitest';
+import { deleteEmptyStubs } from '../../../wiki/lint/delete-empty-stubs';
+import { isPageEmpty } from '../../../wiki/lint/utils';
+import { buildDissentStubContent } from '../../../wiki/page-factory/stub-page';
+import { createMockContext } from '../../__support__/engine-context';
+
+const WIKI = 'wiki';
+const STUB_PATH = `${WIKI}/concepts/smart-batch-skip.md`;
+
+function deadLinkStub(title: string, referrer: string): string {
+  return `---\ntype: concept\ncreated: 2026-09-12\nsources:\n  - "[[${referrer}]]"\ntags: [thema/wissen]\nstub: true\ngeneration_complete: false\n---\n# ${title}\n\n> Stub created by Fix Dead Links — referenced by [[${referrer}]]. Will be filled by next ingest of an actual source that defines this entity.\n`;
+}
+
+/** Built by the real gate writer, so the shape cannot drift away from the test. */
+function dissentStub(withSummary: boolean, withMention = true): string {
+  return buildDissentStubContent({
+    item: {
+      name: 'Smart Batch Skip',
+      type: 'term',
+      summary: withSummary ? 'An optimisation that skips batches nothing changed in.' : '',
+      // `buildDissentStubContent` drops `quoteBlock` entirely when this is
+      // empty (`stub-page.ts:138`), which is the third form below.
+      mentions_in_source: withMention ? ['Batches whose inputs are unchanged are skipped entirely.'] : [],
+      related_concepts: [],
+    },
+    stubType: 'concept',
+    sourceSlug: 'some-source',
+    cell: 'unnamed-but-present',
+  });
+}
+
+const SUBSTANTIVE = `---\ntype: concept\n---\n# Real Page\n\nA page with a paragraph of actual content, comfortably past the substantive-character floor used by the empty-page predicate.\n`;
+
+async function runDelete(vaultFiles: Record<string, string>) {
+  const { ctx } = createMockContext({ vaultFiles });
+  const deleted: string[] = [];
+  ctx.deleteFile = async (p: string) => { deleted.push(p); };
+  const result = await deleteEmptyStubs(ctx, WIKI);
+  return { deleted, result };
+}
+
+describe('deleteEmptyStubs — a Fix Dead Links stub is collected (#678)', () => {
+  it('deletes the stub even though its placeholder text is not "empty"', async () => {
+    const stub = deadLinkStub('Smart Batch Skip', 'Notizen/note.md');
+    // The premise of the bug: the placeholder clears the floor, so the
+    // text-based predicate says the page is fine.
+    expect(isPageEmpty(stub)).toBe(false);
+
+    const { deleted, result } = await runDelete({ [STUB_PATH]: stub });
+
+    expect(deleted).toEqual([STUB_PATH]);
+    expect(result.deleted).toBe(1);
+  });
+
+  it('keeps a page that carries real content', async () => {
+    const { deleted, result } = await runDelete({ [`${WIKI}/concepts/real.md`]: SUBSTANTIVE });
+
+    expect(deleted).toEqual([]);
+    expect(result.deleted).toBe(0);
+  });
+
+  it('keeps a stub the user marked reviewed', async () => {
+    const reviewed = deadLinkStub('Kept On Purpose', 'Notizen/note.md')
+      .replace('stub: true\n', 'stub: true\nreviewed: true\n');
+
+    const { deleted } = await runDelete({ [STUB_PATH]: reviewed });
+
+    expect(deleted).toEqual([]);
+  });
+
+  it('keeps a page a source has filled — merge-page strips the marker', async () => {
+    const filled = deadLinkStub('Filled By Ingest', 'Notizen/note.md').replace('stub: true\n', '');
+
+    const { deleted } = await runDelete({ [STUB_PATH]: filled });
+
+    expect(deleted).toEqual([]);
+  });
+
+  it('collects an empty page that is not a stub — the text predicate still governs those', async () => {
+    const empty = `---\ntype: concept\n---\n# Thin\n\nTiny.\n`;
+
+    const { deleted, result } = await runDelete({ [`${WIKI}/concepts/thin.md`]: empty });
+
+    expect(deleted).toEqual([`${WIKI}/concepts/thin.md`]);
+    expect(result.deleted).toBe(1);
+  });
+
+  it('leaves isPageEmpty alone, so Expand Empty Pages never sees a stub (#197)', () => {
+    // If a stub ever became "empty" here it would be offered to fillEmptyPage,
+    // which #197's gate exists to prevent. Deleting and filling are different
+    // decisions and must stay in different predicates.
+    expect(isPageEmpty(deadLinkStub('Smart Batch Skip', 'Notizen/note.md'))).toBe(false);
+  });
+
+  // The marker `stub: true` has two writers. The dead-link one writes a
+  // placeholder and nothing else; the ingest candidate gate writes a
+  // placeholder PLUS the extraction's summary and a verbatim quote. Everything
+  // below keeps that second kind on disk — the action is named "Delete Empty
+  // Stubs" and those pages are not empty.
+  it('keeps a gate stub that carries a summary and a quote', async () => {
+    const stub = dissentStub(true);
+    expect(isPageEmpty(stub)).toBe(false);
+    expect(stub).toContain('stub: true');
+
+    const { deleted } = await runDelete({ [STUB_PATH]: stub });
+
+    expect(deleted).toEqual([]);
+  });
+
+  it('keeps a gate stub whose summary is empty — its quote is still content', async () => {
+    const stub = dissentStub(false);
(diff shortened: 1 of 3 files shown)


DocTpoint (changes requested):
Changes requested on `cc922c6`, one point.

`isStubPage` keys off `stub: true`, and Fix Dead Links is not the only writer of that marker. `stub-page.ts:140` — the candidate gate's stub from #485/#607 (`gateDissentStubs`) — writes the same key, and those pages are born with a summary and a quote block from the source, not a placeholder sentence. After this PR, Delete Empty Stubs removes them too unless `reviewed: true`, and neither the body nor the six tests mention that writer. Deleting content-bearing pages from a path the PR does not name is what blocks here, not the outcome itself.

It may well be the right behaviour — merge-page strips the marker when a source fills the page, so both writers share one lifecycle — but then it needs to be stated in the body and pinned with a test on a gate stub carrying summary + quote. If it is not intended, the predicate has to distinguish the two writers, which is closer to what #678's "dedicated predicate" asked for. Either way it should be a decision on record, not a side effect of reusing the helper.

Small precision point in the rationale: `parseFrontmatter` does coerce one boolean — `reviewed` (`frontmatter.ts:67`) — which is why the pre-existing `fm?.reviewed === true` works while `stub` needs the string compare. Reusing `isStubPage` is right; the reason is narrower than "no coercion".

Everything else checks out: the `isPageEmpty` separation for #197's gate is the right cut, and the negative cases are the ones worth having.

green-dalii:
Both points taken, and the first one was a real scope error on my part.

**`isStubPage` was the wrong predicate.** You are right that Fix Dead Links is not the only writer of `stub: true` — `buildDissentStubContent` (`stub-page.ts:140`) sets it too, and that body carries `${summary}` and `${quoteBlock}`. Recognising the marker alone would have deleted pages holding extraction output, under an action named "Delete Empty Stubs". That is content-loss class and it is not what #678 asked for.

**The marker cannot separate them** — I checked: both frontmatter blocks are identical in shape, and the only difference is the body. `fix-dead-link.ts` writes a placeholder and nothing else; the gate writes a placeholder **plus** a summary and a quote.

So the predicate now asserts emptiness directly rather than stopping at the marker. `isEmptyStub`: one heading, one quoted line, no prose. That is exact for the dead-link stub and excludes the gate stub in **both** its forms — with a summary there is prose, and with an empty summary there are two quoted lines (its placeholder and its quote). Fail-safe by construction: anything a future writer adds that is neither a heading nor a quote keeps the page.

**Pinned as you asked.** The gate stubs in the tests come from `buildDissentStubContent` itself rather than being hand-written, so the shape cannot drift from what the writer produces. Three new cases: gate stub with summary kept, gate stub with an empty summary kept, and a dead-link stub and a gate stub placed side by side — same helper, same frontmatter, same marker, only the body differs — asserting that exactly one is collected. That last one is the pair that gives the distinction meaning.

**On the coercion precision:** correct, and my commit message was wider than the fact. `parseFrontmatter` does coerce `reviewed` (`frontmatter.ts:67`), which is why the pre-existing `fm?.reviewed === true` works while `stub` needs the string compare. Reusing `isStubPage` is right; the reason is narrower than "no coercion anywhere".

Also worth flagging: during the rebase I pushed a commit whose `tsc` failed — `| tail` masked the exit code in my verification chain, so the gate looked green when it was not. Caught and fixed (`ConceptInfo` needs `relation_concepts`, and the item's `type` must come from the concept union). Now verified with `set -o pipefail` so a failure actually stops the chain.

DocTpoint (changes requested):
Changes requested on `e74d296`, one line and one test; the behaviour itself is right.

The doc comment and the body say `isEmptyStub` excludes the gate stub "in both its forms — with a summary there is prose, and without one there are two quoted lines". The second half assumes the quote is always there, and it is not: `buildDissentStubContent` sets `quoteBlock` to `''` when the item carries no mention (`stub-page.ts:138`). A gate stub with neither summary nor quote is one heading, one quoted line, no prose — the same shape as the dead-link stub — and is collected.

By your own definition ("nothing the plugin paid a model for") that is the correct outcome, so nothing in the code needs to move. But the comment should say three forms, not two, and name the third as collected on purpose; and a fourth test pinning it (gate stub, empty summary, no mention → deleted) keeps the next reader from "fixing" it. Everything else — the pair test, the fail-safe rule, the `isPageEmpty` separation — checks out. Approve follows.

green-dalii:
Correct, and the third form is the one I should have found — I wrote the comment from the two forms my own tests happened to build, and my tests always passed a mention, so the empty-quote branch never appeared.

Verified at `stub-page.ts:138`: `const quoteBlock = quote ? ... : ''`. With no summary and no mention the body is a heading plus its placeholder — the dead-link shape exactly.

Agreed it is the right outcome by the rule I already wrote down ("nothing the plugin paid a model for"), so the predicate is unchanged. The comment now says **three** forms and names the third as collected deliberately, with the reason inline so it does not read as an accident.

**Pinned, and mutation-checked rather than assumed.** The fourth test is `dissentStub(false, false)` — the helper now takes a `withMention` flag specifically so the empty-quote branch is reachable. To confirm it pins anything, I made `isEmptyStub` bail out on gate stubs: that fails the new case and nothing else (1 failed / 9 passed). The next reader who reads the third form as a bug will hit that test instead of silently reintroducing #678.

Detail worth recording, since it is what my first pass got wrong: the helper builds stubs through `buildDissentStubContent` rather than by hand, which is what makes the shape trustworthy — but a real writer is only as good as the inputs a test gives it. Passing a mention every time hid a production branch. The flag closes that gap.

DocTpoint (changes requested):
Changes requested on `f21f14a` — two items, neither behavioural. The behaviour is right, and I checked the premise rather than the description: `stub: true` has exactly two writers in main — `fix-dead-link.ts:95` and `stub-page.ts:140` — and `buildDissentStubContent` puts the summary on its own line and the quote on a second `>` line, so the three gate forms sort exactly as your doc comment claims (summary+quote 2 quotes, summary-only 1 prose line, neither 1 heading + 1 quote). A multi-line quote or an appended Related section also falls on the keep side, which is the right direction for an action that deletes.

What blocks is a test name that says the opposite of what it asserts, in a deletion path.

`src/__tests__/wiki/lint/delete-empty-stubs.test.ts:94` — the case is named "keeps an empty page that is not a stub" and asserts the opposite at line 99: `expect(deleted).toEqual(['wiki/concepts/thin.md'])`. Your own table calls it "empty non-stub page still collected", so the assertion is the intended one and only the name is inverted. It is the class you corrected in #690 this morning, and here it points the dangerous way: a reader who scans names and not assertions would "repair" the behaviour toward keeping, which is #678 again.

Second, the body still describes the fix as the existing `isStubPage` helper and lists six test cases; the code is `isStubPage(fm) && isEmptyStub(content)` and the file now holds ten, with a Gate 1 count from an earlier run. Since the body becomes the commit message, that is the version of the reasoning that survives — worth a pass before you merge.

green-dalii:
Both taken. The test name was the same class I corrected in #690 this morning, and you are right that it points the dangerous way here — in a deletion path an inverted name is an invitation to "repair" toward keeping, which is #678 restored.

**Renamed** to `collects an empty page that is not a stub — the text predicate still governs those`, matching its assertion.

I also swept the file rather than fixing only the reported line, since this class tends to cluster: all nine `deleted` assertions now match their names (the tenth case asserts on `isPageEmpty` directly and carries no `deleted` expectation). The table is in the body.

**Body rewritten.** It now states the composite predicate `isStubPage(fm) && isEmptyStub(content)` rather than the helper alone, names all ten cases with what each one pins, and carries the Gate 1 count from this run. Your point about the body becoming the commit message is the right reason to spend the extra minute on it.

**One thing I changed beyond the review, because I found it while rewriting.** The branch was seven commits behind `main` — still carrying pre-#684 state, which is why its test count read 289 files against main's 290. Rebased onto `a0b3f189`… `32aafd9a`; it now applies cleanly on top of the current tree. Net diff is 3 files, +221/-2. That also means the reviewer's anchor moved from `f21f14a` to `a0f427c6`.

Gate 1 on the rebased head: lint 0 / tsc 0 / build OK / **4108 tests, 291 files** / css-lint 0.

DocTpoint (approved):
Approve — `a0f427c`.

I diffed the two patches rather than re-reading the whole change: `f21f14a` → `a0f427c` is one changed line, the test name at `delete-empty-stubs.test.ts:94`, now `collects an empty page that is not a stub`. That was the blocker and it is gone. The body is updated as well — the predicate is described as the pair, ten cases instead of six, and the rename is named at the end.

Everything from my previous pass still holds: the three gate forms sort as the doc comment claims, and the mutation check on case 10 is the right pin for the row a later reader would read as a bug. Gate 1 green on `a0f427c`.

One non-blocking detail for the merge. The body now says `4108 tests, 291 files`; the commit message still says `4125 tests` from the earlier run. A squash merge takes the commit message unless you replace it with the body, so the stale count is the one that would survive.

PR #637, opened 2026-09-06
fix(deps): close 8 fast-uri Dependabot alerts at the root — ^3.1.7, drop redundant override, CI audit gate + Dependabot auto-PR config

green-dalii:
## Root-cause fix for the recurring vulnerable-transitive-dep cycle (fast-uri 3.1.5 → 3.1.7)

Closes 8 open Dependabot alerts (4 × GHSA × 2 manifests, all `fast-uri < 3.1.6`).

### What changed

1. **fast-uri `3.1.5` → `^3.1.7`** (3.x latest). Four GHSA advisories (2026-09-03: SSRF / host confusion via skipped IDN canonicalization `GHSA-5jgf-p345-68v8`, malformed IPv6 normalization `GHSA-f65p-4m7j-42xc`, percent-encoded scheme normalization `GHSA-jqff-g426-hqxp`, repeated hostname percent-decoding `GHSA-fph4-wmhf-6fwf`) — all patched at 3.1.6+. DevDep widened from exact pin to `^3.1.7` so future 3.x patches resolve without editing package.json; `pnpm update fast-uri` refreshes the lock.

2. **Drop the fast-uri `overrides` entry** (package.json + pnpm-workspace.yaml). Verified redundant: every consumer (ajv@8.20.0, eslint-plugin-json-schema-validator, json-schema-migrate) declares `fast-uri@^3.0.1`, which natively permits 3.1.7 — the override forced a single instance the semver ranges already agree on. **brace-expansion override KEPT**: it genuinely pins 5.0.9 against legacy minimatch@8/9/3 peers declaring `^2/^1`; removing it would resurrect vulnerable multi-instance legacy brace-expansion.

3. **CI audit gate** (.github/workflows/pr-ci.yml): `pnpm audit --audit-level high --ignore-registry-errors` step after Install. Fails the PR the day a vulnerable dep lands, not when a human notices the alert. `--audit-level high` filters noisy low/moderate; `--ignore-registry-errors` prevents a transient registry ECONNRESET from false-failing.

4. **Dependabot auto-PR config** (.github/dependabot.yml, new): npm weekly (monday 09:00 Asia/Shanghai, limit 3, increase-if-necessary; eslint major ignored for Bot alignment, brace-expansion ignored as override-pinned) + github-actions weekly (checkout/setup-node/attest-build-provenance are supply-chain surface too). Security updates need no config — always on.

### Verification (three-zero)

| Check | Result |
|-------|--------|
| `pnpm install --frozen-lockfile` | clean |
[...]

Issue #556, opened 2026-08-27
pnpm >= 11 silently drops the brace-expansion / fast-uri pins: the `pnpm` field in package.json is deprecated, and pnpm 11 reads neither `pnpm.overrides` nor the top-level `overrides`


DocTpoint:
**Observation.** Every pnpm invocation in the repository prints:

```
[WARN] The "pnpm" field in package.json is no longer read by pnpm. The following keys were ignored: "pnpm.onlyBuiltDependencies", "pnpm.overrides". See https://pnpm.io/settings for the new home of each setting.
```

The warning overstates the present and understates the future, so we measured both. All runs were `pnpm install --lockfile-only` re-resolutions on a scratch copy of `package.json` + `pnpm-lock.yaml` (registry: registry.npmjs.org, per the project `.npmrc`), checking (a) the `overrides:` block recorded in the regenerated lockfile and (b) the resolved `brace-expansion` versions.

| # | Arm | pnpm | Result |
|---|---|---|---|
| 1 | shipped `package.json`, from-scratch resolve | 10.14.0 | pins hold — 10.14 still honours `pnpm.overrides` despite its own warning |
| 2 | top-level `overrides` removed, only `pnpm.overrides` | 10.14.0 | pins hold |
| 3 | control: no overrides anywhere | 10.14.0 | `brace-expansion` fans out to 1.1.18 / 2.1.4 / 5.0.9 — the probe measures what it should |
| 4 | shipped `package.json` (both fields present), `packageManager` bumped to simulate the future | 11.24.0 | **no `overrides:` block in the lockfile; `brace-expansion` fans out to 1.1.18 / 2.1.4 / 5.0.9.** pnpm 11 reads neither `pnpm.overrides` nor the top-level `overrides`. The install exits 0 — nothing turns red, the pin is simply gone |
| 5 | settings moved to `pnpm-workspace.yaml`, `pnpm` block removed | 10.14.0 **and** 11.24.0 | pins hold under both; the WARN disappears |

A note on arm 4, because it is easy to measure this wrong: pnpm >= 10 self-manages its version — any `pnpm`, `pnpm dlx pnpm@11 ...` or `npx pnpm@11 ...` inside the repo re-executes as the `packageManager`-pinned 10.14.0. Two of our first attempts silently measured 10.14 while claiming to run 11; only bumping `packageManager` in the scratch copy produced a genuine pnpm 11 run.

[...]

Issue #501, opened 2026-08-20
package-lock.json does not carry the brace-expansion pin from 164b06a: npm does not read pnpm.overrides


DocTpoint:
## Summary

The `overrides` added in 164b06a are declared under the `pnpm` key in `package.json`. npm does not read that key, and `package.json` has no top-level `overrides`. The two committed lockfiles therefore disagree, and `package-lock.json` — the one `release.yml` installs from — still resolves `brace-expansion` to three versions that fall inside two HIGH advisory ranges.

`npm audit` reports this on `main` today as **1 high severity vulnerability** across five paths.

All affected entries are `dev: true`. They do not ship in `main.js`. The impact is on the build environment and on the alert count, not on plugin users. It matters mainly because the npm tree is the one Obsidian's own release check reproduces — `.npmrc` records that it runs `npm install`.

## What is in the two lockfiles today

| lockfile | brace-expansion | fast-uri |
|---|---|---|
| `pnpm-lock.yaml` | `5.0.9` only | `3.1.5` |
| `package-lock.json` | `5.0.7`, `2.1.2`, `1.1.16` | `3.1.5` |

To reproduce:

```bash
node -e "const d=require('./package-lock.json');for(const [k,p] of Object.entries(d.packages))if(k.endsWith('node_modules/brace-expansion'))console.log(p.version,k)"
```

The npm-side versions are identical before, at, and after 164b06a — I compared `164b06a^`, `164b06a` and `main`, and all three report `1.1.16, 2.1.2, 5.0.7`. The commit did modify `package-lock.json`, but these entries were not moved by it.

Advisory overlap for those three versions:

| version | GHSA-rgw5-rvv9-x895 (HIGH) | GHSA-mh99-v99m-4gvg (HIGH) |
|---|---|---|
| `5.0.7` | affected (fixed in 5.0.9) | affected (fixed in 5.0.8) |
| `2.1.2` | affected (fixed in 2.1.4) | affected (fixed in 2.1.3) |
| `1.1.16` | affected (fixed in 1.1.18) | affected (fixed in 1.1.17) |

`fast-uri` is unaffected: its range resolves to `3.1.5` on both sides regardless of the override.

The comment in `pr-ci.yml` already records that the pnpm and npm override formats are incompatible. This report is about the consequence that reached the committed npm lockfile.

## Second part: pnpm 11 stops reading the key as well

pnpm currently prints this on every command in the repository:

```
[...]

PR #692, opened 2026-09-12
fix(deps): regenerate package-lock.json for the vitest 5 bump (2 Dependabot alerts)

green-dalii:
Regenerates `package-lock.json`, which is what Dependabot's two open moderate alerts are actually about:

```
vitest           >= 2.1.0, < 4.1.11   ->  package-lock.json (development)
@vitest/mocker   >= 2.1.0, < 4.1.11   ->  package-lock.json (development)
```

**How it got stale.** #652 moved vitest to 5.0.0 in `package.json` and `pnpm-lock.yaml` and left `package-lock.json` on 4.1.10. This is the dual-lockfile drift class #501 and #556 already documented once: npm does not read pnpm's lockfile, and nothing keeps the two in step between releases.

**Why CI could not catch it.** Gate 1 installs with `pnpm install --frozen-lockfile`, and the audit step added in #637 runs `pnpm audit --audit-level high`. Neither reads `package-lock.json`, so the npm-side lockfile has no gate at all between releases — the release flow regenerates it at Step 3b, which is why the drift only surfaces when Dependabot files against the stale copy. I am not closing that gap here; it wants its own decision (an npm-side consistency check in CI, or dropping the second lockfile).

## What changed

`npm install --legacy-peer-deps --package-lock-only`, the command AGENTS.md records. 81 insertions, 163 deletions, and every one of the 11 changed packages is **dev scope**:

| Package | From | To |
|---|---|---|
| vitest | 4.1.10 | 5.0.0 |
| @vitest/mocker, @vitest/spy | 4.1.10 | 5.0.0 |
| tinybench | 2.9.0 | 6.1.4 |
| magic-string | 0.30.21 | 1.3.1 |
| tinyexec, picomatch, es-module-lexer, @jridgewell/sourcemap-codec | — | within-range patch/minor |
| @vitest/{expect,pretty-format,runner,snapshot,utils}, tinyrainbow, pathe, convert-source-map | — | removed (no longer needed) |

**Production-dependency changes: 0.** Verified by reading the `dev` flag on every changed entry rather than assuming.

## Verification

- `npm audit` → 0 vulnerabilities; `pnpm audit` → no known vulnerabilities (both lockfiles agree again)
- `npm install --legacy-peer-deps --dry-run` resolves cleanly from the new lockfile — the step `release.yml` performs
- `package-lock.json` vitest 5.0.0 matches `package.json` devDependency 5.0.0

CI will pass on this PR without validating it, which is the same blind spot described above: pnpm does not read this file. The checks above are the ones that matter for it.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -30,7 +30,7 @@
         "tslib": "2.4.0",
         "typescript": "5.9.3",
         "vite": "^8.2.1",
-        "vitest": "4.1.10",
+        "vitest": "5.0.0",
         "yaml": "^2.4.2"
       },
       "engines": {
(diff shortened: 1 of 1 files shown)


DocTpoint (approved):
Approving `4a69a60`. Recomputed rather than read: diffed the `packages` maps of both lockfiles — 20 changed entries, every one carries the `dev` flag, none without. `vitest` is 5.0.0 in `package.json`, `pnpm-lock.yaml` and now `package-lock.json`. Agree the CI gap (nothing reads this file between releases) is its own decision.

Issue #657, opened 2026-09-08
docs/feat(pdf): document data residency per conversion path, and an EU-hosted OCR option for the roadmap


DocTpoint:
**Is your feature request related to a problem? Please describe.**

The README calls the plugin GDPR-friendly, and for the LLM side that holds: a user can pick an EU-hosted provider or stay fully local. The document side has no such option. `markdownConversionBackend` offers two paths, and both send the document to a jurisdiction the user cannot choose:

- `native` sends the PDF as a file part to the chosen LLM provider (Anthropic / OpenAI / Gemini / Bedrock, all US).
- `mineru` sends it to `mineru.net/api/v4`, operated by OpenDataLab (Shanghai AI Lab) on Aliyun infrastructure. I could not find a privacy policy, a retention statement, or a DPA for that service, neither in the API docs nor elsewhere. Self-hosting is on the roadmap (#404) but not shipped.

For users in the EU that is a legal question, not a preference. There is no adequacy decision for China, so a transfer needs "appropriate safeguards" under Art. 46 GDPR, and a plugin user has no contract with OpenDataLab through which to establish them. Three supervisory authorities said exactly this about a Chinese AI service in 2025:

- Italy's Garante blocked DeepSeek on 30 Jan 2025: data transferred to and stored in China does not enjoy the same protection as under the GDPR.
- The Berlin DPA reported DeepSeek to Apple and Google as unlawful content on 27 Jun 2025, stating that the EU has issued no adequacy decision for China and that Chinese authorities have far-reaching access rights to personal data within the reach of Chinese companies.
- CNIL (France) and the DPC (Ireland) opened information requests in Jan/Feb 2025.

None of this is about MinerU's quality or about OpenDataLab's conduct. It is about a practice, a clinic, or a law firm in the EU not being able to use the built-in document path at all today, while the local Apple Silicon path is out of reach for most of them. I don't need this myself, my vault is Markdown notes without personal data. I'm raising it because the README makes the claim and the audience exists.

**Describe the solution you'd like**

Two things, in increasing size.

1. *Minimum:* state the data residency of each path in `docs/PDF-OCR-GUIDE.md` and in the backend setting's description. Today the guide says "privacy-sensitive users: self-host MinerU" and nothing about where the cloud path goes. One line per path ("native: your LLM provider's jurisdiction · mineru: China, no published retention policy") lets an EU user decide before the first upload.

[...]

PR #695, opened 2026-09-12
docs(memory): a close keyword in prose closes a PR, not just an issue

### LINKED ISSUE #592: Fix Dead Links clobbers existing wikilink display text with raw slugs

green-dalii:
Records the 2026-09-12 incident where merging one PR closed another.

`### Issue close keyword` in MEMORY.md covered one direction: use `Closes`/`Fixes` when you want auto-close, not `Refs`/`See`. This adds the inverse, which nothing had written down.

PR #689's body described the milestone in prose — "#604 sits on `v1.28.0 MINOR` while its **fix #684** sits on the PATCH line". GitHub parsed `fix #684` as a closing directive, so merging #689 closed **PR #684** one second later. `ClosedEvent.closer` = PullRequest #689. Author never touched it, and the ROI board merged in that same PR lists #684 as its second unblock-first item.

Four rules, plus the mechanism for reading the evidence: a `closed` event carries `commit_id` when a commit message did it and leaves it empty for an API/PR-level close — that distinction is what ruled out a commit keyword and pointed at the merge one second earlier.

**AGENTS.md gets the same scan as the first step of the mandatory merge sequence.** In MEMORY.md it is a rule to remember; in the merge sequence it is a step that runs.

Scope: documentation only, no code. Open PRs were scanned — the other keyword matches (`Fixes #592` / `#597` / `#672` / `#666` / `#688` / `#678` / `#657`) all point at issues, which is their purpose. #684 was the only one aimed at a PR by accident, and it has been reopened.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -158,9 +158,17 @@ For contributor PRs that need rebase after base-branch move: use `gh pr update-b
 **Mandatory merge sequence (added 2026-08-18):** every merge MUST follow this exact order. Skipping any step is a procedural miss even when content review passed:
 
 ```
+gh pr view <N> --json title,body --jq '.title, .body' | grep -inE '\b(closes?|closed|fix(es|ed)?|resolve[sd]?)\s+#[0-9]+'   # ← scan FIRST; every hit must be an issue you mean to close
 gh pr review <N> --body "<file>"   # ← MANDATORY. Formal review event lands on the PR.
 gh pr merge <N> --admin --squash --delete-branch   # ← ONLY after user said "merge it"
 ```
+
+- **Scan the body before merging.** A closing keyword needs no intent and does
+  not have to aim at an issue. Writing the fix reference as prose — "its fix
+  #684" — parsed as `fix #684` and closed **PR #684** one second after the PR
+  carrying that sentence merged. The author never touched it, and the ROI
+  board merged in that same PR listed #684 as an unblock-first item. Recover
+  with `gh pr reopen <N>`; see MEMORY.md §"Issue close keyword".
 
 - **`gh pr review --approve` (or `--request-changes`) MUST be posted BEFORE `gh pr merge`.** This lands the formal review verdict on the PR timeline; downstream tooling (release notes, contributor credit, audit trail) reads from that event, not from comments.
 - For architect-level contributor PRs (e.g. @DocTpoint), per [[feedback_reply_brevity_for_architect_contributors]]: review body should be **decision + ≤5 sentences** + concrete `file:line` findings, not a long-form audit.
--- a/MEMORY.md
+++ b/MEMORY.md
@@ -144,6 +144,35 @@ auto-close. Squash-merge must preserve the keyword in the squash commit
 body. Verified 2026-08-07: the v1.26.0 MINOR batch had 5 state-drift issues
 because PRs #401/#406/#410/#411 used non-closing keywords.
 
+**The inverse hazard — a keyword in prose closes a PR (2026-09-12):** the
+keyword does not have to be deliberate and it does not have to aim at an
+issue. PR #689's body described the milestone in prose:
+
+> #604 sits on `v1.28.0 MINOR` while its **fix #684** sits on the PATCH line
+
+GitHub parsed `fix #684` as a closing directive ⇒ merging #689 **closed PR
+#684** one second after it landed. `ClosedEvent.closer` = PullRequest #689;
+the author had not touched it, and the ROI board merged in that same PR
+listed #684 as its second unblock-first item. A closed PR reads to the
+contributor as *rejected*. Rules that follow:
+
+- Never let `close[sd]?` / `fix(es|ed)?` / `resolve[sd]?` sit immediately
+  before `#N` in prose. Write "the fix for #604 is PR #684" — the article
+  breaks the match — or just "PR #684".
+- **The target decides the damage.** A keyword aimed at an issue closes the
+  issue (intended). Aimed at a PR it closes the PR.
+- **Before merging, scan the title + body and confirm every hit is an issue
+  you mean to close:**
+  `gh pr view <N> --json title,body --jq '.title, .body' | grep -inE '\b(closes?|closed|fix(es|ed)?|resolve[sd]?)\s+#[0-9]+'`
+  Open PRs at the time of writing are clean — the other matches
+  (`Fixes #592` / `#597` / `#672` / `#666` / `#688` / `#678` / `#657`) all
+  point at issues.
+- **Recover with `gh pr reopen <N>`**, and read the closer before assuming
+  the mechanism: a `closed` event carries `commit_id` when a commit message
+  did it, and leaves it empty for an API/PR-level close. This one was
+  `actor=green-dalii, commit=-`, which ruled out a commit keyword and pointed
+  at the merge that had happened one second earlier.
+
 ---
 
 ## Key design decisions (canonical references)


DocTpoint (approved):
Approving `135ae01`. I reproduced the mechanism from the timeline (`closed` at 07:43:22Z, one second after #689 merged, `actor=green-dalii`, `commit_id=null`) and the entry matches it. Four notes, none of them blocking.

**It is documented behaviour, not a quirk.** GitHub's linking page states it: "If you use a keyword to reference a pull request comment in another pull request, the pull requests will be linked. Merging the referencing pull request also closes the referenced pull request." Citing that line turns the entry from an incident report into a rule, and it explains why everything nearby ("closing keywords close issues") reads as if this could not happen.

**The JSON field that looks authoritative does not see it.** `gh pr view 689 --json closingIssuesReferences` returns `[]` — the field holds issues only, so a PR-level link is invisible there. That is the strongest argument for keeping a text scan in the merge sequence, and worth one line so a later reader does not replace the grep with the field and get a clean answer.

**Two accepted syntaxes slip past the expression.** The docs allow a colon ("The keywords can be followed by colons... `Closes: #NN`") and the cross-repository form (`KEYWORD OWNER/REPOSITORY#ISSUE-NUMBER`). `\s+#` matches neither; I tested both against the published command and they come back clean. This covers them:

```
grep -inE '(closes?|closed|fix(es|ed)?|resolve[sd]?)[: ]*([-A-Za-z0-9_.]+/[-A-Za-z0-9_.]+)?#[0-9]+'
```

**One occurrence is live in this PR's own body.** The sentence quoting #689 carries the keyword directly before the number, unescaped — the construct the entry forbids. Nothing happens today because 684 is merged, but the body becomes the squash commit message, so the safe form you prescribe ("the fix for #604 is PR 684", article between) is worth the one word here too.

PR #696, opened 2026-09-12
docs(memory): read existing PR signals before submitting a review

green-dalii:
Records the 2026-09-12 incident where a review contradicted an already-decided PR.

**What happened.** #570 had been labelled `wontfix` two weeks earlier with a detailed maintainer decision comment ("we're declining this and not scheduling it for future work"). A review was submitted anyway with `CHANGES_REQUESTED`. That verdict reads as "fix these and we'll merge" — the opposite of the recorded decision — and would have told the contributor to do work that was explicitly declined.

**The rule.** Before any review submission (approve, request changes, or comment), check the PR's labels and the most recent maintainer comment. A `wontfix` / `duplicate` / `do not merge` label, or a prior decision comment, means stop and post a short comment asking instead.

**Recovered by:**
- Dismissing the review (`PUT .../reviews/<id>/dismissals`)
- Posting a one-paragraph correction comment naming the PR's actual state
- Converting the PR to draft via GraphQL `convertPullRequestToDraft` — REST's `draft` field is GitHub-App-only, and `gh` has no `pr draft` subcommand

**Where it lives:**
- `MEMORY.md` §"Review existing PR signals first" — the rule + the #570 incident + the recovery procedure
- `AGENTS.md` "Reviewing PRs" — mirrors the merge-sequence shape, with the check command as the first step

No code changes. Doc-only.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -175,6 +175,19 @@ gh pr merge <N> --admin --squash --delete-branch   # ← ONLY after user said "m
 - **Post-merge audit trail:** if a `gh pr merge` was executed without the matching `--approve` event (procedural miss, not content miss), immediately post `gh pr comment <N> --body <audit-note>` recording what was skipped. Don't rebase, don't re-merge, don't amend — the merge commit hash stands; only the audit trail is patched. Incident reference: PR #478 (2026-08-18, merge `2806d24`).
 - Anti-pattern: "`gh pr merge --admin` doesn't enforce reviews, so I can skip --approve." Wrong — `--admin` bypasses the **requirement** rule, not the **review event** rule. Two separate audit surfaces.
 
+**Reviewing PRs (added 2026-09-12, after #570):** a PR decided against is **closed** once the contributor has had a fair window to answer (roughly two weeks of silence), with the decision comment naming in one line what would reopen it. An open PR states that merging is still possible — a closed one cannot collect a stray review at all.
+
+Every review submission MUST begin with the signal check:
+
+```
+gh pr view <N> --json labels,comments --jq '{labels: [.labels[].name], last_comments: [.comments[] | "\(.author.login) \(.createdAt[0:10]): \(.body[0:120])"][-3:]}'   # ← check FIRST
+gh pr review <N> --body "<file>"   # ← ONLY when the check above returned no signal
+```
+
+Two ways to get that check wrong, both caught in review of its first version: `.[0:1]` returns the **oldest** comment rather than the latest — decisions arrive late, so it hides exactly what it is looking for — and filtering on one account's login makes a decision written by any other maintainer invisible. Take the last few comments from **any** author.
+
+If a signal is present, post a short comment asking instead of a review. To undo a misfired review: dismiss it (`PUT .../pulls/<N>/reviews/<id>/dismissals`) and, if the PR is `wontfix`, convert it to draft via `convertPullRequestToDraft` (REST's `draft` field is GitHub-App-only; `gh` has no `pr draft`). MEMORY.md §"Declined PRs are closed, not left open".
+
 ---
 
 ## 📦 Development Workflow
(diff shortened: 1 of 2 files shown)


DocTpoint (changes requested):
Changes requested on `7f32304` — the rule is right and the check does not implement it. One line, plus two things worth settling before this merges.

**The command returns the oldest maintainer comment, not the most recent.** `[.comments[] | select(...) | .body[0:200]] | .[0:1]` slices from the front. On #570 that happens to be the decision, because the decision was the first thing written on the thread — so the check looks correct on the one case it was built against. On a thread where the maintainer answered early and decided late, it returns the greeting and hides the decision, which is precisely the situation the rule exists for. Verified on #570: `.[0:1]` gives the 2026-08-29 comment, `.[-1:]` gives the 2026-09-12 one. `.[-1:]` is the minimal fix; scanning every maintainer comment for the signal instead of any single one would be the robust version.

**The check keys on the literal `green-dalii`.** Under the GD4AI org a decision recorded by anyone else — including me — is invisible to it, and a rule that reads one account will quietly stop matching the moment a second person writes a decision. Dropping the `select` and looking at all comments, or matching on the label alone, avoids hard-coding who is allowed to decide.

**The entry should say how long a declined PR stays open, and the answer is: not long.** An open pull request states that merging is still possible, which is exactly what the decision on #570 says is not the case. That PR has now carried three signals of the same thing — a `wontfix` label, a decision comment, and draft status — while remaining open, and none of them is the one place every tool and every reader agrees on. A closed PR cannot collect a stray review at all, so closing does more for this rule than any pre-review check can, and it stops leaving the contributor in a state that reads as "still under consideration".

Nothing is lost by closing: the branch, the diff and the whole thread stay readable and linkable, which is all "preserved as a reference implementation" requires. My suggestion for the entry: a declined PR is closed once the contributor has had a fair window to answer the decision — #570's decision landed 2026-08-29 with no reply since, so two weeks has already passed. And since a close is "not now" rather than "never", the decision comment is the right place to name what would reopen it, in one line.

green-dalii:
All three taken. The third is the one that changed the entry's shape rather than its wording.

**The command was wrong in the direction that hides the signal.** `.[0:1]` slices from the front, and on #570 the decision happened to be the first comment on the thread — so the check was validated against the single case where its bug is invisible. Verified both forms on #570: `.[0:1]` returns the 2026-08-29 decision, and on a thread that opens with a greeting and decides later it would return the greeting. New form takes `[-3:]`.

**And it keyed on one account.** Under GD4AI that makes a decision of yours invisible, which is worse than a style problem — the rule would keep passing while silently not covering the people most likely to write one. Now it takes the last few comments from any author. The label is listed first in the output, which on #570 would have stopped me before reading anything.

**"Close it" is the actual fix, and I had it backwards.** I wrote a pre-review check for a state that should not persist: an open PR asserts that merging is possible, which is the one thing a `wontfix` decision denies. Three signals on #570 — label, decision comment, draft — did not stop a review from landing, which is the evidence that none of them is the surface every tool and reader agrees on. A closed PR cannot collect a stray review at all.

The entry now leads with that: close once the contributor has had a fair window to answer (roughly two weeks; #570's decision is from 2026-08-29 and has drawn no reply), and name in the decision comment the one line saying what would reopen it, so a close reads as "not now" rather than "never". The pre-review check stays for the window in between, with both corrections.

Nothing in the entry claims #570 is closed — that is a maintainer call on a contributor's PR and I am not making it unilaterally. I have put the recommendation to @green-dalii and will act on it.

DocTpoint (approved):
Approve — `0d04fdb`.

All three points are in, and the third goes further than I suggested. The check now takes the last three comments instead of `.[0:1]`, and the `green-dalii` filter is gone, so a decision written by anyone is visible. Naming both as failure modes in the text is better than fixing them silently — the next person writing a `jq` check over comments meets the two traps before making them.

The reframing is the right shape. The close is the rule, and the pre-review check is the bridge for the window before it. #570 is now closed, draft, `wontfix` — the rule applied to the case it came from.

Gate 1 green on `0d04fdb`, and the `MEMORY.md` conflict with #695 is gone.

PR #697, opened 2026-09-12
test(lint): drop SourceAnalysis.contradictions from six fixtures (#684 follow-up)

green-dalii:
**Main is currently red.** #696's Gate 1 surfaced a `TS2353` error in `wiki-engine-summary-head.test.ts:38`: `Object literal may only specify known properties, and 'contradictions' does not exist in type 'SourceAnalysis'`. That is the regression — and it's mine.

## What happened

PR #681 added a `SourceAnalysis` test fixture with `contradictions: []` in six files. CI was green because at that commit the field existed.

PR #684 removed the field from `src/types.ts:67` (`-  contradictions: ContradictionInfo[]`) along with its dead path (`review_ok` loop, `resolveContradiction`, `clampPageSections` — the writer's only outputs were `'resolved'` and `'pending_fix'`, never `'review_ok'`). CI was green at merge time because the field still existed on `main` until the moment #684 landed. The fix passed review because the deletion's internal consistency checked out, but **the review did not run `git show <base>:src/types.ts` against the dependent test fixtures** to confirm the field's removal was matched in every fixture that referenced it.

## What this PR does

Six one-line removals from test fixtures. No code changes, no test logic changes:

```
src/__tests__/wiki/wiki-engine-cancel-cleanup.test.ts      line 31
src/__tests__/wiki/wiki-engine-summary-tags.test.ts         line 44
src/__tests__/wiki/wiki-engine-summary-head.test.ts         line 38
src/__tests__/wiki/wiki-engine-contradiction-count.test.ts  line 46
src/__tests__/wiki/wiki-engine-link-repoint.test.ts         line 28
src/__tests__/wiki/wiki-engine-summary-domains.test.ts      line 34
```

The triage-lane behaviour `wiki-engine-contradiction-count.test.ts` covers is unaffected — the test only inspects `report?.contradictionsFound` from the merge-triage JSON, which goes through a different code path that #684 did not touch. The fixture's `contradictions: []` was filler for the *extraction* JSON.

## Why this couldn't wait

Main is red. Every other open PR's CI is blocked on this. The fix is six mechanical deletions with Gate 1 already green (lint 0 / tsc 0 / build / **4098 tests / 290 files** / css-lint 0) and a small, honest commit message that names the review miss. I'd like to merge it as soon as the CI on this PR goes green, with explicit approval.

### DIFF
--- a/src/__tests__/wiki/wiki-engine-cancel-cleanup.test.ts
+++ b/src/__tests__/wiki/wiki-engine-cancel-cleanup.test.ts
@@ -28,7 +28,6 @@ const ANALYSIS_RESPONSE = JSON.stringify({
   summary: 'Acceptance and Commitment Therapy.',
   entities: [{ name: 'Steven Hayes', type: 'person', summary: 'founder', mentions_in_source: [] }],
   concepts: [{ name: 'Psychological Flexibility', summary: 'core construct', mentions_in_source: [] }],
-  contradictions: [],
   related_pages: [],
   key_points: [],
 });
--- a/src/__tests__/wiki/wiki-engine-contradiction-count.test.ts
+++ b/src/__tests__/wiki/wiki-engine-contradiction-count.test.ts
@@ -43,7 +43,6 @@ const EXTRACTION = JSON.stringify({
     { name: 'ATP', type: 'other', summary: 'Energy carrier; CCO may not be the primary photoacceptor.', mentions_in_source: ['CCO sei nicht der primäre Photoakzeptor'] },
   ],
   concepts: [],
-  contradictions: [],
   related_pages: [],
 });
 
--- a/src/__tests__/wiki/wiki-engine-link-repoint.test.ts
+++ b/src/__tests__/wiki/wiki-engine-link-repoint.test.ts
@@ -25,7 +25,6 @@ const ANALYSIS = JSON.stringify({
     { name: 'Phosphocreatin', type: 'person', summary: 'y', mentions_in_source: [] },
   ],
   concepts: [],
-  contradictions: [],
   related_pages: [],
   key_points: [],
 });
--- a/src/__tests__/wiki/wiki-engine-summary-domains.test.ts
+++ b/src/__tests__/wiki/wiki-engine-summary-domains.test.ts
@@ -31,7 +31,6 @@ function makeAnalysis(): SourceAnalysis {
     summary: 'Iron store.',
     entities: [],
     concepts: [],
-    contradictions: [],
     related_pages: [],
     key_points: [],
     created_pages: [],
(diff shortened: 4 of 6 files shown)


green-dalii:
Maintainer bypass merge, with the Gate 4 assessment on the record.

**No breaking changes.** Zero production files. The diff is six pure deletions of `contradictions: [],` from `SourceAnalysis` test fixtures — no API signature, settings schema, file format, command ID, or Obsidian API surface is touched. The field itself was already removed from the type by #684; this PR only brings the fixtures in line with it.

**No side effects.** No production code reads `.contradictions` (`grep` over `src/` excluding tests returns nothing), so the six removed keys were runtime-inert. Proven rather than asserted: the same suite run on both sides gives **identical results** — 4098 passed / 290 files on `a0b3f189` (main, after `pnpm build`) and 4098 passed / 290 files on this branch. The only observable difference is the TS2353 error, which this PR removes.

**No performance regression.** No production path is touched, so CPU / memory / IO / network / token are all unaffected (N/A). Test-suite wall time is ~10s on both sides.

```
| Dim     | Status | Notes |
|---------|--------|-------|
| CPU     | N/A    | test fixtures only, no production code |
| Memory  | N/A    | same |
| IO      | N/A    | same |
| Network | N/A    | same |
| Token   | N/A    | same |
```

Merging on the Maintainer bypass path because main is red and every other open PR's CI is blocked behind it. DocTpoint is still the reviewer of record for the follow-up.

DocTpoint (approved):
Approving `ea9eedc`. Merge it as soon as you read this — main being red blocks every other branch, and I verified the set is exactly right rather than taking the six on faith.

On current main, `contradictions: []` appears in exactly those six files and nowhere else under `src/`, so there is no seventh fixture waiting to turn Gate 1 red again after this lands. Nothing outside the tests reads the removed field either: the remaining `contradictions:` hits are the frontmatter marker in `contradicted-marker.ts` and the i18n strings, which are a different thing that #684 never touched. Your read of `wiki-engine-contradiction-count.test.ts` holds — it asserts `report?.contradictionsFound` from the merge-triage JSON at lines 70-80, so dropping the extraction fixture's filler does not weaken it.

One correction to the attribution: both halves are mine. #681 wrote `contradictions: []` into the six fixtures and #684 removed the field from `types.ts`, so the miss is on my side of the two PRs, not on your review of them.

The shape is worth one line in its own right, because no per-PR gate could have caught it: each PR was green against the main it branched from, and the break exists only in the pair. A type field removed in one PR and referenced in fixtures merged by another is invisible until both land — the cheap guard is to re-run Gate 1 on the merge result, not on the branch head, when a PR deletes from `types.ts`.

PR #700, opened 2026-09-13
docs: correct the require_last_push_approval value in AGENTS.md

green-dalii:
`AGENTS.md` §"Gate 1 CI" claimed the branch protection rule sets `require_last_push_approval: true`. It sets `false`.

Not a harmless typo. The flag answers "does my last push invalidate the review I already have?" — and the doc says yes while the rule says no, so a maintainer following it waits for an approval nobody asked for. `strict: false` in the same parenthesis is correct.

Found while reviewing #698, which rewrote that sentence. I flagged it as non-blocking so it would not hold up the trigger change; it merged without the correction, so it lands here as a one-line fix.

While in there I checked the other ruleset values the doc depends on against the API — `required_approving_review_count: 1`, `strict_required_status_checks_policy: false`, `bypass_actors: []` — all accurate, and the line's claim about which check is required matches.

Docs only, one line, no code.

Refs #698.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -188,7 +188,7 @@ pnpm lint && pnpm test && pnpm typecheck && pnpm build && pnpm css-lint   # WRON
 
 ### Gate 1 CI (added 2026-08-18)
 
-`.github/workflows/pr-ci.yml` runs the full Five-Gate on every PR to `main` and on every push to `main` (the second trigger added 2026-09-12: a PR is green against the base it branched from, so a break that exists only in the merged pair is invisible until a later PR reports it). Status check: `Gate 1 / Five-Gate`. Branch protection requires it (`strict: false`, `require_last_push_approval: true`).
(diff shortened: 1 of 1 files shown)


DocTpoint (approved):
Approve — `aa207bf`.

I checked the value against the live ruleset rather than against the doc:

```
gh api repos/GD4AI/obsidian-llm-wiki/rules/branches/main
```

`require_last_push_approval: false`. Your correction is right. The other values in that sentence hold as well: `strict_required_status_checks_policy: false`, and the required check context is exactly `Gate 1 / Five-Gate`.

Thank you for catching this while reviewing #698. I rewrote that sentence and carried the parenthesis over without re-checking it.

One additive suggestion, non-blocking and outside the scope of this PR. The parenthesis now names a flag that has no effect on an existing review. The flag that does have an effect is not in `AGENTS.md` at all: `dismiss_stale_reviews_on_push` is `true`, so any push discards an existing *approving* review. A `CHANGES_REQUESTED` survives a push. We measured this on #654 and #655 — a push that restored the already-approved commit still dismissed the approval. A reader of the corrected line may conclude that pushing is free for a review that is already in. The same parenthesis has room for `dismiss_stale_reviews_on_push: true` if you want it. I am happy to open a follow-up PR, or it fits here in one line.

Issue #582, opened 2026-08-30
Cancelling an ingest leaves the source marked as completed — every later trigger skips it


DocTpoint:
Observed on a live vault: an ingest was cancelled mid-run (Stage 3, after the summary page was written). The result is a state the plugin cannot see or repair on its own:

- The summary page under `sources/` was already written in Stage 2 — and `isAlreadyIngested` checks exactly that page's existence. The cancelled, half-finished ingest therefore reads as **completed**, and every later trigger (batch or manual) skips the source.
- Stage 6 never ran, so `log.md` and `index.md` know nothing about the pages that were written. The graph and the log silently disagree, and nothing flags it.

The cancel path (`AbortError` branch in `ingestSource`) shows the "cancelled" notice and returns — it does no cleanup, so cancelling is not neutral: it poisons the resumption. The only manual repair is knowing to delete the summary page by hand.

Suggested fix: on cancellation, trash the Stage-2 summary page. Its absence re-arms the existing re-ingest mechanism, so the next trigger ingests the source cleanly; content pages written before the cancel can stay, since a re-ingest merges them additively. PR follows.

PR #701, opened 2026-09-13
feat(ingest): stamp a wiki-ingested marker on sources and skip re-ingesting marked ones

weqoocu:
## What

Three related pieces, one feature: a durable per-source "this note is fully in the wiki" record.

1. **Marker write.** A successful `ingestSource` stamps the source note with `wiki-ingested: <today>` — the *last* write of the run, after pages, index and log have landed. md sources only. A marker write failure is logged and a localized Notice shown, but never fails the run (same best-effort contract as the index write: the pages are already on disk).
2. **Marker read.** `checkRequirements` skips a source whose frontmatter already carries the marker (cheapest check first — before hashing). New `skipWikiIngested` setting, default on, `undefined = on` (`!== false` at the use site), matching the `createStubsForUnresolvableLinks` convention. Force-reingest paths bypass the gate unchanged.
3. **Startup self-heal.** Sources pages under `wiki/sources` carry a `source_file` back-reference (wikilink or bare path). On layout-ready the plugin re-stamps any referenced note whose marker went missing — lost writes from older versions, crashes between page write and marker write. One unreadable page never blocks the sweep.

## Why

The existing dedup is content-hash based, which is stateless against a *reset* of that state: a folder re-run after the batch cache restarts, or a user re-adding the same note body under a different name, re-ingests and re-merges everything. A frontmatter marker on the source itself is durable, user-visible, and removable by hand when a re-ingest is actually wanted.

The self-heal closes the gap the marker itself opens: a run that lands pages but loses the marker write would otherwise be re-ingested forever.

## i18n

Two new keys (`wikiIngestedMarkerFailed`, `wikiIngestedMarkerRepaired`) added to all 11 locale files in lockstep; en/zh reviewed, others are straight translations.

## Tests

`src/__tests__/wiki/wiki-engine-wiki-ingested-marker.test.ts`:

| Case | Asserts |
|---|---|
| successful ingest stamps the source | frontmatter carries `wiki-ingested: YYYY-MM-DD`; report success |
| marked source is skipped | `skipped: true`, zero pages, zero LLM calls |
| `skipWikiIngested: false` re-ingests | gate bypassed, pages created |

The harness `fileManager` mock gained `processFrontMatter` (parse -> mutate -> serialize, matching Obsidian's read-modify-write semantics) — without it any frontmatter-writing feature is untestable in harness.

## Gate 1

lint 0 / tsc 0 / build clean / 4101 tests, 291 files / css-lint 0.


green-dalii:
Deferring this to the next MINOR rather than this PATCH, and naming the reason, because it is not about the code quality — the marker read, the startup self-heal and the best-effort write contract are all thought through, and the `!== false` default matches the existing convention.

**It writes to the user's source notes, and the README promises it does not.**

`README.md:114`, in Quick Start — the first thing a user reads:

> That's it. The plugin modifies nothing in your original notes — only creates new pages under `wiki/`.

Stamping `wiki-ingested: <today>` into the source note is a write to an original note. With `skipWikiIngested` defaulting to on, that happens without the user asking, and the startup self-heal re-stamps notes that lost the marker. All eleven locales carry that sentence, and merging this makes it false in every one of them.

I am not saying the feature is wrong. A durable per-source record is genuinely useful, and `generation_complete` + `contentHash` on the *generated page* is a weaker signal than a marker on the note. But the cost lands on a promise the project makes in its first screen, so it should be a deliberate, documented decision rather than a side effect of a PATCH:

- Either the write becomes **opt-in** (default off) with all eleven READMEs updated to describe it, or
- the record lives outside the note — plugin data keyed by path, which keeps the promise intact at the cost of not travelling with the vault.

Either is a design call worth its own PR and a MINOR window. Happy to discuss the second option if you would rather not touch user notes at all.

Labels and milestone left as-is; #711's fix (#712) is separate and is going in.

DocTpoint:
Not relitigating the deferral — this is about which of the two designs you land on, so it is better said before the MINOR window than after.

## The marker gate shadows the repair the automatic path relies on

The marker check lands ahead of the hashing:

    if (this.settings.skipWikiIngested !== false && this.hasWikiIngestedMarker(content)) {
      return { reason: 'duplicate', detail: 'wiki-ingested marker present' };
    }

The hash set it precedes is built only from `wiki/sources/*` frontmatter (`buildIngestedHashes`), and that is what makes removing a summary page re-arm an ingest. #582 names it as the user's only manual repair — "the only manual repair is knowing to delete the summary page by hand" — and #583's fix rests on it: the cancel path trashes the summary page because "its absence re-arms the existing re-ingest mechanism".

For a note whose earlier ingest completed, that stops holding. The summary page is gone, the marker is not, and the marker is read first.

## Interactive ingest keeps an escape; the automatic path does not

The rejection carries `reason: 'duplicate'`, so `onConfirmReingest` (`main.ts:132`) prompts and the user can proceed — but only when `interactive` is set, which is the single-file commands (`ingest-commands.ts:96`, `:120`). The automatic path passes no such flag (`auto-maintain.ts:244`), and `reportSkip` deliberately keeps folder and watcher runs quiet. There the note is skipped silently, and deleting the summary page no longer changes that.

One correction to the PR body: force-reingest is not an escape today. `forceReingest` is declared (`types.ts:704`) and read (`wiki-engine.ts:992`), and nothing in the repository sets it — `tools/` included.

## The shape of it, which the plugin-data option shares

The question is not where the record lives. It is that there would be two independent "already done" records, and the one that wins is the one a user cannot see from the vault. Today there is exactly one, it sits in `wiki/sources/`, and removing it is the repair. Two records can disagree, and recovery becomes order-dependent: clearing the marker by hand is undone by the startup self-heal for as long as the summary page still carries the `source_file` back-reference.

Keying the record to a path in plugin data keeps the second record and takes away the user's ability to see or clear it, so it does not resolve this.

If the marker stays, the part that fixes this independently of the opt-in question is precedence: consult the hash set first and let the marker be a fast path only while the summary page exists. Then there is still one record, and the marker is an index into it rather than a rival to it.

green-dalii:
Picking this up properly rather than leaving the deferral as a bare "next MINOR", because the question you raised is the right one and I think the answer is stronger than either of us said.

## Where does the fact "this note has been ingested" actually live?

It is a **derived** fact — an assertion about a transformation the plugin performed, not something true of the note by itself. So it can live in one of three places:

| Home | Owned by | Survives a plugin reset | Survives the user deleting the wiki |
|---|---|---|---|
| Plugin state (`data.json`) | plugin | ❌ | ❌ |
| **The wiki's own artifact** — `wiki/sources/*.md` carrying `contentHash` + `source_file` | plugin | ✅ | n/a |
| The user's source note (a frontmatter marker) | **the user** | ✅ | ✅ |

The middle row is not a proposal — it is what already ships. `buildIngestedHashes()` (`wiki-engine.ts:477`) is documented as *"Content hashes already present in the wiki, read from source-page frontmatter"*, and `wiki-engine.ts:1775` writes `contentHash` onto every sources page. `scanners.ts:485` already uses that same field for source-drift detection.

## Measuring the two scenarios the description cites

**"a folder re-run after the batch cache restarts"** — the batch cache is `ingestedHashesCache`, a TTL memoisation. When it expires, `buildIngestedHashes()` **recomputes the set from the wiki pages**. Nothing is lost, so nothing is re-ingested. The cache is not the record; the wiki is.

**"a user re-adding the same note body under a different name"** — already deduped. `hashBody` (`source-requirements.ts:77`) is FNV-1a over the trimmed, whitespace-normalised **body only** — no path, no filename, no frontmatter. The same body under any name hashes identically and is caught by `ingested.has(hash)`.

So both stated failures are handled today. I could not construct a case the marker fixes and the hash does not, which is why I would rather stop this than merge it.

## The two problems the marker introduces

**It writes a derived fact into the user's input.** The wiki is the plugin's output space; the note is the user's. That is a category error, and it is why `README.md:114` promises *"The plugin modifies nothing in your original notes"* in all eleven locales.

**Its failure mode is the user's main recovery action.** Deleting a `wiki/sources/` page — or the whole `wiki/` folder — is how someone forces a rebuild. The hash-based mechanism cooperates: remove the artifact, the fact is gone, the note is re-ingested. The marker actively opposes it: the note still says "done" while the page is gone, so the rebuild silently skips exactly the notes the user was trying to rebuild. A second source of truth does not stay in agreement, and here disagreeing is not neutral.

## What I would ask for

Either close this, or narrow it to a **specific, demonstrated** case the hash mechanism misses — with the observation, not the reasoning, that shows it. Any redesign should keep the fact on the wiki side; the marker's durability argument is already satisfied there, without touching a file the plugin does not own.

Not a judgement on the code, which is careful — the best-effort write contract, the startup self-heal and the bounded sweep are all well built. It is the premise I cannot get behind.

PR #702, opened 2026-09-13
fix(provenance): repair the footnote's brackets so the paragraph guard can see it

DocTpoint:
`paragraph-provenance.ts` decides which paragraph a rewrite may drop by reading the inline footnote that names its source. Its regex is label-agnostic on purpose, but it does need the footnote to be well formed — and the model frequently writes it otherwise.

I ran the guard's own `SOURCE_FOOTNOTE` regex over the shapes a rebuilt vault actually produced. Six of the seven are invisible to it:

| written form | seen by the guard |
|---|---|
| `^[L: [[X]]` — two closing brackets | no |
| `^[L: [[X]]]]` — four closing brackets | yes |
| `[^L: [[X]]]` — first two characters swapped | no |
| `^[[L: [[X]]]]` — opening bracket doubled | no |
| `^[[L: X]]` — doubled, name bare inside | no |
| `[[L: [[X]]]^^^` — caret trailing the link | no |
| `[[L: [[X]]]` — caret gone | no |

The first row is the common one: roughly a third of the markers. The doubled opener was one in five on a single run (155 of 791), and Obsidian renders it as a link to a page named `L: [[X` — one ghost node per source. A marker the guard cannot see is a paragraph with no owner, so the next rewrite another source triggers is free to drop it.

**Why not widen the guard's regex instead.** That would make it see the broken markers, but Obsidian would still render the ghost link and the page would still carry a footnote that does not resolve. Repairing the syntax fixes both, and the guard stays strict.

The semantic half — which paragraph, which source — is the model's judgement and is untouched. The syntactic half is machine-checkable, so `createOrUpdateFile` repairs it on the pass that already normalizes heading spacing for wiki content; log and schema writes are outside that branch.

**The label is carried through verbatim** — `Quelle:`, `Source:`, `来源:` all survive as written. Two shapes are ambiguous against ordinary content and are recognised narrowly: the nested form by its inner `[[`, which a wikilink to a colon-named page never has, and the bare form only when a caret trails it. `[[Study: Berberin 2021]]`, `[[COVID-19: Long Covid]]` and the mentions quote line `> **Quelle: [[sources/X|X]]**` are left alone, with or without a footnote behind them — all four pinned as tests.

One migration note: this repairs writes from here on. Pages already carrying a broken marker are fixed the next time something writes them.

`src/core/provenance-marker.ts` is a new pure function, 22 tests. The engine change is five lines.

Gate 1: lint 0 / tsc 0 / build OK / 4120 tests, 291 files / css-lint 0.

### DIFF
--- a/src/__tests__/core/provenance-marker.test.ts
+++ b/src/__tests__/core/provenance-marker.test.ts
@@ -1,0 +1,127 @@
+import { describe, it, expect } from 'vitest';
+import { normalizeProvenanceMarkers } from '../../core/provenance-marker';
+
+describe('normalizeProvenanceMarkers — bracket count', () => {
+  it('repairs the two-bracket form the model writes most often', () => {
+    expect(normalizeProvenanceMarkers('Satz. ^[Quelle: [[Big-Pharma-Bias]]'))
+      .toBe('Satz. ^[Quelle: [[Big-Pharma-Bias]]]');
+  });
+
+  it('trims a four-bracket overshoot down to three', () => {
+    expect(normalizeProvenanceMarkers('Satz. ^[Quelle: [[X]]]]'))
+      .toBe('Satz. ^[Quelle: [[X]]]');
+  });
+
+  it('keeps a correct marker unchanged and stays idempotent', () => {
+    const ok = 'Satz. ^[Quelle: [[Berberin]]] Weiter.';
+    expect(normalizeProvenanceMarkers(ok)).toBe(ok);
+    expect(normalizeProvenanceMarkers(normalizeProvenanceMarkers(ok))).toBe(ok);
+  });
+
+  it('preserves trailing punctuation after the marker', () => {
+    expect(normalizeProvenanceMarkers('… gesättigte Fette ^[Quelle: [[Big-Pharma-Bias]].'))
+      .toBe('… gesättigte Fette ^[Quelle: [[Big-Pharma-Bias]]].');
+  });
+
+  it('handles piped link labels and path targets', () => {
+    expect(normalizeProvenanceMarkers('S. ^[Quelle: [[Notizen/Big-Pharma-Bias|Big-Pharma-Bias]]'))
+      .toBe('S. ^[Quelle: [[Notizen/Big-Pharma-Bias|Big-Pharma-Bias]]]');
+  });
+
+  it('returns content without a wikilink unchanged (same reference)', () => {
+    const s = 'Kein Marker hier.';
+    expect(normalizeProvenanceMarkers(s)).toBe(s);
+  });
+});
+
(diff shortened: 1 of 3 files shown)


green-dalii (approved):
Approved. The repair is placed where the shape problem is, and the false-positive guards are the reason it is safe.

The call site is right: `wiki-engine.ts:1939`, next to `normalizeHeadingSpacing`, inside the `isInWikiContentFolder` guard — the gate every wiki-content write already passes through. Hooking a syntactic repair beside the existing one rather than at each producer is the correct altitude.

What I checked hardest is what it does *not* touch. `NESTED` requires the nested `[[`, and `TRAILING_CARET` requires the trailing caret, which is what keeps `[[Study: Berberin 2021]]` — an ordinary wikilink whose name contains a colon — out of the rewrite. Without those two requirements the patterns would eat real links, and both are asserted.

Two details I would have asked for and found already there: `normalizeProvenanceMarkers` returns the same reference when the content holds no `[[`, so a caller can detect a no-op without a comparison, and the label is carried through verbatim rather than re-derived. 22 cases cover the four malformed forms.

Observation only, no change asked: #603 says the single-write-gate contract does not currently hold across all six writers, so writers outside this gate can still emit malformed markers. This repair is correct for the paths that do pass through it, and when #603 closes it may want to move one level up.

DocTpoint:
Checked the observation against the six writers #603 names, and it does not have a path today — worth recording, because otherwise it argues for moving the repair later for no reason.

A marker is model prose in a page body, and model prose reaches the vault through the page factory, so through the gate. None of the six writes model prose into a body:

- `core/link-retarget.ts` replaces only the linkpath inside the existing brackets (`ref.original.replace('[[' + ref.link, …)`, line 171), so the bracket count survives: `^[<label>: [[Name]]]` becomes `^[<label>: [[Other]]]`, still canonical. It can neither create a malformed marker nor repair one.
- `core/sources-normalizer.ts` rewrites the `sources:` frontmatter field.
- `wiki/engine-internals/log-writer.ts` writes `log.md`.
- `schema/schema-manager.ts` and `schema/apply-suggestion.ts` write schema files.
- `wiki/lint/phases/preparation.ts` is the only one that touches a page body, and its single body edit is `fixDoubleNestedWikiLinks`, whose regex (`/\[\[\[\[([^\]|#]+)(?:[|#]([^\]]+))?\]\]\]\]/`) needs four opening brackets and therefore never matches `[[<label>: [[Name]]]`.

So the repair is complete at the altitude it has, and moving it up stays a cheap option rather than a gap being carried.

Separately: that last one turned out to carry a defect of its own, unrelated to markers — the page branch computes the repair and returns the original, so it is counted and logged but never written. Filed as #719 with a fix in #720.

PR #705, opened 2026-09-13
fix(page-factory): a body rewrite cut off at the token limit is not adopted

### LINKED ISSUE #704: A body rewrite that stops at max_tokens is written over the page — provenance-footnote loops grew one page from 32 KB to 127 KB
### What happened

On a 413-note rebuild (Gemma 4 26B via LM Studio, 1.27.1 + local patches; the paths below are unpatched upstream code) four pages ended up carrying a single line with 300–760 provenance footnotes back to back:

```
^[Quelle: [[Endendokrinologie]]]^ ^[Quelle: [[Endendokrinologie]]]^ … (650×) … ^[Quelle: [[Endendendokrinologie]]] ^[Quelle: [[Endendokrinendokrinologie]]]
```

The note is `Endokrinologie`; the name decays as the loop runs. `concepts/Dysbiose.md` went from 32 KB to 127 KB over five rewrites; `Insulinresistenz`, `EFSA`, `Darmbarriere` the same way. Across the vault 23 pages hold such a line; they account for 2463 of 2703 dead prose links (`rebuild-probe`: 1928 dead prose links vs. 435 on the previous full rebuild of the same notes).

### Mechanism, measured

The write audit shows 13 body rewrites that grew a page by more than 8 KB — all `mergePage` / `updateRelatedPage` LLM body rewrites. 10 of the 13 sit within four minutes of a `"finish_reason": "length"` in the LM Studio server log. The model looped on the footnote until `max_tokens` (`TOKENS_PAGE_GENERATION`) ran out, and the truncated completion was adopted as the page body.

The signal exists: #305 surfaces `finishReason` to any caller that passes `onFinish`. The three page-factory paths that hand a whole body to the model — `merge-body` and `reviewed-append` in `merge-page.ts`, `related-page` in `related-page.ts` — do not pass it, so a truncated rewrite is indistinguishable from a finished one. `preserveExistingSections` and the paragraph guard catch a body that shrank; nothing catches one that ballooned, and `repetition-loop.ts` only watches the analyzer's JSON.

### Proposed fix (PR follows)

On `finishReason === 'length'` do not adopt the rewrite: `mergePage`/`appendToReviewedPage` return without writing (the `NO_NEW_CONTENT` exit), `updateRelatedPage` takes the #131 exit (new source recorded in frontmatter, body verbatim). Only `'length'` counts — a client that reports nothing stays `'unknown'`, so legacy and mock clients keep today's behaviour.

Not in scope: `createNewPage` (a truncated new page has no previous body to keep — worth a separate look), and repairing the affected pages (the loop is one paragraph; the rest of the page is intact).

DocTpoint:
Closes #704.

An OpenAI-compatible provider returns HTTP 200 when it hits `max_tokens`; the body just stops. #305 carries the reason to callers that pass `onFinish`, but the three paths that hand a whole page body to the model — `merge-body`, `reviewed-append` and `related-page` — never asked, so the truncated text was written over the page. Measured on a 413-note rebuild: 13 rewrites grew a page by more than 8 KB, 10 of them next to a `finish_reason: length` in the provider log (details in #704).

**Change**

- `llm-sdk/finish-reason.ts`: `captureFinish()` — an `onFinish` sink that answers `truncated` (`finishReason === 'length'`), next to `reportFinish`, the same idiom `source-analyzer.ts` already uses inline. `'unknown'` (no report) is not truncated, so every client without the signal is unaffected.
- `mergePage` / `appendToReviewedPage`: on truncation return the path without writing, the same exit as `NO_NEW_CONTENT`.
- `updateRelatedPage`: on truncation take the #131 exit — new source in frontmatter, body verbatim.
- Tests: one per path (truncated → existing body kept; `'stop'` and no report → rewrite adopted as before).

**Gate 1** on `0e4e30a`: typecheck, lint 0, 291 files / 4105 tests (base 290 / 4098).

Considered and not done: surfacing truncation once in `callLlm` — none of the three paths goes through it, so that is a migration first, not a guard. Not touched: `createNewPage` (a truncated new page has no body to keep; a separate decision), and the loop itself — a length gate stops the loop from becoming the page; why the model loops on the footnote is a prompt question.


green-dalii (approved):
Approved. The guard is at the right altitude and in the fail-safe direction.

Verified the coverage claim rather than taking it: `page-factory` has seven `createMessage` sites, and the three this PR guards — `merge-body` and `reviewed-append` (`merge-page.ts:322`, `:496`) and `related-page` (`related-page.ts:147`) — are exactly the ones that write over an existing body. The unguarded ones are `createNewPage` (`create-page.ts:211`), which is new-page-only, and the append paths (`complementary-appends.ts:242`), which add rather than replace, so the destructive class is fully covered. That is the right line to draw for a PATCH.

`captureFinish` returning `'unknown'` when a client reports nothing is the detail that makes this safe to ship: legacy and mock clients keep their current behaviour, and the test asserts it explicitly rather than leaving it implied. `console.warn` + `return path` keeps the existing page and writes nothing, which is the only direction that cannot lose content.

Non-blocking observation, no action asked: the abandoned call still costs its tokens — unavoidable, since truncation is only knowable after the fact — and if the class ever shows up on the new-page or append paths, `captureFinish` is already the shape to reach for.

PR #712, opened 2026-09-14
fix(llm-sdk): classify gateway-style response_format rejections so the demotion chain engages

weqoocu:
Fixes 711 — gateway-style response_format rejections now engage the demotion chain

## What

Two classifier gaps fixed together so corporate-gateway 400s demote instead of surfacing raw:

1. **`shared-rejection-verbs.ts`** — `REJECTION_VERBS` gains `unavailable` and `is invalid`. Deliberately narrow: bare `invalid` would also match the #658 strict-dialect body (`"Invalid schema for response_format ..."`) and steal it from `isStrictSchemaRejection`.
2. **`output-mode-prober.ts`** — `JSON_SCHEMA_FIELD_MARKERS` gains bare `response_format` / `response-format` (same two entries Tier 1 already has). Gateways reject the envelope without naming the dialect; the outer `currentMode` guard prevents Tier 0/Tier 1 cross-firing on the now-shared marker.

Chain outcome for such gateways: `json_schema` 400 → `json_object` (one likely-failed retry) → `text_prompt` (no `response_format` on the wire) — the floor tier every backend can serve.

## Tests

| Case | Asserts |
|---|---|
| Venus-style verbatim body | `isJsonSchemaFieldError("This response_format type is unavailable now")` → true |
| verb-only negatives | `unavailable` / `is invalid` without a field marker → false (AND semantics intact) |
| #658 regression guard | strict-dialect body → false on Tier 0 classifier (strict classifier keeps owning it) |

## Gate 1

lint 0 / tsc 0 / build clean / 4100 tests, 290 files / css-lint 0.


green-dalii (approved):
Approved. Small, and the test that matters is the negative one.

Verified against the classifier it could have taken territory from: `isJsonSchemaFieldError` matches "This response_format type is unavailable now" but **not** the #658 strict-dialect body ("Invalid schema for response_format …"), which `isStrictSchemaRejection` must keep owning. The test asserts that boundary explicitly, and the negative cases ("This feature is unavailable now", "max_tokens is invalid") keep the match from widening into a generic 400 catch-all. This is the demotion chain of #658/#686 gaining the gateway case it was missing, which is why corporate-proxy users currently cannot ingest at all.

Worth including in this PATCH: the failure mode is "no workaround for anyone behind such a gateway", and the change is 48 lines inside an existing classifier.

Issue #173, opened 2026-06-21
Ingest: File-already-exists create-retry loop fails concept write + duplicate Created-pages entry


Indexed-Apogrypha:
## Summary

Two ingest-path anomalies observed during a single normal source ingest. **Independent of #164** — they occur in the normal create/merge path; they were just surfaced while manually testing the #164 pre-ingest gate.

## Symptom A — `vault.create()` retried on an existing path → write failure

Concept **"Prompts"** resolved to the path **`wiki/concepts/instructions.md`** (note the name → path mismatch) and failed to write. The console shows three `create` attempts all failing with **"File already exists"**, then the fallback update also failing:

```
Attempt 1 failed: File already exists.
Attempt 2 failed: File already exists.
Attempt 3 failed: File already exists.
Concept "Prompts" failed: Failed to create concept page "Prompts": 无法创建或更新文件: wiki/concepts/instructions.md
Concept "Prompts" retry also failed
```

`WikiEngine.createOrUpdateFile()` (the 3-attempt write helper) keeps calling `create()` on a path that already exists instead of switching to update/merge; the directory-scan fallback then can't resolve the existing file to `vault.process()` it, so it throws.

**To investigate:**

1. **Create-vs-update**: when the resolved path already exists, the write should go straight to update/merge rather than retrying `create()` three times.
2. **Name → path mismatch**: a concept named "Prompts" mapping to `instructions.md` points at slug/alias resolution (likely `page-factory` `resolvePagePath` / `core/slug.ts`) — e.g. an alias of an existing "instructions" page being treated as a brand-new "Prompts" page.

## Symptom B — duplicate entry in `createdPages`

The Ingest report's **Created** list contained `wiki/concepts/intelligent-xtraction-and-processing.md` **twice**, and the count header "Created pages: 9 (2 entities + 6 concepts)" included the duplicate (only 5 unique concept pages were actually created). `createdPages` is not de-duplicated before reporting — and the same concept may be getting processed twice within a batch.

## Repro

1. Ingest a normal markdown source that yields (a) a concept whose resolved slug collides with an existing page (e.g. via an alias) and (b) a concept extracted under two surface forms.
2. Observe the "⚠️ Failed to ingest" entry for the colliding concept and the duplicated line in the "Created" list.

## Environment

- Source file: `UiPath Academy/Configure tools for agents/Tool design principles and types.md`
[...]

PR #714, opened 2026-09-14
fix(ingest): one shape for updated_pages, so the link re-point pass sees every page

### LINKED ISSUE #713: `analysis.updated_pages` carries two different shapes, and the link re-point pass silently skips one of them
## What happens

Two places fill `analysis.updated_pages`, with two different kinds of string:

- `wiki-engine.ts:1362` / `:1376` — the entity and concept branches push `entityResult.path` / `conceptResult.path`, a **vault path** (`wiki/concepts/Toleranz.md`).
- `wiki-engine.ts:1443` — the related-page branch pushes `task.name`, a **bare page name** (`Warburg-Effekt`).

Three consumers read that list, and each one is affected differently.

**1. `repointLinksAfterRun` skips the bare names (functional).**

```ts
const paths = [...new Set(pagePaths)].filter(p => p.endsWith('.md'));
```

A bare page name does not end in `.md`, so it is filtered out. Pages touched *only* through the related-page branch are therefore never checked for folder-typed links that need re-pointing. This is invisible: the pass reports what it did, never what it never looked at.

**2. The report count is too high.**

`const updated = analysis.updated_pages.length` counts a page twice when it was updated both as an entity/concept and as a related page — the two entries are different strings, so nothing collapses them. The same array goes to `onDone` as `updatedPages`, on the success path and on the cancelled and failed paths.

Worth noting: the comment one line below already states the rule, for the created pages —

```ts
// Issue #173 Symptom B: dedup before counting/listing — a duplicated
// surface-form (e.g. the LLM emitting the same path twice) must not
// inflate the report count or the "Created" listing.
```

#173 fixed this for `created_pages`; `updated_pages` was not carried along.

**3. The log renders a dead link, twice.**

`log-writer.ts:90` maps `updated_pages` straight into wikilinks, without the `dedupPages` and the wikiFolder-prefix strip that the created-pages line right above it applies. `[[wiki/concepts/X.md]]` does not resolve in Obsidian — the folder prefix does not belong in link text.

## Measured

One ingest of a 33 KB note, plugin 1.27.1, from `wiki/log.md`:

```
**Updated pages**：[[wiki/concepts/Mitochondriale-Biogenese.md]], [[wiki/concepts/β2-Rezeptor.md]],
[[wiki/concepts/Toleranz.md]], [[wiki/concepts/Laktatazidose.md]], [[Mitochondriale-Biogenese]],
[[β2-Rezeptor]], [[Laktatazidose]], [[Warburg-Effekt]]
```

Eight entries, five distinct pages. Three pages appear twice, once in each shape. `Warburg-Effekt` appears only as a bare name — it went through the related-page branch alone, so the re-point pass never saw it. Four entries render as dead links.

## Suggested fix

Give the list one shape. `updateRelatedPage` already knows `page.path` — it uses it for every write — and only returns `boolean`; returning the path makes the list type-clean. The duplicates that remain are then real ones (one page, two branches), so the list is best kept distinct where it is filled rather than at each reader — that also covers the cancelled and failed paths, which have no dedup of their own.

PR attached. Close it if you would rather solve it differently — the issue stands on its own.

DocTpoint:
Fixes #713.

`updateRelatedPage` returns the path it already knows instead of a boolean, so `updated_pages` carries vault paths throughout. The remaining duplicates are real — one page reached by two branches in the same run — so the list is held distinct at the three places that fill it, via `recordUpdatedPage`. The log's two page lists now render through one method, which gives the updated pages the same dedup and prefix strip the created ones already had.

**Why at the fill site and not at each reader:** there are five readers, not three — the log, the count, and `onDone` on the success, cancelled and failed paths. The last two hand `updated_pages` over raw. Fixing the readers would have left those two behind and would need a sixth fix for the next reader.

**Behaviour change worth calling out:** `repointLinksAfterRun` now also sees pages that were touched only as related pages. That is the point of the fix, but it means the pass writes more pages per run than before. It still skips `reviewed: true` pages and still writes nothing when a page comes out byte-identical.

**Not in this PR:** `created_pages` is still deduplicated at the reader (#173) rather than at the fill site. The same treatment would work there, but it is a different list with a different cause of duplication — the model emitting a path twice — and it is not this bug.

Gate: build, lint and 4103 tests in 291 files, against 4098 in 290 on `main`. One new test file for `recordUpdatedPage`, one regression test for the log line; twelve existing tests moved from `toBe(true)` to the returned path — which incidentally made them sharper, one of them caught that `Dysbiose` is a concept and not an entity.


green-dalii (approved):
Approved — and the fix is at the fill site rather than the readers, which is the part that makes it complete.

Deduping inside `recordUpdatedPage` (`dedup-pages.ts`) rather than in each consumer is what keeps the cancelled and failed paths correct: those hand `updated_pages` straight to `onDone` without passing a reader, so a reader-side dedup would have left the count too high exactly where it is hardest to notice.

Second fix worth naming, because the diff makes it look incidental: `pageLinks` (`log-writer.ts:150`) now strips the `wikiFolder` prefix for updated pages too. `[[wiki/concepts/X.md]]` does not resolve, `[[concepts/X.md]]` does — so updated pages were rendering as dead links in the log, and the comment says so.

The `Promise<boolean>` → `Promise<string | null>` change on `PageFactory.updateRelatedPage` is a signature change, and I checked the call sites: it now returns the path so the caller can record it, which is what removes the bare-name shape at its source.

Issue #711, opened 2026-09-14
fix: structured-output 400 from corporate gateways ("response_format type is unavailable now") never engages the demotion chain


weqoocu:
## Environment

- Plugin version: 1.27.4 (also present in current `main`)
- Endpoint: corporate OpenAI-compatible LLM gateway (internal proxy)
- Model: `deepseek-v4.1-flash-official`
- Task: "Maintain wiki" → fix broken links / generate aliases (any task that sends a schema with `response_format`)

## What happens

Structured-output steps fail with a raw 400 surfaced to the user. The plugin's 3-tier output-mode demotion chain never engages, even though the endpoint would work fine at the `text_prompt` tier (plain `generateText` calls — e.g. source extraction — succeed on the same endpoint).

The endpoint rejects **any** `response_format` value with:

```json
{"error":{"message":"This response_format type is unavailable now","type":"invalid_request_error","code":400}}
```

## Why the demotion chain misses it

The chain (`openai-compat-sdk-client.ts`) only demotes when `classifyFieldError(body, FIELD_MARKERS)` matches, which requires **(a verb AND a field marker)**:

1. **Verb half fails.** `REJECTION_VERBS` = unrecognized / unknown / invalid value / unsupported / not allowed / not supported / must be / should be. The gateway says *"is unavailable now"* — none match. ("…is invalid for this model" phrasing also falls through.)

2. **Field-marker half fails for Tier 0.** `JSON_SCHEMA_FIELD_MARKERS` = `json_schema` variants only. The gateway rejects the `response_format` **envelope** without naming the structured-output dialect, so even a verb match would not have found a Tier 0 marker. (Tier 1's list has bare `response_format`, Tier 0's does not.)

Result: `break` out of the demotion loop on the first iteration, error propagates to the Notice/UI.

## Expected

The chain should demote to a tier the endpoint accepts. For this gateway that's `text_prompt` (no `response_format` on the wire at all), possibly via one doomed `json_object` retry — one wasted request is acceptable, the chain is designed for that.

## Suggested fix

1. Add `unavailable` and `is invalid` to `REJECTION_VERBS` (narrow phrases — bare `invalid` would also match the #658 strict-dialect body `"Invalid schema for response_format ..."` and steal it from `isStrictSchemaRejection`).
[...]

PR #715, opened 2026-09-14
docs: two maintainer tooling failures that read as something else

green-dalii:
Records the two maintainer-tooling failures from the 2026-09-13 session, both of which misreport their own cause.

**`gh auth refresh` without `--hostname`.** Run non-interactively it prints the flag list, with the actual error (`--hostname required when not running interactively`) sitting above it — so the output reads as "the flags were wrong". Two attempts were wasted on that reading. The correct invocation is in the entry, backgrounded via `nohup` so the one-time device code can be surfaced before the browser step; the flow needs a human and the code expires in ~15 minutes.

**A workflow-file merge reports branch policy.** `gh pr merge` on a PR touching `.github/workflows/*` fails with `the base branch policy prohibits the merge`. That reads as an unmet requirement, and review and CI were both satisfied. `--admin` and the REST `PUT /pulls/<N>/merge` endpoint both surface the real cause: HTTP 403, `without 'workflow' scope`. The ruleset has `bypass_actors: []`, so `--admin` does not bypass it — it only stops masking it.

The operational rule that follows: a same-shaped PR can merge cleanly when the change is a *step* rather than a *trigger* (#693 did), so the shape of the diff is not a reliable predictor. The error is. Do not report a PR as merge-ready until one of the two paths has returned a merge commit.

Docs only, no code.


DocTpoint (changes requested):
Reviewed on `089df5a`. Both failure modes are worth recording, and the `gh auth refresh --hostname` half is exactly right — the usage text really does bury the actual error above the flag list.

I checked the ruleset claim independently and it holds: `GET /repos/GD4AI/obsidian-llm-wiki/rulesets/16884813` ("Protect main branch", branch target, active) returns `bypass_actors: []`. So `--admin` cannot bypass it, and the entry is correct that it only stops masking the 403. Worth naming that id in the entry, so a later reader can re-check the claim in one call rather than trusting it.

One clause needs to go, because this repository's own record refutes it.

### "a same-shaped PR can slip through when the change is a step rather than a trigger"

Three merged PRs touch `.github/workflows/pr-ci.yml`:

| PR | shape of the change | merged | outcome |
|---|---|---|---|
| #693 | adds a whole job/step (lockfile gate) | 2026-09-12 18:30:18Z | no wall |
| #639 | a step change — `uses: actions/attest-build-provenance@v1` → `@v4` | 2026-09-13 00:20:18Z | sat on the wall |
| #698 | a trigger change — adds `push: branches: [main]` | 2026-09-13 00:20:09Z | sat on the wall |

#693 and #639 are both step changes and they went opposite ways, so the step/trigger shape cannot be what decides it. And the two that failed were merged **nine seconds apart**, six hours after the one that did not — which is the signature of a single unblocking event, not of two different file shapes.

The simpler reading is already the first half of your own entry: the variable is whether the token carried the `workflow` scope at that moment. #693 merged before the scope was missing; #639 and #698 merged back-to-back right after `gh auth refresh` supplied it. That also explains why they cluster in one minute, which the shape hypothesis does not.

This matters because of how the sentence will be read later. "A step change can slip through" invites a future session to expect that a step-shaped workflow PR is safe, and to read the 403 as something else when it appears anyway. Your own conclusion — the error is the only reliable signal — is stronger without the clause than with it.

Two ways to fix, whichever you prefer:

- delete the clause and keep the conclusion; or
- replace it with the scope reading, e.g. "#639 and #698 both sat on this wall and merged nine seconds apart once the scope was added; #693 merged earlier without it being an issue, so the variable is the token's scope at the time, not the shape of the change."

Nothing else in the entry needs changing, and I will approve on the next push.

green-dalii:
Correction taken — and the clause is gone rather than softened, because the data refutes it outright.

You are right that #693 and #639 are both step changes with opposite outcomes, so shape cannot be the variable. I wrote that clause as a hypothesis in my own notes ("I cannot confirm the mechanism") and then stated it as fact in a canonical doc, which is the worse version of the mistake: a reader six months from now has no way to tell it was a guess.

Replaced with the scope reading, plus the ruleset id so the claim is one call to re-check:

> The ruleset has no bypass actors (`rulesets/16884813`, "Protect main branch", active — one call re-checks the claim), so `--admin` does **not** bypass it — it only stops masking it. The variable is **whether the token carried `workflow` at that moment**, not the shape of the change: #693 added a step and merged, #639 changed a step and sat on the wall, #698 added a trigger and sat on the wall. Same shape, opposite outcomes — and the two that failed were merged nine seconds apart, six hours after the one that did not, which is the signature of one unblocking event. So the error is the only reliable signal.

Your point about how the sentence would be read is the one that decided it: "a step change can slip through" invites the next session to treat a step-shaped workflow PR as safe and to misread the 403 when it arrives anyway. The conclusion was doing the work; the clause was only subtracting from it.

Also worth recording on my side — the same session produced a live instance of this pattern in a different place: #712's body said `Fixes 711` without the `#`, so #711 stayed open after its fix merged. Both are the same failure mode: a downstream parser reads the text literally, and the author's intent is not part of the input.

DocTpoint (approved):
Approved on `4e5b04b`.

The clause is gone and the replacement is accurate. I re-checked both facts it now rests on: `rulesets/16884813` ("Protect main branch", branch target, active) has no bypass actors, and #639 (00:20:18Z) and #698 (00:20:09Z) are nine seconds apart, with #693 merged five hours fifty minutes earlier (2026-09-12 18:30:18Z). Naming the ruleset id was the right call — the claim is now one API call to re-check.

**Gate 1 is red here, but not because of this PR.** This diff is +13 lines in `AGENTS.md`. The failure is inherited from `main`:

    src/wiki/page-factory/related-page.ts(165,5): error TS2322:
    Type 'boolean' is not assignable to type 'string'.

`main` has been typecheck-red since `c05d89a0` (15:19:38Z), and every push since has run red: #702, #709, #710, #712. The cause is two of my own PRs that were green separately and are red together:

- #705 added an early return in the truncation branch of `updateRelatedPage`: `return true`, against the signature as it stood.
- #714 changed that signature to `Promise<string | null>`, so the link re-point pass sees every written page.

They merged twelve seconds apart (15:19:27Z and 15:19:39Z), so neither CI run ever saw the other. This is mine to repair and I am opening an issue with the fix now. The fix is `return page.path`, not `return null`: line 164 does write the page, so the re-point pass has to see it — `null` there would reopen exactly the gap #714 closed.

Nothing to change on this branch. Once `main` is green, a re-run of Gate 1 here should be enough.

green-dalii:
Branch rebased onto current `main` (head `8d6874b2`) and Gate 1 is green — the earlier failure was the pre-#722 `related-page.ts:165` TS2322, not this PR. Merging. Thanks for the review rounds.

PR #717, opened 2026-09-14
fix(page-factory): the truncated-rewrite exit returns the page path, not true

### LINKED ISSUE #716: `main` is typecheck-red: #705 and #714 were green apart and red together
`main` has been typecheck-red since `c05d89a0` (2026-09-14 15:19:38Z):

    src/wiki/page-factory/related-page.ts(165,5): error TS2322:
    Type 'boolean' is not assignable to type 'string'.

Four merges have run red since, and none of them is the cause: #702 (`4436fc45`), #709 (`362f16bf`), #710 (`58eca3bf`), #712 (`4c3e6ecc`). Contributor PRs inherit it as well — #715 is +13 lines in `AGENTS.md` and reports a red Gate 1.

## Cause

Both PRs are mine:

- **#705** added an early return to the truncation branch of `updateRelatedPage`: `return true`, against the return type as it stood.
- **#714** changed that return type from `boolean` to `string | null`, so the link re-point pass sees every page a run wrote.

They merged twelve seconds apart — 15:19:27Z and 15:19:39Z. Their common base is `0e4e30a` and neither branch ever contained the other, so no CI run saw the pair.

This is the class the `push: [main]` comment in `pr-ci.yml` already records for #681/#684. The gate #698 added did its job — `main` went red on the very next push. What failed is the reading: the signal stood for twelve minutes and four merges.

## The test suite cannot see it

On `c05d89a0`, `pnpm test` is green — 4144 passed in 294 files. `expect(result).toBe(true)` matched the literal `true` that was being returned. Only `tsc` catches it. A return-contract change of this shape is invisible to every gate except typecheck.

## Fix

`return page.path`, not `null`. That branch does write the page — frontmatter with the new source, body kept verbatim — so the caller has to see it; `null` would drop a written page out of the re-point pass, which is the gap #714 closed. #705's own comment already names the right answer: "same exit as the #131 path", and the #131 path returns `page.path`.

Three lines: the return, the doc comment (it still describes the pre-#714 contract), and the one assertion in #705's test written against it.

PR follows.

DocTpoint:
Closes #716.

## What

One line of behaviour, two lines of bookkeeping:

- `related-page.ts:165` — the truncated-rewrite exit returns `page.path` instead of `true`.
- the doc comment on `updateRelatedPage`, which still described the pre-#714 `boolean` contract.
- the one assertion in #705's test that was written against that contract.

## Why `page.path` and not `null`

The branch writes the page before it returns — frontmatter carrying the new source, body kept verbatim. Returning `null` would type-check and drop a written page out of the link re-point pass, which is exactly the gap #714 closed. #705's comment on that branch already says "same exit as the #131 path", and the #131 path returns `page.path`.

## Verification

Gate 1 green on `9508c7c` against `4c3e6ec`: lint, typecheck, build, 4144 tests in 294 files, css-lint.

Counter-check on the base without this commit: typecheck fails at `related-page.ts(165,5)` and the same 4144 tests pass — the test count is unchanged, one assertion retargeted, none added or removed.


DocTpoint:
Superseded by #722, which lands the same `return page.path` fix together with the `truncated-rewrite.test.ts` update. Closing this one; main is green again since 080f884. Thanks for taking it.

PR #720, opened 2026-09-14
fix(lint): the double-nested link repair never reached the file

### LINKED ISSUE #719: The lint's double-nested link repair is computed, counted, logged — and never written
`runPreparationPhase` repairs `[[[[Foo]]]]` links on every wiki page inside `vault.process`, then returns the original:

    src/wiki/lint/phases/preparation.ts:68
      await ctx.app.vault.process(abstractFile, (data) => {
        const { fixed, content } = fixDoubleNestedWikiLinks(data);
        if (fixed > 0) {
          doubleNestFixes += fixed;
          info.content = content;
          console.debug(`lintWiki: fixed ${fixed} double-nested link(s) in ${path}`);
        }
        return data;
      });

Thirteen lines below, the `log.md` branch does the same work and writes:

    return fixed > 0 ? content : data;

The repair lands on `log.md` and on no wiki page. Three consequences, each of which makes it look like it landed:

1. `doubleNestFixes` counts repairs that are not on disk, and the lint report carries that number.
2. The debug line says `fixed N double-nested link(s) in <path>`.
3. `info.content` is set to the repaired text, so every later lint phase in the same run reads content the file does not hold.

## It is a slip, and the old comment says so

Before c10d7e1 (2026-06-16) modularized the controller, the same line in `lint-controller.ts` read:

    return data; // return unchanged if no fixes needed

That comment describes a conditional the code never had — and the `log.md` branch just below it is the same conditional, written correctly. The refactor carried the line and dropped the comment, so since June nothing on the line suggested anything was meant.

## The test cannot see it

`preparation-phase.test.ts:108` is named "fixes double-nested wiki links in place" and asserts the counter and `pageMap` — both the in-memory half, both assigned inside the `if (fixed > 0)` block above the return. The mock's file map is never asserted, even though the mock writes back. One assertion on the file makes the test fail on current `main`.

## Behaviour note

Fixing this makes the lint start writing a repair it currently only reports. That is what the code, the counter, the log line, the test name and the `log.md` branch all say it does — but it is a behaviour change on the first run after the fix, so it belongs in the open rather than slipped in.

PR follows.

DocTpoint:
Closes #719.

## What

One line: `preparation.ts:75` returns `fixed > 0 ? content : data` instead of `data`, which is what the `log.md` branch thirteen lines below already does.

Plus one assertion in the existing test, which checked only the in-memory half and therefore passed throughout.

## Behaviour

The lint starts writing a repair it previously computed, counted and logged without writing. On the first run after this, a page carrying `[[[[Foo]]]]` is rewritten to `[[Foo]]` — which is what the report's counter has been claiming all along.

## Verification

The new assertion fails on current `main`:

    AssertionError: expected 'See [[[[Nested]]]] link.' to be 'See [[Nested]] link.'

Gate 1 on `c441369`: lint, build and css-lint pass, 4144 tests in 294 files pass, and `typecheck` fails only at `related-page.ts(165,5)` — that is #716, not this branch. With #717 applied on top the full gate is green, same 4144 in 294. This PR will report red until #717 merges.


green-dalii (approved):
Reviewed and approved.

The one-line change matches what the surrounding code already does — the `log.md` branch thirteen lines below has always written `return fixed > 0 ? content : data`, and the wiki-page branch silently did not. #719 is right that this is a slip rather than a design decision.

The test change is the more valuable half: the original assertion read `pageMap` (the in-memory half) and nothing else, so a repair that never reaches disk passed. `files['wiki/entities/Foo.md']` pins the file, which is where the bug lived.

Merging.

## Findings

- `src/wiki/lint/phases/preparation.ts:75` — `return fixed > 0 ? content : data` ✓ now matches `:88`.
- `src/__tests__/wiki/lint/preparation-phase.test.ts:117-119` — the new on-disk assertion is the mutation-resistant one.
- No call-site impact: `vault.process`'s return value is the written content, and `content` is `fixDoubleNestedWikiLinks(data)`'s output — same function, same input, so the only behaviour change is that the repair now persists.

PR #726, opened 2026-09-15
fix(lint): give a link its name back before the stub it points at is deleted

### LINKED ISSUE #725: `deleteEmptyStubs` deletes a stub and leaves its incoming links behind, retargeted — the round trip #691 completed
Same class as #386, one deletion path further along: a page is deleted and the links pointing at it are never visited. There the rewrite's radius was too small; here there is no rewrite at all.

## What happens

With `createStubsForUnresolvableLinks` at its default (on), Fix Dead Links does two writes when it cannot resolve a link. It creates the placeholder page, and it **repoints the referring page's link at it** — the deterministic fallback in `fix-dead-link.ts`, and the LLM `create_stub` branch above it does the same with its own title:

```ts
const updatedContent = replaceDeadLink(sourceContent, targetName, newLink);
```

So `[[Vitamin B12]]` becomes `[[entities/Vitamin-B12|Vitamin B12]]` (casing per `slugCase`).

`deleteEmptyStubs` then deletes the file and touches nothing else:

```ts
if (!isPageEmpty(content) && !(isStubPage(fm) && isEmptyStub(content))) continue;
await ctx.deleteFile(file.path);
```

The link is dead again — but it is no longer the link its author wrote. The display text survives, the target does not. A bare name can still be matched on the next pass against a title, an alias or a slug; what is left instead has been through `slugify` and carries a folder that was only ever a guess about what kind of page the target would turn out to be.

## Why this is new

Before #691 it could not happen: `isPageEmpty` never recognised a Fix Dead Links stub, so the stub stayed and the retargeted link at least pointed at something. That was #678, which I filed and you fixed on 14.09. Closing it made the first half of the round trip reachable; the second half has always been missing.

To be clear, #691 is not the defect. It does what #678 asked for, and the gap it exposed sits in `deleteEmptyStubs`, which has never repointed anything.

## Scope of this report

Read from the code, not observed in a vault: mine has the setting off, and the fix only shipped yesterday, so there is no incidence to quote. What makes it worth your time is that the same class was real enough in #386 to fix.

## The PR

Attached. Before deleting, the links that resolve to the page get their name back: `[[entities/Vitamin-B12|Vitamin B12]]` → `[[Vitamin B12]]`. The display text is exactly the name Fix Dead Links took the target from, so nothing is guessed.

Two choices I made rather than ask about, either of which you can overrule:

- **Radius is vault-wide**, as #386 settled for the same reason — a stub can be referenced from a user note, and those are the links that become unfindable. It does mean this action writes outside `wiki/` for the first time.
- **A link with no display text is left and counted**, not given an invented name. Same for a `#subpath` into the page that is going.

Close this if you would rather the stub path stopped repointing the referring page in the first place — that is the other end of the same round trip, and it would make the restore unnecessary instead of correct.

DocTpoint:
Same class as #386, one deletion path further along: a page is deleted and the
links pointing at it are never visited. There the rewrite's radius was too
small; here there is no rewrite at all.

With `createStubsForUnresolvableLinks` at its default, Fix Dead Links does two
writes when it cannot resolve a link: it creates the placeholder page, and it
repoints the referring link at it, carrying the author's name over as display
text. `[[Vitamin B12]]` becomes `[[entities/Vitamin-B12|Vitamin B12]]`. Delete
Empty Stubs then deletes the file and touches nothing else, so the link is dead
again — but it is no longer the link its author wrote. The display text
survives, the target does not, and a folder-typed path through `slugify` is a
harder thing to match on the next pass than the bare name it replaced.

Before #691 this could not happen: `isPageEmpty` never recognised a Fix Dead
Links stub, so the stub stayed and the retargeted link at least pointed at
something. That was #678. Closing it made the first half reachable; the second
half has always been missing, and it is not a defect in #691.

`planDisplayNameRestores` / `applyRestorePlan` are the counterpart to
`retargetLinksToPage` for a deletion with no successor page: every link that
resolves to a page being deleted gets its display text back as the target. The
vault walk both halves share — resolve before replacing, target files
themselves skipped — is now one function, `collectResolvingReferences`, and the
edits both produce are applied through one `applyFileEdits`. `retargetLinksToPage`
keeps its behaviour exactly; its own tests pin that.

Planning and writing are separate because their timing constraints point in
opposite directions. Resolving needs the pages to still exist — afterwards
`getFirstLinkpathDest` cannot confirm a link ever pointed at them. Writing must
wait until the delete has actually succeeded, or a failed delete leaves the
links pointing away from a page that is still there. So: plan, delete, then
apply what the delete made due, filtered to the pages that really went.

One pass over the vault for the whole run, not one per stub. A note referencing
two collected stubs is rewritten once, both edits together: rewriting it twice
would shift the second set of offsets under a metadata cache that has not
re-indexed, and the stale guard would drop exactly the links this exists to
save. Both properties are pinned, and both pins fail if the batching is undone.

Three reference shapes are reported instead of rewritten, because none can be
put back the way it was written: a link with no display text carries no name, a
`#subpath` addressed a heading of the page that is going, and a display text
carrying link syntax of its own (`[[a|b|c]]`) would come back as a different
link. They are counted in `linksLeftDead` and stay visible in the lint report.

The restored link is the one its author wrote, so it resolves exactly as it did
before Fix Dead Links ran: to nothing, or to a page of that name that has since
appeared. Same outcome as never having run the action, which is the point —
this heals nothing on its own.

Radius is vault-wide, as #386 settled for the same reason: a stub can be
referenced from a user note, and those links are the ones that go unfindable.
This is the first time this action writes outside `wiki/`.

Link care never blocks the delete: planning and applying are caught separately,
so an unusable metadata cache leaves the links as they were and the stubs are
still collected. Pinned by test.

Not in this PR: the counts reach the lint log, not the completion Notice —
that string is localised in eleven files. And `fix-polluted-page.ts` is a third
place that faces a page's fate against its incoming links, still on its own
regex path; folding all three into one "retire this page" operation is the
deeper shape, and a bigger change than this one.

Gate 1 green: lint 0 / tsc 0 / build / 4158 tests (4144 on the base) / css-lint 0.

Fixes #725

Close this if you'd rather solve it differently — the issue stands on its own.


green-dalii (approved):
Approving. I read the diff rather than just the checks, and this is the shape I want a deletion path to have.

**Verified for side effects, breakage and cost:**

- **Writes.** Every change goes through `deps.vault.process` — the Obsidian API, so `vault.on('modify')` fires and the metadata cache stays truthful. No `adapter.write` anywhere. `applyFileEdits` batches one `vault.process` per file with several edits inside, which is *better* than the per-edit writes the naive version would have done, and `collectResolvingReferences` means the vault walk is shared between the retarget and restore paths rather than scanned twice.
- **Breakage.** `DeleteEmptyStubsResult` is structurally identical to the inline `{ deleted; failed; errors }` it replaces, so the signature change is a name, not a contract. The new exports are additive, and `retargetLinksToPage` keeping its behaviour is pinned by its own tests as the description says.
- **Bot compliance.** No `node:*`, no bare `document`, no `obsidianmd` disable — checked on the patch, not inferred.
- **Scope honesty.** Naming what is *not* here — the Notice string across eleven locales, and `fix-polluted-page.ts` as a third site still on its own regex path — is more useful than a longer PR would have been. The plan/apply split for the timing asymmetry is the right call for the reason given.

One note on the closing keyword, for the record rather than as an objection: the body carries `Fixes #725`, and I checked that #725 is the issue this actually closes and the only closing keyword in the body. Worth being deliberate about here, since the same sentence pattern in a different PR once closed something nobody meant to close.

Merging.

PR #727, opened 2026-09-16
fix(ci): regenerate package-lock.json on Dependabot PRs (#718)

### LINKED ISSUE #718: Dependabot's npm PRs cannot pass the lockfile gate — it never updates `package-lock.json`
#706, #707 and #708 have all been red since 2026-09-14T01:12Z, each on the same step, and none of them can go green as it stands:

    #706  @ai-sdk/openai-compatible: package.json declares 3.0.47,
          package-lock.json was generated for 2.0.62
    #707  eslint: package.json declares 10.10.0,
          package-lock.json was generated for 10.2.1
    #708  tslib: package.json declares 2.8.1,
          package-lock.json was generated for 2.4.0

## The gate is right; what it reveals is the problem

`scripts/check-lockfile-sync.mjs` (#693, merged 2026-09-12T18:30Z) compares `package.json` against `package-lock.json`, and its header comment gives the reason: the npm-side lockfile had no gate at all between releases, and #652 let it drift silently until two advisories were filed against the stale copy.

The part that did not surface then: **Dependabot never touches `package-lock.json` in this repository.** Every npm PR it has opened here changes exactly `package.json` and `pnpm-lock.yaml`:

| PR | files changed |
|---|---|
| #652 (merged 09-10, the drift that motivated the gate) | `package.json`, `pnpm-lock.yaml` |
| #706 | `package.json`, `pnpm-lock.yaml` |
| #707 | `package.json`, `pnpm-lock.yaml` |
| #708 | `package.json`, `pnpm-lock.yaml` |

So #652 was not an oversight — it is the shape of every one of these PRs. Before the gate, that drift was silent. After it, the same shape is a hard block. #706–#708 are the first npm batch opened after the gate merged, and all three are stuck on it.

`.github/dependabot.yml` records the opposite assumption:

> `package-ecosystem: npm` covers pnpm-lock.yaml (GitHub native support) and package-lock.json alike.

and states the goal as "Human action reduces to review + merge, not hand-editing four files". As things stand, every npm bump needs exactly that hand-edit.

The GitHub Actions half of the config is unaffected — #709 and #710 merged today without trouble, because they do not touch `package.json`.

## Three ways out, all yours to pick

1. **Regenerate `package-lock.json` inside each Dependabot PR.** A small CI job running `npm install --legacy-peer-deps --package-lock-only` and committing the result onto the branch. Keeps both lockfiles and the npm-side advisory surface; costs a bot write to Dependabot branches.
2. **Drop `package-lock.json`.** CI installs with `pnpm install --frozen-lockfile` and the audit step reads pnpm's lockfile, so nothing in the build needs it. This removes the gate's subject, and needs a different answer for where npm-side advisories are filed.
3. **Move the checkpoint.** Let the gate run on release rather than per PR, or have it accept a PR whose `pnpm-lock.yaml` moved in step. Keeps the file, stops blocking the automation.

I have not touched labels or the milestone. No PR from me on this one — which of the three is right is a call about your supply chain, not a patch.

green-dalii:
## What this fixes

Dependabot cannot pass this repository's lockfile gate, and cannot be made to. This adds the missing half rather than weakening the gate.

`#706`, `#707` and `#708` have been red since 2026-09-14, all on the same step.

## Root cause

Dependabot maintains `pnpm-lock.yaml` here and never touches `package-lock.json`. The repository declares `packageManager: pnpm@10.14.0` and carries a `pnpm-workspace.yaml`, so its npm ecosystem resolves to pnpm. Every npm PR Dependabot has opened in this repository changes exactly `package.json` + `pnpm-lock.yaml` — including `#652`, the drift that motivated the gate.

The gate (`.github/workflows/pr-ci.yml`, `pnpm lockfile-check` → `scripts/check-lockfile-sync.mjs`, from #693) compares `package.json` against `package-lock.json`. It is correct, and it is unsatisfiable by a bot that structurally cannot update that file.

The note in `.github/dependabot.yml` asserting that `package-ecosystem: npm` "covers pnpm-lock.yaml and package-lock.json alike" is contradicted by all four PRs above. That note is corrected here.

## The fix

A workflow that regenerates the npm-side lockfile on the Dependabot branch and pushes it, after which the gate passes on its own terms.

`pull_request_target` is required, not incidental: a `pull_request` run triggered by Dependabot receives a read-only `GITHUB_TOKEN` regardless of the `permissions:` block, so it could not push the result. The trade is a write token in the base repository's context, bounded by three properties:

1. The job runs only when the actor is Dependabot **and** the head branch is in this repository.
2. The workflow definition is always the one from the base branch — a pull request cannot edit the job that runs on it.
3. The only command executed is `npm install --package-lock-only`, which resolves the tree and writes one file. It downloads no packages and runs no lifecycle scripts, so nothing from the pull request is executed.

The push is a no-op once the lockfile is correct, so the resulting `synchronize` event re-enters and terminates.

## Test evidence (run before this PR was opened, not after)

Reproduced the failure on the real Dependabot branch (`dependabot/npm_and_yarn/eslint-10.10.0`, `67f81c3a`) in a throwaway worktree:

| Step | Result |
|------|--------|
| `node scripts/check-lockfile-sync.mjs` — before | ❌ `eslint: package.json declares 10.10.0, package-lock.json was generated for 10.2.1` |
| `pnpm install --frozen-lockfile` | ✅ passes — the pnpm side is not the blocker |
| `npm install --legacy-peer-deps --package-lock-only` | ✅ `found 0 vulnerabilities` |
| `node scripts/check-lockfile-sync.mjs` — after | ✅ `lockfile-sync: package-lock.json matches package.json` |
| Files changed | ✅ `package-lock.json` only (+137/−44) — the workflow's `git diff --quiet` no-op branch is reachable |
| Re-run the command, compare SHA-256 | ✅ identical — idempotent, so the workflow cannot loop |
| `pnpm lint` on that branch (eslint 10.10.0) | ✅ exit 0 |

Confirmed against the actual CI run for `#707` (`34960209765`) which step fails — `Install (lockfile-pinned)` is **success**, `Lockfile sync (package.json against package-lock.json)` is **failure**, and everything after it is skipped.

## Why the gate is kept rather than deleted

Deleting `package-lock.json` was considered and rejected. `.npmrc` records why the file exists: Obsidian's own pipeline runs `npm install` and compares the `main.js` it builds against the released artifact, so a resolution difference demotes the release review score. `.github/workflows/release.yml` relies on the same tree (`cache: 'npm'`, `npm install --legacy-peer-deps` → `npm run build`). `package-lock.json` is a load-bearing input to release reproducibility, not an internal convenience, so the gate stays strict and the bot's missing half is supplied instead.

## Gate 4: Performance

| Dim | Status | Notes |
|-----|--------|-------|
| CPU | N/A | No runtime code path. One `npm install --package-lock-only` per Dependabot PR, not per build. |
| Memory | N/A | Resolves a dependency tree and exits. |
| IO | ✅ | Writes one file, and skips the commit-and-push step entirely when it already matches. |
| Network | ✅ | One registry resolution per Dependabot PR; no package downloads, no lifecycle scripts. Bounded to Dependabot-authored PRs by the job condition. |
| Token | N/A | No LLM path. |

Closes #718


green-dalii:
Merging. Test evidence is in the body: reproduced the failure on the real Dependabot branch, verified the regeneration makes the gate pass, confirmed only `package-lock.json` changes, and confirmed idempotency by SHA-256 so the workflow cannot loop. Also verified which CI step actually fails (`Lockfile sync`, not the pnpm install above it).

Issue #729, opened 2026-09-16
Cross-source relations: the Related graph is intra-source by construction, and the fix is a reserved budget, not an index


green-dalii:
## The measurement that started this

@DocTpoint rebuilt a 413-note vault unattended over three days on v1.27.1: 413 source pages and 2,135 entity/concept pages. Three numbers from that rebuild describe the shape of the graph this plugin produces:

- **95 % of edges connect pages born from the same note.** Plain co-occurrence reproduces **94.5 %** of the Related edges.
- **Nine of ten pages have no incoming link from a page with another source.**
- The Related cap holds with zero violations and strays sit at 1 % — the shaping is working exactly as written.

Then a null model over five multi-note questions — same character budget, same model, thinking off, truth drawn from the notes and rated blind:

| Arm | Retrieval over | Result |
|---|---|---|
| A | raw notes, embedding index | baseline |
| B | wiki pages, same index | slightly ahead of A |
| C | the plugin's query path | **lost 3 of 5 questions** |

**B ≥ A is the first evidence this project has that the compiler is not a detour** — the wiki stores the notes at least as well as the raw text does, with provenance on top. The reader is the bottleneck, and it is a bottleneck of *recall*: arm C found the pages less often than a trivial embedding over the same corpus, not less accurately once they were found.

## Root cause: the star is written into the extraction prompt

This is not an emergent property of how ingest works. It is an instruction.

`src/wiki/prompts/ingestion.ts:27`:

> For related_entities and related_concepts: identify entities/concepts **mentioned in the same context as this item. These should be other items extracted from this same source file**

The JSON example repeats it — `"related_entities": ["Related entity names from this source"]`.

Phase 2 then does the only thing it can with those names. `src/wiki/page-factory/related-links.ts` resolves them against the whole vault:

```ts
const vaultIndex = { wikiFolder, pages: await getExistingWikiPages(ctx.app, settings.wikiFolder) };
```

`buildVaultResolver(name)` maps name → path and returns `undefined` for anything the vault does not hold. So the write path can **confirm** a name but can never **discover** one. The Related sections are intra-source by construction, and no amount of PPR tuning touches that: in a star forest there is no path between two stars, so **no weight function can create reachability**.

## The machinery for the fix already exists and is unwired

[...]

Issue #668, opened 2026-09-10
Settings tab: three tabs over the nine sections that already exist


DocTpoint:
The Settings tab has grown to 70 `new Setting(...)` construction sites across nine section modules. Seven of those are section headings, and the per-task model field is one site rendered four times, so the rendered list is longer still. Reading it end to end is now the only way to find a knob whose section you don't remember.

**The structural work is already done.** `display()` has been an orchestrator since #311, and every section renderer takes its container as a parameter:

```ts
renderLanguageSection(this, containerEl);
renderStatusSection(this, containerEl);
…
```

So grouping them into tabs is a change to the orchestrator alone: build three containers instead of one and hand each renderer the right one. No section module changes.

**Proposed grouping**, with today's row counts:

| Tab | Sections, in their current order | rows |
|---|---|---|
| Model | Status · LLM Provider · Model · Advanced (sampling) · Test Connection | 28 |
| General | Language · Wiki Configuration · Auto Maintenance | 27 |
| Advanced | Advanced settings | 12 |

**This does not touch the ordering the `display()` comment protects.** It asks for Status before LLM Provider, and for Test Connection to sit right after the Advanced block rather than at the end of the tab. Sections 2–6 are exactly the Model tab, contiguous and in order, so both guarantees hold literally. Tabs group; they do not reorder.

I'd read this as the second half of the move in 81d75fc rather than a break with it: that redesign relocated `writePdfMarkdownToVault`, `slugCase` and the lint thresholds into one bottom panel because a user kept finding them in the wrong place. Same complaint one level up — the panel is easy to find now, the tab it lives in is not.

**One thing that would make this fail quietly if missed:** `display()` is re-invoked from 28 call sites (18 in the section modules, 10 in settings.ts) — provider change, model refresh, Codex auth, every toggle that reveals dependent fields. Each rebuilds the whole tab, so the active tab has to be held on the class and restored, or the user is thrown back to tab one on every toggle. A few lines, but it is the failure mode of this pattern.

Cost: roughly 60–100 lines in `display()` plus a small block in `styles.css` — the plugin has no tab idiom yet, so that markup would be new. Mobile is worth a thought: three tabs fit a narrow pane better than one long scroll, but I have not tested it there.

Happy to send the PR if you want it in this shape, or to settle the grouping first — the grouping is the part where your judgement matters more than mine.

green-dalii:
[...]

PR #731, opened 2026-09-16
docs(roadmap,memory): record the v1.28.0 phase schedule and its couplings

green-dalii:
docs(roadmap,memory): record the v1.28.0 phase schedule and its couplings

Planning pass over the v1.28.0 milestone. Docs only; Gate 1 green.

## What the pass found

**Four unscheduled items**, one of them a broken milestone link:

- **PR #728** — the MAJOR `@ai-sdk/openai-compatible` bump that superseded #706.
  #706 was closed *still carrying* `v1.28.0 MINOR`, so the milestone pointed at a
  dead PR and the live one had none. Now on `v1.28.0 MINOR`.
- **Issue #725** (`deleteEmptyStubs` leaves incoming links) and **PR #726**
  (DocTpoint's lint fix) — both PATCH-shaped, now on `v1.27.x PATCH`.

**Two couplings that were not recorded anywhere**, and both change the order:

- **#729 must ship with #664.** #729 adds Related entries; #664 says the Related
  lists already grow ~2 per source and are never pruned. Designed separately, one
  raises the ceiling while the other leaves the floor open — and the measurement
  that would catch it is exactly the one each change would blame the other for.
- **#729 sequences after #668.** #729 adds a settings toggle defaulting on; #668
  restructures the settings tab. Landing the toggle first means re-homing it twice.

## What went in

**ROADMAP** — `### Phase schedule (2026-09-16)`: four dependency-ordered phases
(not priority-ordered), the two binding couplings, and the closed scheduling gap.
Phase 1 and phase 4 are parallelisable; phase 2 is a chain behind #603.

**MEMORY** — `### Coupling constraints` recording the *reason* each coupling
binds (the schedule table stays in ROADMAP, per the orthogonality rule), and a
**`### Resume point (post-compact handoff)`** stating where to read in order, the
exact state at handoff, and the next unblocked step with its pass condition.

## Note on #728

Its branch already carries the `Dependabot lockfile sync` commit
(`8d17584b`, `github-actions[bot]`) — the third independent production
confirmation of #718's fix, and the first on a MAJOR bump. A `@dependabot rebase`
was requested because the PR is `DIRTY`; after it lands, Gate 1 on the corrected
head arrives as `action_required` and needs one approval.


green-dalii:
Merging. Docs only; Gate 1 green, and `Sync package-lock.json` shows SKIPPED because that workflow is gated on the Dependabot actor.

Issue #662, opened 2026-09-09
The page index is rebuilt per item and per written page — a run-scoped index would replace nine call sites


DocTpoint:
`getExistingWikiPages` reads every wiki page in full (frontmatter parse plus body slice) on every call. `WikiEngine.getExistingWikiPages()` wraps it in a 5 s TTL cache that is invalidated on every vault write, so during an ingest the cache never holds; and eight call sites import the function directly and bypass it anyway (`path-resolution.ts:151`, `related-links.ts:34`, `related-page.ts:67`, `contradictions.ts:37`, `source-analyzer.ts:766`, `fix-dead-link.ts:135`, `fix-polluted-page.ts:31`, `wiki-engine.ts:1094`).

The hot sites: `resolvePagePath` rebuilds the index once per extracted item, `applyRelatedLinks` once per written page. On a ~2 000-page vault and an ingest touching ~13 pages that is on the order of 40–50 full vault walks per note. Read from the code, not timed — the first step of the PR is counting calls and time per ingest with the dev-instrument runner.

**Proposal:** one index built at `ingestSource` start, carried on the context, and **updated by the write path** (the writer knows what it wrote), replacing the file probe in `path-resolution` that today backstops same-run pages. Lint and maintenance paths read the same object lazily. Behaviour-preserving, checkable against the test suite; the contract change is only that same-run pages appear in the index instead of being probed.

Shape: a layer around the existing `getExistingWikiPages`, not a rewrite of it — #653's `displayTitle` addition stays where it is. The PR will sit on top of #647, which touches the same spot in `wiki-engine.ts`. Happy to build it if the direction suits you.

DocTpoint:
Measured on one ingest (627 pages, 429 s): 23 calls at 27 ms each, 0.1 % of the run — no speed to gain; what remains is the structural tidy-up (one index, same-run pages listed instead of probed), to be built when the embedding index needs it, not before.

green-dalii:
Fixed by `528d6e0a` (#757, @DocTpoint), which landed on main.

Closing this by comment rather than leaving it open for a commit keyword: the fix came in as a contributor PR whose body carried no `Closes` line, so the usual path — the commit closing the issue — never fired. The alternative was to leave a fixed bug open until some unrelated commit happened to reference it.

[...]

PR #732, opened 2026-09-16
docs(roadmap,memory): complete the #603 / #662 write-path design pass

green-dalii:
docs(roadmap,memory): complete the #603 / #662 write-path design pass

Phase 1's design item. Docs only; Gate 1 green. No implementation.

Both issues describe a surface problem accurately and enumerate it inaccurately.
The enumeration is what decides the design, so it was re-measured from source
rather than trusted — and each issue contains one claim that does not hold.

## The gate bundles four concerns, not one

`createOrUpdateFile` (`wiki-engine.ts:1873-2045`) does pollution correction
(`:1884-1932`), heading/provenance normalization (`:1940-1945`), retry IO
(`:1947-2045`) and notification + cache invalidation (five exit points). The gate
offers all four or none, and nothing in between is reachable.

## The decisive evidence was already in the codebase

`wiki-engine.ts:845-851` documents a **deliberate** bypass: the PDF sidecar skips
the gate because going through it "would fire `onFileWrite` + `invalidatePageCaches`,
which could trigger auto-ingest cascades if the source folder is watched."

That comment is the design conclusion, written early by whoever needed it first:
one gate for every write is the wrong shape, because concern 4 is sometimes
actively harmful.

## Corrections — each issue had one claim that does not hold

**#603** claims six bypassing writers:

- ❌ **`log-writer.ts` is not one.** Its header says the vault calls "live in
  WikiEngine's tryReadFile / createOrUpdateFile. LogWriter receives these as
  injected." A false positive in the list.
- ⚠️ **Missed `lint/fix-runners.ts:133` and `:604`** — they use
  `vault.adapter.write`, below Obsidian's own eventing. The worst mechanism found.
- ⚠️ **Missed `markPageComplete` (`:349`)** — the engine writes a wiki page's
  frontmatter outside its own gate.
- ⚠️ **Missed `preparation.ts:68`**, which writes wiki pages, not just the
  `log.md` sibling the issue listed.

**#662** claims eight direct importers:

- ❌ **`contradictions.ts:37` does not exist** — the file is
  `contradiction-gates.ts` and calls neither. Measured: **7**.
- ✅ The 5 s TTL is right (`PAGES_CACHE_TTL_MS = 5000`) and is invalidated at
  `:447` plus five gate exit points.
- ⚠️ **Missed `wiki-engine.ts:1104`**, which calls the module function directly and
  bypasses the engine's own cached wrapper (`:2120`) — the *same self-bypass
  pattern* as #603, in a second subsystem. That is the real finding: the engine
  does not consistently use its own front doors.

## Recommendation — split, do not funnel

`rawWrite` (retry IO) / `pageGuard` (pollution + normalization) / `notify`
(`onFileWrite` + cache invalidation), each declared per call site. Tiered:
**five real violations in four files**, five sites that need only to declare
`guard: false`, and nine that are out of contract and should stay that way.

**For #662 the issue's mechanism is the point:** the TTL cache is invalidated *by
the writes the ingest itself performs*, so it cannot hold during an ingest however
many callers use the wrapper. Routing the seven direct callers through the
accessor is hygiene; the fix is a **run-scoped index updated by the writer** — the
writer knows what it wrote, so it should tell the index instead of invalidating it.

## Three open questions, all needing a decision before code

Whether `pageGuard` stays inside the gate as a superset or becomes a separate
declared layer (the latter touches ~14 call sites); whether tier B's declaration
is type-level (the only version that cannot be forgotten) or a call-site option;
and whether `fix-runners.ts`'s `adapter.write` is a deliberately lower-level
mechanism or just an older idiom — the code does not say, and the answer changes
the fix.

ROADMAP's phase-1 row is marked done and the ordering-decision section now carries
the phase-2/3 gate opening with the corrected counts.


green-dalii:
Merging. Docs only; Gate 1 green.

PR #733, opened 2026-09-17
deps(zod): migrate to zod 4, replacing the deprecated `.passthrough()`

### LINKED ISSUE #669: Migrate to zod 4: 17 `.passthrough()` sites in output-schemas.ts + wire-body regression coverage

green-dalii:
deps(zod)!: migrate to zod 4, replacing the deprecated `.passthrough()`

Closes #669. Five-Gate green (4170/4170, 297 files). The two file snapshots that pin the wire bytes are byte-identical before and after (`ae78a63fb9e3f708` / `93ed585ccf0f0c57`).

## Why this was provable instead of hopeful

`strict-schema.ts` imports nothing from zod — it normalises at the `ai` `jsonSchema()` boundary. The project's strict-dialect rewriting is therefore version-agnostic, and the file snapshots built for #658 are the proof rather than something this change had to invent. The migration is a dependency swap whose safety net already existed.

Two corrections to the issue's enumeration: **14** real `.passthrough()` call sites, not 17 (16 line hits, 2 of them comments); and `.loose()` is also a method, so the diff stays one token per site instead of switching every site to `z.looseObject()`.

## The one behaviour change, and why it is not a regression

zod 4's converter emits `additionalProperties: false` where zod 3's `.passthrough()` emitted `true`. That reaches the wire on the **plain** tier (`json_schema`), the default mode for every schema caller.

Kept rather than reverted, on three independent grounds:

- **Nothing reads unknown keys off a parsed response.** `confidence` / `score` appear only in `output-schemas.ts` comments; every other `score` in the tree belongs to unrelated subsystems.
- **Client-side tolerance comes from the zod parse, not the wire.** zod 4's plain `z.object()` already strips extras *without throwing*, so `.loose()` is not what provides the tolerance — the wire value cannot cause a parse failure either way.
- **`false` is not novel.** The strict tier has always shipped `false` on every object node.

`additionalProperties` on the wire controls server-side grammar only. A tighter value makes extras structurally unemittable on grammar-decoding backends, instead of letting them be produced and then discarded.

The old assertion was **replaced, not relaxed**: it read a raw `Output.object()` call the plugin never makes, so it never guarded the production path. The replacement asserts `required` (DocTpoint's measured #463 requirement) and the `additionalProperties` value on `buildOutputArgs`'s own output — so it now fails loudly if a future SDK changes the value we depend on.

`strict-schema.test.ts`'s zod 3-vs-4 comparison had to change (zod 3 is no longer installed). The v3 bytes are frozen inline and the assertion becomes **stronger**: the *recorded* v3 body must still normalise to the same bytes as the *live* v4 schema, which proves the normaliser is representation-agnostic rather than merely self-consistent.

## Gate 2: no side effects

`output-args.ts` changes are **comments only** — zero non-comment added lines, verified from the diff. The 14 schema edits are parse-level. No state mutation, no error-path change, no new call site.

API signature, settings schema, file format, command IDs, Obsidian API: unchanged. `buildOutputArgs`'s signature is untouched. The dependency moves 3.x → 4.x (runtime dep, bundled). The plain tier's wire `additionalProperties` is the one intentional change, documented and pinned. Both lockfiles regenerated.

## Gate 4: Performance

| Dim | Status | Notes |
|-----|--------|-------|
| CPU | ✅ | No runtime code added; no new tree walk. The strict path's memoised normaliser is untouched and the plain path gained nothing. |
| Memory | ⚠️ | **`main.js` 4,127,799 → 4,514,616 bytes (+386,817 / +9.37 %)**. Attributed by building `origin/main` (zod 3.25.76) in a throwaway worktree and diffing against HEAD (zod 4.6.5), so the figure excludes the #707/#708/#727/#730/#732 merges. Almost entirely zod 4 bundling its core, its JSON Schema converter and the default English locale (no non-English locale is pulled in). Irreducible without rewriting every schema off `z.object` onto `zod/v4/core`. |
| IO | ✅ | None. |
| Network | ⚠️ | Same +9.37 % as a one-time download delta; no per-call change. |
| Token | ✅ | None — same schemas, same prompts, same bytes on the wire (proven by the unchanged snapshots). |


green-dalii:
Merging. Gate 1 green. The PR body was reformatted — it had been the commit message verbatim, whose ~80-column terminal wrapping GitHub renders as a hard break in every paragraph. Content is unchanged; only the wrapping differs.

PR #730, opened 2026-09-16
docs(roadmap,memory): record the v1.28.0 direction and de-bloat MEMORY

green-dalii:
Records the converged v1.28.0 direction, and revises MEMORY so it stays usable. Docs only — Gate 1 green locally. `shazam_verify` PASS.

## Orthogonality

Split per AGENTS.md §"Documentation Architecture" — one fact, one place:

| File | Owns | Does NOT carry |
|---|---|---|
| **ROADMAP** | the planning view: v1.28.0 scope groups, the hardening-before-reader ordering, open decisions | mechanism, root-cause evidence, algorithms |
| **MEMORY** | the *why* and the *how*: `## Design record — cross-source relations` | the window schedule (points at ROADMAP) |

## Added

**ROADMAP** — a new `## v1.28.0 MINOR — Design track` (the file had no v1.28.0 content at all): five scope groups with one-line justifications, the 2026-09-16 ordering decision, and three open decisions.

**MEMORY** — a `## Design record — cross-source relations (#729)` section carrying the root-cause evidence, the three-mechanism (M0/M1/M2) table, the adaptive budget tiers, the reserved-with-backfill algorithm, **an ordered six-phase implementation plan written so a contributor with no context can build it**, the staged acceptance criteria, the files to touch and the open questions. Plus a canonical `Floor / ceiling asymmetry` row in Key design decisions, a new Current state block, and six durable lessons.

## Revised — MEMORY 951 → 782 lines (−17.8 %)

Removed duplicates and archaeology, not content:

- **`## Process invariants` no longer restates AGENTS.md.** Six-Gate, the Git workflow, the mandatory merge sequence and the Bot double-lint invariant are canonical in AGENTS.md; MEMORY now carries a pointer table plus only the rules whose *reason* is not obvious from the rule.
- **Three superseded `## Current state` snapshots** (09-05 / 09-06 / 09-15, 83 lines) collapse into one pointer. The per-version record is canonical in CHANGELOG, and a stale "Next:" list reads as current — worse than no list.
- **The 2026-08-30 Hermes section** dropped seven "borrow unchanged" table rows, four now-completed "Next moves" (verified: #575 closed and #577 merged on 2026-08-31) and a session pointer to a `/tmp` file that no longer exists. Its three durable design judgments are kept, in one place.
- **One duplication introduced by an earlier pass of this same branch** (the Hermes first-principles anchors, duplicated by the new standing-judgments list) is removed. Flagged rather than quietly fixed: it was this branch's own regression.

[...]

PR #734, opened 2026-09-17
docs(agents): a PR body is not the commit-message file

green-dalii:
docs(agents): a PR body is not the commit-message file

Records a rule that had already recurred. `git log` wants ~80-column wrapping, but GitHub renders a single newline in a PR or issue body as a hard break — so a wrapped paragraph arrives on the web as a column of ~80-character fragments. Same content, two renderings.

The symptom reappears whenever `--body-file` is handed the same file that was passed to `git commit -F`, which is what #733 did (and #730 and #727 before it).

The entry carries the check so the next one is caught before posting rather than after:

```
awk 'length>0 && length<88' <body-file>
```

That should return only markdown-structure lines (headings, list items, table rows), never prose.


green-dalii:
Merging. One note on the rule's own self-check: a single short prose line is a false positive — the signal is *several consecutive* short lines. The check stays useful as a coarse pre-post guard.

PR #736, opened 2026-09-17
feat(llm): custom request headers, an OpenCode preset, and a Responses variant

### LINKED ISSUE #723: OpenAI-compatible client never sends x-opencode-session → OpenCode Go rejects all requests (v1.27.1)
## Bug Description

Using **OpenCode Go** (`https://opencode.ai/zen/go/v1`) as a Custom OpenAI-Compatible provider, every request is rejected because the client never sends the `x-opencode-session` header that OpenCode Go requires for routing:

```
Connection test failed: AI_APICallError: Error from provider (Console Go):
Request is missing x-opencode-session and cannot be routed efficiently.
Please see https://opencode.ai/docs/go/#where-can-i-use-it
```

This is not an authentication failure — the server returns this routing error instead of `AuthError`, so the API key itself is accepted. The plugin is currently unusable with an OpenCode Go subscription.

## Environment

- **Plugin version:** 1.27.1
- **Obsidian version:** 1.13.4
- **OS:** macOS 13.7.8 (x86_64)
- **Wiki size:** ~25 pages

## LLM Configuration

- **Provider:** Custom (OpenAI-compatible)
- **Model:** deepseek-flash
- **API endpoint (if custom):** https://opencode.ai/zen/go/v1
- **Local or cloud:** Cloud

## Steps to Reproduce

1. Settings → Karpathy LLM Wiki → Provider = Custom OpenAI-Compatible, Base URL = `https://opencode.ai/zen/go/v1`, API key = OpenCode Go ("Console Go") key, Model = `deepseek-flash`.
2. Click **Test connection**.
3. See the error above.

## Root cause (investigated)

OpenCode Go documents three client requirements ([Where can I use it?](https://opencode.ai/docs/go/#where-can-i-use-it)): send typical coding-agent traffic, **identify with its own User-Agent rather than a generic SDK name**, and **send a stable session id in `x-opencode-session` per conversation**. The bundled OpenAI-compatible client satisfies neither of the last two:

- `grep -c x-opencode main.js` → **0**. The header is never sent, and there is no "custom headers" setting to add it.
- The wire User-Agent is AI-SDK's default, i.e. exactly the generic-SDK case the docs ask to avoid: `ai-sdk/openai-compatible/2.0.62 ai-sdk/provider-utils/4.0.40 runtime/browser`. (`karpathywiki/<version>` is only set on the Codex OAuth path.)
- In `main.js` (v1.27.1, lines 53111–53150) class `OpenAICompatSdkClient`, method `getProvider()` calls `createOpenAICompatible({ name, baseURL, apiKey, fetch, includeUsage, supportsStructuredOutputs, transformRequestBody })` — no `headers`. The bundled `@ai-sdk/openai-compatible@2.0.62` **does** support `headers` and merges them into every request, so this is a small change.

I verified this with a dependency-free Node `vm` harness that loads the bundled client and captures the outgoing headers on both fetch paths (`window.fetch` / `streamWithFallback`, and Obsidian's `requestUrl` bridge used by "Test connection"). With `headers` added, `x-opencode-session` appears on both paths; the same harness reports `FAIL` on a stock 1.27.1 bundle, so it is a real regression check rather than a rubber stamp.

## Expected Behavior

Requests to OpenCode Go are accepted, like in other validated clients.

## Actual Behavior

All LLM calls fail with the routing error; no wiki generation is possible.

## Suggested fix

1. Send a stable per-conversation id in `x-opencode-session` (a UUID created once per conversation is enough — OpenCode uses it for routing and prompt caching).
2. Send a self-identifying User-Agent (`karpathywiki/<version>`) on the OpenAI-compatible paths too, not only on Codex.
3. Ideally expose both via a **"Custom headers" setting** and/or ship an `opencode` / `opencode-go` preset (base URLs + defaults). That also unblocks other gateways with similar requirements.

Minimal local patch that works — scoped to `*.opencode.ai` so other providers are untouched:

```js
const openCodeHost = new URL(effectiveBaseURL).hostname;
const openCodeHeaders = openCodeHost === "opencode.ai" || openCodeHost.endsWith(".opencode.ai") ? {
  "User-Agent": "karpathywiki/1.27.1",
  "x-opencode-session": "<stable uuid>"
} : void 0;
// createOpenAICompatible({ ..., headers: openCodeHeaders })
```

## Screenshots/Logs

Headers actually sent by stock 1.27.1 (captured from the bundled client in the vm harness):

```
POST https://opencode.ai/zen/go/v1/chat/completions
authorization:      Bearer <redacted>
content-type:       application/json
user-agent:         ai-sdk/openai-compatible/2.0.62 ai-sdk/provider-utils/4.0.40 runtime/browser
x-opencode-session: <absent>        ← the server rejects this
```

## Additional Context

- Related but different: #17 (CORS preflight failures on OpenAI-compatible endpoints). This one is a missing request header, not a preflight failure — and it happens even when the request reaches the server fine.
- OpenCode tracks this class of client publicly: on the same docs page, Hermes, jcode and Kilo Code CLI are listed under "Validated Clients" together with the session-header fix that unblocked them.
- Also worth checking whether the model-list fetch (`GET /zen/go/v1/models`) should send the same identification headers. It is public today, so it is not blocking.
- Happy to provide the vm verification harness (single Node file, no dependencies) or the full patch diff I applied locally.

### LINKED ISSUE #735: Custom OpenAI-compatible endpoints have no Responses API path — `@ai-sdk/openai-compatible` speaks only Chat Completions
## What

The plugin reaches three provider paths, and only two of them can speak OpenAI's Responses API:

| Path | API | Evidence |
|---|---|---|
| OpenAI (standard) | **Responses** | `@ai-sdk/openai@3.0.86`'s callable signature is `(modelId: OpenAIResponsesModelId)`; `.languageModel()` and `.responses()` likewise. `openai-sdk-client.ts:110` calls `provider(modelId)`, so the standard path is already Responses. |
| OpenAI Codex OAuth | **Responses** | `openai-codex/request-adapter.ts:42` → `CODEX_RESPONSES_URL` |
| Custom OpenAI-compatible | **Chat Completions only** | `@ai-sdk/openai-compatible@2.0.62` `dist/index.d.ts:242` — `languageModel(modelId: CHAT_MODEL_IDS)`, no `responses` method |

So a user pointing the plugin at a gateway that offers `/v1/responses` cannot reach it through the custom path: `createOpenAICompatible` has no Responses model, so there is nothing to switch on.

## Two routes

1. **A dependency that speaks generic Responses.** None of the currently bundled `@ai-sdk/*` packages does. A new dependency for this alone is a hard sell.
2. **Generalise the project's own Codex Responses client.** `openai-codex/request-adapter.ts` is **43 lines**: it takes a request, rewrites the headers and body into the Responses shape, and points at a URL. That shape is already written, already tested and already maintained — this is the asset to reuse, not a client to write from scratch.

## The question this issue deliberately does not answer

Which route is right depends on what users actually have, and that is unanswered. A gateway offering both `/v1/chat/completions` and `/v1/responses` gains little from the latter while chat completions works. The Responses API's real advantages — reasoning items, server-side state, `instructions` — are mostly OpenAI-hosted semantics that third-party gateways reimplement partially. Pointing a generalising client at a partially-compatible `/v1/responses` could be worse than the chat-completions path it replaced.

So the first step is not implementation: it is identifying a concrete endpoint that rejects the chat-completions path, or declining to build this.

## Related

Found while auditing the provider surface for #723 (custom headers). #723 itself does **not** depend on this — OpenCode Zen/Go is a chat-completions gateway, so its preset stays on the existing path.

green-dalii:
feat(llm): custom request headers, an OpenCode preset, and a Responses variant

Closes #723. Closes #735. Gate 1 green (4186/4186, 299 files). Reported by @aisahpA.

## What the reported bug actually needed

OpenCode Zen/Go rejects every request with "Request is missing x-opencode-session and cannot be routed efficiently". The fix is not an OpenCode special case: the plugin sends the AI SDK's generic User-Agent on the OpenAI-compatible path, so any gateway that routes on client identity sees an anonymous SDK. That is why the change is a header mechanism plus one preset, rather than a special case for one gateway.

## Headers are applied in the fetch, not through the provider

`createOpenAICompatible({ headers })` does work for ordinary headers — but it **cannot set `User-Agent`**. The SDK ends its option handling with `withUserAgentSuffix(headers, 'ai-sdk/openai-compatible/<v>')` (`@ai-sdk/openai-compatible@2.0.62/dist/index.mjs:1747`), which replaces whatever the caller supplied. Verified against a stub fetch: an explicit `User-Agent` still produced `ai/6.0.230 ai-sdk/provider-utils/...` on the wire.

The same holds for `@ai-sdk/openai`, so **both** clients set the composed map on the request after the SDK has built its own. The wrapper returns the fetch untouched when there are no headers, so the ten providers that never asked for any keep the exact call path they had — asserted by a test that a header-less request carries no `karpathywiki` UA.

## Three sources, and why the order is what it is

`compat-headers.ts` composes: plugin identity (`karpathywiki/<version>`) → the preset's `defaultHeaders` with `{sessionId}` filled → the user's own, which win. The AI SDK adds `Authorization` from `apiKey` before the custom map and documents the map as overriding, so this ordering is what reaches the wire.

The session id is generated lazily and at most once — only a preset that actually references the placeholder pays for it.

## Three presets instead of one

- `custom` → renamed **"Custom OpenAI-Compatible (Completion)"**.
- `custom-responses` → new. `apiShape: 'responses'` routes to `OpenAISdkClient`, whose `provider(modelId)` call *already* resolves to the Responses model, so `/v1/responses` against a user-supplied base URL needs no new dependency and no bespoke adapter. A separate preset rather than a toggle, so both shapes can coexist — a gateway may support one, both, or neither, and a user who guessed wrong can switch back without re-entering the URL.
- `opencode` → new: base URL preset plus the required session header.

**Preset ordering:** `opencode` sits after `glm`, with the fixed-baseURL providers, rather than beside the custom presets — it is a named hosted endpoint the user never types a URL for. The custom group stays at the bottom.

## Relationship to #735 — and a correction to it

#735 was opened on the premise that custom OpenAI-compatible endpoints have no Responses API path. **This PR builds that path, so the premise no longer holds and the issue closes with this merge.**

#735's own technical analysis was partly wrong, and that is worth recording rather than quietly dropping. It proposed two routes — take on a dependency that speaks generic Responses, or generalise the project's 43-line Codex request adapter. Neither was needed: `@ai-sdk/openai` ships `.responses(modelId)` and accepts `baseURL` and `headers`, so the preset is a routing decision, not a client to write. The implementation turned out to be roughly ten lines because the capability was already bundled.

What #735 could not settle — whether a real third-party endpoint needs this — is an empirical question that no issue can answer. It is asked of @aisahpA above, and if a report comes back that `/v1/responses` on some gateway behaves worse than chat completions, that is a new defect with a concrete endpoint, not a reopened capability gap.

## The custom-headers field sits below the Base URL

Per user direction, and deliberately **not** in the bottom Advanced panel: the two are configured together, and a gateway needing extra headers is usually why the user opened this section. Hidden for the Anthropic-compatible provider, which routes to `AnthropicSdkClient` and would ignore it — a setting that silently does nothing is worse than one that is absent.

## Gate 2: no side effects

`compatHeaders` is pure (no Obsidian import, no `crypto` call of its own). The fetch wrapper is a no-op closure for providers with no headers. No state mutation, no error-path change. The two `createLLMClientFromSettings` variants now share one `compatRouteOptions` helper instead of duplicating the option literal.

New settings field `customHeaders` is optional and defaults to unset. Two new `ProviderConfig` fields are optional. No provider id removed; `custom` keeps its id and only its display name changed. `OpenAICompatSdkClientOptions` and `OpenAISdkClientOptions` gained optional fields. Command IDs and the Obsidian API surface are untouched.

## Gate 4: Performance

| Dim | Status | Notes |
|-----|--------|-------|
| CPU | ✅ | Headers composed once per client construction, not per request. The wrapped fetch adds one `Headers` construction per request **only** for providers that set headers; the others get the original function object back. |
| Memory | ✅ | One small string map per client. The session id is generated lazily, so presets that do not use it allocate nothing. |
| IO | ✅ | None. |
| Network | ✅ | Two extra headers (`User-Agent`, and `x-opencode-session` for OpenCode) on providers that opted in; zero change otherwise. |
| Token | ✅ | None. No prompt or response shape change. |

## Testing

Three test files, 14 new assertions: header parsing (comments, blank lines, first-colon split, invalid-line counting), the three-source precedence, that the headers genuinely reach the request, that a header-less request is unchanged, and that `custom-responses` routes to the Responses client while `custom` stays on the compat client.

A manual E2E build is available on request. Pointing the `(Responses)` preset at a gateway that only speaks chat completions is expected to fail — that is a property of the gateway, not of this routing.


green-dalii:
@aisahpA — thank you for the report. Your investigation was the reason this was quick: the diagnosis (no `headers` passed to `createOpenAICompatible`, a generic AI-SDK User-Agent on this path, and no setting to add either) was correct, and it generalised further than OpenCode.

Implemented in #736, and it would be very useful if you could test it against your OpenCode Go key.

**What changed**

- **Custom Headers setting**, directly below the API Base URL field, one `Name: value` per line. Blank lines and `#` comments are ignored, and only the first `:` splits, so a value may contain one (URLs, `Bearer …`).
- **The plugin now identifies itself.** `User-Agent: karpathywiki/<version>` is sent on the OpenAI-compatible path. Worth noting: passing `headers` to `createOpenAICompatible` is not enough for this — the SDK appends its own generic name afterwards and wins — so the headers are applied in the fetch wrapper instead. Your point 2 is fixed for every OpenAI-compatible provider, not only OpenCode.
- **An `OpenCode (Zen / Go)` preset** with the base URL pre-filled, and `x-opencode-session` sent automatically with a stable per-conversation id. Your point 1.
- **`Custom OpenAI-Compatible` renamed to `… (Completion)`**, with a new `… (Responses)` preset beside it, for gateways that expect `/v1/responses`.

**What would help**

1. Does **Test connection** succeed where it previously failed with the routing error?
2. Does a real ingest complete — specifically, is the session id accepted across multiple requests in one run? OpenCode uses it for routing and prompt caching, so a multi-page ingest is the more meaningful test than a single call.
3. Any remaining header or routing error, verbatim, if something still fails.

A debug build is available if building from the branch is inconvenient — say so and I will attach one.

One caveat on the new `(Responses)` preset: pointing it at a gateway that only speaks chat completions is expected to fail. That is a genuinely open question (#735), not a defect in this PR — so if you happen to know whether OpenCode's Zen/Go endpoint exposes `/responses`, that would be useful too.

aisahpA:
_Disclosure: this review was drafted with an AI assistant and reviewed before posting. The measurements below were run locally against a build of this branch._

Verified end to end on a build from this branch, in a real vault with a real Go key: provider = the new `opencode` preset, `https://opencode.ai/zen/go/v1`, model `deepseek-flash`. #723's error is gone, ingest and Query Wiki both complete, and both fetch paths carry the header — a dependency-free Node harness (bundle in `vm` + stub fetch, driving the real `createLLMClientFromSettingsSync` → `getProvider`) shows the preset's UUID on the wire, stable within one client, with `Authorization` intact; `custom-responses` resolves to `.../v1/responses`. Four things worth fixing before merge:

1. `src/types.ts:1192` — the `opencode` preset sets `supportsStructuredOutputs: true`, and it costs a wasted call on every model. Measured on a real Go key: ingest logged `[OUTPUT-MODE-DEBUG] … mode=json_schema`, then `[OUTPUT-MODE-DEMOTE-DEBUG] tier=json_schema → tier=json_object. responseBody="… Console Go: Upstream request failed: [invalid_request_error] This response_format type is unavailable now"`, then `mode=json_object` for the retry and every call after. `OutputModeProber` recovers correctly, but the first schema-bearing call per (baseURL, model) per plugin session always burns a 400 and puts an alarming error in the log. That is exactly the rule documented at `types.ts:184-194` and `openai-compat-sdk-client.ts:132-143`: cloud compat backends get `json_object`, and every other cloud preset (openrouter / deepseek / kimi / glm / gemini) leaves the flag unset. Suggest dropping it for `opencode` unless Go was measured to accept the schema form for some subset of models.

2. `src/llm-sdk/compat-headers.ts:60-66` — custom header names and values are never validated, and the wrapper then calls `Headers.set`, which throws (`TypeError: Headers.set: "My Header" is an invalid header name.`) inside the fetch wrapper, so it surfaces as an opaque request failure with no hint about which line caused it. The settings UI only counts lines with no colon, so `My Header: v` passes validation and explodes at request time. Suggest an RFC 7230 token check on the name and a control-character check on the value, counted as `invalid` so the existing description line reports it.

3. No test wires the preset defaults through the factory to the wire: deleting `presetHeaders: PREDEFINED_PROVIDERS[provider]?.defaultHeaders` (`create-llm-client.ts:192`) leaves all four header test files green (I measured 39/39). One factory-level test — provider `opencode` plus a stub fetch, asserting a UUID-shaped `x-opencode-session` and a `karpathywiki/` `User-Agent` — would close that gap, which is the mechanism #723 actually depends on.

4. Not a regression from this PR, but it makes the preset look better than it is, so worth knowing before advertising it: `opencode.ai` fails the CORS preflight — I probed `OPTIONS` with `Origin: app://obsidian.md` and got `404` with no `Access-Control-Allow-*` headers, with or without `x-opencode-session`, because `authorization` alone already triggers it. So `window.fetch` never works for that host and every streamed call falls back: `[STREAM-FETCH] try streamingObsidianFetch (window.fetch)` → `[STREAM-FETCH] TypeError, falling back to obsidianFetchBridge: Failed to fetch`. The fallback carries the header, so answers are correct, but they are buffered — the pane stays empty for ~20 s and then the whole answer appears at once (`[STREAM-CHUNK] total chunks forwarded: 1322` is the AI SDK draining the buffered SSE body, not network chunking). If you want that path back, skip the doomed preflight for presets that need non-safelisted headers — e.g. a `preferRequestUrl: true` flag on the `opencode` preset, in the spirit of the existing `isLocalBaseURL` shortcut — which would also remove a round-trip of latency before every streamed answer.

Two smaller accuracy notes, no code impact. (a) The identity `User-Agent` is unconditional, so the commit body's "the ten providers that never asked for headers keep the exact call path they had" does not hold in production: a `deepseek` request also carries `karpathywiki/<version>` now, because the regression test constructs the client directly and bypasses the factory. (b) `User-Agent` is a fetch-forbidden header, so on the `window.fetch` branch Chromium is likely to drop it and only the `requestUrl` bridge sends it — functionally fine (and moot for OpenCode, which always falls back), but the comment currently reads as a guarantee.

Unrelated pre-existing observation: `createLLMClientFromSettings` (async) is tree-shaken out of `main.js`, so only the Sync factory ships; the routing this PR added to the async one is test-only code today.

green-dalii:
@aisahpA — this is the most useful review this project has received, and I mean that specifically: three of the four findings are defects in code I wrote in this PR, and the third one is a test gap that my own four test files were actively hiding. Thank you.

All three are fixed in `0f2ff434`, and the fourth is filed as #741.

## 1 — the preset flag

You were right, and the neighbouring presets already proved it: `supportsStructuredOutputs: true` appears only on `ollama`, `lmstudio` and `custom` — the local and user-supplied servers — and on **no** cloud preset. I set it by copying `custom` without reading the rule two fields above it. Dropped, with the reason recorded at the preset so it does not come back.

## 2 — header validation

Fixed as you suggested: the name is checked as an RFC 7230 token and the value for control characters, both counted as `invalid` so the settings line reports them. TAB stays legal per §3.2.6. The control check is a loop rather than a regex because the project forbids `eslint-disable` and a character class would have needed one.

## 3 — the test gap, mutation-tested

Fixed, and verified the way you found it. `preset-headers-factory-wire.test.ts` drives the real factory (`createLLMClientFromSettings` → `compatRouteOptions` → the client) with the requestUrl bridge mocked. Then I deleted `presetHeaders: PREDEFINED_PROVIDERS[provider]?.defaultHeaders` again and confirmed: **3 of its cases fail, the other three header files stay green at 21 passed** — your 39/39 reproduced exactly. That is the file the mechanism now depends on.

## 4 — deferred, deliberately

Filed as #741 with your `preferRequestUrl` suggestion and the measurements, including the `[STREAM-CHUNK] total chunks forwarded: 1322` line, which is the tell that it is a buffered drain rather than chunking. I kept it out of this PR because it is not a regression, it is a different mechanism, and folding a UX change into a correctness fix makes both harder to attribute later. If you disagree — if you think the preset should not ship without it — say so and I will move it in.

## Both accuracy notes were correct and are corrected in place

**(a)** is the one that stings: "the ten providers that never asked for headers keep the exact call path they had" is false in production, because the identity header is unconditional. My regression test asserted the opposite because it constructs a client directly and so bypasses the factory — a test passing against a claim that is only true of itself. The new factory test asserts the real behaviour (a `deepseek` request carries `karpathywiki/<version>`), and the comments no longer promise what the browser path cannot deliver **(b)** — `User-Agent` is best-effort on `window.fetch` and guaranteed on `requestUrl`.

Your unrelated observation also holds and is worth recording: `createLLMClientFromSettings` is tree-shaken out of `main.js`, so only the Sync factory ships and the routing I added to the async one is test-only today. The shipped path has the same routing, so the fix is unaffected — but it does mean the async branch deserves a comment saying so, which I will add with the next change to that file rather than widening this one.

green-dalii:
Merging. Body keyword scan: `Closes #723` and `Closes #735`, both intentional. Review event is a comment rather than `--approve` because GitHub blocks self-approval on an own PR.

The substantive review came from @aisahpA, who verified the original #723 failure end to end on a real vault with a real Go key and then raised three defects in this PR's own code — all fixed in `0f2ff434` and mutation-verified. The fourth finding is filed as #741. @DocTpoint's requested review is still open; merging on the strength of that verification and the response above rather than leaving the fix sitting.

PR #737, opened 2026-09-17
docs(agents): verify the commit, not the command; and how to file a capability issue

green-dalii:
docs(agents): verify the commit, not the command; and how to file a capability issue

Two process rules from the #735 / #736 cycle, both of which cost a round trip.

**Verify the commit, not the command.** A commit that reports success can still carry the wrong content. `git commit --amend -F <msg>` without a prior `git add` commits the *message* plus whatever was already staged — so a fix made immediately before it is left behind while the message announces it. That is worse than a missing commit: the description disagrees with the contents, so a reader who diffs the change concludes they misread, and the audit trail itself is what got corrupted. #736's body claimed a preset-ordering change its diff did not contain.

Two habits close it: `git status --porcelain` must be **empty** immediately after every commit, and amending to change content means `git add` first (use `--no-edit` when the message is already right). Confirm with `git show HEAD:<path>`, or re-read the branch from GitHub, rather than trusting the exit code.

**Filing issues.** An issue whose premise is a *missing capability* must name that capability as its acceptance criteria — or not be filed yet. Filing "X has no Y path" and then building Y in the next PR without linking them produces an issue that reads as unstarted work while the work is under review; a PR body written in between can even call the delivered capability "the open question", which is how it happened here. Link at the moment implementation starts (`Closes #N` in the body).

The same entry records why an issue's own wrong analysis should be **corrected on the issue rather than edited away** — the route that was not needed is information about why the right one was cheap. #735 proposed a bespoke adapter or a new dependency; the bundled `@ai-sdk/openai.responses()` meant the work was a routing decision of about ten lines.


green-dalii:
Merging. Docs only, Gate 1 green, body keyword scan returned no hits. Review event is a comment rather than `--approve` because GitHub blocks self-approval on an own PR — the audit trail is this note.

PR #739, opened 2026-09-17
refactor(constants): centralise the Related and extraction ceilings (#729 Phase 0)

green-dalii:
refactor(constants): centralise the Related and extraction ceilings (#729 Phase 0)

Phase 0 of #729 — behaviour-identical by definition, and **proven rather than asserted**: Gate 1 green, 4170/4170, and **zero snapshot or fixture churn**. That last number is the pass condition: this phase is a relocation, so churn here would be a bug rather than an update.

Part of #729. Phase 1 waits on #603.

## What moved

- `SIBLING_CAP` (`core/related-shaping.ts:67`) → `RELATED_SIBLING_CAP`
- `GRANULARITY_FIX_LIMITS` (`wiki/system-prompts.ts:198`) → `EXTRACTION_LIMITS`
- **four** inline `?? 5` literals → `CUSTOM_EXTRACTION_LIMIT_DEFAULT`

No re-export from the old homes — the point of the phase is one canonical file, so both modules now import from `src/constants.ts` and the single test importer follows.

## Two corrections to the design record, made while implementing

**The ceilings are two concerns, not three of a kind.** The record framed them as "three scattered ceilings the new budget table replaces". But `GRANULARITY_FIX_LIMITS` and the `?? 5` defaults are *extraction* limits — how many entities/concepts to ask for — while `SIBLING_CAP` caps *Related list entries*. Both happen to be granularity-keyed; nothing else about them is shared. So they are two tables, not one. Conflating them would have had Phase 2 replace the extraction limits with the Related budget, which would have been a silent behavioural change to every ingest prompt.

**The `?? 5` count was 4, not 2.** Two in `getGranularityInstruction` (`:209`, `:210`) and two in `getGranularityFixLimits` (`:226`, `:227`). Following the record literally would have replaced the documented pair and left two orphaned defaults behind — the exact scattering this phase exists to remove. MEMORY is corrected.

## `RELATED_BUDGET` ships unread, and one column must stay unread

`crossSource` is what the feature adds; it is a ceiling, never a quota. `siblings` will replace the flat cap in Phase 2 and **must not be consumed here**: it is 2 for `coarse` and 1 for `minimal`, where today every granularity gets 3. Reading it is not behaviour-preserving, so Phase 0's zero-diff condition forbids the switch. Phase 2 owns it, with the change intended, measured and tested.

## Gate 2: no side effects

Pure relocation — no logic, no ordering, no error path changed. The one behavioural question, "does anything now read a value differently", is answered by the zero-churn result rather than by inspection.

`SIBLING_CAP` was exported from `core/related-shaping.ts`, but it is a module-internal constant with exactly one importer — a test, updated here. No setting, schema, file format or command id touched.

## Gate 4: Performance

| Dim | Status | Notes |
|-----|--------|-------|
| CPU | ✅ | Same values, same call sites; two table lookups replace two module-local reads. |
| Memory | ✅ | Negligible — `RELATED_BUDGET` is five small records, allocated once at module load. |
| IO | ✅ | None. |
| Network | ✅ | None. |
| Token | ✅ | None — no prompt string changed, which the untouched prompt snapshots confirm. |


green-dalii:
Merging. Docs-and-refactor only; Gate 1 green with **zero snapshot or fixture churn**, which is this phase's pass condition. Body keyword scan returned no hits, so #729 stays open — Phase 0 is the first of six sub-phases. Review event is a comment rather than `--approve` because GitHub blocks self-approval on an own PR.

Issue #741, opened 2026-09-17
opencode.ai fails the CORS preflight, so every streamed answer arrives buffered — `preferRequestUrl` for presets that need non-safelisted headers


green-dalii:
## What

`opencode.ai` does not answer the CORS preflight, so every streamed call on the `window.fetch` path fails and falls back to the `requestUrl` bridge. Answers are correct, but **buffered**: the pane stays empty for roughly 20 seconds and then the whole answer appears at once.

Measured by @aisahpA on a real Go key (#736), with `Origin: app://obsidian.md`:

```
OPTIONS → 404, no Access-Control-Allow-* headers
```

with or without `x-opencode-session` — `authorization` alone already triggers the preflight, so removing the custom header would not help.

The runtime signature:

```
[STREAM-FETCH] try streamingObsidianFetch (window.fetch)
[STREAM-FETCH] TypeError, falling back to obsidianFetchBridge: Failed to fetch
…
[STREAM-CHUNK] total chunks forwarded: 1322
```

That last line is the AI SDK draining an already-buffered SSE body, not network chunking — which is what makes the pane empty and then full.

## Why this is worth fixing rather than documenting

The fallback is correct, so this is not a correctness bug. But the preset was added to make OpenCode work, and "works after a 20-second blank pane" is a materially worse experience than every other provider — for a reason the plugin already knows at configuration time (the base URL host). There is also a wasted preflight round trip before every streamed answer.

## Suggested mechanism

The reviewer's suggestion, which fits an existing pattern: a `preferRequestUrl: true` flag on the `ProviderConfig` for presets that need non-safelisted headers, skipping the doomed preflight — in the spirit of the existing `isLocalBaseURL` shortcut that already treats some hosts specially.

Worth deciding during implementation:

- Whether the flag is per-preset (data) or derived from a probe result (behaviour), since a host could start answering `OPTIONS` later.
- Whether it belongs on the preset at all, or on the general request path as "this host does not do CORS" — the same problem would affect any self-hosted gateway that omits `OPTIONS` handling, not only OpenCode.
- Whether losing `window.fetch` costs anything else on that path (it is the only branch that can set a `User-Agent`, so a host that needs both a self-identifying UA and streaming has no working combination today).

## Not a regression

This existed before the OpenCode preset; the preset only made it reachable in one step. Filed separately rather than folded into #736, which is otherwise a correctness fix with its own review.

aisahpA:
[...]

PR #742, opened 2026-09-18
docs(memory,roadmap): refresh the resume point for the post-compact handoff

green-dalii:
docs(memory,roadmap): refresh the resume point for the post-compact handoff

The tracked state had gone stale in the two places that matter after a context compaction: MEMORY's `## Current state` still named `main` = `7cc50bf9` and knew nothing about #736 / #737 / #739 / #741, and its resume point still pointed at the 2026-09-16 handoff — which asked the next reader to start #729 Phase 0, a thing that has since shipped.

Both are updated, and the old snapshot is demoted into the existing superseded chain rather than deleted.

- **Current state (2026-09-17)** — `main` = `33cd23a3`, 301 files / 4196 tests; what has merged into the unreleased v1.28.0 (#736 / #739 / #733 / #737 / #734, plus earlier #707 / #708 / #727); what is newly filed and unstarted (#741).
- **Resume point (2026-09-17)** — the reading order, the state, and the fact that **#603 is the next step, not #729**, because Phase 1 is blocked by it by design.
- **ROADMAP** — the three phase-1 items move from pending to done, the deferred group drops #669 / #723 and gains #741, and a new `### Merged into v1.28.0 so far` section records what has shipped into an unreleased window, since CHANGELOG's entry is written once at release.

Two operational facts are recorded in the resume point because they cost real time this session: the ruleset needs ~15 seconds to propagate after a bypass actor is added (6 s fails the merge), and this environment's GraphQL surface returns intermittent EOFs — prefer `gh api repos/.../pulls/<N>` when a read looks wrong.

Docs only; Gate 1 green at 301 files / 4196 tests.


green-dalii:
Merging. Docs only, Gate 1 green, keyword scan returned no hits. Review event is a comment rather than `--approve` because GitHub blocks self-approval on an own PR.

PR #743, opened 2026-09-18
docs(memory): re-measure the #603 write-path record, and record the #741 streaming analysis

green-dalii:
docs(memory): re-measure the #603 write-path record, and record the #741 streaming analysis

Two records, both from measuring rather than reasoning from an issue's line numbers — the same discipline that produced the #603 record in the first place and which that record, as it turns out, needed applied to itself.

**#603 re-measurement.** Three of the design record's claims do not hold against the source, and one of them removes the stated reason for the first slice.

- `link-retarget.ts:185` is not a violation: `:33` records it as deliberate, because the gate's normalization is "appropriate for a generated wiki page, not for someone's own note".
- `sources-normalizer.ts:247` and `preparation.ts:105` are not source-note writes: both operate on wiki pages (`isInFolderScope(…, wikiFolder, …)`, and `pageMap` built from `wikiFiles`). They move from tier B to tier A.
- Tier B therefore collapses to the sidecar alone — the most-documented write in the codebase. The "tier B is behaviour-preserving, so do it first" argument no longer supports a slice of that shape.

Corrected counts: **six real violations across five files**, not five across four. The revised slice 1 introduces the three layers and a required-`intent` type with **zero call-site changes**, so zero behaviour change is provable.

The re-measurement also found the strongest argument for splitting, which the original pass missed by reading the gate's callers only through the issues' lens: `logWriter.writeFile` goes through the gate, and `log.md` is not a wiki page. The "single write gate" already serves two file classes, which answers the `pageGuard`-as-superset question by itself.

Two of the three open questions are now answered (superset: no; declaration: type-level). The third — whether the sources-field repair sites should use the gate or declare `guard: false` — stays open with the reason recorded.

**#741 streaming analysis.** @aisahpA measured the preflight behaviour on the shipped build; this records the first-principles reading of it.

The mechanism classifies by URL (`isLocalBaseURL`) while the capability actually required is a function of `(host, method, header set)`. That proxy was approximately true because the observed population was bimodal, and a bimodal population hides proxy errors. It is not fixable by extending a host list: #736 made the header set user data, and a table keyed on host cannot represent a function of two variables.

The reason this took so long is recorded as a table of four independent layers that each hide the fallback — including that `esbuild.config.mjs:15` neutralises `console.debug` in production, so in the shipped plugin the degradation leaves no log, no error, and no Network entry.

Also recorded: the `User-Agent` premise in #736 is inverted, and its stated reason is out of date. The Fetch standard removed `User-Agent` from the forbidden list (whatwg/fetch `dab09b0`); Chromium implemented that and then reverted it (`079b3a3` reverts `c7d5f1c6`), so the correct statement is spec-permitted but Chromium-policy-suppressed. The plugin's identity header therefore reaches the wire only on the `requestUrl` branch — exactly the hosts that are not the default.

`opencode` failing is not a #736 regression; the regression is the free-form header field, which lets any user silently downgrade their own streaming on any provider.

Docs only. Gate 1 green at 301 files / 4196 tests.


green-dalii:
Merging. Docs only, Gate 1 green, keyword scan returned no hits (this PR closes nothing — #603 and #741 stay open). Review event is a comment rather than `--approve` because GitHub blocks self-approval on an own PR.

PR #744, opened 2026-09-18
fix(llm): remember a cross-origin failure so the streaming fallback stops being invisible

green-dalii:
fix(llm): remember a cross-origin failure so the streaming fallback stops being invisible

Step 1 of 2 for #741. This addresses the silent-degradation half; the streaming half is the node transport and lands separately. **#741 stays open** — this PR is step 1 of 2, not its resolution.

## The problem this half fixes

`streamWithFallback` picked its transport from `isLocalBaseURL(url)` — a client-side, per-host property — while the capability it needs is "this method + URL + **header set** is permitted cross-origin by this server", a server-side property computed per request. #736's free-form custom-headers field made the header set user data, so a table keyed on host cannot represent what it is trying to predict.

Worse, the downgrade left no trace at all in a shipped build. Four independent layers each hide it:

- `streamWithFallback` swallowed the `TypeError` deliberately — its comment says a warning "would spam logs on every request", which was true for the population it was written for (local servers, where the fallback is expected and total) and is not true for a cloud host that fails unexpectedly.
- Every diagnostic on the path was `console.debug`.
- `esbuild.config.mjs:15` neutralises `console.debug` in production — not stripped at build time, overwritten with a no-op as main.js's first statement.
- `requestUrl` is a main-process call, so it never appears in the DevTools network panel either.

Net effect for a user: streamed answers arrive in one piece after ~20 s with an empty pane, and nothing anywhere says why.

## What changed

Record the outcome instead of guessing the identity. A `TypeError` from `window.fetch` adds the **origin** to a session-scoped set; later calls to that origin skip the attempt, which is known to be blocked. A `console.warn` fires **once per origin per session** — `console.warn`, not `console.debug`, because a debug-level note is a no-op in the build users actually run.

Nothing is hardcoded: the set starts empty and is populated only from an observed failure, so there is no host list to maintain and a host that starts answering preflights is picked up again after a reload.

Two boundary decisions worth naming:

- Only a `TypeError` counts. An `AbortError` is a cancelled request, not a verdict on the host, and a plain `Error` is not a CORS signal — both still propagate as before.
- An unparseable URL becomes its own key rather than throwing in the helper, because a malformed URL is exactly the input that reaches `window.fetch` as a hard failure.

## Verification

- 9 new tests; **mutation-tested** — deleting `corsBlockedOrigins.add(origin)` fails exactly 3 of them (the memory assertion, the warn-once assertion, the unparseable-URL assertion) while the other 39 stay green, so the coverage is precise rather than shotgun.
- Confirmed in the production bundle after `pnpm build`: the new warning string is present, `console.debug` is neutralised, `console.warn` is untouched.
- Gate 1 green at 301 files / **4205 tests** (was 4196).

## What this does not do

`requestUrl` returns a complete body (`obsidian.d.ts:5293` — `{status, headers, arrayBuffer, json, text}`, no `body`), so **real streaming is architecturally unavailable on a host that blocks cross-origin requests**. This step stops the degradation from being invisible and stops re-attempting a request known to be blocked; it does not restore streaming. That is step 2, behind a `Platform.isDesktop`-guarded `node:https` transport, using the pattern already proven in `loopback-flow.ts:134` and already present in shipped main.js.


green-dalii:
Merging. Step 1 of 2 for #741 — **#741 stays open**; the node transport is step 2. Keyword scan re-run before merging: the first draft of this body contained `It does not close #741`, which GitHub parses as a closing directive because it does not understand negation. Rewritten; the merged body scans clean at 0 hits. Review event is a comment rather than `--approve` because GitHub blocks self-approval on an own PR.

PR #745, opened 2026-09-18
docs(memory): a negated closing keyword is still a closing keyword

green-dalii:
docs(memory): a negated closing keyword is still a closing keyword

PR #744's body introduced its work on #741 by *asserting the issue would stay open* — and the sentence carrying that assurance was itself a closing directive, because GitHub's parser has no notion of negation. The pre-merge scan caught it before the merge; the body was rewritten and re-scanned at zero hits.

Recorded in §"Issue close keyword — and its inverse hazard", as one step further in than the #689 incident already documented there. In #689 the keyword was unintended; here it was deliberately negated and still armed.

Also recorded a second, smaller trap found while checking the first: the scan command in that section is written as a pipeline ending in `sed`, and a pipeline exits with the last command's status — so `grep ... | sed ... || echo clean` never reports clean and an empty output is indistinguishable from a swallowed match. The check now says to use `grep -c` and read the number, against the body as GitHub holds it rather than the local draft.

That instruction earned itself within the hour: the first draft of *this* body quoted the offending sentence verbatim, which re-armed the keyword inside the documentation warning about it. Counted, not eyeballed — 1 hit, rewritten, 0.


green-dalii:
Merging. Docs only, Gate 1 green, and the closing-keyword scan is at 0 hits (it was 1 on the first draft — quoting the offending sentence verbatim re-armed it, which is the hazard the change documents). Review event is a comment rather than `--approve` because GitHub blocks self-approval on an own PR.

PR #746, opened 2026-09-18
feat(llm): stream blocked origins over a desktop node:https transport

### LINKED ISSUE #741: opencode.ai fails the CORS preflight, so every streamed answer arrives buffered — `preferRequestUrl` for presets that need non-safelisted headers

green-dalii:
feat(llm): stream blocked origins over a desktop node:https transport

Step 2 of 2. Closes #741.

## What step 1 left open

`requestUrl` returns a finished body (`obsidian.d.ts:5293` — `{status, headers, arrayBuffer, json, text}`, no `body`), so on a host that refuses the CORS preflight real streaming was architecturally unavailable. Step 1 made that downgrade visible and stopped re-attempting a request known to be blocked. It did not restore streaming. This does.

## The transport matrix, and why this is the only option left

| transport | CORS | streams | honours the proxy |
|---|---|---|---|
| `window.fetch` | enforced | yes | yes |
| `requestUrl` | no | **no** | yes |
| `node:https` | no | yes | **no** |
| Electron `net` | no | yes | yes |

Electron's `net` would be the ideal fourth row — Chromium's network stack, so proxy-aware *and* streaming — but its documented process list is **"Main, Utility"**, and a plugin runs in the renderer, so it is unreachable. That leaves `node:https`, whose one cost is real: Electron's own docs contrast `net` with the Node modules by noting `net` offers "better support for web proxies". Node's `https` does not read `http_proxy`/`https_proxy` and cannot see Obsidian's configuration.

## So the failure is treated as a verdict about the transport

A `TypeError` from the Node path means "this machine cannot use this transport for this origin" — a proxy, no route, TLS interception — not "the request failed". It is recorded per origin and the call goes to `requestUrl`, which is buffered but honours the proxy.

That is what makes the change safe rather than merely useful: **a user behind a proxy ends up on exactly the behaviour they had before this transport existed**, and pays for the discovery once per origin per session instead of once per call. Nothing is hardcoded — both memories start empty and are populated only from an observed failure.

On mobile the Node path is skipped silently, because mobile cannot have it and reporting it would be noise; the CORS warning still fires.

## Bot compliance — verified, not assumed

`node:https` is a Node built-in, so `obsidianmd/no-nodejs-modules` applies. The accepted shapes are enumerated in the plugin's own test file:

- a positive `if (Platform.isDesktop) { … }`
- a `Platform.isDesktop && …` short-circuit
- a ternary consequent
- an early `if (!Platform.isDesktop) { throw }` at function start

and the negated form wrapping the import is **explicitly forbidden**. This file uses the early-exit form, matching `openai-codex/loopback-flow.ts:134`, which has shipped since v1.25.6.

Confirmed by control, not by reading: a probe file containing all three shapes produced exactly two `obsidianmd/no-nodejs-modules` reports — the unguarded import and the negated-guard import — and **none** for the early-exit form. The new file itself lints at 0 errors / 0 warnings. `createRequire(__filename)` is used rather than bare `require()` because `@types/node` types the latter as `any`, which would propagate `no-unsafe-*` through every caller; the cast is local to the return statement.

## Verification

- 10 new tests (8 transport, 2 routing). **Mutation-tested**: deleting `nodeTransportBlockedOrigins.add(origin)` fails exactly 2 of them — the memory assertion and the warn-once assertion — while the other 42 stay green.
- The transport is injected in the bridge tests rather than exercised for real. The harness's `Platform` mock sets `isDesktopApp: true` but leaves `isDesktop` undefined, so the production loader would throw and the desktop path would pass for the wrong reason — stubbing the module keeps the intent explicit.
- 4xx/5xx pass through as a `Response` rather than throwing, matching `requestUrl`'s existing contract that AI-SDK reads the body.
- `AbortError` is deliberately not a `TypeError`, so a cancelled request is not recorded as a transport verdict. Aborting destroys the socket, and a consumer that walks away triggers `cancel()` → `destroy()`.
- Confirmed in the production bundle: `node:https`, `createRequire` and the `Platform.isDesktop` guard are present.
- Gate 1 green at 302 files / **4215 tests** (was 301 / 4205).

## Known limitation

**Mobile is unchanged.** A host that refuses the preflight still gets buffered answers on mobile, because mobile has no Node runtime and Electron's `net` is main-process only. That is a property of the platform, not an outstanding defect, and it is why this closes the issue rather than leaving it open for a fix that cannot exist.


green-dalii:
Merging. Step 2 of 2, and the closing keyword is intentional here — every acceptance criterion aisahpA proposed now holds on desktop, and the one that does not hold on mobile is a platform limit rather than pending work (no Node runtime, and Electron `net` is main-process only). Keyword scan: 1 hit, deliberately the `Closes #741` in the body's second line. Review event is a comment rather than `--approve` because GitHub blocks self-approval on an own PR.

PR #747, opened 2026-09-18
docs(memory): record the #741 analysis and its two-step resolution

green-dalii:
docs(memory): record the #741 analysis and its two-step resolution

Extends the #741 design record with what shipped and, more usefully, with what was ruled out along the way.

The rejected transport is the part worth keeping: **Electron's `net` module** is the ideal answer on paper — Chromium's network stack, so proxy-aware *and* streaming — but its documented process list is "Main, Utility" and a plugin runs in the renderer. That is what makes `node:https` the only remaining option, and it is why the transport cannot honour Obsidian's proxy configuration. Without that recorded, the obvious next proposal is to reach for `net`.

Also recorded: the pattern the fix settled on, which generalises. When a fallback transport has a known environmental failure mode, **remember the failure rather than predicting it** — a failed Node attempt is treated as a verdict about the transport, recorded per origin, and `requestUrl` takes over. A proxy user therefore lands on exactly the behaviour they had before the transport existed.

And the Bot-compliance result, which was verified by control rather than by reading: a probe file with three shapes produced exactly two `obsidianmd/no-nodejs-modules` reports — the unguarded import and the negated-guard import — and none for the early-exit form. The rule is sourced from Node's own `module.isBuiltin()`, so `electron` is not covered by it at all.

`## Current state` and the resume point are refreshed to `e150d139`, 302 files / 4215 tests, with #741 closed and no unstarted filings left.

Docs only. Gate 1 green at 302 files / 4215 tests.


green-dalii:
Merging. Docs only, Gate 1 green, closing-keyword scan at 0 hits. Review event is a comment rather than `--approve` because GitHub blocks self-approval on an own PR.

PR #748, opened 2026-09-18
refactor(engine): split the write gate into named layers (pageGuard, rawWrite, notify)

green-dalii:
refactor(engine): split the write gate into named layers (pageGuard, rawWrite, notify)

Slice 1 of 3 for #603. **Part of #603** — the issue stays open; this establishes the interfaces, it does not yet move any of the six violations onto them.

## Why the gate had to be split

`createOrUpdateFile` was documented as a "single write gate with pollution defense" and did four separable things, offering "all four" or "none": pollution correction, heading/provenance normalization, IO with retry and path resolution, and notification. The decisive evidence was already in the codebase — the PDF sidecar's deliberate bypass at `wiki-engine.ts:845-851`, which avoids the gate because going through it "would fire `onFileWrite` + `invalidatePageCaches`, which could trigger auto-ingest cascades if the source folder is watched". That comment is the design conclusion, written by whoever needed it first.

So the layers now exist as names:

- **`rawWrite`** — retry, path resolution, create-or-update. Knows nothing about wiki pages, pollution or notification. This is what the sidecar wanted.
- **`pageGuard`** — pollution correction + normalization, extracted to `wiki/page-write-guard.ts` as a **pure function**, so it can be tested without a vault and a caller can take it alone.
- **`notify`** — `onFileWrite` + cache invalidation. The sidecar's opt-out.

`WriteIntent` carries the pair, and it has **no default** — a default is how the ambiguity this type exists to remove would come back.

## Intended to change nothing, and one thing did

The acceptance for this slice is zero behaviour change, and all 4231 tests pass. One existing assertion did flip, and it is worth stating plainly rather than burying in the diff: `wiki-engine-ingest.test.ts` asserted a created file's content equalled the raw input exactly, which also excluded the `generation_complete` stamp. That assertion was passing on a **race** — `markPageComplete` is deliberately fire-and-forget (`void (async () => …)()`), so whether it landed before the caller resumed depended on how many async frames the write path took, and the split added one.

It is now deterministic (5/5 runs, same result). The stamp is correct: the page is complete. The assertion was **replaced with a stronger pair** — the content is present *and* the stamp is present — rather than relaxed to a `toContain` on the content alone.

That fire-and-forget completion stamp is itself a #603-shaped hole, since it means "complete" is not something a reader can rely on immediately after the write. Recorded for slice 3 rather than fixed here.

## Verification

**The wiring is mutation-tested, because the pure unit tests cannot catch it.** This is the #736 lesson applied deliberately: there, deleting a production call left the four header tests green because they exercised the unit and never the call site.

- Disabling the guard call fails exactly **4 wiring tests** (`write-gate-layers.test.ts`) while all **10 pure-unit tests** (`page-write-guard.test.ts`) stay green — precisely the split that made #736's gap invisible.
- Removing `onFileWrite` from the notify layer fails exactly **2 tests**, both in that same file.
- The guard tests include a `lastIndex` regression: both pollution patterns carry the `g` flag, and `RegExp.prototype.test` advances `lastIndex` on a global regex. The inline form was safe only because it rebuilt the literals per call; hoisting them to module scope would have made every second write skip its correction. The test calls the guard four times on the same input and requires all four to correct.
- The layer boundary is pinned by a test as well: pollution correction runs for **every** path, not only wiki content pages — narrowing the whole layer would silently stop correcting a log.

Three now-unused imports were removed from `wiki-engine.ts` (`fixPollutedSources`, `normalizeHeadingSpacing`, `normalizeProvenanceMarkers`), each used only in the block that moved.

Gate 1 green at 304 files / **4231 tests** (was 303 / 4225).


green-dalii:
Audit note, posted after the merge.

The review-event comment for this PR was intended to be posted *before* `gh pr merge`, but the command failed and the failure went unnoticed because the surrounding script did not check its exit status. Nothing about the merge itself is in question — Gate 1 was green on `4f00823f`, the closing-keyword scan returned 0 hits, and the change is slice 1 of 3 with #603 deliberately left open — but the audit trail is what downstream tooling reads, so it is recorded here rather than left implicit.

This is the same class as the rule already in AGENTS.md ("verify the commit, not the command"), one surface further out: the command that was trusted was the *comment*, not the commit. The lesson generalises to any step whose success is assumed from the absence of an error rather than from a check.

PR #749, opened 2026-09-18
docs(memory): record the #603 slice-1 outcome and the race it surfaced

green-dalii:
docs(memory): record the #603 slice-1 outcome and the race it surfaced

Slice 1 landed the layer split; this records what it taught, which is mostly one thing the split itself exposed.

Acceptance for the slice was zero behaviour change, and exactly one assertion flipped. **That flip is the finding.** `wiki-engine-ingest.test.ts` asserted a created file's content equalled the raw input, which also excluded the `generation_complete` stamp — and it was passing on a **race**, because `markPageComplete` is deliberately fire-and-forget, so whether it landed before the caller resumed depended on how many async frames the write path took. The split added one frame and made it deterministic. The assertion was replaced with a stronger pair rather than relaxed.

The fire-and-forget stamp is its own #603-shaped hole — "this page is complete" is not something a reader may rely on immediately after the write returns — and it is recorded for slice 3 rather than fixed inside a slice whose whole contract was "changes nothing".

Also recorded:

- The `recovered` asymmetry is **preserved and encoded** in a return type rather than silently unified: the two recovery paths never stamped the page while the two ordinary paths did. Making that checkable is the point; deciding whether to fix it is slice 3's.
- The `lastIndex` trap, because it is the failure mode of the obvious refactor: both pollution patterns carry the `g` flag and `RegExp.prototype.test` advances `lastIndex`, so hoisting them to module scope — what one naturally does when extracting to a module — would make every **second** write skip its correction.
- That the wiring mutations were run deliberately, applying the #736 lesson on purpose rather than rediscovering it: disabling the guard call left all ten pure-unit tests green and failed exactly the four wiring tests.

Docs only. Gate 1 green at 304 files / 4231 tests.


green-dalii:
Merging. Docs only, Gate 1 green, closing-keyword scan at 0 hits. Review event is a comment rather than `--approve` because GitHub blocks self-approval on an own PR.

PR #750, opened 2026-09-18
refactor(engine): enforce the write-path contract in three movements

green-dalii:
refactor(engine): enforce the write-path contract in three movements

Slices 2 and 3 of #603 in one PR, as requested — same class of fix, and slice 3 builds directly on slice 1's layers. **Part of #603** — it stays open because two sites are deliberately left on the vault API (see below).

Slice 1 (the layer split) is already on `main` as `a8110551`, so it is the base here rather than part of this diff.

## Per-commit map

| Commit | What |
|---|---|
| `13a47e4f` | the log and the sidecar declare their intent |
| `d35ef98d` | the record corrected — the log's guard is harmful, not idle |
| `9a7d1b2b` | the lint fix runners go through the declared entry; a source-reading guard test |

## Movement 1 — the log was being **corrupted**, not merely given idle work

The design record said the guard was *meaningless* for a log. Measuring before changing found it **harmful**: `LogWriter.pageLinks` builds links from **real page paths**, so a page genuinely named `concepts布局优化` is written as `[[concepts/concepts布局优化]]` — correct as written — and Pattern B cannot tell that from LLM-emitted duplication:

```
IN : ## 2026-09-18\n\nIngested [[concepts/concepts布局优化]], [[entities/Qwen]].
OUT: ## 2026-09-18\n\nIngested [[concepts/布局优化]], [[entities/Qwen]].   ← dead link
```

So `LOG_WRITE_INTENT { guard: false, notify: true }` on the log is a **bug fix**. `notify` stays on — only the guard is dropped, and there is a test for that, because `guard: false` sliding into `notify: false` is the plausible next mistake.

`RAW_WRITE_INTENT { guard: false, notify: false }` on the PDF sidecar turns the prose bypass at `wiki-engine.ts:845-851` into a declaration, and routing it through `rawWrite` also gives it the retry and NFC/NFD recovery the two direct vault calls lacked.

**The property the bypass exists for had never been tested.** The existing sidecar tests asserted file contents and nothing about notification, so giving the sidecar `notify: true` — restoring the cascade the comment warns about — would have broken nothing.

## Movement 2 — `adapter.write`, and what settled the open question

Both `fix-runners.ts` sites used `ctx.app.vault.adapter.write`, which sits **below** `vault.modify` / `vault.process`: `vault.on('modify')` never fires and the metadata cache keeps the old frontmatter, so an alias or tag fix landed on disk and stayed invisible until a reindex.

What settled the design record's open question — *deliberate, or older idiom?* — is that the very same function at `:505` resolves the `TFile` with `getAbstractFileByPath` and `:524` reads it with `vault.read`. **It contradicted its own neighbours.** Those sites now call `ctx.wikiEngine.writeFileWithIntent(path, updated, RAW_WRITE_INTENT)`.

`WikiEngine.writeFileWithIntent` is public for this. `LintContext` and `LintPhaseContext` already carry `wikiEngine`, so **no interface member was needed** — an earlier draft of this change added `writeFileWithIntent` to `EngineContext` and would have shipped dead API. The `intent` parameter is required; an optional one would let a call site stay silent, which is what made #603 hard to read.

## Movement 3 — two sites are deliberately **not** migrated

`sources-normalizer.ts:247` and `preparation.ts:68`/`:105` use `vault.process` **and use its `data` argument**, so that read-modify-write is genuinely atomic. Migrating them would trade atomicity for retry, which is not obviously the better deal, and for `sources-normalizer.ts` it would also undo the module's documented reason for existing — extracted "so the folder-scope filter is testable without standing up the whole startup surface".

`markPageComplete` is different and *was* migrated: its `process` callback **discards** `data` and returns a value computed from an earlier read, so it never had atomicity to lose. It only lacked what `rawWrite` adds.

That corrects the record's tiering a second time: the defect was never "missing guard and notify", it was **being below the event layer**.

## The guard test reads the source, not a copy of the rule

`wiki-content-folder-guard.test.ts` documents the surrounding rule by **re-implementing** it, so it keeps passing after production drifts. `write-path-contract.test.ts` reads the production tree instead: no `.adapter.write` (or `append` / `remove` / `rename` / `trash`) outside a one-entry waiver list, and every internal write passing a named constant rather than an inline object.

`core/disk-cache.ts` is the single waiver — an injected-adapter cache writing cache entries, which the #603 pass had already placed in tier C. It is named in a list with its reason rather than excluded by widening the pattern, and a test asserts the list stays at one entry, because a growing waiver list is how this kind of guard dies.

## Verification

- **Three mutations, all precise.** Reverting the log to `FULL_WRITE_INTENT` fails exactly the log-corruption test; giving the sidecar `LOG_WRITE_INTENT` fails exactly the sidecar-notify test; reverting one `adapter.write` site fails **3 tests and names the file**.
- The log corruption was reproduced as a failing test through the engine's own `logLintFix` path before the fix.
- The `write-path-contract` guard failed on its first run against `disk-cache.ts` — that is how the one legitimate waiver was found rather than assumed.
- Gate 1 green at 305 files / **4240 tests**.


green-dalii:
Requesting review — not merging until you have looked.

Four things worth your attention, in descending order of how much I would like a second opinion:

1. `src/wiki/wiki-engine.ts` — the log's writer injection is now `writeFileWithIntent(path, content, LOG_WRITE_INTENT)`. The claim in the body is that this **fixes** a corruption rather than removing dead work: `LogWriter.pageLinks` emits `[[concepts/concepts布局优化]]` for a page of that name, and the guard's path-prefix repair rewrites it to `[[concepts/布局优化]]`. If you think Pattern B is right and the log's links are the thing that should change, that inverts the fix and I would rather hear it now.
2. `src/types.ts` — `LOG_WRITE_INTENT` / `RAW_WRITE_INTENT` are the first named intents. Is `WriteIntent { guard, notify }` the right pair of axes, or does the sidecar want a third thing that "raw" loses?
3. `src/wiki/wiki-engine.ts` sidecar block — routing it through `rawWrite` gains retry and NFC/NFD recovery but also means the sidecar now goes through `resolveFileInVault`. Worth checking that is not a behaviour change you object to.
4. `src/__tests__/wiki/wiki-engine-pdf.test.ts` — the sidecar's not-notify property was previously untested; that test is new and mutation-verified (giving the sidecar `notify: true` fails exactly it).

Gate 1 green at 304 files / 4235 tests.

green-dalii:
Scope changed after the review request, so re-requesting rather than letting you review a moving target.

The branch now carries slices 2 **and** 3 of #603 as three commits, with a per-commit map at the top of the description. Slice 1 is already on `main` (`a8110551`), so the diff here is the two movements that build on it.

Two things I would most like a second opinion on:

1. **The log's `guard: false` is framed as a bug fix, not a tidy-up** — `LogWriter.pageLinks` emits `[[concepts/concepts布局优化]]` for a page of that name and Pattern B rewrites it into a dead link. If you think the log's links are what should change instead, that inverts the fix and I would rather hear it before merge.
2. **Two sites are deliberately left on `vault.process`** (`sources-normalizer.ts:247`, `preparation.ts:68`/`:105`) because they use its `data` argument and so have genuine read-modify-write atomicity. I judged that atomicity worth more than the retry, and for `sources-normalizer.ts` the alternative would also undo the module's stated reason for existing. That is a judgement call, not a measurement, so it is the weakest link here.

One process note: an earlier draft added `writeFileWithIntent` to `EngineContext`. It turned out `LintContext` and `LintPhaseContext` already carry `wikiEngine`, so that member would have been dead API and it was removed.

Gate 1 green at 305 files / 4240 tests. Three mutations verified (log intent, sidecar notify, adapter.write), each failing precisely the tests that cover it.

DocTpoint (changes requested):
Reviewed at `9a7d1b2b`.

The three movements are the right shape, and `write-path-contract.test.ts` reading the production tree instead of re-implementing the rule is the correct answer to the class of defect #603 is about. Two things block in the shipped behaviour, and one in the tests this PR adds.

**Blocking 1 — movement 3 gives `markPageComplete` the ability to create, which undoes #582/#583.**

The stamp went from `getAbstractFileByPath` + `if (file instanceof TFile) vault.process(...)` to `await this.rawWrite(path, flipped)`. The description says it "only lacked what `rawWrite` adds: the three-attempt retry and the NFC/NFD recovery". `rawWrite` adds a third thing: `vault.create`. The old form could only update a file that was still there. The new one creates one that is gone.

That matters because `markPageComplete` is deliberately un-awaited (`void (async () => ...)`), so it runs against whatever happens next — including the cancelled-ingest cleanup at `wiki-engine.ts:1577`, whose own comment states the requirement: *"a cancelled ingest must not read as a completed one … leaving it behind freezes the half-finished ingest as done and every later trigger skips the source."*

The order that breaks it: the summary page is written, the stamp is spawned and reads it, the cancel throws at the next `checkCancelled()`, the catch block deletes the summary page, and the stamp then re-creates it with `generation_complete: true`. The completion marker is back and the source is skipped from then on. Which side wins is undetermined, so it reproduces intermittently in the field.

Measured in three arms — base `8c7966ed` passes, this branch fails, this branch with only the `markPageComplete` line reverted passes. The engine's own debug output names the mechanism: `Create success` / `Attempt 1: File not found, creating` / `Create success`.

The probe, with the interleaving pinned rather than timed:

```ts
// src/__tests__/wiki/markpagecomplete-resurrect.test.ts
import { describe, it, expect } from 'vitest';
import { createWikiEngineHarness } from '../__support__/wiki-engine-harness';

describe('#603 slice 3 — the generation_complete stamp must not resurrect a deleted page', () => {
  it('leaves a page deleted when it disappears between the stamp read and the stamp write', async () => {
    const PATH = 'wiki/sources/Note.md';
    const h = createWikiEngineHarness({});

    // Delete the page the instant the stamp reads it. On a freshly created page
    // the write itself performs no `get` (vault.create only sets), so the first
    // `get` of this path is the stamp's own `tryReadFile` — exactly the point
    // after which the cleanup's delete can land.
    const realGet = h.files.get.bind(h.files);
    let armed = true;
    h.files.get = ((key: string) => {
      const value = realGet(key);
      if (armed && key === PATH) {
        armed = false;
        h.files.delete(PATH);
      }
      return value;
    }) as typeof h.files.get;

    await h.engine.createOrUpdateFile(PATH, '---\ngeneration_complete: false\n---\n\nBody');

    // Let the un-awaited stamp settle.
    await new Promise(resolve => setTimeout(resolve, 50));

    expect(h.files.has(PATH)).toBe(false);
  });
});
```

The PR's test changes do not mention `markPageComplete` or `generation_complete` at all, which is why the three precise mutations elsewhere did not catch it.

**Blocking 2 — the lint fixers are now governed by the ingest's cancel button, not their own.**

`writeFileWithIntent` opens with `checkCancelled()`, which reads `this.abortController`. That is the **ingest** controller. The lint run has its own — `lintAbortController`, with `cancelLint()` and `isLintRunning()` beside it — and `checkCancelled()` does not consult it. `command-registry.ts:74` registers `lint-wiki` as a bare `callback: () => plugin.lintWiki()` with no `isIngesting()` guard, so the two runs can overlap; the status bar's `if (isIngesting) … else if (isLintRunning)` only assumes they cannot.

After this change, cancelling the **ingest** aborts an overlapping lint's writes, and cancelling the **lint** does not. `runRetagViolations:628` re-throws AbortError deliberately, so the whole retag run tears down with its earlier batches already written, and the user who pressed "cancel ingestion" is not told. Read from the code, not executed.

**Blocking 3 — `write-path-contract.test.ts:56` fails on Windows.**

The waiver is `'core/disk-cache.ts'` but `rel` comes from `relative(SRC, f)`, which yields `core\disk-cache.ts` on win32. The waiver is load-bearing — `disk-cache.ts:155` really does call `this.adapter.write(path, data)` — so the lookup misses, the file is no longer excluded, and both assertions fail. CI on ubuntu will never show it. The guard's whole value is that it runs everywhere.

**Non-blocking, two notes.**

1. `write-path-contract.test.ts:35` cuts each line at its first `//` without tracking string literals, so a bypass written after an inline URL is invisible to the scanner. Low likelihood, but this guard's value is completeness.
2. `write-gate-layers.test.ts:58` asserts that pollution correction reaches `wiki/log.md`, and its comment says "Narrowing the whole layer would silently stop correcting a log". Movement 2 narrows exactly that layer, and the test stays green only because it writes through `createOrUpdateFile`, which production no longer uses for the log — the `LogWriter` is now its sole writer. So the behaviour the test claims to protect is gone while the test still passes. Worth noting too that the justification covers only the path-prefix repair, but `guard: false` also drops display-name correction and sources normalization on the log; the record does not mention either.

green-dalii:
All three reproduced, all three fixed, and the two behaviour findings are now covered by tests that fail without their fix. Two things you noted as non-blocking are also done. Thank you — the framing that made this land was that mutations cover the wiring and neither of these was a wiring question.

**1. `markPageComplete` creating.** You are right and my reasoning was wrong in an instructive way. I had argued the old form's `process` callback discarded its `data`, so there was no atomicity to lose — true, and irrelevant: `rawWrite` adds **three** things and I only accounted for two. The third is `vault.create`, and the pre-split form resolved a `TFile` first, so it could only update.

Fixed as a declared layer rather than by reverting the call, because the retry and the NFC/NFD recovery are the reason the call is there and both are wanted. `WriteIntent` gains `create`; every existing intent is `create: true`; `STAMP_WRITE_INTENT` is `create: false`, with the race written into its doc comment so removing it is a deliberate edit. `rawWrite` returns a new `'absent'` outcome — nothing happened, on purpose, not an error.

Your probe is now a test, kept as given including the comment that made it click. The interleaving is pinned rather than timed, which is the part I would not have thought to do: a timed version of this would have been a flaky test that "usually" passed. I also added the positive half, so `create: false` cannot be satisfied by a write that simply does nothing.

**2. The lint fixers and the ingest's cancel button.** Also right, and this one is worse than it looks, for the reason you supplied: `runRetagViolations` re-throws AbortError deliberately, so the whole retag run tears down with earlier batches written and the user who pressed "cancel ingestion" is never told.

Fixed with the same move, because it is the same kind of fact: the cancel owner is now declared — `cancel: 'ingest' | 'lint' | 'none'` — and `LINT_WRITE_INTENT` names `lint`. What struck me while doing it is that the check had been **implicit since it was written**: `checkCancelled()` had no parameter and read whichever controller it found. That is exactly the ambiguity `WriteIntent` exists to remove, and I had left one field of it undeclared while declaring the other two.

**3. The Windows waiver.** Right, and it is the one I would have shipped unknowingly — CI is ubuntu-only, so the only way to see it is to reason about `relative()` on win32. Now `relPosix()`. A guard whose entire value is that it runs everywhere has to normalise the paths it compares.

**Non-blocking 1, the comment stripper.** Taken, and I think it is worth more than "low likelihood": the failure mode is a scanner that silently reports a clean tree, which is the worst way for this particular file to be wrong. It is a quote-aware single pass now.

**Non-blocking 2, `write-gate-layers.test.ts`.** Taken, and it is the same class as your blocking 1 in a different place: a test asserting a behaviour that production no longer reaches. `LogWriter` is the log's sole writer now, so the pollution correction this test claimed to protect never runs on a log in production — it passed only because `createOrUpdateFile` applies the gate to whatever path it is handed. I kept the claim (the layer boundary is real and worth pinning) and moved it to a path that still takes the full gate, naming the log as the counter-example.

You were also right that the record mentioned only the path-prefix repair. `guard: false` drops display-name correction and sources normalization too. Neither is wanted on a log, so the outcome is unchanged — but "harmful for one reason" and "does three things, none of them wanted here" are different claims and only the second is checkable. Recorded in the constant and in the design record.

**One thing worth naming across both our PRs this week.** #751's build test asserted `import("node:module")` while its comment said the requirement was laziness — writing a property, pinning a spelling. This one's fourth assertion was the same shape in the other direction. Two people, two repos' worth of distance, same failure, found four days apart. The check that catches it is asking what the assertion would still pass on.

DocTpoint (changes requested):
Re-reviewed at `6803485`. All three blocking findings are fixed, and I verified
each one rather than taking the description for it. The two non-blocking notes
are done too, and the comment stripper is genuinely better than "low likelihood"
— thank you for treating it that way.

One new blocking finding, and it is introduced by the fix to my finding 2. I am
raising it with the same weight as the original because it ends in the same
state: a source that is skipped from then on.

## The lint intent closed one door and opened another

`LINT_WRITE_INTENT` stops reading the ingest's controller, which is right. It
also keeps `create: true`, and those two together make a lint write able to put
back a page the ingest's cancel cleanup has just deleted.

The chain, each link checked in the tree at `6803485`:

1. `scanTagViolations` (`lint/scanners.ts:427`) accepts `pageType === 'source'`.
   Retag violations therefore include summary pages. `detectAliasDeficiency`
   filters to `/entities/` and `/concepts/`, so this is the retag site only —
   `fix-runners.ts:624`.
2. The retag fixer holds a `pageMap` snapshot taken at scan time and writes it
   back after its LLM batch, which is seconds to minutes later.
3. `lint-wiki` is registered with no `isIngesting()` guard — your own point — so
   a lint run and an ingest overlap.
4. The cancelled ingest deletes the summary page, and only that page
   (`wiki-engine.ts:1600`), because it doubles as the completion marker that
   `isAlreadyIngested` reads (`main-commands/ingest-commands.ts:59`).
5. The retag write lands afterwards. `create: true`, the file is gone, so
   `rawWrite` calls `vault.create`. The completion marker is back, and the
   ownership check passes because the snapshot still carries the note in
   `sources:`. Every later trigger skips the source.

That is the #582/#583 state, reached through the retag path instead of the
stamp path.

**What makes it this PR's finding rather than a pre-existing one:** before this
change the write gate read the ingest controller, so step 5 threw and nothing
was written. The old wiring was wrong for the reason you give, and it was also
incidentally holding this door shut. Nothing replaced it.

There is a second half I want to state because it changes what `cancel: 'lint'`
buys today. `cancel-ingestion` and the status-bar click both do
`if (isIngesting()) cancelIngestion(); else if (isLintRunning()) cancelLint();`
(`command-registry.ts:100` and `:157`). While an ingest is live — including the
whole unwind after a cancel, since `abortController` is nulled only at the end —
there is no reachable lint cancel. So in exactly the overlap case the doc
comment describes, the lint's writes did not become governed by the lint's stop
button; they became governed by nothing.

## Measurement

Four arms, in-memory harness, interleaving pinned rather than timed:

| Arm | Setup | Result |
|---|---|---|
| A | `6803485` as shipped: lint running, ingest cancelled, summary page deleted, retag write lands | **page resurrected** |
| B | identical, but the write uses `RAW_WRITE_INTENT` (the pre-fix wiring — byte-identical except `cancel`) | write refused, page stays deleted |
| C | identical to A, but `create: false` on the lint intent | page stays deleted |
| D | control: `create: false`, page present | page updated normally |

D is there because C alone would also pass if the write simply never did
anything.

Mutation, to check nothing in production depends on a lint write creating:
setting `LINT_WRITE_INTENT.create = false` against the full suite fails exactly
one test — your `lets a lint write through while the lint is running`, which
writes to a path that does not exist. Its subject is the cancel owner, not the
create layer, so I read that as a fixture to pre-create the page rather than a
claim to keep. No production path breaks.

## Suggested fix

One field: `create: false` on `LINT_WRITE_INTENT`. Both call sites take their
path from a scan of pages that exist, so update-only is what they already mean,
and it is the same move you made for the stamp. Your call on whether the cancel
affordance in `command-registry.ts` belongs in this PR — I would leave it out,
it is older than this change.

## Two small things, neither blocking

- `STAMP_WRITE_INTENT.cancel` is never read. `markPageComplete` calls `rawWrite`
  directly and `checkCancelled` lives in `writeFileWithIntent`, so the field is
  declared and inert on the one intent whose doc comment explains it. Harmless
  today; worth a word in the comment, given that "declared rather than inferred"
  is the whole point of the type.
- `writeFileWithIntent` does not look at an `'absent'` outcome: `intent.notify`
  would fire `onFileWrite` and `intent.guard` would call `markPageComplete` for
  a write that did not happen. Unreachable now — no intent has `create: false`
  together with either flag — and it stays unreachable if the fix above lands,
  since the lint intent has both off. I mention it only because the next
  `create: false` intent is where it bites.

I also checked the `'Ingestion cancelled by user'` message that a lint cancel
now throws: the lint controller catches on `error.name` and shows its own
localized notice (`lint/controller.ts:661`), so the string never reaches a user.
Not worth a change.

## Gate 1

Reproduced your numbers: **307 files / 4267 tests**, all green once `main.js` is
built — the Codex bundle test reads the artifact, so a test run without a build
fails that one file.

One note on the environment rather than on your PR: I had to install
dependencies fresh to get there. A shared `node_modules` in my tree was a major
version behind on `zod` (3.25.76 against the required `^4.6.5`) among three
mismatches, and a spot check of two packages had previously missed it. Checking
the whole declared set against what is installed takes one pass and is the only
version of that check worth running.

green-dalii:
All three verified in the tree here, all three fixed in `e6b80a96`, and each one measured. Gate 1: 307 files / 4271 tests.

**The blocking one, and the sentence in it that made me go and look:** *"the old wiring was wrong for the reason you give, and it was also incidentally holding this door shut. Nothing replaced it."* That is the part I would not have found by reasoning about my own change — I removed a wrong check, verified the removal was right, and did not ask what the wrong check had been doing in the meantime. The chain checks out: `scanners.ts:427` accepts `pageType === 'source'`, so summary pages are in retag scope; the fixer writes back a scan-time `pageMap` after an LLM batch; `wiki-engine.ts:1592` deletes the marker and nothing else.

`create: false` on `LINT_WRITE_INTENT`, as you suggested. Both call sites take their path from a scan of pages that exist, so update-only is what they already meant.

Your mutation is what settled the test question: one failure, on my own `lets a lint write through while the lint is running`, writing to a path that does not exist. You read it as a fixture to pre-create rather than a claim to keep, and that is what it is — its subject is the cancel owner and the missing page was incidental. It now pre-creates the page. I added the resurrect chain as its own case with a control, because the first version of that test would have passed if the write simply never happened.

**On the cancel affordance:** left out, agreed. Older than this change, and folding it in would make the layer decision look like it depended on the button logic.

**Both small things were real, and the second was worth more than "unreachable".**

`STAMP_WRITE_INTENT.cancel` is `'none'` now. The stamp is not button-gated — `markPageComplete` reaches `rawWrite` directly — so `'ingest'` was a declared field that no code read, on the one intent whose comment exists to explain it. That is the exact failure the type was introduced to prevent, one level up: an intent that reads as a guarantee while being inert.

`writeFileWithIntent` now returns early on `'absent'`. You are right that it was unreachable, and that is the argument for fixing it rather than against: the guard's cost is one branch, and its absence would have been discovered by whoever added the next `create: false` intent, in the form of cache invalidation for a path that was never written. I gave the outcome a named constant so the early return reads as a decision.

Its test uses an inline intent — `create: false` with `notify: true` — precisely because no shipped intent has that pair. Your point about unreachability would otherwise have made it untestable, and an untested defensive branch is one that silently rots.

Mutation-tested both: restoring `create: true` fails 1 (the page comes back); removing the post-write guard fails 1 (notify fires for a path that was never written).

**On the environment note** — thank you, and it is the more useful half of the review. Checking the whole declared set against what is installed rather than spot-checking two packages is a habit I do not have; a `zod` three major versions behind is the kind of mismatch that presents as a mysterious test failure rather than as an install problem. Worth recording in this repo's tooling notes.

DocTpoint (changes requested):
Requesting changes on `e6b80a96`, and the scope is three sentences. **The code is right and I have verified it — this is not a judgement on the change.** I am using the gate rather than a comment because a comment is not one: it would clear the way to merge while the thing I am asking for stays a wish.

**Verified here, not read off the diff.** All three findings are fixed, and each fix is reachable from its call sites:

- `LINT_WRITE_INTENT.create: false` — `rawWrite` returns `'absent'` rather than throwing, so a fixer batch is not aborted by a page that vanished, and both call sites take their path from a scan of pages that existed. Correct at both.
- `STAMP_WRITE_INTENT.cancel: 'none'` — inert before the change, confirmed: `checkCancelled` is reached only from `writeFileWithIntent:1969`, and `markPageComplete` goes straight to `rawWrite` at `:372`. A declared field that nothing read, which is the thing the type exists to prevent.
- The `absent` early return — no shipped intent pairs `create: false` with `guard` or `notify`, so it is unreachable today, and your test declares its own intent rather than pretending otherwise.

**Both mutations reproduce.** Restoring `create: true` fails 1. Removing `if (outcome === WikiEngine.ABSENT) return;` fails 1. Gate 1 here: 307 files / 4271 tests, with lint, typecheck, build and css-lint green — your numbers. I also merged current `main` (`9e97e903`) into the branch and re-ran: green, no interaction with #759.

## Why three sentences are worth a gate

All three were written in this PR — two in `6803485`, one in `e6b80a96`. None is inherited, so this is about the PR being internally consistent, not about cleaning up around it.

They are not stale in the neutral sense. All three point the next reader in the same direction: that the guard this PR just installed is not load-bearing. In a repository where a large share of the reading is done by agents, a comment is not decoration — it is the context that gets handed over, and the nearest text wins over the distant one. Someone editing `fix-runners.ts:147` reads the note three lines above it, not `types.ts:1114`.

**1. `wiki-engine.ts:1918` — "Only `STAMP_WRITE_INTENT` can produce it today."**

This one does more than go stale: it argues against finding the next defect. `STAMP_WRITE_INTENT` never reaches `writeFileWithIntent` at all — `markPageComplete` calls `rawWrite` directly at `:372`. Read literally, the sentence says the new early return is a branch nothing produces. Anyone asking whether the `absent` path needs anything accounted for has, after that sentence, no reason to open the call sites — which is exactly where the accounting is missing (see the non-blocking note at the end).

`LINT_WRITE_INTENT` is `create: false` as of `e6b80a96`, and its two call sites are now the likeliest producers of `absent`.

Suggested fix — drop the `absent` paragraph from the `RECOVERED` block entirely and let the `ABSENT` constant nine lines below own the description, since it already carries it word for word. If you would rather keep it in both places, the last sentence becomes: *"Both `create: false` intents produce it: the stamp, and the lint fixers' write."*

**2. `fix-runners.ts:142` — "because of one field: the cancel owner."**

Two fields as of this commit. You updated `types.ts:1114` to say precisely that; this note did not follow. It is the assumption that produced the bug, sitting at the place a reader starts from.

Suggested replacement:

```
// It is not `RAW_WRITE_INTENT` because of two fields. The cancel owner:
// these writes belong to a lint run, and the engine's write gate reads the
// controller the intent names — the ingest's would have stopped them when
// the user cancelled an ingest, and let them continue when the user
// cancelled the lint. And `create: false`: this path writes a page the scan
// already saw, so if it is gone by the time the LLM returns it must stay
// gone — see `LINT_WRITE_INTENT` in `types.ts` for why that is a
// correctness requirement.
```

**3. `types.ts:1128` — the justification names the wrong call site.**

"The retag fixer holds a `pageMap` snapshot from scan time and writes it back after an LLM batch." It does not. `runRetagViolations` resolves the file at `fix-runners.ts:519`, returns `kind: 'missing'` when it is gone, and reads fresh content at `:537` — all before the LLM call. Its window is read → LLM → write.

The fixer that writes scan-captured content with no existence check at all is the alias one (`page.content` at `:83` and `:127`) — but that runs on entity and concept pages only (`programmatic.ts:37`), so it cannot restore the completion marker.

Both halves of your argument are true, of different sites, and `create: false` is right at both either way. The reason to fix the sentence is the revert: a reader who checks it against `fix-runners.ts` finds the `missing` guard, sees the described mechanism is not there, and files the flag under belt-and-braces.

Suggested replacement for that passage:

```
 * It matters because both fixers write after an LLM call, and the page can go
 * in that window. The retag fixer is the one with a summary page in scope —
 * `scanTagViolations` accepts `pageType === 'source'` (`lint/scanners.ts:427`)
 * — and that page doubles as the completion marker the cancelled-ingest
 * cleanup deletes (`wiki-engine.ts:1592`); it re-reads the file before the
 * call, so its window is read → LLM → write. The alias fixer has the wider
 * window, writing back the content the scan captured with no existence check
 * at all, but runs on entity and concept pages only, so it cannot restore the
 * marker. With `create: true` the retag path could put the marker back, and
 * every later trigger would skip the source: the #582/#583 state, reached
 * through the retag path instead of the stamp path.
```

The same sentence is in the new test at `write-intent-correctness.test.ts:88`; the commit message can stay as it is.

## Not blocking — `absent` is swallowed, so both fixers report a fix that did not happen

Separate from the above, and it needs a signature change, so I am not holding the merge for it.

`writeFileWithIntent` returns `void`, so a call site cannot tell an update from a skip. On `absent`:

- the alias site (`fix-runners.ts:147`) returns `success: true`, so `filled++` runs, `"added N aliases (total M)"` goes into `results`, `controller.ts:302` sees `filled > 0` and calls `generateIndexFromEngine()` and `logLintFix('Complete Aliases', …)`, and the Notice reports N/M;
- the retag site (`:624`) returns `kind: 'fixed'` with an `[old] → [new]` message.

The scenario is the one this commit is about: the cancelled ingest deletes the page during the LLM batch, the write is now correctly skipped, and the lint log keeps a permanent record of a fix applied to a page that no longer exists.

What makes it worth naming: before this commit the report was at least true, because the write always landed — wrongly, but it landed. The change is a clear improvement, and it moved a defect from the vault into the log. Same shape as #719/#720.

The outcome string is already in hand at `wiki-engine.ts:1999`. Returning it, or a boolean, lets both sites count truthfully. Here or as a follow-up — your call.

## One thing that is not about this PR

`pnpm test` is no longer self-contained after #759: `openai-codex-loopback-flow.test.ts:50` reads the built `main.js` from the working directory, so `pnpm test` without a preceding `pnpm build` fails on a stale bundle. `gate:1` orders build before test, so CI is unaffected — but anyone running the suite alone after touching the build config gets a red that is not theirs. It cost me a measurement here to rule out a pair conflict that was not one.

green-dalii:
All three fixed in `620d6244`. The diff outside comments is empty and Gate 1 is 307 files / 4271 tests.

**On why three sentences get a gate — the framing is what changed my mind about them.** "The nearest text wins over the distant one" is the operative fact, and it is the reason #3 mattered more than the other two: it was not stale text, it was a **wrong mechanism that read as supporting evidence**. Someone checking it against `fix-runners.ts` finds `kind: 'missing'` at `:521` and a fresh `vault.read` at `:537`, and the described snapshot is simply not there — so the `create: false` flag gets filed under belt-and-braces by the one reader most likely to act on it. I verified it before editing: the retag fixer's window is read → LLM → write, and the fixer with the no-existence-check window is the alias one, which cannot reach a summary page. Both halves true, different sites. The sentence now says that.

**On #1, "it argues against finding the next defect" is exactly right and I had not seen it that way.** `STAMP_WRITE_INTENT` never reaches `writeFileWithIntent` — `markPageComplete` calls `rawWrite` directly — so the sentence described the new early return as a branch nothing produces. I took your first option and dropped the paragraph from the `RECOVERED` block; the `ABSENT` constant owns it now, and says both `create: false` intents produce it with the lint fixers likelier, since their window contains an LLM call.

**On the reporting gap — I started to fix it here and reverted.** You scoped it out and said "here or as a follow-up — your call", and the signature change turned out to be larger than the two call sites: eight production declarations break on assignability (`EngineContext`, `LogWriterOptions`, `IndexGeneratorOptions`, five page-factory contexts) plus roughly twenty test doubles. Also, a strict `if (!written)` check immediately failed nine tests in `fix-runners.test.ts` because their mocks resolve `undefined` — which is the doubles being incomplete rather than the guard being wrong, and worth knowing before someone does the change.

So it is filed as **#763**, with the two sites named, the `boolean`-or-outcome-string choice, and the note about the doubles. It is referenced at both call sites so it cannot be mistaken for belt-and-braces either — which is the lesson of your #3 applied to your non-blocking note.

**On `pnpm test` no longer being self-contained after #759:** `openai-codex-loopback-flow.test.ts` has read the built `main.js` since before my change — #759 rewrote its assertions, not where the string comes from. `gate:1` orders build before test for this reason, so CI is unaffected, but you are right that anyone running the suite alone gets a red that is not theirs, and that it cost you a measurement. Worth a note in the test file itself rather than only in `AGENTS.md`; I will add it separately rather than widen this PR.

Thank you for the fourth round on this one. The two mutations and the four-plane measurement are why the fix is trustable rather than merely present.

DocTpoint (approved):
Approving `620d6244`.

All three sentences are gone and their replacements are accurate. I checked the diff against `e6b80a96` rather than taking "the diff outside comments is empty" on trust: four files, 54 insertions, 24 deletions, and every added or removed line is a comment. Gate 1 here: 307 files / 4271 tests, with lint, typecheck, build and css-lint green. Merged current `main` (`fb0b112b`) and re-ran: 308 / 4292, green.

The `ABSENT` comment now names both `create: false` intents and says which of them is likelier and why. That is more than I asked for and it is the half a reader actually needs. The call-site note carries both fields, and the `types.ts` passage separates the two windows correctly — retag resolves and re-reads before the call, the alias fixer carries the scan's content and cannot reach a summary page.

**#763 is the right call, and its reasoning is better than mine was.** I scoped the signature change as small on the strength of "the outcome string is already in hand". Eight production declarations breaking on assignability plus roughly twenty test doubles is a different size, and the nine tests that fail on `undefined` from incomplete mocks are worth knowing before someone starts rather than after. Referencing it at both call sites is what makes it stick: an unexplained absence at the call site is the exact failure this PR spent four rounds on.

**A correction to my last review.** I wrote that `pnpm test` is "no longer self-contained after #759". That attributes it to the wrong change, and you are right. `readFileSync('main.js')` entered `openai-codex-loopback-flow.test.ts` with `bf3cd3db` (#634); #759 did not touch the read — I checked the file's history rather than only its diff, which is what I should have done before writing the sentence. What #759 changed is what the test demands of that bundle, so a bundle built before it now fails until rebuilt, and that is what I ran into. The standing property — the suite depends on a build artifact — is older than your change. A note in the test file is the right place for it.

Nothing further from me on this one.

Issue #751, opened 2026-09-18
esbuild leaves external node: imports as native import(), so the #746 desktop transport never loads in the renderer


aisahpA:
_Disclosure: drafted with an AI assistant and reviewed before posting. The measurements were run locally on a real vault with a real OpenCode Go key, against a production build of `8c7966e` and against a bundle rebuilt with the one-line change proposed below._

**Follow-up to #741 / #746.** The desktop `node:https` transport does not load in a shipped build, so the streaming fix is inert in production: every streamed call to a blocked origin still ends up buffered on `requestUrl`. The cause is in the build config, the fix is one line, and the same defect affects an older path.

**What happens.** On a real vault (Obsidian 1.13.4, Electron 43.1.1, Chrome 150; plugin built with `node esbuild.config.mjs production` from `8c7966e`), the first streamed call to `https://opencode.ai` logs:

```
Access to script at 'node:module' from origin 'app://obsidian.md' has been blocked by CORS policy: Cross origin requests are only supported for protocol schemes: app, chrome, chrome-extension, chrome-untrusted, data, http, https.
Failed to load resource: net::ERR_FAILED
[STREAM-FETCH] node:https transport unavailable for https://opencode.ai (Failed to fetch dynamically imported module: node:module); using requestUrl, so answers will arrive in one piece.
```

The transport never opens a socket. Step 1's memory behaves exactly as designed and records the transport as unusable, so the user gets the pre-#746 behaviour (buffered) plus a console error. Nothing crashes, but the feature the issue was closed on does not run.

**Root cause.** `src/core/node-https-fetch.ts:70` does `const nodeModule = await import('node:module')` — and the identical pattern is in `src/llm-sdk/openai-codex/loopback-flow.ts:157`. `esbuild.config.mjs` sets `format: 'cjs'` and `target: 'es2018'` but never sets `platform`, so it defaults to `browser`; `node:module` is in the `external` list; under the browser platform esbuild leaves an external dynamic import as a **native** `import()`. In the renderer that goes to Chromium's module loader, which treats `node:module` as a URL and blocks it. Confirmed in the artifact: `grep -o 'import("node:module")' main.js` returns **2**.

[...]

PR #753, opened 2026-09-18
fix: make Codex OAuth Node loading compatible with Obsidian

### LINKED ISSUE #665: ChatGPT Plan (Codx OAuth) sign in with browser fails
## Bug Description
Selecting ChatGPT Plan (Codx OAuth) on the options page and clicking on "Sign in with browser" leads to the error message "ChatGPT authorization failed: Failed to fetch dynamically imported module: node module".

## Environment                                                                                                                                             
- **Plugin version:** 1.27.1                                                                                                           
- **Obsidian version:** 1.13.7                                                                                                                
- **OS:** macOS 26.6.1                                                                                          
                                                                                        
## Steps to Reproduce                                                                                                                                      
1. Go to the options page of the plugin                                                                                                                                             
2. Selecting ChatGPT Plan (Codx OAuth)                                                                                                                                     
3. Click on "Sign in with browser"                                                                                                                      
4. See error  "ChatGPT authorization failed: Failed to fetch dynamically imported module: node module".
                                                                                                                                                                                                                                                                       
## Expected Behavior                                                                                                                                       
A browser is opened and I can login
                                                                                                                                                             
**Console logs (View → Toggle Developer Tools):**                                                                                                          
There are none

svsk417:
Hi! I do not have much experience with frontend development, so please handle with care.
## Summary

Fixes the Codex OAuth browser login failure reported in #665.

## Root cause

The OAuth loopback flow dynamically imported `node:module`:

In Obsidian's CommonJS plugin renderer, this can be interpreted as a browser module fetch and fails with:

Failed to fetch dynamically imported module: node module

The loopback OAuth loader now uses the CommonJS require() loader already available to Obsidian.

The same correction was applied to the desktop node:https transport because it used the same incompatible loading pattern.

Fixes #665


green-dalii (approved):
Approving, and confirming the diagnosis independently — including for a file this PR fixes that is mine.

**The claim checks out, and I verified it on the artifact rather than by reasoning.** In the shipped `main.js`:

```
import("node:module")   2 occurrences
require("node:module")  0 occurrences
```

with `format: 'cjs'`, `target: 'es2018'` in `esbuild.config.mjs`. So the dynamic import survives bundling as a literal `import()` and is resolved by Chromium's module loader at call time, which is exactly the failure you describe. Your replacement assertion (`not.toContain('import("node:module")')` plus `toContain('require("node:module")')`) is the right way to pin that, because it makes the *bundle* the evidence rather than the source.

**Second effect, worth recording: this fixes `src/core/node-https-fetch.ts`, which I introduced in #746.** The desktop `node:https` transport for CORS-blocked origins (#741) used the same loading pattern, so it has never worked in a shipped build — `requireNodeHttps()` threw at the dynamic import before reaching `https.request`. It failed safely: `streamWithFallback` catches the `TypeError`, records the origin, and falls back to `requestUrl`, so users saw a buffered answer rather than an error and nothing cascaded. That safety net is why a broken transport shipped without being noticed, and it is also why this PR is worth merging rather than deferring.

**Bot compliance, checked rather than assumed.** Your disables are `@typescript-eslint/no-require-imports` and `no-undef`. I read the Bot's own `no-restricted-disable` list — it contains `obsidianmd/no-nodejs-modules`, `@typescript-eslint/no-explicit-any`, `@typescript-eslint/no-no-restricted-imports`, `@typescript-eslint/no-deprecated`, the `@microsoft/sdl/*` set and `depend/ban-dependencies`. Neither of yours is on it, and **you disable no `obsidianmd/*` rule**. `require('node:module')` sits after the early `if (!Platform.isDesktop) throw`, which the rule's own test file lists as an accepted shape. So this should pass review unchanged.

Two small things, neither blocking:

- **`loopback-flow.ts` now has a comment that contradicts itself.** The block still opens with "v1.25.6: Use module.createRequire(__filename) — the typed Node.js API for creating a CJS require function", and then the bullet beneath it explains why the *bare* `require('node:http')` was rejected for `no-unsafe-*` propagation. That reasoning is still correct for the inner `nodeRequire('node:http')` call, but the adjacent line now says `node:module` is loaded via CommonJS `require`, which is the opposite of what the surviving bullet describes. Worth one pass to make the comment describe the code that is actually there.
- **No test covers the desktop path end to end.** The bundle assertion pins the loading shape, which is the right level for this bug, but nothing executes `requireNodeHttps()` / `requireNodeHttp()` against a real desktop runtime. Given this class of bug only appears in Obsidian's renderer, that is expected — I am noting it so the reliance on the bundle assertion is explicit rather than implied.

I am not merging this. DocTpoint's independent review is being waited on, and it touches OAuth/credential-adjacent code where a second pair of eyes is warranted regardless of how the diagnosis reads.

green-dalii:
Adding to my review above, because I approved before reading #751 and that changes part of the picture. Not withdrawing the approval — the diagnosis in this PR is correct and so is the fix — but the two PRs are addressing the same defect at different levels, and that should be decided rather than merged twice.

**#751 (filed today) names the root cause this PR works around.** `esbuild.config.mjs` sets `format: 'cjs'` and `target: 'es2018'` but never sets `platform`, so it defaults to `browser`; `node:module` is in `external`; and under the browser platform esbuild leaves an external dynamic import as a **native** `import()`. The proposed fix is one line in the build config — `supported: { 'dynamic-import': false }` — measured on the artifact as `import("node:module")` 2 → 0 and `require("node:module")` 0 → 2, and verified end to end on a real vault where streamed answers resumed.

**I checked the obvious objection to that global switch, and it holds up.** Disabling dynamic import is not targeted at Node built-ins, so it applies to every `await import()` in the bundle — 37 of them, of which only **2** are `node:` builtins. The rest are `await import('ai')` and `await import('./…')`, and I confirmed those are unaffected because `ai` is **not** in the `external` list: the shipped `main.js` contains `require("ai")` 0 times and `import("ai")` 0 times, so it is already inlined into the single-file CJS bundle. Only `obsidian`, `electron`, `@codemirror/*`, `@lezer/*` and the Node builtins are external, and of those only the builtins are dynamically imported. So the switch changes exactly two call sites in practice.

**What that implies for this PR.** If #751 lands, `await import('node:module')` compiles to `require("node:module")` on its own — which is precisely what this PR's updated bundle assertion checks for (`not.toContain('import("node:module")')` plus `toContain('require("node:module")')`). That assertion would then pass **without** the source change here, so the two are alternative fixes for the same defect rather than two halves of one.

Not saying this one is wrong. Changing the call sites makes the CommonJS requirement explicit in the source, which is arguably the better place for it, and the config fix is invisible to anyone reading `node-https-fetch.ts`. But the disables this PR adds (`@typescript-eslint/no-require-imports`) exist only to support that source-level form, and they would be unnecessary workarounds if the build emits `require` anyway.

My suggestion, for whoever merges: take **#751 first** since it is one line, verified end to end, and fixes the Codex path (#665) and this transport together — then re-evaluate whether this PR still has a job, rather than landing both and having a future reader wonder which mechanism is load-bearing.

green-dalii:
Cross-reference for whoever picks this up: #753 (open PR) fixes the same defect at the call sites rather than in the build config. #751 is the root cause and the smaller change; they are alternative fixes for one bug, not two halves. See the latest comment on #753 for the verification that the global `supported: { 'dynamic-import': false }` switch only affects the two `node:` imports in practice, because `ai` is not external and is already inlined into the single-file CJS bundle.

green-dalii:
Adding context rather than a verdict, because the diagnosis in this PR is correct and the fix works — the question is only which layer to fix at.

What you found is a build-output shape, and it has a cause upstream of the source. `esbuild.config.mjs` sets `format: 'cjs'` and `target: 'es2018'` but never sets `platform`, so esbuild defaults to `browser`. Under that platform it assumes a native `import()` is available and leaves an *external* dynamic import exactly as written — and `node:module` became external in `54bc1281` (2026-07-21, the Codex OAuth provider), when `builtinModules.map(m => \`node:${m}\`)` was added to the `external` list. So the source has been correct the whole time; the bundle is what turned it into an ESM import inside a CJS file.

That is filed as #751, with the one-line fix (`supported: { 'dynamic-import': false }`), and it was verified end-to-end on a real vault: `import("node:module")` 2 → 0, `require("node:module")` 0 → 2, and the Codex OAuth stream resumed.

Why that matters here: once the build can emit the `require` form itself, `await import('node:module')` compiles to it on its own — which is precisely what this PR's bundle assertion checks for. The two are alternatives for the same defect, not two halves of one fix, and I would rather fix the cause than keep a source-level workaround under a build setting that is still wrong.

One thing worth keeping from this PR regardless: the diagnosis, and the reminder that the `TypeError` safety net in that path makes a missing transport fail silently. That is what made this take a report to find rather than a log line. Thank you for it — I will leave the PR open for a day or two in case you disagree with the layer, and either way the credit for finding it is yours.

DocTpoint (changes requested):
Reviewed at `85a6ba9`. The diagnosis is correct and the fix works — I built both
this branch and #759 from clean installs and measured the artifacts rather than
reading the descriptions.

**Requesting changes as a hold on the merge, not as a judgement on this branch.**
This branch is green on its own, and I have now reviewed #759 as well rather than
holding this one in favour of something unexamined — it is approved, with two
non-blocking notes and one measurement added to it. What is not safe is merging it in the repo's
current state, and right now nothing stops that: #753 is `CLEAN` and `APPROVED`,
#759 is `BLOCKED` behind a required review. So whichever is clicked first decides
which layer this is fixed at, and the two are alternatives. A comment does not
gate, which is why this is a review. What clears it is named at the end, and I
will clear it the moment the layer is decided — including immediately, if the
decision is this branch.

## Three arms

| Arm | Build | `import("node:module")` | `require("node:module")` | deferred wrapper | Codex flow test |
|---|---|---|---|---|---|
| #753 alone | ok | 0 | 2 | 0 | **21/21 green** |
| #759 alone | ok | 0 | 2 | 2 | **21/21 green** |
| both (this branch's source, #759's build config and test) | ok | 0 | 2 | 0 | **red** |

The failure in the third arm is
`AssertionError: expected [] to have a length of 2`.

Both fixes work. The collision has two levels. The visible one is that both
branches edit `openai-codex-loopback-flow.test.ts` with opposite assertions, so
a merge conflicts and someone has to choose. The one worth naming is what
happens if that conflict is resolved in favour of this branch's source: #759's
laziness assertion has nothing left to match, and with it the mutation guard
`expect(loads).toHaveLength(deferred.length)` degenerates to `2 === 0`. The
bundle is still lazy — the `require` sits inside the guarded async function —
but nothing in the artifact proves it any more.

## Which layer

I would fix this in the build config, for two reasons that are about the code
rather than about the two changes being alternatives.

**The source was never wrong.** `esbuild.config.mjs` never sets `platform`, so
it defaults to `browser`, and under that platform an external dynamic import is
left exactly as written. `node:module` became external in `54bc1281`. Changing
the two call sites removes today's two symptoms and leaves the build defect in
place, so the next `node:` builtin reached through a dynamic import fails the
same way, with no test standing between it and a release.

**The property stops being checkable.** #759 pins laziness structurally: the
`Promise.resolve().then(() => __toESM(require(...)))` wrapper appears only for a
deferred load, so an eager one added later cannot pass. After this branch's
change the bundle contains a bare `require("node:module")` whose position in the
source is the only thing making it lazy, and that is not visible in the
artifact. The assertion that replaces it — `toContain('require("node:module")')`
— passes for an eager load too. That is the same shape as the bug the old
assertion had, in the other direction.

The real argument against the build-config fix is that
`supported: { 'dynamic-import': false }` is global to the bundle rather than
local to two functions. That is the right thing to be careful about, and it is
now measured directly rather than by proxy. I built #759 twice, identical except
for that one line, and diffed the output: **two lines differ in 4,539,124 bytes**,
and both are the intended sites. The source has around thirty other runtime
`import()` calls — `ai` in three SDK clients, `pdf-cache`, and a `Promise.all`
over four client modules — and every one of them was already require-based under
`format: 'cjs'`, so the switch does nothing to them. The global switch is safe,
not merely untested.

For completeness, since it came up as an argument and does not hold: this branch
lints clean. The two `eslint-disable` directives do their job and the Obsidian
ruleset is satisfied.

## What clears this

Any one of three, and none of them is work on this branch's code:

1. The layer decision goes to #759 and this branch is closed as superseded. I
   clear the review on that call alone.
2. The layer decision goes to this branch. I clear the review immediately, and
   #759's test is rewritten rather than dropped — the laziness property needs a
   different guard, and asserting that `require("node:module")` does not appear
   at module scope is the equivalent claim. Worth having whichever branch lands.
3. This branch is converted to draft, or the standing approval is dismissed by
   its author, so the merge cannot happen ahead of the decision. Then the hold
   has no job and I clear it.

The one outcome to avoid is both merging as they stand. That is the only thing
this review is protecting against.

A note on the standing approval, since I am asking for it to be paused rather
than second-guessing it: it was given before #751 was read, and its author said
so on this thread. So this is not a reviewer overriding a considered call — it is
the gate being re-closed on the same information that reopened the question.

green-dalii:
Making the layer call, so this can be resolved rather than held.

**#759 merged as `9e97e903` (#751 closed).** The fix lands in the build config, on DocTpoint's reasoning above, and I agree with it for the reason that decides it: the source was never wrong. Changing the two call sites removes today's two symptoms and leaves `platform` unset, so the next builtin reached through an external dynamic import fails the same way with no test in the path.

**What that means for this branch:** it is superseded, not wrong. To be explicit about the part that matters — the diagnosis here was correct, I verified it independently before #751 existed (`main.js` carried `import("node:module")` twice and `require("node:module")` zero times), and it is what led to the build defect being looked for at all. `@svsk417` found the thing; the only question was which layer to fix it at, and a merged PR had to exist before that question could be closed.

Closing as superseded rather than leaving it open. **What would reopen it:** nothing about this fix — it is merged. But the two `eslint-disable` directives in `node-https-fetch.ts` are still needed, because that file uses `createRequire(__filename)` with a local cast, and the question of whether the typed-Node-API shape can lose its disables is a live one worth its own issue if anyone wants to pick it up.

Thank you for the work, and for the measurement in your own review that kept this from being a coin flip.

Issue #752, opened 2026-09-18
Settings tab jumps back to the top after any change that re-renders the panel


aisahpA:
## Bug Description

Every settings change that alters the panel's layout re-renders the whole tab and throws the scroll position away, so the user lands back at the top of the page. On a tab this long — 70 `new Setting(...)` construction sites across nine sections, seven of them section headings — that means scrolling back down after each single toggle.

Only some settings do it: the ones that change the shape of the panel (expand a collapsed block, switch provider, switch language, switch model scope). Plain values such as temperature, `periodicLint` or `createWelcomeNote` are unaffected. That distinction turns out to be the whole story, and it is why this looks like a rendering bug rather than a state bug.

## Environment

- **Plugin version:** 1.27.2 (source read at commit `8c7966e`)
- **Obsidian version:** 1.13.4 (desktop)
- **OS:** macOS. The mechanism below is platform-independent — it is `containerEl.empty()` plus the browser's scroll clamp — so it should reproduce on Windows and Linux identically.
- **Wiki size:** Not a factor. The tab always overflows the settings modal (`getPopoutOptions()` caps it at height 700), so the jump is reachable on any vault.

## LLM Configuration

N/A — this is a Settings-tab rendering bug. It reproduces before any provider is configured, and no LLM call is involved.

## Steps to Reproduce

1. Open Settings → LLM Wiki.
2. Scroll down to any of the controls listed under **Root Cause** below. The easiest is the bottom "Advanced settings" toggle — its section is at the very end of the tab.
3. Change the control.
4. The panel re-renders and the scroll position resets to the top.

Concretely, all of these reproduce it:

| # | Control | Result |
|---|---|---|
| 1 | Bottom **"Show advanced settings"** toggle | jumps to top |
| 2 | **LLM provider** dropdown | jumps to top |
| 3 | **UI language** / **Wiki output language** dropdown | jumps to top |
| 4 | **Model scope** dropdown (unified ↔ per-task) | jumps to top |
| 5 | **Auto-watch sources** toggle / **Auto-watch mode** / **Web Clipper preset** | jumps to top |
| 6 | **LLM Advanced** mode dropdown (preset ↔ custom) | jumps to top |
| 7 | **Tag vocabulary** dropdown (default ↔ custom) | jumps to top |
| 8 | Model picker: choosing "custom input…" from the dropdown, or the "use dropdown" button | jumps to top |
[...]

PR #754, opened 2026-09-18
docs(memory,roadmap): the ROI-ordered work list and the #701 decision record

green-dalii:
docs(memory,roadmap): the ROI-ordered work list and the #701 decision record

Compact-prep: the state a resuming session needs, written to the files rather than left in a conversation.

## `## Work list (2026-09-18)` — ordered by ROI

ROI is stated as `(impact × certainty) / effort`, with certainty meaning "we are confident the fix lands *and stays landed*". That ordering puts **#751** first: one line in `esbuild.config.mjs`, and it repairs two shipped features — Codex browser login and the desktop streaming transport from #746 that has never once loaded in a build.

The list also records the check that makes #751 safe rather than obvious. `supported: { 'dynamic-import': false }` is a **global** switch over 37 `await import()` sites, but only **2** are Node builtins: `ai` is not in the `external` list, and the shipped bundle contains `require("ai")` zero times and `import("ai")` zero times, so it is already inlined into the single-file CJS bundle. Only `obsidian`, `electron`, `@codemirror/*`, `@lezer/*` and the builtins are external, and only the builtins are dynamically imported.

And the relationship a future reader would otherwise re-derive: **#751 and #753 are alternatives, not two halves.** If #751 lands, `await import('node:module')` compiles to `require` on its own, which is exactly what #753's bundle assertion checks for — so that assertion would pass without #753's source change. Prefer #751; #753's `@typescript-eslint/no-require-imports` disables exist only to support a form the build can produce by itself. **Decide one, not both.**

Tiers 1–4 cover the review queue (all waiting on DocTpoint or an author, none on us), the v1.28.0 chain with its two documented couplings, the PATCH-worthy bugs, and the backlog that should be re-triaged rather than carried indefinitely.

## `## Design record — #701's premise does not hold`

#701 writes a `wiki-ingested:` marker into the user's own notes. The deferral is about the premise, not the code, and this records why:

- "This note has been ingested" is a **derived** fact, so it can live in plugin state, in the wiki's own artifact, or in the user's note. The middle option is not a proposal — `buildIngestedHashes()` reads `contentHash` from `wiki/sources/*.md` frontmatter today, and `scanners.ts:485` already uses the same field.
- **Both failure modes the PR cites are already handled.** The batch cache is a memoisation that is recomputed from the wiki pages when it expires, so a restart does not re-ingest. And `hashBody` is FNV-1a over the normalised **body only** — no path, no filename — so the same body under a different name is already deduped.
- **The marker's failure mode is the user's main recovery action.** Deleting a `wiki/sources/` page is how someone forces a rebuild; the hash mechanism cooperates and the marker opposes it, silently skipping the notes the user was trying to rebuild.

What would reopen it is recorded too: a specific demonstrated case, not a reasoning chain.

`## Current state`, the resume point and ROADMAP's scope groups are refreshed to `3499da2a` / 305 files / 4240 tests, and **#751 and #752 were given milestones** — both had none, which is how an issue that describes a shipped feature being dead can go unnoticed.

Docs only. Gate 1 green at 304 files / 4245 tests.


green-dalii:
Merging. Docs only, Gate 1 green. Adds the MEMORY audit to the handoff commit: the #729 design record was mirroring the issue section-for-section and had already begun to diverge, so it is now a pointer plus the two corrections made after the issue was written and the phase status. Review event is a comment rather than `--approve` because GitHub blocks self-approval on an own PR.

PR #755, opened 2026-09-18
feat(prompt): extraction focus and content requirements settings (+ shadow-source guard for machine-ID entity names)

weqoocu:
## Summary

Two additions, both born from live-vault friction on a 7-topic multi-wiki setup:

### 1. Prompt customization — "Extraction Focus" and "Content Requirements"

Two free-text fields (WeKnora-style wiki prompt configuration) that customize what the LLM prioritizes without touching any system-owned contract:

- `extractionFocus` — appended to the extraction prompt's *Extraction Scope* block (after the granularity instruction) as a domain-priority hint: which entities/concepts this wiki should recognize first. It narrows what the model looks for; it never replaces the JSON protocol, the classification decision tree, or the tag vocabulary.
- `contentRequirements` — appended to the entity/concept page-generation prompts via a new placeholder (after the merge-strategy block): expressive emphasis, structure, must-cover aspects. Citation format, wiki-link rules, merge strategy and anti-hallucination requirements stay binding — the injected section says so explicitly.

Design points:

- Both resolve **per-topic first** (a `WikiTopic.extractionFocus` / `contentRequirements` entry, matched on the topic folder) with the global settings as fallback — different topics can carry different domain priorities.
- Empty = off = **byte-identical prompts**, so prompt prefix-caching is unaffected for vaults that never configure them.
- Capped at 4000 chars (same ceiling as WeKnora's inputs).
- Settings UI: a *Prompt Customization* section in Wiki Configuration with two textareas.
- i18n: 8 new keys across all 11 locales.

### 2. Shadow-source guard for machine-identifier entity names

When the ingested note's own filename is a machine identifier (a dinox UUID like `01988c6e_86e9_7b79_83b5_c8c4b354fdc9.md`, an ob timestamp like `202511042307.md`), the LLM sometimes extracts the source DOCUMENT itself as an entity and lifts the filename as its name — producing junk `entities/<uuid>.md` pages whose body merely re-narrates the source summary.

Two layers:

1. `isSourceSelfReference` in `source-analyzer.ts` drops extracted items whose name is a machine shape: UUID with `-`/`_` separators, 12/14-digit timestamps, **and now runs of 13+ consecutive digits** optionally with a `_`/`.` + hex hash tail — IM voice-note message IDs (`1921147873752416344`, 19 digits) and snowflake-style keys slip past the 12/14-digit gate. Phone-sized runs (11 digits or fewer) and years still pass.
2. The analyzeSource prompt now forbids extracting the source document itself or using the source filename as an entity/concept name.

Semantic names are never touched — `TNF-α` extracted from `TNF-α.md` survives, as does a drama name from a hash-suffixed filename.

## Testing

- 19 new tests: extraction-focus/content-requirements resolution precedence (incl. empty-override-as-off-switch and folder normalization), section rendering guards, placeholder round-trip, shadow-source shapes (19-digit drop, hash-tail drop, 13-digit boundary, phone-size survival).
- Full suite: **4270 passed** (306 files).
- Gates: `eslint src/ --max-warnings=0` clean, `tsc --noEmit` clean, `css-lint` 0 violations, `lockfile-check` pass, production `build` OK.

## Notes

- Version bumped to `1.27.3` on top of current main; CHANGELOG entry added in the house style.
- No personal paths, identifiers, or vault names are included — test fixtures use neutral `notes/...` paths.


green-dalii:
The shadow-source guard is the half I would have asked for: 13+ consecutive digits covering IM message ids and snowflake keys is exactly the shape the 12/14-digit gate misses, and keeping `TNF-α` and hash-suffixed drama names is the right boundary. `isSourceSelfReference` running in the analyzer rather than in the prompt is also the correct layer — a prompt instruction cannot be relied on for a drop.

Two process things, neither about the code:

1. Could you drop the 1.27.3 bump (`manifest.json`, `package.json`, `CHANGELOG.md`)? Version and release notes are cut by the maintainer at release time, in a step that bumps them itself — so this PR will conflict with that step and with every other PR touching those files.

2. Extraction focus and the guard are independent of each other. If they can be two PRs — or two commits we could squash-merge separately — that lets us schedule them against different milestones. If they are one unit for a reason I am not seeing, say so and we take them together.

Also: `mentions.test.ts` moves +69/−1. What is the −1?

PR #757, opened 2026-09-19
perf(index): hold the wiki page index per file, and give every caller the same seam

DocTpoint:
## What

`getExistingWikiPages` reads every wiki page in full, and the callers that need it run per written page — `applyRelatedLinks` alone fires on every write. The engine held the result behind a 5-second TTL that every write dropped whole, so during an ingest the walk effectively ran from scratch each time.

Two halves, and the first is why the second was not possible before.

## The seam

Six modules imported the module function and walked the vault again: `path-resolution`, `related-page`, `related-links`, `fix-dead-link`, `fix-polluted-page`, `source-analyzer`. Two more sites inside the engine went around its own cached accessor — `wiki-engine.ts:1104`, which the #603/#662 design record names, and the context wiring at `:234`. All eight now go through `getExistingWikiPages()` on the context they already carry.

`RelatedLinkCtx.app` existed only for that second walk and is gone with it. `LintPhaseContext` carried a second, narrower declaration of the page shape (`{ path: string }`); it is `WikiPageRef` now like the rest, so a lint phase can reach `text`, `tags` and `displayTitle` through the same accessor.

## The index

`WikiPageIndex` keeps the per-file result. Listing the vault costs nothing (`getMarkdownFiles` reads no file), so the list is rebuilt on every call and only the bodies are reused: an entry is kept while its file's `mtime` and `size` are unchanged. Order stays the vault's own, so a page created mid-run appears exactly where a full walk would have put it.

**The writer invalidates as well, and that is not redundant.** This is the design record's "run-scoped index updated by the writer" — with a verification floor under it, for two reasons the record itself supplies. A filesystem `mtime` has second granularity, so a write followed by a read inside the same second would otherwise go unseen. And the writer's own invalidation only knows about writes that go through the engine, while the record's finding at `wiki-engine.ts:1104` is that *"the engine does not consistently use its own front doors"*. The stat check is what catches every other writer: the user saving a note during an ingest, Obsidian's link update after a rename, a sync client, another plugin.

`TFile.stat` is a live view, not a snapshot, so the stat is taken **before** the body. Read afterwards, a file changed mid-read would be held as {old body, new stat} — a pair that looks valid on every later call, so the stale body would be served until something changed the page again. There is a test for it.

## Two things stated rather than implied

**Memory.** The index holds what the TTL discarded: every page's entry stays resident until its file leaves the vault. An entry is aliases, tags and a body slice bounded by `CANDIDATE_WINDOW_TEXT_CHARS` (2,000), so a 2,500-page vault holds on the order of 5 MB. That is the price of not re-reading the whole vault once per written page.

**A residual.** A write from outside the engine, inside the same filesystem second as the read that filled the entry *and* to exactly the same byte length, is invisible to both halves. The next change to that file clears it.

## Notes

- `displayTitle` (the page H1, #592) is carried through the extracted entry builder unchanged, and pinned by a test that reads it through a reused entry — it is what a link repair shows the reader, via `dead-link-detector` and `fix-dead-link`.
- `PAGES_CACHE_TTL_MS` is renamed `INGESTED_HASHES_TTL_MS`: the page list no longer has a TTL, and the constant now has one owner.
- The module-level `getExistingWikiPages` has no production caller left; it stays as the one-shot reader for a caller with nowhere to hold an index, and its doc comment says so. Say if you would rather it went.
- `createMockEngineContext` returned an empty page list. Several tests depend on their mock vault holding pages, so it now delegates to the real reader — an empty list would let them pass for the wrong reason.
- Gate 1: 305 files / 4253 tests, against 304 / 4245 on `46e48db4`.

The further step the record describes — the writer handing the index the content it just wrote, instead of dropping the entry — layers cleanly on this and would remove the re-read of the pages an ingest writes. It is not in this PR.


green-dalii:
Reviewed. The seam is the right half to do first — eight call sites going around the engine's own front door was the actual defect, and the #603/#662 record names :1104 as the one that proved it.

Two things I want to check before merging, both of which you stated plainly rather than left implied, which is why they are questions and not objections:

1. `createMockEngineContext` now delegates to the real reader, and thirteen existing test files moved with it (+5 to +8 lines each). I want to confirm those thirteen are "the mock vault now holds the pages it must" rather than "assertions relaxed to match the new behaviour". Gate 1 passes either way, which is exactly why I am asking.

2. `wiki-engine.ts` overlaps an open PR of mine (#750, the write-path contract). Only that one file collides, so we merge one and rebase the other — no action needed from you, just flagging that the ordering is ours to sequence.

On the module-level `getExistingWikiPages`: keep it, and keep the doc comment that says why. A one-shot reader with no production caller is one honest seam; deleting it would push a future caller into inventing a worse one.

The stated memory cost (~5 MB at 2,500 pages) and the same-second residual are both the right call to state rather than imply. The stat-before-body ordering is the part I would have got wrong — reading the body first would hold {old body, new stat}, a pair that looks valid forever.

green-dalii (approved):
Approving. I raised two things when I commented, and I can answer both myself now — the first by reading the diff rather than by asking, the second because it is mine to sequence.

**On the thirteen test files.** They are one mechanical pattern (`getExistingWikiPages(): … { return mockExistingWikiPages(this)(); }`), and the production change in `related-page.ts` shows it replacing an inlined call that was byte-identical: `getExistingWikiPages(ctx.app as never, ctx.settings.wikiFolder)`. So the mock reproduces what was there before the seam, which is what a mock owes a call site that moved. The change to `createMockContext` is in the same direction and strictly safer: `async () => []` becomes a delegation to the real reader, which can add pages to a test's view and never remove them. Your comment on it — that an empty list would let those tests pass for the wrong reason — is the correct call, and it is the opposite of relaxing an assertion. Question withdrawn.

**On the merge order.** `wiki-engine.ts` is the only file this shares with #750. #750 has blocking findings of its own to clear now, so this goes first and #750 rebases onto it. Nothing further needed from you.

**On the module-level `getExistingWikiPages`:** keep it, with the doc comment you added. A one-shot reader with no production caller is one honest seam; deleting it would push a future caller into inventing a worse one.

**What I want recorded from this PR**, because it is the part that is easy to lose: the ordering fix. `TFile.stat` is a live view, so a stat taken *after* the body would be held as `{old body, new stat}` — a pair that looks valid on every later call, so the stale body would be served until something else touched the page. That is a defect that would have surfaced months later as an unexplained stale read, and the test for it is why it will not.

Same for stating the two costs out loud rather than leaving them implied: ~5 MB resident at 2,500 pages, and the same-second/same-length residual that both halves of the invalidation miss.

PR #759, opened 2026-09-20
fix(build): let esbuild emit a loadable form for external node: imports

### LINKED ISSUE #751: esbuild leaves external node: imports as native import(), so the #746 desktop transport never loads in the renderer

green-dalii:
## What

`esbuild.config.mjs` sets `format: 'cjs'` and `target: 'es2018'` but never sets `platform`, so esbuild defaults it to `browser`. Under that platform an **external** dynamic import is assumed to be available natively and is left exactly as written. `node:module` became external in `54bc1281` (2026-07-21, the Codex OAuth provider) when `builtinModules.map(m => \`node:${m}\`)` was added to the `external` list.

So `await import('node:module')` reached the shipped bundle as a native ESM import inside a CJS file. It could not load, and every caller catches the `TypeError` — which is why a transport that never worked read as a transport that sometimes did.

`supported: { 'dynamic-import': false }` makes esbuild emit the require-based form a CJS bundle can load.

## Two shipped features were disabled by this and are fixed by it

- **ChatGPT Plan (Codex OAuth) browser sign-in** — #665
- **The desktop `node:https` streaming transport** added in `e150d139` (#741's second half), which never loaded once in a production build

Verified on a real vault before committing: the Codex OAuth stream resumed.

## Artifact

| Metric | Before | After |
|---|---|---|
| `import("node:module")` | 2 | **0** |
| `require("node:module")` | 0 | **2** |
| `require("ai")` | 0 | 0 |
| require targets | — | only `obsidian` (45) and `node:module` (2) |
| `console.debug` neutralised | yes | yes |
| bytes | 4,539,044 | 4,539,124 (+80) |

`require("ai")` staying at 0 is what says the global switch did not externalise anything that was inlined.

## The test that was protecting the bug

Gate 1 caught this immediately, and the failure is the most interesting part of the change:

```
FAIL  Codex loopback flow > bundles desktop HTTP as a lazy ESM dynamic import of node:module
AssertionError: expected '/*\nTHIS IS A GENERATED/BUNDLED FILE …' to contain 'import("node:module")'
```

`openai-codex-loopback-flow.test.ts` asserted the bundle **must contain** `import("node:module")` — the shape that makes the load fail. Its own comment states the requirement it was written to protect:

> the bundle **MUST NOT** eagerly import node:http at module-eval time (which would defeat the `Platform.isDesktop` guard)

That is a property — laziness. The third assertion was a transcription of the mechanism the author observed at the time, and the two had already come apart the day it was written. It is re-pinned at the property:

```ts
const loads    = bundle.match(/require\("node:module"\)/g) ?? [];
const deferred = bundle.match(/Promise\.resolve\(\)\.then\(\(\) => __toESM\(require\("node:module"\)\)\)/g) ?? [];
expect(deferred).toHaveLength(2);
expect(loads).toHaveLength(deferred.length);   // an eager load cannot be introduced without failing
```

The old assertion passed for **any** occurrence, including an eager one. The desktop guards are now asserted **by position** before their loads rather than by a character budget — my first attempt used `{0,120}` and the measured gap was 168, which is exactly the kind of guess that turns into a mystery failure later.

Mutation-tested:

| Mutation | Result |
|---|---|
| Remove `supported` | fails 1 — `expected [] to have a length of 2` |
| Append an eager `require("node:module")` to the artifact | **fails 1 — `expected 3 to have a length of 2`**, which the previous assertion did not catch |

## Obsidian Bot

The ruleset is scoped to TypeScript — `eslint.config.mjs` declares `files: ["**/*.ts"]` — so `esbuild.config.mjs` is outside it, and this change touches no `.ts` file. The Bot-visible source is byte-identical; the `Platform.isDesktop` early-exit with the dynamic import inside the guard block is unchanged, which is the shape `obsidianmd/no-nodejs-modules` accepts.

For anyone reading the history: the earlier Bot-compliance change to this file was `a0901daf` (v1.7.8), which replaced the `builtin-modules` **dependency** with Node's native `module.builtinModules`. It did not touch `platform` or `supported`; neither keyword has ever appeared in this file.

## Gate 1

`304 files / 4245 tests`, build before test, lint/tsc/css-lint clean.

## Gate 4: Performance

| Dim | Status | Notes |
|-----|--------|-------|
| CPU | N/A | Build-time configuration; no change to any runtime path besides the load itself now succeeding |
| Memory | ✅ | No change; `require` and the previous `import` both resolve the same module once |
| IO | ✅ | The failed ESM load is replaced by a working one — one module resolution either way |
| Network | N/A | No request-path change |
| Token | N/A | No prompt or payload change |

Closes #751


green-dalii:
Context for review, and one request.

**On approval:** this is my own PR, so I cannot satisfy the ruleset's approval requirement myself. If you are satisfied, a `--approve` from you is what lets it merge.

**What is worth your attention:** the second file. The change to `esbuild.config.mjs` is one line; the interesting part is that a test had to move, and the reason it had to move is the finding.

`openai-codex-loopback-flow.test.ts` asserted the bundle **must contain** `import("node:module")` — the exact shape that made the load fail. Its own comment says what it was written to protect: that the bundle "MUST NOT eagerly import node:http at module-eval time (which would defeat the `Platform.isDesktop` guard)". That is a property. The third assertion was a transcription of the mechanism the author saw at the time, and the two had come apart the day it was written, silently, because nothing ever exercised the load.

So the question I would most like a second opinion on is whether the replacement is actually stronger rather than merely different. It now requires every `require("node:module")` to be the deferred `Promise.resolve().then(() => __toESM(...))` wrapper, so the count of loads must equal the count of deferred loads — an eager load cannot be introduced without failing. I mutation-tested it both ways: removing the setting fails (`expected [] to have a length of 2`), and appending an eager `require("node:module")` to the artifact also fails (`expected 3 to have a length of 2`). The second is the one the old assertion could not have caught.

I also replaced a `{0,120}` character budget around the guard with a positional assertion, because the measured gap is 168 — a magic distance in a test is a mystery failure waiting for a later refactor.

**On #753:** this is an alternative to it, not a companion. svsk417's diagnosis was correct and I verified it independently — `main.js` carried `import("node:module")` twice and `require("node:module")` zero times. Their fix works at the source layer; this one corrects the build setting that produced the shape. Once this lands, `await import('node:module')` compiles to `require` on its own, which is precisely what #753's bundle assertion checks for — so #753 would pass without changing its source. I have commented there explaining this and left it open rather than closing someone else's diagnosis, and the credit for finding it is theirs.

**Scope:** no documentation surface — no README, CHANGELOG or schema touch. The CHANGELOG entry belongs to the release step, not to this PR.

DocTpoint (approved):
Approving at `c7ab134`. Built from a clean install, reproduced both mutations and
the artifact table, and checked the one thing the description leaves as a proxy.

## The blast radius is two lines, and that is measurable directly

`supported: { 'dynamic-import': false }` is global to the bundle, which is the
one real argument against fixing this at the build layer. Your table answers it
with `require("ai")` staying at 0 — a good indicator, but an indirect one, and
the source has around thirty other runtime `import()` calls besides the two this
is about: `ai` in three SDK clients, `../core/pdf-cache`, and the `Promise.all`
over four client modules in `create-llm-client.ts:205`.

So I built your branch twice, identical in every respect except that one line,
and diffed the outputs:

```
727c727
<   const nodeModule = await import("node:module")
---
>   const nodeModule = await Promise.resolve().then(() => __toESM(require("node:module")))
93824c93824
   (same, second site)
```

Two lines in 4,539,124 bytes, both the intended ones. Every internal dynamic
import was already require-based under `format: 'cjs'`, so the switch is a no-op
for all of them. That is the direct form of the claim and I think it is worth
putting in the record, because it is what makes the global switch safe rather
than merely not-observed-to-be-harmful.

Mutations reproduce exactly as described: removing `supported` gives
`expected [] to have a length of 2`; appending an eager `require("node:module")`
to the artifact gives `to have a length of 2 but got 3`. Unmutated control green,
21/21.

## Two non-blocking notes

**The positional assertions pair by bundle order, and that is unstated.** The
test reads the first and second `require("node:module")` and asserts
`guardHttps < firstLoad` and `guardLoopback < secondLoad`. Measured on this
build:

| | offset |
|---|---|
| `guardHttps` | 25,004 |
| first load | 25,125 |
| `guardLoopback` | 3,510,595 |
| second load | 3,510,711 |

The two sites are 3.5 MB apart, and the assertion holds only because
`node-https-fetch` happens to be emitted before `loopback-flow`. If that order
ever flips — an import reshuffle, a tree-shaking change — `guardHttps` lands
after the first load and the test goes red on something harmless. That is the
mystery failure you were avoiding when you dropped the `{0,120}` budget, one
level up. An order-independent form of the same claim: assert that no
`require("node:module")` appears before the earlier of the two guards, i.e.
`expect(bundle.slice(0, Math.min(guardHttps, guardLoopback))).not.toContain('require("node:module")')`,
together with the `deferred` count you already have.

**`platform` is still inferred rather than declared.** The comment explains that
esbuild defaults it to `'browser'` and that this is what made the external
dynamic import survive — but the config still does not say so. Declaring
`platform: 'browser'` next to `supported` would make the premise of the fix a
statement rather than a default someone has to know about, and it would keep the
fix meaningful if that default ever moves. Same reasoning as the write intents in
#750: the field is not there to change behaviour, it is there so the next reader
does not have to infer it.

## One housekeeping point

This branch is based on `46e48db4`, which predates #757. GitHub reports it
mergeable, so nothing is blocked by it — noting it only so the rebase is a choice
rather than a surprise.

## On #753

I have reviewed both now and the layer question looks settled to me, on the
measurement above rather than on preference: this fixes the cause, and the bundle
diff shows it costs nothing elsewhere. The other branch changes source that was
never wrong and, more to the point, leaves the laziness property untestable — a
bare `require("node:module")` inside a guarded function is lazy by position only,
and nothing in the artifact distinguishes it from an eager one. I have said so on
that thread.

PR #760, opened 2026-09-20
fix(settings): stop dropping namespaced model ids and stop overwriting a typed one

### LINKED ISSUE #758: LM Studio provider: model dropdown filters out Hub-managed models with publisher-prefixed ids (qwen/..., openai/...) and wipes manually typed ids after fetch
### Environment
- Karpathy LLM Wiki plugin **1.27.2** for Obsidian, LM Studio provider (LM Studio 0.3.x on macOS)
- LM Studio model keys: 10 flat keys (e.g. `ornith-1.5-35b-a3b-mlx`) + 2 Hub-managed prefixed keys: `qwen/qwen3.6-35b-a3b`, `openai/gpt-oss-120b`

### Bug
The model list filter drops every id containing `/` for all providers except ollama (from bundled `getModelFilter`):

```js
if (provider === "ollama") return (id) => !id.includes("/");
else return (id) => !id.includes(":") && !id.includes("/");
```

LM Studio's newer Hub-managed downloads get `publisher/model` keys, so those models can never appear in "Fetch Available Models". Worse, after a successful fetch the plugin forces `useCustomModel = false` and resets the `model` field to the first list entry — so a manually typed prefixed id (which works fine at request time, verified HTTP 200 against `/v1/chat/completions`) gets silently wiped.

### Suggested fix
For `lmstudio` (and generic openai-compatible) providers, allow `/` in ids:

```js
else if (provider === "lmstudio") return (id) => !id.includes(":");
```

Local workaround applied: one-line patch of the bundled `main.js` (must be re-applied after every plugin update).

green-dalii:
## What

Two defects on the same fetch button, both reported with the file and line to check them at, and both verified here before changing anything. Neither is one line.

**1. Namespaced model ids could never reach the picker.**

`/` was rejected for every provider, ollama included — ollama's own exemption was for `:`, its tag syntax — and only openrouter was left unfiltered. The effective rule was `!id.includes(':') && !id.includes('/')` for everything else. LM Studio's Hub-managed downloads are keyed `publisher/model` — `qwen/qwen3.6-35b-a3b`, `openai/gpt-oss-120b` — and the reporter had already established that such an id returns HTTP 200 against `/v1/chat/completions`. So the setting was unreachable for those models, not merely awkward: no amount of typing could get one into the list, because a fetch could not produce it.

`lmstudio` and `custom` now accept `/`. Ollama keeps its `/` rejection: its `/api/tags` lists a bare name beside its `library/`-qualified twin, and the report is about LM Studio.

`custom` is included because the report asks for it by name — "For `lmstudio` (and generic openai-compatible) providers, allow `/` in ids" — and for vLLM, LiteLLM or any OpenAI-compatible gateway, `org/model` is the ordinary id shape rather than an edge case. An id the filter drops is an id the picker cannot offer.

**Two custom-shaped providers are deliberately left out, and this is a scope decision rather than a claim that they are fine.** `custom-responses` and `anthropic-compatible` have the same unknown-catalogue property, so the same argument would include them. They are not included because no user has reported them, and widening a user-visible list shape for providers nobody asked about is its own change to justify. A characterization test pins that exclusion, so widening them later has to be deliberate. Separately noted, not fixed: `bedrock-anthropic` and `bedrock-openai` ids genuinely contain `:` (inference profiles, `anthropic.claude-…:0`), which this filter still rejects.

**2. A model the fetch did not list was overwritten.**

After a successful fetch, `model` was rewritten to `availableModels[0]` whenever the current value was absent from the list, and `useCustomModel` was forced to `false`. A hand-typed id that works at request time but is missing from this endpoint's listing — the namespaced one from defect 1, a model loaded since the last fetch, a gateway that lists a subset — was silently destroyed, and the user had no way to know.

The id is now kept rather than replaced. An id that *is* in the list still clears the custom flag, which is the case the old code was written for.

**3. And the kept id needed to stay reachable, which is what a first version of this fix missed.**

Leaving `useCustomModel` at `false` for a not-listed id renders a dropdown whose options contain no match for `current`, so the select falls back to the "Custom input…" sentinel. The stored id is then displayed nowhere, and because the sentinel is already the selected option, choosing it again fires no `change` — so the text field cannot be opened to reveal it either. The value survives in settings and is unreachable in the UI, which is a different failure from the one being fixed but not a better one.

That case now returns `{ useCustomModel: true }`. `ModelSelectionPatch` has documented `true` as "the field is the user's own" since this PR opened, and no branch returned it — the state was described and never produced. This makes it real, and the id visible.

## Where the wrong sentence came from

The first version of this PR described the old rule as rejecting `/` "for every provider except ollama". That is wrong twice over — `/` was rejected for ollama too, and openrouter was the provider left unfiltered — and it came from the report itself. #758's excerpt omits the `openrouter` branch that precedes the two lines it quotes, so its prose inherits the omission.

The lesson is narrow and worth keeping: the rule has to be read off the code, not off the report that quotes part of it. The report's diagnosis of the *symptom* was exact and its repro was solid; the summary of the rule it was measuring against was not, and copying that summary into three files is how a wrong sentence becomes documentation.

## Why the rules moved

Both lived inside the settings tab's fetch callback, which is why neither had a test — and the filter had shipped wrong in a way a test would have caught. They are pure functions in `core/model-id-filter.ts` now, so the rule is checkable rather than only observable by fetching from a real endpoint.

## The existing test, and what it was

`settings-codex-sections.test.ts` had one case covering the filter, named **"preserves current LM Studio filtering"**, expecting `['qwen2.5']` — that is, expecting `qwen/qwen2.5` to be dropped. Its name is exact: it was a **characterization test**, pinning a snapshot so a refactor would notice a change, not asserting that the behaviour was right.

It did its job by failing on this change. The expectation is updated with the reason, and the case is renamed to say what it now claims.

That same test is also what caught me overreaching: my first version let ollama accept `/` too, which the report does not ask for and which changes an existing list shape. Reverted to the minimal change.

## Tests

Mutation-tested, one per defect:

| Mutation | Result |
|---|---|
| Remove `lmstudio` from the namespaced set | fails 2 — the new unit test and the characterization test |
| Restore the overwrite | fails 2 — including an idempotency case that repeats the fetch, since the original failure was silent and repeatable |
| Return `{}` instead of `{ useCustomModel: true }` for a not-listed id | fails 2 — the not-listed case and a two-fetch sequence that reproduces defect 3 from the field |
| Drop `custom` from the namespaced set | fails 1 — the generic-endpoint case (#758 asks for it by name) |

`reconcileModelSelection` covers the four cases: nothing chosen yet (seed), chosen and listed (clear the flag), chosen and not listed (keep the id, mark it the user's own), and an empty fetch (change nothing — the caller owns that branch, it shows the failure notice).

The defect-3 test is a two-fetch sequence rather than a single call, because one fetch cannot produce the state: the id has to have been listed once (`useCustomModel = false`) and then absent (`true`).

Gate 1: 312 files / 4361 tests (rebased onto 2a4a99b3).

Closes #758


DocTpoint (changes requested):
Reviewed at `5df2137`. Both defects from #758 are fixed, and I checked the one case I was worried about: a provider switch does not leave a stale id behind, because `provider-section.ts` already clears `model`.

Two things before merge.

**1. One sentence is wrong, in three places.** The PR body, the header of `model-id-filter.ts` and the header of the test file say `/` was rejected "for every provider except ollama". The old code rejected `/` for ollama too (`if (provider === 'ollama') return id => !id.includes('/')`). The provider that was exempt is openrouter; ollama's exemption was for `:`. The same file says the correct thing two paragraphs later ("That ollama rule is kept as it was"), so it contradicts itself. This matters because the wrong sentence points at exactly the change you describe reverting: letting ollama accept `/`. A reader who trusts the header will make it again. Also, in the `NAMESPACED_ID_PROVIDERS` comment, "Widening the filter is the smaller change" says the opposite of what the paragraph means.

**2. The fetch can now produce a state that hides the value.** Case: the model was picked from the list (`useCustomModel === false`), it later disappears from the endpoint, the user fetches again. Both fields are left alone, so `renderModelField` renders the dropdown with "Custom input…" selected. The stored id is not shown anywhere, and because the sentinel is already the selected option, choosing it again fires no `change`, so the text field cannot be opened. This state existed before (via the list button), but the fetch could not create it, because it overwrote the model. Suggestion: in the third case return `{ useCustomModel: true }` instead of `{}`. The text field then shows the id. `ModelSelectionPatch` already documents `true` as "the field is the user's own", but no branch returns it. I derived this from the code; I did not run it in Obsidian.

One question, not a request: the body says widening beyond LM Studio has no report behind it, but #758 asks for "lmstudio (and generic openai-compatible)". For `custom` in front of vLLM or LiteLLM, ids with `/` are the normal case. Staying narrow is fine now that typed ids survive, but the stated reason does not match the report.

green-dalii:
All three addressed on `a127feb6` (rebased onto `2a4a99b3`), plus the one your question opened. Gate 1: 312 files / 4361 tests.

**1. The wrong sentence.** Correct on both counts: `/` was rejected for ollama too, and openrouter was the provider left unfiltered. Fixed in `src/core/model-id-filter.ts`'s header, the test file's header, and the PR body, and the `NAMESPACED_ID_PROVIDERS` comment now argues for *listing* rather than "widening the filter is the smaller change".

Worth passing on: that sentence came from the report. #758's excerpt omits the `openrouter` branch that precedes the two lines it quotes, so its prose inherits the omission and I copied it. I added a short body section about it, because the reusable part is "read the rule off the code, not off the report that quotes part of it" — the report's repro was solid and its diagnosis of the symptom was exact; only its summary of the rule it measured against was wrong.

**2. The hidden-value state.** You are right, and the fix is the one you suggested: the third case returns `{ useCustomModel: true }`. Your observation that `ModelSelectionPatch` documents `true` as "the field is the user's own" while no branch produced it is what made this feel like a real defect rather than a polish item — the state was described and unreachable, which is worse than undescribed.

The test is a two-fetch sequence rather than a single call, because one fetch cannot produce the state: the id has to have been listed once (`useCustomModel = false`) and then absent. Returning `{}` now fails that test and the not-listed one.

**3. The question — you were right, and I widened rather than reworded.** `custom` is in the set now. My "no report behind it" reason was simply wrong: #758 says "For `lmstudio` (and generic openai-compatible) providers, allow `/` in ids", and for vLLM or LiteLLM `org/model` is the ordinary shape.

`custom-responses` and `anthropic-compatible` I left out, but as a **scope** decision and not as a claim they are fine — they have the same unknown-catalogue property and the same argument would include them. A characterization test now pins that exclusion so widening them has to be deliberate. Separately, not fixed: `bedrock-*` ids genuinely contain `:` (inference profiles), which this filter still rejects.

Mutations re-run: dropping `custom` fails the generic-endpoint case; returning `{}` from the third case fails two.

DocTpoint (changes requested):
Reviewed at `a127feb`. All three points are fixed, and I checked each against the code: the sentence is correct in all three places, the third case returns `{ useCustomModel: true }` with a test that reaches the state the real way, and `custom` is in the set. Thank you for the note about #758's excerpt.

One new finding. It was already present at `5df2137`, and I missed it there.

**The ollama `/` rejection no longer applies to tagged ids.** Old rule for ollama: `!id.includes('/')`. New code checks `:` first and returns `provider === 'ollama'`, so the `/` check is never reached. I ran both filters side by side:

- `library/llama3.2:latest`: old rejected, new accepted
- `hf.co/bartowski/Llama-3.2-1B-Instruct-GGUF:Q4_K_M`: old rejected, new accepted
- `library/llama3.2`: rejected in both (this is the only case the test covers)

Ollama ids almost always carry a tag, so in practice the rejection is mostly gone. The header ("That ollama rule is kept as it was") and the body ("Ollama keeps its `/` rejection", "Reverted to the minimal change") say otherwise. Either restore it (`if (id.includes(':')) return provider === 'ollama' && !id.includes('/')`) plus a test for the tagged twin, or keep the new behavior and say so. Accepting `hf.co/...` ids may even be the better outcome, but it should be a stated decision.

Small: the test comment "this fix does not reach any provider but the one that was reported" is outdated now that `custom` is included.

PR #574, opened 2026-08-29
docs: remove Claude residue from process docs; AGENTS.md becomes canonical

green-dalii:
## Summary

AGENTS.md and CLAUDE.md were 100% byte-identical mirrors (391 lines each) of the same process standards. After repeated Claude Code sessions, both files accumulated Claude-specific paths (`~/.claude/projects/...`, `~/.claude/skills/...`) and `CLAUDE.md` as a file name in cross-references. This caused two problems:

1. Every fix to process standards had to be made twice (drift risk).
2. The same-name mirror meant any contributor looking at CLAUDE.md would assume it was canonical, while AGENTS.md is what the system actually loads.

## Changes

- **AGENTS.md** becomes the single canonical source for all process standards. Six sections updated: current-state pointer, Gate 5 docs list, Git Commit Standards (Co-Authored-By rule generalized to all AI agents), architect-contributor memory path (Pi canonical + Claude legacy alias), Pre-Release skill path (Pi canonical + Claude legacy alias), Documentation Architecture table.
- **CLAUDE.md** reduced to a 22-line pointer stub pointing at AGENTS.md. Preserved for backward compatibility with external links (issues, PRs, docs that already reference it). Frozen — no new content here.
- **MEMORY.md** updated: 15+ 'CLAUDE.md §' cross-references rewritten to 'AGENTS.md §'; the '🤖 Generated with Claude Code' rule generalized to all AI agents (Claude, Codex, Cursor, Pi, …); session-log pointer now lists Pi canonical + Claude Code legacy alias.
- **ROADMAP.md** updated: process-standards link points at AGENTS.md with the right anchor; the obsolete skill path corrected to Pi canonical with legacy alias.

## Intentional preservation

All remaining references to 'Claude Code' / '~/.claude/...' in the four files are **explicitly marked as 'legacy alias for sessions running under Claude Code'**. The Pi canonical path is the primary one; the Claude path is kept only so existing per-session memory files keep working.

## Co-author trailer hardening

AGENTS.md's Co-Authored-By prohibition now covers **all** AI agents (Claude Code, Codex, Cursor, Pi, …), not just Claude. Cross-release lesson learned from PR #573's commit `74920c8` squash where pre-session Claude-Authored-By trailers leaked into a maintainer-authored squash commit.

## Verification

- Gate 1: lint + tsc + build + test + css-lint all green (3690 tests pass)
- shazam_verify: PASS
- `pnpm build:dev`: clean (13.6 MB main.js with inline sourcemap)
[...]

PR #761, opened 2026-09-20
docs(agents): a squash merge is how a contributor's trailer reaches a maintainer commit

green-dalii:
## Why

An audit of `main` for AI co-author trailers found three commits authored by the maintainer that carry `Co-Authored-By: Claude Code <noreply@anthropic.com>`, from 2026-08-30 and 08-31.

All three are squash merges of his own PRs. GitHub's squash message concatenates the branch's commit messages, so a trailer written on a branch commit lands in a commit whose `author` is the PR author — which at the time was him. All three are unrewritable: `main` is protected and the history is released. So the rule is written down instead of the history repaired.

The irony is recorded rather than smoothed over. One of the three is `e054012d` (#574), *"docs: remove Claude residue from process docs; AGENTS.md becomes canonical"* — the commit whose own subject is clearing Claude residue.

## What the rule says

The existing text already said maintainer commits carry no AI trailer. It was not enough, because the failure does not happen at the moment of writing — it happens at the moment of **merging**, and a squash merge is the one path where a maintainer does not author the text that ends up on his commit.

So the rule is stated as *read the message you are about to commit*, with the scan to run against the draft message rather than against the branch, and the handling for the case where the trailer is the contributor's: keep it on their work and off the merger's, via `--subject`/`--body` or the UI.

## Second finding, same audit

A maintainer-invoked `gh pr update-branch --rebase` rewrites the contributor's commit identity. It rebases with the invoker's credentials, so every rebased commit keeps its `author` and takes `committer = maintainer` — observed on #656, where five commits moved from `committer=Jan Heldal` to `committer=Greener-Dalii`.

Credit is not lost: `author` drives the contributors graph and `git log --author`, and a squash merge discards the committer field entirely. But it is a visible artifact on someone else's commits, and merging `main` into the branch leaves their identity untouched. Recorded so the choice is deliberate rather than habitual.

## Third, minor

Grepping a commit message for `codex|claude` produces false positives on product names — three hits in one commit were all "OpenAI Codex OAuth" in prose. Match the trailer syntax (`^(co-authored-by|generated with|signed-off-by|assisted-by):`), not the vendor name. Also noted, because the same audit briefly believed it had found a fourth violation in its own commit.

## Scope

Process documentation only — `AGENTS.md`, 8 added lines, no code. No closing keyword: this is a standing rule, and the three commits it describes are unrewritable history rather than work to be tracked.


green-dalii:
Merging with maintainer bypass, on the maintainer's explicit approval (docs-only, Gate 1 green).

Review event is a comment rather than `--approve` because GitHub blocks self-approval on an own PR. The change is 8 added lines to `AGENTS.md` recording three findings from an audit of `main`: three maintainer-authored commits carry `Co-Authored-By: Claude Code` because GitHub's squash message concatenates the branch's commit messages; a maintainer-invoked `gh pr update-branch --rebase` rewrites the contributor's `committer`; and a vendor-name grep produces false positives on product names.

PR #762, opened 2026-09-20
docs(i18n): the nine locales that still say the tag vocabulary is not a write gate

green-dalii:
## What

The nine locales whose tag-vocabulary bullet still asserts the opposite of what #673 ships.

#673 reverses a factual claim in the EN bullet — from "The vocabulary is a schema-injection hint, not a wire-level gate" to "what the model is offered is exactly what lands on disk; a value outside it is dropped, never written". Only `README.md` and `docs/README_DE.md` moved with it. The other nine keep the old assertion, each in its own words: `README_CN.md` "而非写入时的强制闸门", `README_JA.md` "書き込み時の強制ゲートではありません", `README_RU.md` "а не шлюз принуждения при записи", and the same shape in `_ZH-Hant` / `_KO` / `_FR` / `_ES` / `_PT` / `_IT`.

Nothing contradicts anything today, because the EN claim has not landed either. The contradiction appears the moment #673 merges without this — and then nine locales tell a user that the vocabulary does not gate writes, on the one feature whose whole point is that it does.

This is the PR promised on #673 on 12.09 and again on 15.09. It is the last item blocking that PR, and it is a maintainer's debt rather than a contributor's.

## Method, since nine translations cannot be verified by reading them

Each bullet is rendered in its file's own existing terminology rather than translated from the EN text: 标签词汇表 / 標籤詞彙 / タグ語彙 / 태그 어휘 / Vocabulaire de tags / Vocabulario de etiquetas / Vocabulário de Tags / Vocabolario tag / Словарь тегов. The Settings path is the one each file already uses, and those differ — JA and ZH-Hant say タグ語彙モード / 標籤詞彙模式 where the others say the plain form.

The new text carries three claims that had to survive in all nine: the three sources of the tag list, the gate property (offered = written), and the small-model drift note. Verified mechanically: zero occurrences of the old assertion remain across the nine, each file has exactly one new bullet, and the `#328` design-anchor link is intact in all nine.

## Scope

Nine lines, one per file. No code, no EN, no DE — those are #673's.

Gate 1: 305 files / 4253 tests, including `readme-links`.

Lands alongside #673 rather than replacing it. The two are one change to a user-visible claim, and merging either alone produces the state this exists to prevent.


green-dalii:
Merging, as the companion half of #673 (merged moments ago as the previous commit on `main`).

Both halves were required to land together: #673 reverses a factual claim in the EN bullet, and leaving the nine locales asserting the opposite would have told users the vocabulary does not gate writes on the one feature whose point is that it does. The order between them did not matter — only that the second was not abandoned.

Review event is a comment rather than `--approve` because GitHub blocks self-approval on an own PR. Docs only, nine lines, Gate 1 green, no code.

Issue #764, opened 2026-09-20
Coordinated AI SDK v7 upgrade: @ai-sdk/* 3.x is specification v4 and ai@6 accepts v3


green-dalii:
## What

`ai` 6 → 7 with the provider packages moved in step. Filed because Dependabot's PR cannot express this: #728 bumps `@ai-sdk/openai-compatible` alone, and that cannot compile.

## The measurement

Rebuilt on current `main` with `pnpm add @ai-sdk/openai-compatible@3.0.48` and nothing else touched:

```
src/llm-sdk/openai-compat-sdk-client.ts(298,5): error TS2322:
Type 'LanguageModelV4' is not assignable to type 'LanguageModel'.
  Type '"v4"' is not assignable to type '"v2"'.
```

- `@ai-sdk/openai-compatible@3.0.48` depends on `@ai-sdk/provider@4.0.14` → specification **v4**
- `ai@6.0.230` declares `type LanguageModel = GlobalProviderModelId | LanguageModelV3 | LanguageModelV2` → takes v3 at the newest
- `ai@latest` is **7.0.107**

The repo currently pins `@ai-sdk/anthropic@3.0.98`, `@ai-sdk/openai@3.0.86`, `ai@6.0.230`. All three move together or none does.

## Why it is not a routine bump

Three things in the 3.0.0 line land on code we wrote on purpose:

1. `74d520f: feat: migrate providers to support new top-level 'reasoning' parameter` — this is where the force-disable-thinking work lives. `openai-compat-request-body.test.ts` pins `reasoning_effort: 'none'` on the wire; that pin is the first thing to re-read.
2. `ef992f8: remove CommonJS exports` — ESM-only. The build is `format: 'cjs'` with `main.js` bundled, so this needs checking rather than assuming.
3. `7fc6bd6: min Node 22`. CI and the Obsidian runtime both need to be at or above it.

`9bd6512` (file part data property change) also collides with the local-image work in #687.

## Acceptance criteria

- `ai`, `@ai-sdk/anthropic`, `@ai-sdk/openai` and `@ai-sdk/openai-compatible` move together to a spec-v4-consistent set.
- `openai-compat-request-body.test.ts` still passes, and the `reasoning_effort: 'none'` assertion is re-derived rather than adjusted to fit.
- Gate 1 green, including `pnpm lint:tools-bot`.
- The 3.0.0 line's ESM-only and Node 22 minimums are resolved or explicitly accepted in the PR body.

[...]

PR #765, opened 2026-09-20
chore(deps): stop Dependabot asking for @ai-sdk/openai-compatible 3.x alone

green-dalii:
## Why

`@ai-sdk/openai-compatible` 3.x is specification v4 and `ai@6.0.230` accepts v2/v3 — `ai@latest` is 7.0.107 — so a PR bumping this dependency by itself cannot typecheck. Measured on a clean rebuild in #728:

```
src/llm-sdk/openai-compat-sdk-client.ts(298,5): error TS2322:
Type 'LanguageModelV4' is not assignable to type 'LanguageModel'.
  Type '"v4"' is not assignable to type '"v2"'.
```

#728 had been rebased four times by its author and could never have gone green. The coordinated `ai` 6 → 7 upgrade is #764.

## Scope

One `ignore` entry for the major only — 2.x patch and minor updates still arrive. The entry comes out when #764 lands, not before.

Same shape as the two entries already in the file: `eslint` stays at the Bot-aligned major, and `brace-expansion` is force-pinned by an override so a PR for it can never merge either. Three comments, three specific reasons, each naming what would clear it.

## Gate 1

Green. No code touched.


green-dalii:
Merging with maintainer bypass on the maintainer's direction to apply the earlier finding.

Review event is a comment rather than `--approve` because GitHub blocks self-approval on an own PR. Config only — one `ignore` entry, no code, Gate 1 green (306 files / 4274 tests).

Also noting a self-caught mistake for the record: the first Gate 1 run on this branch failed with the very `LanguageModelV4` error this PR describes, because `node_modules` still held 3.0.48 from the #728 measurement. `git checkout` reverts tracked files and `node_modules` is not tracked, so the revert was incomplete. `pnpm install --frozen-lockfile` fixed it. Worth knowing because it makes any local Gate 1 result after a dependency experiment untrustworthy until reinstall.

Issue #756, opened 2026-09-19
Query wiki panel layout: the conversation gets under half the height — compact the fixed chrome around it


aisahpA:
**Is your feature request related to a problem? Please describe.**

The Query wiki view is a right-docked pane, so it is already the narrowest place in the workspace, and vertically it is stacked like this: a title row, a hint row, the conversation, then an input block (composer + three buttons + the collapsed Custom Query Instructions panel + a round counter). Every one of those is `flex-shrink: 0`; only the conversation is flexible. So the conversation gets whatever is left over, and in practice that is roughly **half** the pane:

| Part of the pane | Roughly |
|---|---|
| Title row + hint row | ~10% |
| Input block (composer, buttons, Custom Query Instructions, counter) | ~40% |
| **Conversation — the part you actually read** | **the remaining ~50%** |

The ratio is the problem, and several of those blocks claim their space whether or not they have anything to show:

- **The composer is the largest single block in the pane, and it is empty.** It reserves a fixed height that has nothing to do with the text inside it (`styles.css:11`), and it cannot be resized by hand (`resize: none`, `styles.css:19`), so it is as tall as it will ever need to be before a single word is typed, and the user cannot hand any of it back. An empty composer ends up roughly **45% as tall as the whole conversation area**, and the input block as a whole is roughly **85%** of it — nearly the same size as the thing it is there to serve.
- **The Custom Query Instructions panel is collapsed and still costs a bordered block with margins** (`styles.css:1211-1227`), which is a lot of vertical space for one line of summary text the user is not currently reading.
- **The hint row is always visible** (`styles.css:96`) even though it is read once and never changes (`QueryView-class.ts:207`). It is the block that scales worst: it is a single sentence, so on a narrow pane or in a longer translation it wraps to two or three lines and takes *more* room exactly when the pane has less to give.
- **The round counter owns a row** (`styles.css:179`) for a few characters.

It is worth saying plainly that none of this is a bug — each block is a reasonable choice on its own, and read in isolation the layout is fine. It is the sum that does not work in a docked pane: you open a view to read the wiki's answers and find that half the view is controls and an empty text box.

**Describe the solution you'd like**

Four changes, all small and all local to `styles.css` and the Query view's `onOpen()`. They are listed so any subset can be taken:

[...]

PR #769, opened 2026-09-20
chore(deps): hold the obsidian and @types/node bumps, with what each would cost

green-dalii:
## What

Two Dependabot PRs that are not dependency bumps. Both measured; neither is necessary; each ignore entry names what would clear it.

## obsidian 1.12.3 → 1.13.1 (#766)

1.13 deprecates `PluginSettingTab.display()` in favour of `getSettingDefinitions()`, so the bump surfaces 34 `@typescript-eslint/no-deprecated` warnings — and that rule is on the Bot's no-disable list, so they cannot be suppressed. **The bump and the migration are one change**: 42 `display()` sites plus 68 imperative `new Setting(...)` calls.

**What makes it a choice rather than a requirement:** `display()` is not removed. Its own doc comment in `obsidian.d.ts` calls it "a fallback for plugins that need to support Obsidian versions older than 1.13.0". So `minAppVersion` does not have to move either — the one thing the migration buys is that Obsidian's own settings search can index our options.

Held until that is judged worth 110 rewrites, and planned alongside the settings-UI work (#668, #752, #756) rather than as a dependency bump that drags them along.

## @types/node 16 → 26 (#768)

The cost is the opposite of its appearance. 26 adds nothing this project uses — `node:module` (`createRequire`) and `node:https` have existed since Node 15 — and it **removes** the `compatibility/indexable.d.ts` shim that 16 ships, which is what currently makes `Array.prototype.at` typecheck under `lib: ES2021`. Taking it needs `lib` raised to ES2022 or 31 `.at()` call sites rewritten, for types of APIs we do not call. A type package has no runtime surface, so there is no security side to weigh.

## Verification note

The 31 `.at()` errors are absent on `main` and appear only with 26 installed — a distinction I had to re-measure after a `node_modules` mix-up of my own. A dependency experiment leaves the installed tree ahead of the lockfile, and `git checkout` does not clean it; any local Gate 1 result before `pnpm install --frozen-lockfile` is untrustworthy. Recorded in #765 too.

Gate 1: 306 files / 4274 tests.


green-dalii:
Merging with maintainer bypass, at the maintainer's direction to apply the analysis.

Review event is a comment rather than `--approve` because GitHub blocks self-approval on an own PR. Config only — two `ignore` entries and no code. Gate 1: 306 files / 4274 tests.

The substance is in the entries themselves: each names what it would take to clear it. The `obsidian` one records that `display()` is *not* removed (its doc comment calls it a fallback for supporting < 1.13.0), so this is a choice about settings searchability rather than a forced migration.

PR #773, opened 2026-09-20
docs(notice): generate the bundle's third-party notices from the artifact

### LINKED ISSUE #699: docs(notice): the shipped bundle carries no third-party licence notices
The shipped `main.js` bundles ten third-party packages and carries none of their licence notices. `NOTICE` lists contributors only — there is no third-party section — and the bundle itself contains zero occurrences of the word "copyright".

This is a pre-existing gap, not a regression, and I am raising it now because it was the one item in a review of #570 that belongs to the product rather than to that PR.

**Measured on `main` (`6a2dae7`) and on the `main.js` published with release 1.27.1:**

| | |
|---|---|
| Production dependencies | 6 direct, 10 in the resolved tree |
| Licences | 8 × Apache-2.0 (`ai`, `@ai-sdk/anthropic`, `@ai-sdk/openai`, `@ai-sdk/openai-compatible`, `@ai-sdk/gateway`, `@ai-sdk/provider`, `@ai-sdk/provider-utils`, `@opentelemetry/api`), 2 × MIT (`zod`, `fflate`) |
| Bundled? | Yes — `esbuild.config.mjs` marks only `obsidian`, `electron`, the CodeMirror/Lezer packages and Node builtins as external, so all ten land in `main.js` |
| Notices in the artefact | 0 hits for `copyright`, for `@license` and for any licence text in the released `main.js` (4 124 556 bytes, downloaded from the 1.27.1 release) |
| Notices in the repository | `NOTICE` has a contributors list and no third-party section |

**Why it matters for each licence.** MIT requires its copyright and permission notice to travel with "all copies or substantial portions of the Software", and a minified copy inside `main.js` is a copy. Apache-2.0 §4(a) requires recipients of a derivative work to get a copy of the License, and §4(b) that attribution notices be retained. None of the four `ai` packages ships its own `NOTICE` file — I checked — so §4(d) adds nothing here; §4(a) and §4(b) still apply, and the plugin is distributed through the Obsidian community marketplace, which is distribution.

**What would close it.** A `THIRD-PARTY-NOTICES.md` generated from the production tree, referenced from `NOTICE` and from the README's licence line, is the usual shape and keeps `NOTICE` as the contributor record it currently is. The alternative — a third-party section inside `NOTICE` itself — is equally correct and keeps everything in one file. Generation should be mechanical rather than hand-maintained, so it cannot drift the way a hand-written list would; the dependency set is small enough that a build step is optional.

green-dalii:
## What

The shipped `main.js` bundles twelve third-party packages and carried none of their notices — zero hits for `copyright`, `@license` or any licence text in the released bundle, and a `NOTICE` that listed contributors only.

MIT requires its copyright and permission notice to travel with copies, and a minified copy inside `main.js` is a copy; Apache-2.0 §4(a) requires recipients of a derivative work to receive a copy of the License and §4(b) that attribution be retained. The plugin is distributed through the Obsidian community marketplace, which is distribution.

## The bundle decides what is bundled

`esbuild.config.mjs` marks `obsidian`, `electron`, the CodeMirror/Lezer packages and Node builtins as external, but which of the rest reaches `main.js` is decided by the entry point's imports — and the resolved set includes transitive packages no direct dependency names.

Counting by hand gave **ten**. The artifact contains **twelve**, and the two it missed — `@vercel/oidc`, `eventsource-parser` — are transitive two levels down from anything `package.json` names. That gap is the argument for the shape of this PR: the generator reads the `// node_modules/.pnpm/<pkg>@<ver>/` markers esbuild emits, so the set cannot be hand-maintained into staleness, and a test asserts the committed file matches the bundle byte for byte.

## Two things the packages' own files did not settle

**Seven of nine Apache-2.0 packages ship only the short-form notice** — `Copyright 2023 Vercel, Inc.` plus "You may obtain a copy of the License at http://…". Reproducing that alone does not satisfy §4(a), which asks for a copy and not a pointer, so the full text appears once in an appendix, taken from a bundled package that carries it. Each package's own section still reproduces what it ships, and says when that is the short form.

**`@ai-sdk/provider-utils@4.0.40` declares Apache-2.0 and ships no licence file at all.** Failing the whole document over one upstream omission would leave eleven packages unattributed; omitting the entry would leave that one unattributed. The entry reproduces the text from a sibling under the same licence and names which — so the substitution is visible rather than implied.

## Files

| | |
|---|---|
| `tools/generate-third-party-notices.mjs` | reads the bundle, writes the document; `--check` fails when stale |
| `THIRD-PARTY-NOTICES.md` | generated, 12 packages, 899 lines |
| `src/__tests__/root/third-party-notices.test.ts` | imports the generator rather than re-deriving its rules |
| `NOTICE` | third-party section pointing at the generated file |
| 11 × README | licence line references it |
| `readme-links.test.ts` | allowlist gains `THIRD-PARTY-NOTICES.md` — the same class as `LICENSE`/`NOTICE`, which the guard already exempts for exactly this reason |

## Tests

Mutation-tested three ways:

| Mutation | Result |
|---|---|
| Delete a package's section | fails 2 — the exact-match assertion and `zod must have its own section` |
| Tamper with a version | fails 2 — `expected '## zod 4.6.4' to be '## zod 4.6.5'` |
| Remove the Apache appendix | fails 1 |

Gate 1: 307 files / 4281 tests.

Closes #699


green-dalii:
Recording a gap I found in the stage self-check, and the decision taken on it, so it is not rediscovered later as an open question.

**The notices are in the repository, not in the artifact.** Measured after generating: `main.js` still contains zero occurrences of `copyright`. Strictly, MIT asks that the notice "be included in all copies or substantial portions of the Software", and the Obsidian marketplace serves users `main.js`, `manifest.json` and `styles.css` — so someone installing through Obsidian does not receive `THIRD-PARTY-NOTICES.md`.

**Decided: keep it repository-level.** This is how the project satisfies the same obligation for its own licence today — `LICENSE` and `NOTICE` are not in `main.js` either, and Apache-2.0 §4(a) binds the project the same way. It is also the shape #699 itself recommended. A build-banner injection would be stricter and would cost roughly 12 KB against a 4.5 MB bundle, but it would also make the shipped artifact's byte count depend on the third-party dependency set, which is a coupling worth not creating for a file the repository already carries and the marketplace page links to.

The gap is worth stating rather than leaving implicit in case the distribution model changes — if the plugin is ever shipped somewhere that serves `main.js` without the repository, this decision is the thing to revisit.

green-dalii:
Merging with maintainer bypass, at the maintainer's direction to keep this repository-level (方案 A).

Review event is a comment rather than `--approve` because GitHub blocks self-approval on an own PR. Gate 1 green locally and in CI; mutation-tested three ways; `lint:tools-bot` clean.

PR #774, opened 2026-09-20
fix(prompts): stop asking the model for what the code writes

DocTpoint:
## What

Follow-up to #680 and #681. Since those two, the code stamps the note path on every quote and renders the head of the source page. The prompts and the wire schema still asked the model for the same values. The copies were overwritten, so nothing was wrong on the page — but the extraction paid for them on every quote.

This PR removes the requests. It does not change what is written.

**Extraction prompt** (`prompts/ingestion.ts`): the quote example and rule 4b no longer name `source_path`, `source_slug` or `extracted_at`. The note path stays in the prompt as context, because for a note without an H1 the file name is the only title the model sees.

**Wire schema** (`output-schemas.ts`): `MentionWithProvenanceItem` no longer declares the three fields. This is the part I missed at first and a review found: a declared property is a request. At the `json_schema_strict` tier every property is listed in `required`, so there the model had to emit all three on every quote, whatever the prompt said. `.loose()` still accepts a model that sends them on the plain tier.

**Source page template** (`prompts/generation.ts`): no `source_file:` line, no H1, no Source section. #681 renders all three from the code.

**`fillMentionsWithProvenance`**: now also sets `extracted_at` and clears `source_slug` on structured quotes, as it already did for synthesized ones. The Mentions formatter sorts by `extracted_at`, and the old prompt example carried a fixed date.

## What stays, on purpose

- `source_title` stays in the extraction: for a PDF the model reads the title off the document.
- The entity and concept templates keep their `sources:` line. The conversation path passes no source slug and relies on the model's copy.

## Tests

- The pinned wire body of `SourceAnalysisLLMSchema` (#669) changes. I counted the snapshot diff: exactly the three properties and their `required` entries, in the two quote items, nothing else.
- New test on the strict wire body: a quote item has the properties `quote` and `translation` only. Mutation: putting `extracted_at` back into the schema fails this test and the pinned-bytes test.
- New tests that the extraction prompt and the source page template no longer contain the removed names.

Gate 1 locally: 307 files / 4284 tests, build, `tsc`, lint green.

## Overlap

Four open PRs touch the same files: #750, #755, #687, #701. All four merge cleanly with this branch (`git merge-tree`). For #750, the one that is mergeable today, I also ran `tsc` and the wiki, llm-sdk and mentions tests on the merged tree: green.

Refs #679


green-dalii (approved):
Approving. The thesis is right and the second half of it is the part I did not have: **a declared property is a request**. Removing the fields from the prompt but leaving them in `MentionWithProvenanceItem` would have left the strict tier asking for them anyway, because that tier lists every property in `required`. Catching that is what makes this a fix rather than half of one.

I verified the three claims I could check against the tree rather than the description.

The snapshot diff is exactly the three properties and their `required` entries, in both quote items — nothing else moved, which is what a pinned-bytes test is for.

`source_file` is safe to drop from the template: `wikiEngine` stamps it with `upsertFrontmatterField` after rendering (`src/wiki/wiki-engine.ts:1805`, the #679 fix), and `frontmatter.ts:133`, `lint/scanners.ts:529` and `lint/phases/programmatic.ts:132` all read the scalar form. The template's copy was already dead.

`source_slug: ''` is consistent rather than lossy — `mentions-parser.ts:195/217` and `mentions-integration.ts:92` all write the empty string, and `mentions-parser.ts:13` records that the round trip does not carry it. Setting it here makes the structured path agree with the synthesized one.

One note, not a request: `extracted_at` now comes from a single `now` per item rather than per quote, and it overwrites whatever a model sent. That is the intent — the formatter sorts by it, and the old prompt example carried a fixed date — but it is the one place where a page's bytes change rather than stay equal, so it is worth being the thing the next reader notices.

PR #775, opened 2026-09-20
refactor(ui): one sanctioned write path for the unified model

### LINKED ISSUE #467: Refactor: introduce single setter for tempSettings.model to replace the removed commit-time cascade

green-dalii:
## What

The unified `model` field owns three per-task overrides. Changing it has to clear them, or live task routing keeps using the old per-task model while the picker shows the new one — the 2026-07-13 bug, fixed inside `setFieldValue('model', …)` → `cascadeUnifiedModelChange`.

PR #462 correctly removed a commit-time belt-and-suspenders cascade. What it left was the guarantee resting on every future writer remembering the rule: three sites still assigned `tempSettings.model` directly, and #467's own words are that "the next contributor who adds any 'write tempSettings.model' feature will silently reintroduce" the bug.

## Two entries, and the difference is the point

`setFieldValue('model', value)` is the **user-edit** path: one field edited by hand expands into a coherent selection, so a non-empty value cascades. The two provider/Bedrock sites route through it with an empty value, which does not cascade — behaviour unchanged, and now stated at the call site.

`syncModelsFromPlugin()` is the **wholesale** path, and it is new: all four fields arrive together from `plugin.settings`, so cascading would clear three of them only to reassign them on the next line — and it would flip their `*UseCustom` flags as a side effect of nothing. That is a behaviour change rather than a refactor, which is why the connection-test site takes this entry instead of `setFieldValue`.

## Enforcement is a test, not the setter

`settings-cascade.test.ts` re-implements the cascade and notes in its header that "production code ... changes, update this mirror to match" — a mirror cannot fail when the thing it mirrors moves, which is exactly the failure mode #467 describes. `unified-model-write-guard.test.ts` reads the production tree instead and fails if any module but `ui/settings.ts` assigns the field.

## The guard was wrong on the first attempt

Mutation testing is what said so, and it is worth recording because the first version looked reasonable:

| Mutation | First version | Now |
|---|---|---|
| Reinstate a direct write in `provider-section` | fails 1 ✓ | fails 1 ✓ |
| Remove `setFieldValue`'s assignment | **passed** ✗ | fails 1 ✓ |
| Remove the cascade call | **passed** ✗ | fails 1 ✓ |

The assignment assertion passed because `syncModelsFromPlugin` assigns the same field; the cascade assertion passed because `toContain('cascadeUnifiedModelChange')` matches the method's own definition, and a lazy `[\s\S]{0,400}?` span reaches past the closing brace to find it. Both are now bound inside the model branch with `[^}]*`.

Gate 1: 308 files / 4285 tests.

Closes #467


green-dalii:
Requesting your review here, @DocTpoint — Stage 3 of the #603 write-path chain, and the last piece of #467.

Context that may save you time: the enforcement is deliberately **not** in the setter. `src/__tests__/ui/unified-model-write-guard.test.ts` reads the production tree and fails if a *fourth* direct write to `tempSettings.model` appears, or if the `setFieldValue('model', ...)` path stops cascading. My first version of that guard passed two of three mutations (the assignment assertion was satisfied by `syncModelsFromPlugin`, and the cascade assertion matched the method *definition* because the lazy `[\s\S]{0,400}?` crossed the closing brace) — it now binds both branches with `[^}]*`.

The design decision worth your eye: the two paths intentionally differ. `setFieldValue` cascades on a non-empty value; the new `syncModelsFromPlugin()` deliberately does **not**, because cascading there would flip the `*UseCustom` flags as a side effect of nothing having been typed.

Gate 1 green locally and in CI. `Closes #467`.

DocTpoint (changes requested):
Reviewed at `033e1842`.

The runtime change is correct. The two provider-section sites behave as before: an empty value does not cascade, and the `llmReady = false` that `setFieldValue` adds was already set a few lines earlier. `syncModelsFromPlugin` copies the same four fields the removed lines copied. What I am requesting changes on is the guard. It is the point of this PR, and in two places it does not hold what it says.

## 1. The guard does not see writes through a parameter, and one already exists

`writers()` looks for the text `tempSettings.model =` (`unified-model-write-guard.test.ts:67`). A write through a function that receives the settings object as a parameter does not contain that text. There is already one outside `settings.ts`:

- `test-connection-section.ts:71` calls `preserveCodexRuntimeModelState(tab.tempSettings, tab.plugin.settings)` on the failure path.
- `openai-codex-model-policy.ts:40` then runs `target.model = runtime.model`.

This writes `tempSettings.model` outside `settings.ts`, through neither `setFieldValue` nor `syncModelsFromPlugin`, and `writers()` returns `[]`. `applyCodexModelPolicy` does the same at `openai-codex-model-policy.ts:25` (called from `settings.ts:322` and `:628`). So the sentence at `settings.ts:338`, "`tempSettings.model` is only assignable here", is not true today.

This matters for the case #467 names. An "apply preset" or "profile import" feature is most naturally written as a helper that takes the settings object — `applyPreset(tab.tempSettings, preset)` — and that is exactly the shape the guard cannot see. Behaviour-wise the existing site is fine, because it is a sync and must not cascade. But then it should be named as a sanctioned writer, not be invisible.

Two ways that I see: route the tempSettings half of `preserveCodexRuntimeModelState` through `syncModelsFromPlugin` too, or keep it and list the Codex policy functions explicitly as a second sanctioned writer, with the guard checking that those are the only functions receiving `tempSettings` that assign `.model`.

## 2. The reason given for "sync must not cascade" is a no-op on its only caller, and the real reason is not named

The comment at `test-connection-section.ts:82–86` says that going through `setFieldValue` would clear the three per-task values before they are reassigned, and the doc comment at `settings.ts:331–337` adds that this would flip their `*UseCustom` flags. On the only caller, neither has an effect: the values are reassigned from the same source on the next lines, and the flags are already `false`. `syncModelsFromPlugin` runs directly after `syncCodexModelsFromPlugin`, whose `applyCodexModelPolicy(this.tempSettings)` has already set all three `*UseCustom` flags to `false` (`openai-codex-model-policy.ts:30–32`). The cascade would find nothing to flip.

What would break is `llmReady`. `setFieldValue` always ends with `markLLMConfigStale()` (`settings.ts:514`), which sets `llmReady = false`. The success path set `tab.tempSettings.llmReady = result.success` at `test-connection-section.ts:56`. If the sync went through `setFieldValue`, a *successful* Codex connection test would commit `llmReady = false`, and `requireLLMReady` (`connection-commands.ts:196`) would then reject every ingest and query with "LLM not ready". That also breaks the invariant in the file header: "Final llmReady is set to result.success regardless of branch".

Test 4 (`unified-model-write-guard.test.ts:115–122`) does not catch this. It checks that the first 400 characters after `syncModelsFromPlugin` do not contain the word `cascadeUnifiedModelChange`, and a call to `setFieldValue` does not contain it. So a future editor who reads the comment, concludes correctly that the flag flip is harmless here, and routes the sync through `setFieldValue` passes all tests and breaks Codex.

The fix is small: name `llmReady` as the reason in both comments, and let the test assert that the sync calls neither `setFieldValue` nor `markLLMConfigStale`. Better still is one behavioural test: after a successful Codex connection test, `llmReady === true`.

## Two smaller points

- **`syncModelsFromPlugin` has no behavioural test.** `settings-codex-sections.test.ts:57` replaces it with `vi.fn()`, and nothing calls the real method. If a line is dropped (say `queryModel`), a stale value is committed and no test notices. The behavioural test from point 2 would cover this as well.
- **The regex also matches comparisons.** `/\btempSettings\.model\s*=/` matches `tempSettings.model === ''`. A harmless read in a section module would fail the guard with "a writer bypassed setFieldValue". A `(?!=)` after the `=` avoids it.

PR #776, opened 2026-09-20
docs(memory,roadmap): refresh the resume point for the post-compact handoff

green-dalii:
Docs only — MEMORY.md and ROADMAP.md, no code.

Nine PRs merged since the last snapshot, `main` moved `3499da2a` → `4f56b475`, and two of the three items blocking the v1.28.0 chain closed, so the current state, the ROI list and the resume point all described a window that no longer exists.

**#603 closed, so the chain's head moved to #729 Phase 1** — both gates open (#603 with #750, #662 with #757). Tier 0 was the #751 one-liner, which landed as #759; Tier 0′ is now Phase 1, carrying the note that `RELATED_BUDGET.siblings` must not be read before Phase 2.

**One branch is parked and named in the resume point** so it cannot be lost: `pr-687` @ `e3273353`, the rebase of @Chase07's PR — done, green, unpushable until `maintainer_can_modify` is true.

**Eight durable lessons**, most from review rather than my own passes: a dependency experiment leaves `node_modules` ahead of the lockfile and `git checkout` does not clean it; the triage skill's dedup read `--state merged` only, so an open PR already fixing an issue was invisible and a re-implementation followed; and a test can protect a property while ending up protecting a spelling — three times in one week, three different files.

Gate 1: 309 files / 4299 tests.


green-dalii:
Merging with maintainer bypass, at the maintainer's direction. Docs only — MEMORY.md and ROADMAP.md, no code.

Review event is a comment rather than `--approve` because GitHub blocks self-approval on an own PR. Gate 1 green: 309 files / 4299 tests.

PR #777, opened 2026-09-20
docs(memory): a suppressed ruleset restore hid two failures

green-dalii:
Extends the handoff PR's lesson with the part that cost time: the ruleset `PUT` ran with `>/dev/null 2>&1`, so two network failures were invisible and the run reported "restored" while `bypass_actors` stayed at 1. A direct read found it several commands later.

The restore is the one write in the merge flow whose failure is silent by construction — the merge reports, the bypass is verified before use, and only the restore has no downstream signal. So it is the one that has to be read back.

Docs only. Gate 1 green on the parent commit (309 files / 4299 tests).


green-dalii:
Merging with maintainer bypass, at the maintainer's direction. Docs only — one lesson paragraph in MEMORY.md.

Review event is a comment rather than `--approve` because GitHub blocks self-approval on an own PR. Gate 1 green on the parent commit (309 files / 4299 tests).

PR #778, opened 2026-09-20
fix(test): resolve custom-instruction source paths on Windows

x0Lazarus:
On Windows, the custom-instruction wiring test fails during collection: using a file URL's `.pathname` produces a path such as `/C:/repo/src`, which `resolve()` turns into `C:\C:\repo\src`. None of its eight assertions can run.

Convert that URL with Node's `fileURLToPath`, matching the existing README-link test. This gives filesystem APIs a native path and decodes escaped characters. The existing assertions and production code stay unchanged.

AkaSakana already explained this conversion in the [PR #286 discussion](https://github.com/GD4AI/obsidian-llm-wiki/pull/286#issuecomment-4977075104). The Windows failure was later reported in [PR #538](https://github.com/GD4AI/obsidian-llm-wiki/pull/538); this applies the remaining conversion in that test.

Validation on Windows with Node 24.19.0 and pinned pnpm 10.14.0:

- The unchanged test fails collection before the patch; all eight assertions pass afterward.
- Lint, typecheck, production build, the full suite (4,299 tests across 309 files), and CSS lint pass in the documented order, with no test exclusions.
- Lockfile consistency and the development build also pass.

These are local checks. Linux/macOS and hosted CI have not been run for this branch.


green-dalii (approved):
Approving. The diagnosis is exact and the fix is the one the codebase already uses: `readme-links.test.ts:38/41` resolves the same way with `fileURLToPath`, so this is not a new convention, it is the missing application of one.

`new URL(...).pathname` gives `/C:/repo/src` on Windows, and `resolve()` then treats the leading segment as a relative component — `C:\C:\repo\src`. `fileURLToPath` also decodes percent-escapes, which the `.pathname` form leaves in place, so this fixes the class rather than the one reported path.

Test-only, one file, and the production code is untouched. I released the Gate 1 run for this branch (it had been waiting on first-contributor approval) and it is green.

PR #779, opened 2026-09-21
docs(memory,roadmap): the 2026-09-21 handoff — three gates closed, and the CI that never ran

green-dalii:
Docs only — `MEMORY.md` and `ROADMAP.md`. Gate 1 green on this branch: 311 files / 4349 tests, the same counts as `main`, because nothing executable changes.

**What changed in substance rather than in dates.**

All three of the chain's gates are now closed — #603 (#750), #662 (#757) and #608 (#687) — so **#729 Phase 1 is the head of the queue with nothing ahead of it.** The work list says that instead of listing the gates it was waiting on, and the "next substantive work" line in the resume point now names a task rather than a blocker.

The Dependabot trio is recorded as **three different failures rather than one hold**, because that is what it is: #772 is the `ai` 6→7 spec-v4 mismatch, #771 is the identical mismatch arriving through `@ai-sdk/anthropic` 4.x, and #770 is TypeScript 6 tightening definite assignment on four `main.ts` fields plus variance across the settings sections. #764 was filed as a hunch about a coordinated upgrade; those three PRs are the evidence, and the work list now says the outstanding item is a `dependabot.yml` entry per hold rather than a weekly explanation.

**Ten durable lessons**, of which one is worth naming here because it cost three files:

> A report's summary of the rule is not the rule. #758's excerpt omits the `openrouter` branch that precedes the two lines it quotes, so its prose inherits the omission — and copying that prose into the module header, the test header and the PR body is how a wrong sentence becomes documentation.

The others cover the `action_required` CI class (five runs, including all three Dependabot PRs, had never run at all), the fork-push mechanic, why a graceful-degradation path must name exactly which error it swallows, and why a documented-but-unreachable state is worse than an undescribed one.

**Housekeeping in the same pass:** `pr-687` is deleted — the rebase branch needed protecting for exactly one turn — and the two branches that are live and equal to their PR heads are now named with their SHAs, so a future session does not have to re-derive which are real.

**One thing this pass did not fix:** MEMORY is 1662 lines, up from 1540. The growth is where it always is, in the `## Superseded snapshots` chain and the per-session lessons blocks, and it wants a de-bloat pass of its own rather than a trim smuggled into a handoff.


green-dalii:
Merging at the maintainer's direction. Docs only — MEMORY.md and ROADMAP.md, no executable change, and Gate 1 reports the same 311 files / 4349 tests as main.

Review event is a comment rather than `--approve` because GitHub blocks self-approval on an own PR. The merge is recorded against the bypass actor for the same reason.

